PR #32, opened 2014-06-07
Add PostgreSQL database driver

peace-maker:
This is regarding [bug 3849](https://bugs.alliedmods.net/show_bug.cgi?id=3849).
Adds a "pgsql" driver to the dbi system.

See https://forums.alliedmods.net/showthread.php?t=236415

The AMBuild changes are very likely wrong. I don't know how to use it ;)

Here are some instructions on how to create static libs of the libpq out of the [source](http://ftp.postgresql.org/pub/source/v9.3.4/postgresql-9.3.4.tar.gz)

Get the latest postgresql source.
extract it.
# 

Windows:

In a VS-commandline, navigate to src/tools/msvc and run
  perl mkvcbuild.pl
to generate the VS solution in the root.
1. Open the pgsql.sln in VS2010 
2. change the configuration type of the libpq project from "Dynamic library (.dll)" to "Static library (.lib)"
3. under C/C++ -> codegeneration change runtimelibrary from "Multithreaded-Debug-DLL (/MDd)" to "Multithreaded-Debug (/MTd)" for the libpq and libpgport projects
4. compile the libpq project
5. the static .lib will be in Release/libpq/libpq.lib and Release/libpgport/libpgport.lib
# 

Linux:

in the extracted postgresql source run
  ./configure LDFLAGS='-static -m32' CFLAGS='-m32' --prefix=`pwd`/build
You might need to add --without-readline. This is not needed for the libpq library.
  cd src/interfaces/libpq
  make install
  rm libpq.a
  make install enable_shared=no
The library will be in the build/lib directory.

About that double linking to get a good static library: http://www.postgresql.org/message-id/1902207738.20070920111734@comodo.com


peace-maker:
I've rebased this against master.
libpq can be [build natively statically](https://www.postgresql.org/docs/current/static/install-windows-libpq.html) on windows now. They don't include the static .lib in their binary distributions though.
To build on windows it's enough to run the following in the downloaded source
```
cd src
nmake /f win32.mak
```
that creates the static lib in `interfaces\libpq\Release\libpq.lib`
The compilation on linux is covered in the first and second comment.

I'd now vote for a self-hosted archive with the libs and headers.

extensions/pgsql/pgsql/PgDatabase.cpp
```diff
@@ -184,12 +200,20 @@ const DatabaseInfo &PgDatabase::GetInfo()
 
 unsigned int PgDatabase::GetInsertID()
 {
-	return m_lastInsertID;
+	unsigned int lastInsertID;
+	LockForListQueryInfoAccess();
```
KyleSanderson: You're better off using ke::AutoLock.

spumer:
Any plans to merge?

peace-maker:
I've rebased this against master again and fixed not using all of databases.cfg config options.

extensions/pgsql/pgsql/PgDriver.cpp
```diff
@@ -0,0 +1,236 @@
...
+const char *PgDriver::GetProductName()
+{
+	return "PostgreSQL";
+}
+
+PGconn *Connect(const DatabaseInfo *info, char *error, size_t maxlength)
+{
+	/* https://www.postgresql.org/docs/9.6/libpq-connect.html#LIBPQ-CONNSTRING */
+	/* TODO: Switch to PQconnectdbParams to prevent escaping issues. */
+	char *options = new char[1024];
```
KyleSanderson: memleak

extensions/pgsql/pgsql/PgDriver.cpp
```diff
@@ -0,0 +1,236 @@
...
+	if (info->maxTimeout > 0)
+	{
+		offs += snprintf(&options[offs], 1024 - offs, " connect_timeout=%d", info->maxTimeout);
+	}
+
+	/* Make a connection to the database */
+	PGconn *conn = PQconnectdb(options);
+
+	/* Check to see that the backend connection was successfully made */
+	if(PQstatus(conn) != CONNECTION_OK)
```
KyleSanderson: crunch(!) (entire func has this problem)

extensions/pgsql/pgsql/PgStatement.cpp
```diff
@@ -0,0 +1,351 @@
...
+	PGresult *res;
+	/* Bind the parameters */
+	if (m_Params)
+	{
+		// Put bound params into a nice array
+		const char **paramValues = new const char*[m_Params];
+		int *paramLengths = new int[m_Params];
+		int *paramFormats = new int[m_Params];
+
+		for(unsigned int i=0;i<m_Params;i++)
```
KyleSanderson: crunch

extensions/pgsql/pgsql/PgStatement.cpp
```diff
@@ -0,0 +1,351 @@
...
+		delete [] paramValues;
+	}
+	// There are no parameters to be bound!
+	else
+	{
+		res = PQexecPrepared(m_pgsql, m_stmtName, 0, NULL, NULL, NULL, 0);
+	}
+
+	ExecStatusType status = PQresultStatus(res);
+	if(status != PGRES_COMMAND_OK && status != PGRES_TUPLES_OK)
```
KyleSanderson: crunch (it's elsewhere here too).

Andreychik32:
Ah, I thought that you started developing of this extension, but that was a lie. Sad.

KyleSanderson:
@dvander are you okay if we merge this one?

drozdowsky:
Guys would love to help if I had a knowledge here, would love to have postgresql in SM.

KyleSanderson:
@peace-maker https://github.com/alliedmodders/sourcemod/pull/32#issuecomment-678743860

Can you please address this (we just went through it with regex, I'm sure it's going to be the same with others). Tag me when it's done please. Please keep in mind we keep MySQL out of tree.

peace-maker:
@KyleSanderson Here's the missing win64 libpq build.

KyleSanderson (approved):

peace-maker:
Lets get this in the tree to poke at further 💯

Alienmario:
Hi
Don't know if you're aware, the driver is unable to connect to more recent psql versions due to this error:

`SCRAM authentication requires libpq version 10 or above`

I'm guessing that's corresponding to the DB version. I'm on windows.

psychonic:
It can connect to those versions without issue, just not with the default authentication type. You are able to change that though.

The issue is being tracked on https://github.com/alliedmodders/sourcemod/issues/1802

PR #118, opened 2014-07-31
Change checkout-deps script to do a user install of AMBuild

psychonic:
Installing a development dependency system-wide seems silly.


KyleSanderson:
I'm all for build.py as well fwiw. The change should make the entrance lower for users and have a lesser footprint on their system.

KyleSanderson:
This landed in PR 311.

PR #129, opened 2014-08-09
Enable SDK independence with MM:S enabled extensions.

KyleSanderson:
I had to do this for SteamWorks, Drifter hit it with enabling SDK independence with DHooks2.


KyleSanderson:
@dvander is including ISmmPluginExt.h, or including the needed enum if we're core-legacy kosher?

PR #160, opened 2014-09-24
Add OnLogMessage and OnLogError

DoctorMcKay:
<VoiDeD> you know you can do the PR yourself right
    <McKay> but then I'll get yelled at for your bad code
    <VoiDeD> or you'll get a push started for a better/different implementation

### DIFF
--- a/core/logic/Logger.cpp
+++ b/core/logic/Logger.cpp
@@ -550,6 +550,15 @@ bool Logger::LogToErrorForward(int handle, int identity, const char *msg)
 	return result >= Pl_Handled;
 }
 
+bool Logger::SetInErrorForward(bool inForward)
+{
+	bool oldValue = m_InErrorForward;
+
+	m_InErrorForward = inForward;
+
+	return oldValue;
+}
+
 void Logger::EnableLogging()
 {
 	if (m_Active)
--- a/core/logic/Logger.cpp
+++ b/core/logic/Logger.cpp
@@ -95,7 +95,8 @@ void Logger::OnSourceModStartup(bool late)
 
 void Logger::OnSourceModAllInitialized()
 {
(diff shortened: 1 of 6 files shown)


KyleSanderson:
@DoctorMcKay is this still desired?

asherkin:
I'm not a fan of the ability to suppress the logging (or the SetInForward hack required to support it).

What's the use case for these forwards?

DoctorMcKay:
Yes, this is still desired. My use case is logging stuff to a central database for easy review. 

Ability to suppress provides parity with `OnLogAction`, although it's not terribly important to me.

peace-maker:
I'd rather have people fix bugs than suppressing the error logs!

DoctorMcKay:
Ability to suppress error logs is probably not the best idea, but I don't see any big problems with suppressing regular logs.

dvander:
Yeah I'm confused too. This seems like something asking for trouble. Logging is designed to *not* be suppressed. If users don't want it they can turn it off.

VoiDeD:
The goal (at least for the implementation I needed) was to be able to suppress logging for two reasons: to entirely pipe it somewhere else (mysql, etc), and to add some rate limiting to avoid trashing the disk when plugins spew errors in places such as `OnGameFrame` and friends.

dvander:
Seems like that could just be Core functionality (and SM probs shouldn't be doing main-thread I/O either).

KyleSanderson:
@DoctorMcKay are you able to drive this PR?

KyleSanderson:
The relay portion of this is an excellent idea, but we can't add an API to hide plugin errors from first party logging.

<VoiDeD> fork it -> new branch off logmessage -> make ur commits -> new PR
<McKay> that's about 3 steps beyond my caring point

PR #165, opened 2014-10-09
Fix MySQL blocking forever on read

Thordin:
MySQL will block forever on read even if the connection died. Setting the connect timeout does not help.

If you use threaded queries and you try to unload a plugin in this state, the server will freeze.

You may test this by blocking mysql traffic and then typing sm_reloadadmin and waiting for a minute or two.

sudo iptables -A INPUT -p tcp --sport 3306 -j DROP

You can then remove the block. If you attempt to type sm_reloadadmins or some other DB using plugin it will never reconnect. With this patch, it seems to reconnect in a few seconds.

I am not sure if a write timeout is necessary because it should only take a few seconds to detect a broken socket if it is being written to because it will expect an ACK very soon.


KyleSanderson:
@Thordin do you need any help driving this to completion?

Thordin:
We don't need any help, as far as I am concerned we have already provided the only 2 feasible solutions, one of the being adjustable settings that Asherkin requested.

Since it does not look like either one is  going to get accepted any time soon, we hard coded the timeout into the mysql.ext.so module, and copied it over every time we update sourcemod.

We also took advantage of this process to fix charsets. You can permanently set the charset of a connection, but it will block. If you try to use SET NAMES, it will reset to latin-1 every time there is a reconnect, and you will never know when that happens because it happens behind the scenes.

dvander:
What Kyle meant was, in terms of getting this PR checked into master, we can't take a hardcoded timeout. Were you intending to fix that, or should the PR be closed?

Thordin:
I already provided a version with adjustable timeouts.

dvander:
Am I looking in the wrong place? I don't see that in the commit log.

dvander:
Oh - I missed the separate link. My last comment of preferring a single unified timeout is still what I'd like to see. But I don't have a strong preferences.

KyleSanderson:
This PR has been unmaintained since October 9th 2014. Please re-open, or create a new request if you'd like to contribute code to the project.

Thordin:
I don't understand why I would need to maintain it. The code hasn't been made obsolete, and your pull request #248 is quite similar to the solution I offered.

If I knew that's all that was required to be accepted, I would have done that, but from this discussion it wasn't clear how many iterations it would take to satisfy everyone.

While it would have been nice to get the credit, at this point I've already spent way more time than I wanted to on this bug.

dvander:
I apologize we weren't more clear on what we wanted to see. Communication on open-source stuff is tricky since we can't turn around and yell out reminders or questions to each other.

Thank you for bringing the issue to us, regardless.

PR #167, opened 2014-10-21
Added Vector2D Length for Vectors

ghost:


asherkin:
As I explained in the previous PR for this, I don't think it has a place in the SM API (especially not with how trivial these are to calculate manually). This is also unusable as-is as it doesn't specify which elements are used for the calculation.

PR #195, opened 2014-11-09
Fix extension makefile

Thordin:
Added TF2 as an engine to the extension Makefile.


KyleSanderson:
@Thordin are you okay with making AMBuildScript files for the Sample extensions?

Thordin:
I don't know how to do that.

powerlord:
@KyleSanderson You might have better luck asking @Drifter321.

Having said that, how much of the logic can be reused from the AMBuildScript that ships with SourceMod?

Thordin:
I have some basic knowledge of Makefiles but I have no idea how AMbuild works. I glanced at the wiki and it would probably take me a significant amount of time to make one from scratch through trial and error.

If you are asking if I am fine with AMBuild as the cross-platform solution then yes. But if you are asking me to create one, you should probably ask someone else that is familiar with AMBuild.

powerlord:
I took matters into my own hands and generated an AMBuild2 script for the Sample Extension as pull request 341.

PR #200, opened 2014-11-11
Make TravisCI builds use Clang 3.5 to match our Buildbot builds.

psychonic:


KyleSanderson:
https://github.com/broadinstitute/gamgee/blob/e0d54509f66c50a1f82d4e09368cc8aaa83ee36a/.travis_scripts/clang.sh

It looks like the common thing is to just force it in with --force-yes. It isn't ideal, but these are temporary machines.

KyleSanderson:
Done. Your build exited with 0.

PR #203, opened 2014-11-15
Remove faulty g_FlagSet array in adminsys (bug 6248).

psychonic:
As suggested by @Fyren in [bug 6248](https://bugs.alliedmods.net/show_bug.cgi?id=6248), this removes the g_FlagSet array all together.

### DIFF
--- a/core/logic/AdminCache.cpp
+++ b/core/logic/AdminCache.cpp
@@ -46,7 +46,7 @@
 #define LEVEL_STATE_FLAGS		2
 
 AdminCache g_Admins;
-char g_ReverseFlags[26];
+char g_ReverseFlags[AdminFlags_TOTAL];
 AdminFlag g_FlagLetters[26];
 bool g_FlagCharSet[26];
 
--- a/core/logic/AdminCache.cpp
+++ b/core/logic/AdminCache.cpp
@@ -48,7 +48,7 @@
 AdminCache g_Admins;
 char g_ReverseFlags[26];
 AdminFlag g_FlagLetters[26];
-bool g_FlagSet[26];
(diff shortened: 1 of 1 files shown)


KyleSanderson:
r=me,f=fyren?

:shipit:

Fyren:
I only vaguely remember this, but if it the example code from the original reporter works right with it, I guess this is good.  It exercises some of the related natives.

psychonic:
After there was still a lingering issue with this, as well as an admittedly-questionable behavior change, I did it over with a more surgical fix.

New output from the test plugin on bug 6248.
```
0 a 1
1 b 2
2 c 4
3 d 8
4 e 16
5 f 32
6 g 64
7 h 128
8 i 256
9 j 512
10 k 1024
11 l 2048
12 m 4096
13 n 8192
14 z 16384
15 o 32768
16 p 65536
17 q 131072
18 r 262144
19 s 524288
20 t 1048576
abcdefghijklmnopqrstz 2097151
```

core/logic/AdminCache.cpp
```diff
@@ -72,9 +72,9 @@ class FlagReader : public ITextListener_SMC
 			memcpy(g_FlagLetters, g_DefaultFlags, sizeof(AdminFlag) * 26);
 			for (unsigned int i=0; i<20; i++)
```
asherkin: Should this be AdminFlags_TOTAL?

psychonic:
I don't believe so. Both of those arrays are (intentionally) indexed by the flag char, not the flag.

asherkin:
I see, I should've expanded out the context.

PR #212, opened 2014-12-04
Added GetEntityRenderColor

Bara20:


KyleSanderson:
Not that this is on you, but I was under the impression we were deprecating GetEntData* as it is inherently unsafe with GetEntProp* being the preferred alternative.

asherkin:
Why are the params explicitly tagged with the default tag?

Bara20:
#321

PR #229, opened 2014-12-19
Stop MENU_ACTIONS_ALL from tossing warnings with #pragma newdecls required

powerlord:
Basically, this fixes a warning with the `MENU_ACTIONS_ALL` define triggered by commit 154d846.


KyleSanderson:
Thank you for the fix.

PR #238, opened 2015-01-02
Add stocks to properly use TFCond_TmpDamageBonus

FlaminSarge:
Alternative to #233, using tf2_stocks.inc instead.


KyleSanderson:
@psychonic plate or platter in regards to this and #233 ?

FlaminSarge:
There was a rework of the TF2 condition stuff that psy was working on. It would make #233 much simpler, so the plan was to wait for that. However, he lost the stash.

PR #240, opened 2015-01-05
Allow capturing non-public functions as values.

dvander:
This patch allows using non-public functions as callbacks. The JIT is pretty heavily designed around the public function table as an entry point, so the trick here was to just repurpose that table: it now contains all functions that were generated, regardless of whether or not they're public. Non-public ones get some name decoration so they won't cause name collisions (for example, with `static`).

The tricky part is that Pawn is fairly gross, as usual, and it's hard to tell what the final public index will be during codegen. To make this work I added a pseudo-opcode that references the function by name. During assembly this gets rewritten to a normal `const.pri`, after we've computed all the public indices.

As a result, function values no longer result in constant expressions (which is crazy in the first place).

### DIFF
--- a/sourcepawn/compiler/sc.h
+++ b/sourcepawn/compiler/sc.h
@@ -58,7 +58,7 @@
 #define sDEF_LITMAX  500    /* initial size of the literal pool, in "cells" */
 #define sDEF_AMXSTACK 4096  /* default stack size for AMX files */
 #define PREPROC_TERM  '\x7f'/* termination character for preprocessor expressions (the "DEL" code) */
-#define sDEF_PREFIX   "sourcemod.inc" /* default prefix filename */
+#define sDEF_PREFIX   "xsourcemod.inc" /* default prefix filename */
 #define sARGS_MAX		32	/* number of arguments a function can have, max */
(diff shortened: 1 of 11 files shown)


asherkin:
What's the situation with this and default parameter values? Currently `public` prohibits their use (as the calling code wouldn't know about or respect them anyway. I'm going to guess "nothing", and we're just losing the differentiator? (Not that this is really a problem, I don't think anyone could sanely expect them to have any effect)

What about returning arrays? spcomp allows this for non-public functions, what would the calling code end up with? Just the memory address?

dvander:
@asherkin This patch doesn't actually make functions public - it just emits a `.publics` table that happens to have every function in it. The other significant change is just in how function names are evaluated. Before, they were a constant. Now, they are evaluated at run-time. (The runtime code still just pushes a constant, at least until we get lambdas.)

dvander:
Yes, it does technically expose functions that cannot be safely invoked via `IPluginFunction`. That's a good observation. I guess I should prevent those from being added to typesets or typedefs, since there's no way to invoke them anyway.

dvander:
New version will error if evaluating a function name as a value, if the function can return an array.

PR #241, opened 2015-01-06
Fix un-joined threads leaking. (bug 3460)

asherkin:
We currently leak threads if Join() is never called (such as IThreader threads using Thread_AutoRelease or the simple MakeThread overload). Also, calling Join() more than once could cause a crash / the wrong thread to be joined. This changes the POSIX thread implementation to match the behaviour of the Windows one and let the OS clean up resources if the Thread instance is released.

@dvander


asherkin:
Whoops, PRing the upstream repo instead.

PR #242, opened 2015-01-14
Use BfRead tag in umsg hook typedef, rather than a generic Handle.

VoiDeD:


dvander:
Looks good - we should take this for 1.7.

psychonic:
Doesn't this break usage on CS:GO which does have BfReads?

Maybe a typeset is needed with a separate prototype.

dvander:
Whoops yeah. Thanks I'll revert that.

PR #243, opened 2015-01-15
Use BfRead or Protobuf tags in umsg hook callbacks, rather than a generic Handle.

VoiDeD:
Alright, take 2. It's surprisingly easy to forget that CS:GO exists.

Two separate commits to make cherrypicking to 1.7 easier.


dvander:
Looks good, thanks!

PR #247, opened 2015-01-20
Improvements for NormalSHook functionality

AnthonyIacono:
This is my first pull request in a while, hopefully I did it correctly.

Please see the following for notes: https://bugs.alliedmods.net/show_bug.cgi?id=6297

### DIFF
--- a/extensions/sdktools/vsound.cpp
+++ b/extensions/sdktools/vsound.cpp
@@ -1130,6 +1130,147 @@ static cell_t EmitSound(IPluginContext *pContext, const cell_t *params)
 	return 1;
 }
 
+static cell_t EmitSoundEntry(IPluginContext *pContext, const cell_t *params)
+{
+#if SOURCE_ENGINE < SE_PORTAL2
+	return pContext->ThrowNativeError("EmitSoundEntry is not available in this game.");
+#else
+	cell_t *addr, *cl_array;
+	CellRecipientFilter crf;
+	unsigned int numClients;
+	int client;
+	IGamePlayer *pPlayer = NULL;
+
+	pContext->LocalToPhysAddr(params[1], &cl_array);
+	numClients = params[2];
+
+	/* Client validation */
+	for (unsigned int i = 0; i < numClients; i++)
+	{
+		client = cl_array[i];
+		pPlayer = playerhelpers->GetGamePlayer(client);
+
+		if (!pPlayer)
+		{
+			return pContext->ThrowNativeError("Client index %d is invalid", client);
+		}
+		else if (!pPlayer->IsInGame()) {
+			return pContext->ThrowNativeError("Client %d is not connected", client);
+		}
+	}
+
+	crf.Initialize(cl_array, numClients);
+
+	char *soundEntry;
+	pContext->LocalToString(params[3], &soundEntry);
+
+	char *sample;
+	pContext->LocalToString(params[4], &sample);
+
+	// By default, we want the hash to equal maxint
+	unsigned int soundEntryHash = -1;
+
+	// We only generate a hash if the sample is not the same as the sound entry and the sound entry is not empty.
+	if (strcmp(soundEntry, sample) != 0 && strcmp(soundEntry, "") != 0)
+		soundEntryHash = GenerateSoundEntryHash(soundEntry);
+
+	int entity = SoundReferenceToIndex(params[5]);
+	int channel = params[6];
+	int level = params[7];
+	int seed = params[8];
+	int flags = params[9];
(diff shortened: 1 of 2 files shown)


AnthonyIacono:
Woops, forgot something. I'll open a new pull request shortly.

extensions/sdktools/vsound.cpp
```diff
@@ -306,6 +357,15 @@ void SoundHooks::OnEmitSound(IRecipientFilter &filter, int iEntIndex, int iChann
 	char buffer[PLATFORM_MAX_PATH];
 	strcpy(buffer, pSample);
 
+	char soundEntry[PLATFORM_MAX_PATH];
+#if SOURCE_ENGINE >= SE_PORTAL2
+	strcpy(soundEntry, pSoundEntry);
```
psychonic: Need something safer than strcpy here and in OnEmitSound2.

extensions/sdktools/vsound.cpp
```diff
@@ -306,6 +357,15 @@ void SoundHooks::OnEmitSound(IRecipientFilter &filter, int iEntIndex, int iChann
 	char buffer[PLATFORM_MAX_PATH];
 	strcpy(buffer, pSample);
 
+	char soundEntry[PLATFORM_MAX_PATH];
```
psychonic: For < Portal 2, should init this to "", at least null-ing the first char before sending it to sp.

extensions/sdktools/vsound.cpp
```diff
@@ -321,6 +381,8 @@ void SoundHooks::OnEmitSound(IRecipientFilter &filter, int iEntIndex, int iChann
 		pFunc->PushCellByRef(reinterpret_cast<cell_t *>(&iSoundlevel));
 		pFunc->PushCellByRef(&iPitch);
 		pFunc->PushCellByRef(&iFlags);
+		pFunc->PushStringEx(soundEntry, sizeof(soundEntry), SM_PARAM_STRING_COPY, SM_PARAM_COPYBACK);
+		pFunc->PushCellByRef(&nSeed);
```
spumer: As i can see this variable not declareted for SOURCE_ENGINE >= SE_PORTAL2, but passed in both cases

extensions/sdktools/vsound.cpp
```diff
@@ -436,6 +517,8 @@ void SoundHooks::OnEmitSound2(IRecipientFilter &filter, int iEntIndex, int iChan
 		pFunc->PushCellByRef(&sndlevel);
 		pFunc->PushCellByRef(&iPitch);
 		pFunc->PushCellByRef(&iFlags);
+		pFunc->PushStringEx(soundEntry, sizeof(soundEntry), SM_PARAM_STRING_COPY, SM_PARAM_COPYBACK);
+		pFunc->PushCellByRef(&nSeed);
```
spumer: Same way for `nSeed` var. Not declared for all cases.

extensions/sdktools/vsound.cpp
```diff
@@ -1037,6 +1130,147 @@ static cell_t EmitSound(IPluginContext *pContext, const cell_t *params)
...
+	{
+		client = cl_array[i];
+		pPlayer = playerhelpers->GetGamePlayer(client);
+
+		if (!pPlayer)
+		{
+			return pContext->ThrowNativeError("Client index %d is invalid", client);
+		}
+		else if (!pPlayer->IsInGame()) {
+			return pContext->ThrowNativeError("Client %d is not connected", client);
```
spumer: `not connected` -> `not in game` ?

extensions/sdktools/vsound.cpp
psychonic: It is one of the OnEmitSound function variables in >= SE_PORTAL2. It only needs to be defined in others since it does not already exist in those.

extensions/sdktools/vsound.cpp
psychonic: I believe that this was just copied from EmitSound and EmitSentence. We also only check isConnected there too. I do agree however that all of these should probably check InGame instead. Feel free to submit a Pull Request for that.

50Wliu:
So sorry to comment so late on this pull request but I just noticed that this doesn't seem to be backwards-compatible with previous versions of SM.

Using the existing typedef for <=1.7 I get a `error 100: function prototypes do not match` when trying to compile on 1.8.  Can the two new args be made optional?

asherkin:
You'll need to update the plugin or use the include files from 1.7. Compiled binaries are backwards-compatible.

PR #248, opened 2015-01-21
Establish a default timeout for MySQL connectivity.

KyleSanderson:
Users presently experience dead locks from how we lock our DB structures, and how MySQL establishes a connection to a foreign server. If there's a blip of packet loss, the server is almost always toast on level change.

This bundles ideas together from https://github.com/alliedmodders/sourcemod/pull/165 into something we can hopefully merge.

### DIFF
--- a/extensions/mysql/mysql/MyDriver.cpp
+++ b/extensions/mysql/mysql/MyDriver.cpp
@@ -96,10 +96,15 @@ MYSQL *Connect(const DatabaseInfo *info, char *error, size_t maxlength)
 	MYSQL *mysql = mysql_init(NULL);
 	const char *host = NULL, *socket = NULL;
 
+	decltype(info->maxTimeout) timeout = 60;
 	if (info->maxTimeout > 0)
 	{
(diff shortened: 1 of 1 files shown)


asherkin:
The config should keep a (commented out) timeout of 0, we document 0 as being "use the default".

I think the default should be more like 60 as well, 15 is likely too low for a reasonable percentage of users, any timeout at all provides a safety net.

KyleSanderson:
The 15s limit was based around cl_timeout being 30s with the idea being we don't want to drop players at level change if something horrendous happens when the map comes to a conclusion. When the round ends is usually when people dump a ton of data (old SM stats plugins). Alternatively if someone does something really bad with table locking, which is hopefully now deprecated in script but not in compiled plugins. If there are server owners who hit this timeout, they should investigate their database.

I wouldn't be greatly opposed to 60s, but the bite is losing players, not that our query didn't complete. 15s to die, then to load the map on rotational media seems reasonable.

KyleSanderson:
Removed the databases.cfg change, the timeout is now 60s.

PR #253, opened 2015-01-28
Convert adminmenu to use newdecls and new classes/methodmaps.

psychonic:
To keep myself familiar with the new syntax and classes, as well as get more of the incs converted to newdecls, I ported the adminmenu plugin to the transitional syntax, including using ArrayList and DataPack methodmaps.

It now can compile with #pragma newdecls required (although I did not add the pragma). Despite this, the hacky enum structs were left with the old colon syntax (not affecting the pragma) as converting them seemed to generate errors.

### DIFF
--- a/plugins/adminmenu.sp
+++ b/plugins/adminmenu.sp
@@ -36,7 +36,7 @@
 #include <sourcemod>
 #include <topmenus>
 
-public Plugin:myinfo = 
+public Plugin myinfo = 
 {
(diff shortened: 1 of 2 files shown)


psychonic:
Note that this will fail on Travis until #251 is landed due to opting into the more specific prototype in the topmenu handler.

dvander:
sweet, LGTM :+1:

PR #254, opened 2015-01-28
Fix incorrect error line show for incorrect return value on forwards (bug 6226).

psychonic:
As described in [bug 6226](https://bugs.alliedmods.net/show_bug.cgi?id=6226).

When a function implementing a forward has a return type that doesn't match, the line in the error message is the line of the forward declaration, rather than the implementing function. This is a regression in 1.7+.

This seems to fix it.


asherkin:
Been saving this one for a special patch, touching spcomp qualifies.

PR #257, opened 2015-01-31
Fix "sm plugins refresh" not refreshing changed plugins.

psychonic:
RefreshAll is normally only called in LevelShutdown to unload any plugins loaded for that map only (obsolete) or any that have been change since load. LevelInit then calls DoGlobalPluginLoads.

By adding the RefreshAll call in "sm plugins refresh", we replicate this. Tested and it seems to work as documented now.


dvander:
:ship:

PR #258, opened 2015-01-31
GetWeaponPrice fix for CS:GO

Drifter321:
This fixes GetWeaponPrice for CS:GO after valves latest update. For windows there is no way to create a unique signature so the function address is grabbed from another function that calls it. In doing this I had to change CDetour to be able to pass a address instead of a gamedata string. I added some safety measures and also added a method for GDC to check that the offset within the function is correct.


psychonic:
This is a bit gross, but given the very limited number of ways to get at this function in the CS:GO Windows bin, it will have to do.

If it works, :ship:

PR #259, opened 2015-02-03
Remove sizeof() as a special-case default argument value.

dvander:
This craptacular feature has been a minor thorn in my side. I don't like it, it doesn't reflect where we want the API to be, and it's not usable in situations where the array size is indeterminate or controlled by a dynamic value. It also computes its value wrong for character arrays.

From a language design perspective, it's very odd that the `sizeof` in this context is magic and binds to the callsite rather than the in-scope argument. 

As far as I know, it's rare for SourceMod code to use this, so let's try removing it. Other special cases removed as well are `tagof` and `countof` - the former being particularly useless, and we can get rid of a bunch of code as a result.

### DIFF
--- a/sourcepawn/compiler/sc.h
+++ b/sourcepawn/compiler/sc.h
@@ -80,10 +80,6 @@ typedef struct s_arginfo {  /* function argument info */
   union {
     cell val;           /* default value */
     struct {
-      char *symname;    /* name of another symbol */
-      short level;      /* indirection level for that symbol */
-    } size;             /* used for "sizeof" default value */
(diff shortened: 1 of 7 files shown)


50Wliu:
So if this goes through is the expected way to update hardcoding the array lengths?

asherkin:
No, to call sizeof at the callsite if appropriate, else passing in an already computed size (in the case of a dynamic array), exactly the same as you currently would for calling all natives.

50Wliu:
Oh, ok.  Thanks.

PR #261, opened 2015-02-07
Make handle dump location relative to gamedir to match other sm_dump commands.

psychonic:
This.


KyleSanderson:
:shipit: 

It may be worth it in another pull request to unify the 'Prop' dumps as well as I believe those are mismatched as well?

psychonic:
At least sm_dump_netprops and sm_dump_datamap dump to the game directory.

PR #262, opened 2015-02-16
Fix tag mismatch warning when using SQLite_UseDatabase. (bug 6310)

VoiDeD:


dvander:
Thanks for fixing this!

Sarabveer:
Thanks!

PR #264, opened 2015-02-18
Updated Visual Studio sample_ext files

davidskuza:
Project files in sample_ext for Visual Studio pointed to not existing files.


psychonic:
lgtm, but there are other issues with these sample files, such as missing SE_ defines, missing engines, and still pointing to MMSOURCE19.

I'd love to see them eventually replaced with a sample AMBuild script (which can also generate vcproj files in most cases), rather than having to keep all of these configurations up to date separately.

PR #265, opened 2015-02-18
Add missing impl for File.ReadUint16.

psychonic:
This exists in the File methodmap, but there is no implementation for it.


dvander:
rude. :ship: :camel:

PR #266, opened 2015-02-18
Replace frameloop for player condition checks in TF2 ext with sendproxies

psychonic:
For many TF2-specfiic plugins, it's useful to know when a "player condition" is added or removed. We don't/can't hook the functions for add/removing them directly due to them getting inlined in places. We, instead, currently have a loop over all player condition variables (5 of them any counting) for each player, every frame.

Instead, we can insert ourselves as a sendproxy for these sendprops, letting use know when the variable has changed (as the proxies are called when it serializes the data to send to the client).

This is still pending some final testing, but it's ready to be looked at.

@VoiDeD @asherkin :japanese_goblin:

### DIFF
--- a/extensions/tf2/conditions.cpp
+++ b/extensions/tf2/conditions.cpp
@@ -2,7 +2,7 @@
  * vim: set ts=4 :
  * =============================================================================
  * SourceMod Team Fortress 2 Extension
- * Copyright (C) 2004-2011 AlliedModders LLC.  All rights reserved.
+ * Copyright (C) 2004-2015 AlliedModders LLC.  All rights reserved.
  * =============================================================================
  *
  * This program is free software; you can redistribute it and/or modify it under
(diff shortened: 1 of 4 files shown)


KyleSanderson:
I'm unhappy with HandleCondChange (should be static) and PlayerConditionsMgr::ProcessCondChange(CondChangeData_t *pCondData). It's not clear who's responsible for freeing memory.

684b59a92179e4b1556116e62a60acb0c7be38ed introduced a memory leak if the client isn't ingame. ke::AutoPtr should be used in either the caller or callee, somewhere.

psychonic:
> I'm unhappy with HandleCondChange (should be static)
Why should it be static? It uses member vars that are and should be tied to the class rather than floating loose.

> PlayerConditionsMgr::ProcessCondChange(CondChangeData_t *pCondData). It's not clear who's responsible for freeing memory.
> 684b59a introduced a memory leak if the client isn't ingame. ke::AutoPtr should be used in either the caller or callee, somewhere.
The "caller" is SM, since it's a frame callback. The only place it really can be freed it in that function. That oversight in 684b59a sucks but is very easily corrected.

KyleSanderson:
I am happy now.

KyleSanderson:
:shipit:

PR #267, opened 2015-02-21
Add TF2_SetClientTeam to provide symmetry to TF2_GetClientTeam

50Wliu:
Continues #201, which added `TF2_GetClientTeam`.  This PR adds `TF2_SetClientTeam` in order to provide full symmetry.  I was thinking about doing `TF2_ChangeClientTeam`, since it wraps around `ChangeClientTeam`, but I decided against it because of `TF2_{Get|Set}PlayerClass`.

Was also thinking about updating `tf2_stocks.inc` to new-style syntax but I have no idea how to do the C++ side of methodmaps.

And...removes some trailing whitespace.

### DIFF
--- a/plugins/include/tf2_stocks.inc
+++ b/plugins/include/tf2_stocks.inc
@@ -331,14 +331,14 @@ stock TFTeam:TF2_GetClientTeam(client)
 }
 
 /**
- * Sets a client's current team.
+ * Changes a client's current team.
  *
  * @param client        Client index.
  * @param team          TFTeam team symbol.
  * @noreturn
  * @error               Invalid client index.
  */
-stock TF2_SetClientTeam(client, TFTeam:team)
+stock TF2_ChangeClientTeam(client, TFTeam:team)
 {
     ChangeClientTeam(client, _:team);
 }
--- a/plugins/include/tf2_stocks.inc
+++ b/plugins/include/tf2_stocks.inc
@@ -9,7 +9,7 @@
  * This program is free software; you can redistribute it and/or modify it under
  * the terms of the GNU General Public License, version 3.0, as published by the
  * Free Software Foundation.
- * 
+ *
  * This program is distributed in the hope that it will be useful, but WITHOUT
  * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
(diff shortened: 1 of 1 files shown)


KyleSanderson:
> And...removes some trailing whitespace.

Definitely include that in a different commit then the one where you've added something.

50Wliu:
Thanks, will definitely remember that for next time.

psychonic:
I don't know if others agree, but I'd prefer it as TF2_ChangeClientTeam, to keep the name similar to the current API. "Set" can also imply that it's just setting a value, rather than taking action (and changing a clients team does do more behind the scenes than flipping a var).

50Wliu:
I've posted a poll in the Scripting subforum-I guess we'll look and see after a while.

50Wliu:
So it's pretty obvious which one's going to win at this point.  Changed.

In addition, Powerlord pointed out that there's a new `CTFPlayer::ChangeTeam( int iTeamNum, bool bAutoTeam, bool bSilent );`, so I guess if that could be implemented in Sourcemod, that would be great.

psychonic:
> In addition, Powerlord pointed out that there's a new CTFPlayer::ChangeTeam( int iTeamNum, bool bAutoTeam, bool bSilent );

If by new, you mean eight years old :wink:. It's existed since the Orangebox engine debuted.

The beauty of the one we're using now is that it is on a fairly stable interface. The other one would require gamedata as it would occasionally break. Additionally, automatically selecting a team and blocking the player_team event can be done manually without much hassle.

50Wliu:
New™!
I've been working fine with ChangeClientTeam and AddCommandListenering jointeam, so I guess it's ok then ;P.

Also, the Travis CI build seems to have timed out :cry:.

psychonic:
> Also, the Travis CI build seems to have timed out :cry:.

I have convinced it to reconsider...

50Wliu:
Is there any chance this can be merged into the 1.7-dev branch, or will this have to wait until 1.8?

KyleSanderson:
Do you have a more compressed changeset with just the changes you need to support this feature?

asherkin:
It'll probably get picked up in @psychonic's next cherry pick run, generally stuff gets left to bake in master for a few days.

PR #269, opened 2015-02-22
Add Wno-error=sign-compare for gcc to Compiler.

KyleSanderson:
Obviously this isn't ideal, but this should allow newer versions of g++ to successfully compile the SourceMod compiler once again.


asherkin:
If the warning is superfluous, the warning should be suppressed (and the compiler version it was added needs to be checked first), if it isn't, the code should be fixed. Just suppressing the error is not a good way forward.

KyleSanderson:
The warning is superfluous to clang. @psychonic has a more complete (read as: appropriate) cset in progress.

psychonic:
I don't have anything in progress to fix the warnings.

PR #271, opened 2015-02-24
Rename some horrible classes.

dvander:
`JitFunction` is a confusing term but it's really just a Function. They're created lazily though, so I'm renaming it to `CompiledFunction`. `CFunction` is an implementation of the bizarre `IPluginFunction` API. It has nothing to do with functions, it's more a wrapper around a (Context, Function) pair with some local VM state to make sure it's completely unsafe. I'm renaming this to `ScriptedInvoker`. Not the best name but it better describes what it does.


psychonic:
:goberserk:

PR #274, opened 2015-02-24
Refactor the public API entrypoint for SourcePawn.

dvander:
It is very weird that SourcePawn's library has two APIs, both with separate entry points, both with separate global shared state. Part 1: Move the shared state into a new Environment class, which provides access to both API sets. I changed the DLL entrypoint as well.

These changes (should) change no behavior.


dvander:
Second commit merges engine2 and sp_vm_engine into api.h/cpp. This split was very confusing.

asherkin:
Any plans to "completely" kill off API1 and API2?

dvander:
Yeah I would love to kill them both, but I was worried about breaking extensions.

asherkin:
I think we can get away with it if we provide a wrapper via SM, at least extensions aren't gonna be accessing them directly. CDetour / BinTools probably shouldn't be using the SourcePawn memory allocator anyway (which is the only reason they need ISourcePawnEngine).

PR #275, opened 2015-02-24
De-singleton-ify the watchdog timer.

dvander:
Remove the watchdog singleton and tie it to an Environment instead. There should probably be only one watchdog timer in theory (no matter the number of environments), but right now the environment count is fixed to 1.


dvander:
Merged via command-line.

PR #276, opened 2015-02-24
Move the code allocator into SourcePawn, and remove knight/shared.

dvander:
This refactoring has three changes:
 - The code allocator gets moved into sourcepawn/jit where it belongs. I didn't restyle it. I don't want to know what it looks like or what it's doing.
 - The code allocator is now part of Environment, rather than a global singleton.
 - `knight/shared` is removed, since that's like 1,550 lines of code that are totally unused and superceded by AMTL.


asherkin:
The `Knight` namespace / `Ke` prefix still existing is a little odd, but at least its internal only.

dvander:
Merged via command-line.

PR #277, opened 2015-02-24
Remove the JITX86 class.

dvander:
The JITx86 class is basically platform independent, and should be merged into Environment. First patch moves watchdog/runtime interaction into Environment.


dvander:
Patches 2 & 3: Some BaseContext stuff has been moved out of JITX86, and the ICompilation infrastructure has been removed. This is technically a breaking API change, since `StartCompilation` now always returns null. Core is the only caller so I just fixed it.

dvander:
Patch 4: Factor out all the stub stuff out of JITX86 into a new `CodeStubs` class.

dvander:
Merged via command-line.

PR #278, opened 2015-02-24
Tf2 change team

WildCard65:
Adds a call to CTFPlayer::ChangeTeam as it has 2 bools. 1 of them I believe silents the team change.  
Tell me how I did with the C++ backend of native?

@powerlord Mind telling me what bAutoTeam and bSilent do?

### DIFF
--- a/extensions/tf2/natives.cpp
+++ b/extensions/tf2/natives.cpp
@@ -648,6 +648,59 @@ cell_t TF2_RemoveWearable(IPluginContext *pContext, const cell_t *params)
 	return 1;
 }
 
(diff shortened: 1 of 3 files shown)


psychonic:
This would be better in SDKTools than the TF2 extension as that function exists in all supported games other than EP1 mods, The Ship, and Dark Messiah.

But that said, I don't think that it's worth the extra maintenance code of having it per my comment on #267

> The beauty of the one we're using now is that it is on a fairly stable interface. The other one would require gamedata as it would occasionally break. Additionally, automatically selecting a team and blocking the player_team event can be done manually without much hassle.

PR #280, opened 2015-02-25
Remove sp_context_t.

dvander:
This is a holdover from SourceMod 1.0, when the VM was in core and the JIT was a closed-source implementation. sp_context_t allowed the two to share vital VM state. These days, it's confusing and irrelevant.

Once again this is largely a syntactic refactoring. Anything that accessed sp_context_t now accesses PluginContext instead. Many functions that had to take in some horrible combination of pointers now is a helper function directly on PluginContext.


Fyren:
r+

dvander:
Thanks! m_pContext goes away in a later patch and so does the interpreter, praise be.

psychonic:
hallelujah

PR #281, opened 2015-02-25
Don't expose mutable sp_native_t.

dvander:
The `GetNativeByIndex` API call is problematic since it provides a mutable pointer into the VM state. If - for example - a refactoring eliminated the existence of `sp_native_t`, we would not be able to provide a read-only view.

This patch deprecates and nops out the original API call, replacing it with two natives. `GetNative` returns the immutable state and `UpdateNativeBinding` pushes a state change.


dvander:
Merged via command-line.

PR #282, opened 2015-02-25
Simplify the JIT function cache.

dvander:
The function cache is stupid. It wastes about 1000X more memory than is needed. Hashtable lookup should be fine, moving the public-id cache to `ScriptedInvoker`.


dvander:
Merged via command-line.

PR #283, opened 2015-02-25
Remove the interpreter.

dvander:
The interpreter was mildly useful for verifying the JIT rewrite, but it's buggy and hard to maintain since it's not used. Might as well delete it.


Fyren:
r+

dvander:
Merged via command-line.

PR #284, opened 2015-02-26
Rewrite the .smx parser.

dvander:
This removes one the last remnants of the SourceMod 1.0 VM implementation.

The new parser introduces a number of design changes in the VM. First, the VM now takes greater responsibility for validating and sanity checking the structure of the SMX container format. Previously, malformed SMX files could easily crash SourcePawn. The loader now rejects files that have out-of-bounds offsets or incomplete sections. Complex sections, like debug info or the code stream, are verified lazily.

Internally, the sp_plugin_t structre has been removed. It has been replaced by a new LegacyImage class, designed to be independent from the SPVM API. This potentially lets us load code streams from non-.smx containers. More importantly, it removes a lot of bookkeeping and pre-computed state from PluginRuntime. The LegacyImage class is now responsible for handling debug info as well.

PluginRuntime is now intended to hold only cached or immutable data, and PluginContext holds all VM state. As such PluginContext is now responsible for allocating a plugin's runtime memory, not PluginRuntime.

Finally, some aspects of the loading process have been cleaned up. The decompression and image handoff logic should now be easier to understand.


sourcepawn/jit/api.cpp
```diff
@@ -176,124 +177,93 @@ SourcePawnEngine2::SourcePawnEngine2()
...
-
-  if (!fp) {
-    error = SP_ERROR_NOT_FOUND;
-    goto return_error;
+  IPluginRuntime *rt = LoadBinaryFromFile(file, nullptr, 0);
+  if (!rt) {
+    if (err) {
+      if (FILE *fp = fopen(file, "rb")) {
+        fclose(fp);
+        *err = SP_ERROR_NOT_FOUND;
```
KyleSanderson: Inversed with L212?

sourcepawn/jit/smx-v1-image.cpp
```diff
@@ -0,0 +1,617 @@
...
+       iter < buffer() + hdr_->dataoffs;
+       iter++)
+  {
+    if (*iter == '\0') {
+      found_terminator = true;
+      break;
+    }
+  }
+  if (!found_terminator)
+    return error("malformated section name");
```
Fyren: MALFORMED
MISFORMATTED

sourcepawn/jit/api.cpp
dvander: Indeed, thanks!

dvander:
Merged via command-line w/ review comments addressed.

PR #286, opened 2015-03-01
Fix crash regression in games that don't support radio style menus

peace-maker:
Fix regression in ad7d920ce0f57cfec0a35bc942d2a6a8583498e7
GetMenuStyleHandle(MenuStyle_Radio) crashes games, which don't support
the radio menu style. The style is never added to the menu manager, if
it's not supported, so GetMenuStyleHandle tries to call IsSupported on a
nullptr.

### DIFF
--- a/core/MenuStyle_Radio.cpp
+++ b/core/MenuStyle_Radio.cpp
@@ -83,6 +83,10 @@ void CRadioStyle::OnSourceModLevelChange(const char *mapName)
 	}
 
 	g_bRadioInit = true;
+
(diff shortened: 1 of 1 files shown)


psychonic:
Is this broken in master as well? It looks like the PR is directly against 1.7-dev.

peace-maker:
Oh, yes, it is.

psychonic:
Seems like an okay fix for this. It doesn't help that the menu system is... :goat:

PR #287, opened 2015-03-03
Add TE_SendToAllInRange

peace-maker:
Add helper stock to mimic PVS and PAS recipient filters for tempents in
the SDK.


psychonic:
Thanks!

PR #290, opened 2015-03-05
Documentation oddities for Call_PushStringEx

InstantMuffin:
See changes for details, should be obvious. What also strikes me as odd is that there is only one cpflag atm. Has this always been the case? Will there be more in the future? Or why is this designed to be a flagstring?


dvander:
Yup, there's only one flag.

PR #291, opened 2015-03-06
Improve the diagnostic given when a function prototype doesn't match an existing definition

VoiDeD:
Currently, as forwards are being slowly ported to the new transitional syntax, there is a situation where new transitional code is being written against non-ported forwards, ex:

```c
forward OnPluginStart();

// in order for this newdecl function to compile
// it must be declared as returning int because the old syntax always implicitly returned cells
public int OnPluginStart()
{
}
```
Now later in the future when the forward is ported (and the return type changes to `void`), the following occurs:

```c
forward void OnPluginStart();

// error 025: function heading differs from prototype
public int OnPluginStart()
{
}
```

This PR improves the diagnostic message to be: `error 180: function return type differs from prototype. expected 'void', but got 'int'`.

Additionally, it improves the error message given when the argument of a function doesn't match the definition of a forward:

```c

forward void MyFunction(const char[] someString)

// error 181: function argument 'theFirstArgument' differs from prototype.
public void MyFunction(char[] theFirstArgument)
{
}
```

Lastly, there are a few cases where the old error 25 has been retained:
  * When the return tags don't match for a symbol that has `uPROTOTYPED` usage, which I'm unfamiliar with.
  * Functions argument mismatches that involve var args.

### DIFF
--- a/sourcepawn/compiler/sc.h
+++ b/sourcepawn/compiler/sc.h
@@ -981,7 +981,7 @@ typedef struct array_info_s
 } array_info_t;
 
 enum FatalError {
-  FIRST_FATAL_ERROR = 180,
+  FIRST_FATAL_ERROR = 182,
 
   FATAL_ERROR_READ  = FIRST_FATAL_ERROR,
   FATAL_ERROR_WRITE,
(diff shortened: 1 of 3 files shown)


dvander:
Looks good, thanks for doing this!

PR #292, opened 2015-03-06
Signed comparison warning fixes.

Fyren:
My testing involved starting a server and diffing the assembly for mapchooser!

### DIFF
--- a/sourcepawn/jit/file-utils.cpp
+++ b/sourcepawn/jit/file-utils.cpp
@@ -38,7 +38,7 @@ FileReader::FileReader(FILE *fp)
     return;
 
   ke::AutoArray<uint8_t> bytes(new uint8_t[size]);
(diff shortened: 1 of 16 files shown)


sourcepawn/jit/file-utils.cpp
```diff
@@ -31,7 +31,7 @@ FileReader::FileReader(FILE *fp)
 {
   if (fseek(fp, 0, SEEK_END) != 0)
     return;
-  long size = ftell(fp);
+  size_t size = ftell(fp);
   if (size < 0)
```
dvander: ftell() does return `long`, so either the failure check has to change or we need a separate variable for the size.

dvander:
Touche. Thanks for doing this. r=me with that one comment addressed.

PR #293, opened 2015-03-06
Reapply ancient patch to allow spaces in spcomp options.

Fyren:
See https://bugs.alliedmods.net/show_bug.cgi?id=3729 and https://bugs.alliedmods.net/show_bug.cgi?id=6307.


dvander:
r=me thanks for tracking this down.

50Wliu:
:smile: Thanks.

PR #294, opened 2015-03-06
Add explicit return types to forwards missing them.

psychonic:
I believe that this is all of the remaining ones.

@dvander


dvander:
:ship:

PR #296, opened 2015-03-07
Fixed regex.inc

WildCard65:
Makes compiler happy with regex.inc now


psychonic:
This isn't really a fix, but thanks for porting the rest of it to transitional syntax.

PR #298, opened 2015-03-08
Update Plugins/Include for new transitional Syntax

TheJelle:
Converted every *.inc file in plugins/include directory, to the new
transitional syntax.

Don't blame me if I missed one or two.
It compiled seamlessly.


WildCard65:
I believe adminmenu.sp uses TopMenu methodmap

PR #301, opened 2015-03-10
Add GetTeamEntity native to SDKTools.

psychonic:


KyleSanderson:
cpp r=me.

The include hasn't been ported, everything around it is still sp1 traditional syntax.

psychonic:
> The include hasn't been ported, everything around it is still sp1 traditional syntax.
Porting the rest of the include is outside of the scope of the PR.

KyleSanderson:
> Porting the rest of the include is outside of the scope of the PR.
Exactly why I'm saying it should match :-)

If not, it looks fine.

PR #302, opened 2015-03-11
More detailed GetClientLanguage(..) api

TheMadSword:
Title + Also trying a first fork request w/ SM

also : ugh; I didn't expect people to see 8 commits :$ lol


Fyren:
You can squash them into a single commit, but would probably have to use an actual git client and not just the GitHub site (if that's what you did in the first place).  You can Google for something like 'github squash pr' for more info.

PR #303, opened 2015-03-12
FIx some TF2 natives not having the correct vstk size.

Drifter321:
The vstk for some of the function calls is missing some of the sizes of some of the params. I think this is a bug as far as I know everything else has the size of all params.


KyleSanderson:
>FIx

r=me

Fyren:
The comments for the signatures of the functions being called should either be updated or removed.

psychonic:
I concur.

Drifter321:
Good catch, I didn't notice the function signature comments.

Drifter321:
Its the new comma

PR #305, opened 2015-03-13
Add File Transfer hooks to SDKTools.

KyleSanderson:
I've ported this from a pretty dated bug on BZ, not even sure if it still builds...

### DIFF
--- a/extensions/sdktools/hooks.cpp
+++ b/extensions/sdktools/hooks.cpp
@@ -37,18 +37,32 @@
 #include "utlvector.h"
 #include <shareddefs.h>
 #include "usercmd.h"
+#include "filesystem.h"
 
 #define FEATURECAP_PLAYERRUNCMD_11PARAMS	"SDKTools PlayerRunCmd 11Params"
 
 CHookManager g_Hooks;
 static bool PRCH_enabled = false;
 static bool PRCH_used = false;
+static bool FILE_used = false;
 
 SH_DECL_MANUALHOOK2_void(PlayerRunCmdHook, 0, 0, 0, CUserCmd *, IMoveHelper *);
+SH_DECL_HOOK2(IBaseFileSystem, FileExists, SH_NOATTRIB, 0, bool, const char*, const char *);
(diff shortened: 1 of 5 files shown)


WildCard65:
Is this change suppose to include changes to the inc file?

psychonic:
It would be great if the hook was disabled when there are no forward listeners. That way, if interfaces change, the plugin(s) using it can just be disabled temporarily, rather than all of SDKTools to avoid crashing until a fix is available.

extensions/sdktools/hooks.cpp
```diff
@@ -45,10 +46,22 @@ static bool PRCH_enabled = false;
 static bool PRCH_used = false;
 
 SH_DECL_MANUALHOOK2_void(PlayerRunCmdHook, 0, 0, 0, CUserCmd *, IMoveHelper *);
+SH_DECL_HOOK2(IBaseFileSystem, FileExists, SH_NOATTRIB, 0, bool, const char*, const char *);
+#if SOURCE_ENGINE >= SE_ALIENSWARM
+SH_DECL_HOOK3(INetChannel, SendFile, SH_NOATTRIB, 0, bool, const char *, unsigned int, bool);
+#else
+SH_DECL_HOOK2(INetChannel, SendFile, SH_NOATTRIB, 0, bool, const char *, unsigned int);
+#endif
```
psychonic: I believe that L4D2 has the bool as well (but not Nuclear Dawn). I'll look into making sure that our SDKs are up to date there.

KyleSanderson:
> It would be great if the hook was disabled when there are no forward listeners.

Extensions race against core and CForward, there's no clean way to do this. When I was looking at the code again, I too thought of the same thing. Adding callbacks to the IForward iface is my intention to reduce the complexity and opt other forwards into this, but that's another PR.

KyleSanderson:
> It would be great if the hook was disabled when there are no forward listeners.

https://github.com/alliedmodders/sourcemod/pull/307

psychonic:
I think that there would be value in putting this into a separate extension (can still be first-party). With the filesystem interface being a little "breaky" on some games, it would be nice to be able to easily disable this functionality by moving the file out without having to disable all of SDKTools.

KyleSanderson:
I've added an include file, hooks are not used until a plugin asks for one.

Drifter321:
:ship:

PR #306, opened 2015-03-13
Specifying more in details what GetClientLanguage does

TheMadSword:
Specifying more in details what GetClientLanguage does


TheMadSword:
This makes me realize... Should I PR to master or to 1.7-dev ?

50Wliu:
Always master I believe.

TheMadSword:
fail once more from me

PR #307, opened 2015-03-15
Add notification callback for when CForward is modified.

KyleSanderson:
Extensions presently guess at if their callbacks are modified, this should help enable them to become more accurate.


peace-maker:
I had to guess what you'd call a "modification of CForward". Since it's all about listeners being added/removed, i'd prefer to say that explicitly in the docs :)

KyleSanderson:
> I had to guess

Yeah, that's not good at all. The intention was for when CForward is modified that the callback be called. However, you can't change the name or other params at the moment after the forward is created; so there's no real point in having it be vague.

PR #310, opened 2015-03-15
Fixed missing include in psyfork/memoryutils.cpp

InstantMuffin:
Error:
MemoryUtils.cpp: In member function ‘void* MemoryUtils::ResolveSymbol(void*, const char*)’:
MemoryUtils.cpp:249:43: error: ‘fstat’ was not declared in this scope
  if (dlfile == -1 || fstat(dlfile, &dlstat) == -1)

Fixed by adding missing include: #include <sys/stat.h>


psychonic:
Thanks.

PR #311, opened 2015-03-16
Use the new Travis-CI Container environment.

KyleSanderson:
So... The new travis environment doesn't allow you to sudo. So I've added a virtual arch environment using juju ( https://github.com/fsquillace/juju ). The Ubuntu 12.04 image that Travis uses is still ancient. Benefit here is we can maintain our build environment completely, and our own repository if we're going this route...

It's actually pretty nice to work with in general.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ addons:
     - lib32stdc++6
     - lib32z1-dev
     - libc6-dev-i386
+	- linux-libc-dev
     sources:
     - llvm-toolchain-precise-3.5
     - ubuntu-toolchain-r-test
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,8 @@ addons:
     - lib32z1-dev
     - libc6-dev-i386
     - linux-libc-dev
-    - gcc-multilib
+    - g++-multilib
+    - g++4.8
     sources:
     - llvm-toolchain-precise-3.5
     - ubuntu-toolchain-r-test
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ addons:
     - lib32z1-dev
     - libc6-dev-i386
     - linux-libc-dev
+    - gcc-multilib
     sources:
     - llvm-toolchain-precise-3.5
     - ubuntu-toolchain-r-test
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ addons:
     - lib32stdc++6
     - lib32z1-dev
     - libc6-dev-i386
-	- linux-libc-dev
+    - linux-libc-dev
     sources:
     - llvm-toolchain-precise-3.5
     - ubuntu-toolchain-r-test
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,21 +1,24 @@
+addons:
+  apt:
+    packages:
+    - clang-3.5
+    - lib32stdc++6
+    - lib32z1-dev
+    - libc6-dev-i386
+    sources:
+    - llvm-toolchain-precise-3.5
+    - ubuntu-toolchain-r-test
+  cache:
+    directories:
+    - ../mysql-5.0
 language: cpp
+sudo: false
 compiler:
   - clang
 before_script:
-  - sudo apt-get update -qq
-  - sudo apt-get install -y python-software-properties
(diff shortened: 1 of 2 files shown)


KyleSanderson:
For reference, the build is presently failing because of the clang-3.6 change (missing -Wno-inconsistent-missing-override).

.travis.yml
```diff
@@ -1,21 +1,23 @@
...
+install:
+# We're using Juju (Arch Linux). It's actually pretty nice.
+  - git clone https://github.com/fsquillace/juju
+  - PATH=$PWD/juju/bin:$PATH
+  - juju -f echo ':)'
+  - echo Server = 'https://mirrors.kernel.org/archlinux/$repo/os/$arch' | juju -f tee -a /etc/pacman.d/mirrorlist
+  - printf "\n[multilib]\nInclude = /etc/pacman.d/mirrorlist\n" | juju -f tee -a /etc/pacman.conf
+  - yes | juju -f pacman -Syy
+  - juju -f pacman-optimize
+  - yes | juju -f pacman -S lib32-glibc lib32-clang git python tar python-apsw sudo grep
```
psychonic: Do we need to download sudo? It looks like it's not being used here (and cannot be).

KyleSanderson: We actually are using sudo in checkout-deps. When we pass the -f arg to JuJu we're "root" in that instance.

.travis.yml
```diff
@@ -1,21 +1,22 @@
-language: cpp
-compiler:
-  - clang
+cache: apt
+language: c
+sudo: false
```
WildCard65: Could you make this required to allow sudo usage? Travis-ci said this for ff2:  
This job is running on container-based infrastructure, which does not allow use of 'sudo', setuid and setguid executables.
If you require sudo, add 'sudo: required' to your .travis.yml
See http://docs.travis-ci.com/user/workers/container-based-infrastructure/ for details.

WildCard65:
Might it possible to use the auto apt install travis provides for container based infrastructures?

KyleSanderson:
@WildCard65 could you clarify what this means?

WildCard65:
+KyleSanderson Take a look at how VSH and FF2's repos install packages in the container environment. If you don't want to look at that, then ask @50Wliu since he did the change for FF2, VSH and SourceBans(the forked one by @Sarabveer)

asherkin:
Travis have a whitelist of apt source and packages you can install from the
yml file, if the custom source and additional packages we use are on the
list (or we make a request to add them and it is accepted), we wouldn't
need sudo or any magic. There are instructions in the Travis docs, it took
a bit of digging to find.

KyleSanderson:
https://github.com/travis-ci/travis-ci/issues/3548 should track the progress of this (provided I didn't mess anything up).

WildCard65:
@KyleSanderson You may want to trigger another travis build.

WildCard65:
@KyleSanderson I managed to fix the compile error this PR is experiencing in this PR: https://github.com/Arkshine/SteamTools/pull/1

Sarabveer:
I don't even know what is the difference between the old and the new container infrastructure.

WildCard65:
From what @50Wliu said, it should result in faster build start up times.

50Wliu:
Correct.  It's also going to slowly become the default if no sudo/ping usage is detected.

WildCard65:
isn't it default for no sudo/setuid? And isn't that already applied?

Sarabveer:
Apparently, if you don't have a Travis.yml, it used the container infrastructure and tries to compile using Ruby.

50Wliu:
References:
[Original blog post about Docker](http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/)
[Default...soon™](http://blog.travis-ci.com/2015-03-31-docker-default-on-the-way/)

And yes, @Sarabveer, Travis tries to compile with sane™ defaults if it doesn't find travis.yml.  Which isn't that sane for SM projects.

WildCard65:
@KyleSanderson To fix travis builds for this, g++-multilib is needed which is now whitelisted.

psychonic:
> The Travis CI build passed

Sarabveer:
Wow! After 3 months, Lol.

PR #312, opened 2015-03-21
Implement formula parsing.

WildCard65:
Also ported float.inc to new syntax.

### DIFF
--- a/translations/core.phrases.txt
+++ b/translations/core.phrases.txt
@@ -83,6 +83,17 @@
 		"en"			"Your name is reserved by SourceMod; set your password to use it."
 	}
 	
+	"Unknown Variable"
+	{
+		"#format"		"{1:s}"
+		"en"			"Unknown variable {1}"	
+	}
+	
+	"Unknown Error"
+	{
+		"en"			"Unknown error occurred parsing formula"	
+	}
+	
 	/* This is a special "pass-thru" translation */
 	"_s"
 	{
--- a/translations/core.phrases.txt
+++ b/translations/core.phrases.txt
@@ -91,7 +91,7 @@
 	
 	"Unknown Error"
 	{
-		"en"			"Unknown error occurred parsing formula"	
+		"en"			"Unknown error occurred while parsing formula"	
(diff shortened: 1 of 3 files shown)


translations/core.phrases.txt
```diff
@@ -83,6 +83,17 @@
...
 	
+	"Unknown Variable"
+	{
+		"#format"		"{1:s}"
+		"en"			"Unknown variable {1}"	
+	}
+	
+	"Unknown Error"
+	{
+		"en"			"Unknown error occurred parsing formula"
```
50Wliu: `while parsing formula`?

WildCard65: Derp, I'll fix that in the morning alone with compile errors

asherkin:
What's the use case for this? Why does it need to be in SourceMod Core?

WildCard65:
So plugins don't have to re-invent a math parser for usage(like FF2 and VSH have their own) while at same time allow for way to have custom variables be used for other plugins in parsing.

dvander:
I like the idea of a simple eval-string tool, but I'd like to understand more about the plugins that need this.

I do think it's worth making it work the same way the language does, i.e., "5+6*3" or something should evaluate as you'd expect it to in SourcePawn. It's not that difficult with only 5-6 operators, you can either do recursive-descent or something like http://en.wikipedia.org/wiki/Shunting-yard_algorithm is also very simple.

asherkin:
Plugins can provide natives and the extension API exists if it has to be C++, I really don't think we need this in core. I can't see it being useful for more than a tiny handful of plugins (well, 1, exactly). Maybe you don't want to maintain this in FF2, but that doesn't mean punting that burden off to SM.

WildCard65:
@dvander FF2 takes a string from a boss keyvalue file and then parses it to a float(rounded to int later) atm ff2 hardcodes the max complexity of a formula to 128 where this uses ke::Vector. Yes SM exposes ADT Array but it uses CellArray which last I heard it's going to be replaced in the near future right?

PR #313, opened 2015-03-22
Add SetClientName native.

psychonic:
Our current method for changing names (SetUserInfo), as implemented in the sm_rename command, does not work on the top three games by SM install count - TF2 (temporarily due to game bug, this works around), CS:S, and CS:GO. There may be others as well.

By implementing a new SetClientName native to wrap CBaseClient::SetName, we gain better game compatibility as well as bypass name change cooldown restrictions in the engine. SetName is one function earlier in the same vtable as SetUserInfo for every supported game, so there's very little extra support burden.

I've also updated the playercommands plugin to use the new native for sm_rename. I removed the old, ep1-specific logic and message for name changes while dead, too. The game logic it was considering was actually specific to CS:S, not the ep1 base, and hasn't been present in CS:S since the Orangebox port years ago.

However, due to some of the legacy code in CS:S, a hook on IVEngineServer::ClientCommand was added to block changes of the client "name" convar sent from the server. Any change to that client convar in Orangebox or later now resets the client's name to their Steam name.

@GoD-Tony 
@KyleSanderson

### DIFF
--- a/plugins/include/sdktools_functions.inc
+++ b/plugins/include/sdktools_functions.inc
@@ -333,7 +333,7 @@ native ActivateEntity(entity);
 native SetClientInfo(client, const String:key[], const String:value[]);
 
 /**
- * Something
+ * Changes a client's name.
  *
  * @param client		Player's index.
  * @param name			New name.
--- a/plugins/include/sdktools_functions.inc
+++ b/plugins/include/sdktools_functions.inc
@@ -333,6 +333,15 @@ native ActivateEntity(entity);
 native SetClientInfo(client, const String:key[], const String:value[]);
 
 /**
+ * Something
+ *
+ * @param client		Player's index.
+ * @param name			New name.
+ * @error               Invalid client index, or client not connected.
+ */
+native void SetClientName(int client, const char[] name);
+
+/**
  * Gives ammo of a certain type to a player.
  * This natives obeys the maximum amount of ammo a player can carry per ammo type.
  *
--- a/extensions/sdktools/extension.cpp
+++ b/extensions/sdktools/extension.cpp
@@ -50,6 +50,9 @@
  */
 
 SH_DECL_HOOK6(IServerGameDLL, LevelInit, SH_NOATTRIB, false, bool, const char *, const char *, const char *, const char *, bool, bool);
+#if SOURCE_ENGINE == SE_CSS
+SH_DECL_HOOK1_void_vafmt(IVEngineServer, ClientCommand, SH_NOATTRIB, 0, edict_t *);
+#endif
 
 SDKTools g_SdkTools;		/**< Global singleton for extension's main interface */
 IServerGameEnts *gameents = NULL;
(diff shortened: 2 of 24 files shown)


plugins/include/sdktools_functions.inc
```diff
@@ -333,6 +333,15 @@ native ActivateEntity(entity);
 native SetClientInfo(client, const String:key[], const String:value[]);
 
 /**
+ * Something
```
KyleSanderson: Something appears to be outstanding.

psychonic: Thanks. This is fixed now.

Bara20:
Is this for 1.7 possible or only 1.8+?

GoD-Tony:
The old bug report for reference: https://bugs.alliedmods.net/show_bug.cgi?id=5441

> However, due to some of the legacy code in CS:S, a hook on IVEngineServer::ClientCommand was added to block changes of the client "name" convar sent from the server.

Finally some closure on that problem. Nice find!

psychonic:
> Is this for 1.7 possible or only 1.8+?

I don't have a firm answer for that yet. We generally try to avoid new features in stable releases, especially ones that add new gamedata. However, this could also be considered a bug fix.

> The old bug report for reference: https://bugs.alliedmods.net/show_bug.cgi?id=5441

Thanks. I had forgotten about that. I've left a comment there to point here.

Bara20:
So 1.8+?

psychonic:
> So 1.8+?

It has been less than 24 hours since it landed on the dev branch. That is not an indication that it will or will not end up on the stable branch before the next stable release.

One would think that we might want to intentionally wait some time and let early-adopters test it before we determine it to be "stable".

PlayBoy31:
can you had this for Day of Defeat source too please ?

KyleSanderson:
What issues are you seeing on Day of Defeat: Source?

psychonic:
> can you had this for Day of Defeat source too please ?

This works for all games that SM supports. (The old sm_rename should have worked fine too for DoD:S).

PlayBoy31:
I tested on dods but no success: SetClientName not supported by this mod

psychonic:
The gamedata updater hasn't been updated for this game yet and is probably overwriting your gamedata files with ones from before this was added.

PlayBoy31:
ok, so it will be updated in a few minutes/hours?

psychonic:
The updater is updated now.

heshengjie:
It seems that sm_rename(SetClientName) doesn't work properly in CS:GO, the player name will be reverse to steam name after sm_rename command executed
I tested "1.7.1-git5182.tar.gz" and "1.8.0-git5393.tar.gz" on CS:GO server with Metamod:Source version 1.10.4 and OS version centos 6.5 32bit,

psychonic:
CS:GO was one of the game explicitly tested with this. Unless the last game update happened it break it, it should work.

I'll try to take a look in the next day or two.

heshengjie:
I notic that in extension.cpp, some code under the compile switch "SOURCE_ENGINE == SE_CSS" is added to fix the name reverting in CSS.
so, I change "SOURCE_ENGINE == SE_CSGO" here and recompile, the sm_rename works in CS:GO

psychonic:
Good investigative work! I can confirm now looking at CS:GO's server.so that CCSPlayer::ChangeName does have similar to or the same logic as CS:S's version. I must have not added the ifdef until after testing.

psychonic:
This is fixed now.

PR #314, opened 2015-03-22
Pass a handle instead of entptr to ProcessCondChange for stability.

psychonic:
@VoiDeD had seen some crashes on his TF2 servers after the new TF2 player conditions change detection was implemented. The crashes were in CHalfLife2::EntityToBCompatRef, called from PlayerConditionsMgr::ProcessCondChange.

This change makes it so that we pass an entity handle around instead of the raw player pointer, and checking it for validity before use in an attempt to quash the crash from the player no longer being valid.

### DIFF
--- a/extensions/tf2/conditions.cpp
+++ b/extensions/tf2/conditions.cpp
@@ -37,7 +37,7 @@ IForward *g_removeCondForward = NULL;
 
 struct CondChangeData_t
 {
-	CBaseEntity *pPlayer;
+	CBaseHandle hPlayer;
 	PlayerConditionsMgr::CondVar var;
 	int newConds;
 };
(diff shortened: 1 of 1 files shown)


extensions/tf2/conditions.cpp
```diff
@@ -51,10 +51,12 @@ static void HandleCondChange(void *pData)
 
 void PlayerConditionsMgr::ProcessCondChange(CondChangeData_t *pCondData)
 {
-	int client = gamehelpers->EntityToBCompatRef(pCondData->pPlayer);
-	if (!playerhelpers->GetGamePlayer(client)->IsInGame())
+	auto *pEdict = gamehelpers->GetHandleEntity(pCondData->hPlayer);
+	auto *pPlayer = playerhelpers->GetGamePlayer(pEdict);
```
KyleSanderson: This is not safe on >= L4D as pEdict can be NULL (compat_wrappers doesn't check, I'm not sure about in the engine). Otherwise, as this is for TF only, everything seems more safe.

psychonic: I will make a note to fix it when we port the TF2 extension to L4D.

PR #315, opened 2015-03-22
Make DirExists("") throw an error.

psychonic:
Apparently some (very few) plugins were relying on some undocumented behavior regarding passing an empty string to DirExists. SM was returning true, assuming that it referred to the current directory.

This change causes an error to be thrown in the case of an empty string being passed, similar to what we do for OpenDirectory. "." is still valid to explicitly work with the current directory.

### DIFF
--- a/core/logic/smn_filesystem.cpp
+++ b/core/logic/smn_filesystem.cpp
@@ -309,7 +309,7 @@ static cell_t sm_OpenDirectory(IPluginContext *pContext, const cell_t *params)
 	
 	if (!path[0])
 	{
-		return pContext->ThrowNativeError("Invalid file path");
+		return pContext->ThrowNativeError("Invalid path. An empty path string is not valid, use \".\" to refer to the current working directory.");
 	}
 	
 	Handle_t handle = 0;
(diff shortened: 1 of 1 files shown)


core/logic/smn_filesystem.cpp
```diff
@@ -616,7 +616,12 @@ static cell_t sm_DirExists(IPluginContext *pContext, const cell_t *params)
 {
 	char *name;
 	pContext->LocalToString(params[1], &name);
-	
+
+	if (!name[0])
+	{
+		return pContext->ThrowNativeError("Invalid path. An empty path string is not valid, use "." to refer to the current working directory.");
```
KyleSanderson: \" or '.

KyleSanderson:
I still don't think it's ideal we throw an error on DirExists anywhere, but this is better then silent failures.

psychonic:
Invalid input seems like a great reason for throwing an error. After all, you cannot name a directory "" on any common file system that I'm aware of.

dvander:
:ship:
1. :egg:
2. :egg:
3. :egg:

PR #316, opened 2015-03-28
Fix not being able to block CS_OnCSWeaponDrop and clarify include file.

Drifter321:
It appears that CS_OnCSWeaponDrop has always been broken and it was impossible to actually block the function even though the include said it could be done :smiley: The bug report https://bugs.alliedmods.net/show_bug.cgi?id=6334. I've also changed the wording in the include file as it  wasnt that clear what "deny" exactly meant.


KyleSanderson:
:shipit:

PR #319, opened 2015-04-02
Allow GetMapHistorySize to return numbers greater than 20.

KyleSanderson:
I've had a plugin that iterates GetMapHistory for quite a while, seeing the list limited to 20 entries when the server is running for over a month saddens me :(


peace-maker:
I doubt this will get changed often or even midgame, but you don't remove all the entries at the start until the array matches the size again, if the convar has been lowered and the array was already "full".

That TODO: indicator should die too.

KyleSanderson:
r=Dr!fter

PR #320, opened 2015-04-02
Remove Project Files from older outdated build systems.

KyleSanderson:
None of these project files have been maintained for quite some time (since before I even compiled SM, ever). Removing them at this point seems to be the best course of action. If they don't have an AMBuildScript I've left their project files intact (sample_ext, for instance).


psychonic:
I still use the msvc10 ones at times. Until --vs=gen is fixed, I'd really rather we keep these.

The other msvc ones and the Makefiles can still go.

KyleSanderson:
https://github.com/alliedmodders/sourcepawn/pull/9

PR #323, opened 2015-04-09
Eval stock

WildCard65:
Added a stock for plugins to parse a string into a float result. Note: I had to add a wrapper stock to get around a compiler bug.


asherkin:
Previous discussion: #312

KyleSanderson:
STL (std::) is not permitted anywhere in SM. Similar to #312 this seems completely pawn-able with no intricate IPC requirements.

KyleSanderson:
It looks like this is being converted from Pawn to C++... https://github.com/WildCard65/FF2-Official/blob/stable/addons/sourcemod/scripting/freak_fortress_2.sp#L6580

Is there something you need exposed to Pawn to keep it there?

WildCard65:
https://github.com/50DKP/FF2-Official/pull/239

KyleSanderson:
> 50DKP/FF2-Official#239
> Seems like ArrayList breaks floats.

In which version of SM did this break? Does this work on 1.6 and not 1.7-8? Is it just the transitional syntax that's broken?

50Wliu:
The only question that I can answer here is no, it's not just the traditional syntax (if anything even is broken...).  FF2 doesn't use transitional syntax yet.

As a side note, I'm perfectly fine with maintaining a formula parser within FF2 and not outsourcing it to SM.  But it's up to Wildcard in this case since he's handling formula-related stuff I guess.

KyleSanderson:
Can you file a bug please with a minimal test case? Lets get this fixed.

WildCard65:
@KyleSanderson I'm now believing it's a problem from conversion from new to old syntax that @50Wliu made me do and I derped it somewhere. I written a test plugin in new syntax and everything worked as expected, I'm going to change this from a native to a stock.

plugins/include/float.inc
```diff
@@ -237,11 +242,213 @@ native Float:ArcTangent2(Float:x, Float:y);
...
+ * @param variableCallback	Function to call when an invalid character is found.
+ * @return					Result of operation.
+ */
+stock float Eval(const char[] formula, OnError errorCallback, OnVariable variableCallback)
+{
+	int bracket;
+	char buffer[2], strValue[1024];
+	ArrayList sumArray = new ArrayList(.startsize=1);
+	ArrayList operatorArray = new ArrayList(.startsize=1);
+	for (int i = 0; i <= strlen(formula); i++)
```
KyleSanderson: This is extraordinarily expensive in comparison to other constants. Why are you using strlen against a variable that never changes in your for loop?

WildCard65: I may change it to break on null term.

plugins/include/float.inc
```diff
@@ -237,11 +242,213 @@ native Float:ArcTangent2(Float:x, Float:y);
...
+}
+
+public void OnEvalError(const char[] error)
+{
+	LogError(error);
+}
+
+/**
+ * A wrapper function to get around a compiler bug.
+ * Bug in question: Function params causing compile error when passed a default value. Compile error: <variable> prototype doesn't match.
```
KyleSanderson: Can you please file a bug for this along with your minimal repro case?

WildCard65:
Ok, I changed the stocks back to use strlen but only have it called once instead of it originally calling it more then once.

PR #324, opened 2015-04-09
1.7 dev bug with nextmap function

PlayBoy31:
first of all, sorry because i know its not the place here to say the bug i have found.

there is a bug with nextmap function. on my server i disabled lots of plugin of map management from sourcemod because i dont need them.

But what is interessting in my configuration is that when i type nextmap in console:

Williams killed Wolff with thompson.
[SM] Next Map: 9/2015 - 23:07:45

a date on the next map ? Maybe a memory leak issue ?

and sometime i got nothing (i think its normal because i have no plugin to set next map)
Requiem4destruction killed Tremblay with mg42.
[SM] Next Map: 

i think you have something to look in sourcemod code of nextmap function


asherkin:
This is not the place to file a bug. The bug tracker can be found here: https://bugs.alliedmods.net/enter_bug.cgi?product=SourceMod

PR #325, opened 2015-04-13
Notify plugin listeners of SetFailState'd plugins on unload. (bug 6347)

VoiDeD:
Following #93, `SetFailState` was changed to put plugins in the `Plugin_Failed` state rather than `Plugin_Error`.

Unfortunately this had the side effect of not notifying plugin listeners when the failed plugin was then later unloaded which is described in [bug 6347](https://bugs.alliedmods.net/show_bug.cgi?id=6347).

#93 also introduced some cases where some conditional paths were enabled for plugins that are in `Plugin_Failed`, and this patch adds that for notifying plugin listeners.

### DIFF
--- a/core/logic/PluginSys.cpp
+++ b/core/logic/PluginSys.cpp
@@ -1510,7 +1510,7 @@ bool CPluginManager::UnloadPlugin(IPlugin *plugin)
 	List<IPluginsListener *>::iterator iter;
 	IPluginsListener *pListener;
 
-	if (pPlugin->GetStatus() <= Plugin_Error)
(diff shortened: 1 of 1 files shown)


peace-maker:
Good catch :)

KyleSanderson:
Thanks!

dvander:
Hrm... I'm not sure this is correct yet.

dvander:
Okay, sorry, this looks okay. The contract on the extension listener is it's paired with OnPluginLoaded, which happens before the plugin can reach a Failed state.

PR #328, opened 2015-04-23
Add web related natives.

WildCard65:


KyleSanderson:
I'm a little concerned in regards to your handles, it looks like the pointers are never set from ReadHandle.

In DownloadString you're saving references to the present pawn stack which expires when it leaves scope.

You've left comments in the scripting API questioning what you've done.

Have you tested this?

asherkin:
Nice work!

We're trying to avoid having further blocking APIs though (and are slowly deprecating the ones that exist), it would be especially bad for HTTP given the remote nature. You can look at the Updater extension for an example of using IWebternet from a thread. There also appears to be a lot of debug logging left in.

WildCard65:
Woops must of committed debugging messages when debugging random report
error

WildCard65:
I'll have to thread up this tommorrow.

WildCard65:
Alright, threaded the natives

asherkin:
Have you tested this? It has no synchronization and will almost certainly crash.

WildCard65:
Any suggestions for synchronization?

extensions/curl/web_callback.cpp
```diff
@@ -0,0 +1,27 @@
...
+DownloadWriteStatus WebCallback::OnDownloadWrite(IWebTransfer *session,
+	void *userdata,
+	void *ptr,
+	size_t size,
+	size_t nmemb)
+{
+	//All but the WebQueueObject call was taken from SM Gamedata Updater MemoryDownloader.cpp
+	char* buffer = nullptr;
+	size_t bufsize = 0, bufpos = 0, total = size * nmemb;
+	if (bufpos + total > bufsize)
```
KyleSanderson: This entire branch will always be hit, and will crash when total == bufsize.

extensions/curl/web_thread.cpp
```diff
@@ -0,0 +1,52 @@
...
+	queue.remove(0);
+}
+
+void WebThread::OnTerminate(IThreadHandle *pHandle, bool cancel)
+{
+	if (!queue.length())
+		return;
+	while (queue.length() > 0)
+	{
+		queue[0].SkipDownload();
```
KyleSanderson: The DBI layer blocks to finish queries, this should do the same.

plugins/include/web.inc
```diff
@@ -0,0 +1,155 @@
...
+	 * Note: Call this once and only once before DownloadString().
+	 *
+	 * @param form		Handle to web form.
+	 * @return			True on success, false if a form is already set.
+	 * @error			Invalid web session handle.
+	 */
+	public native bool SetPostParam(WebForm form);
+	
+	/*
+	 * Should DownloadString() return false when an http error code is returned?
```
KyleSanderson: You still have messages questioning the API in the most public facing file possible.

plugins/include/web.inc
```diff
@@ -0,0 +1,155 @@
...
+	 *
+	 * @param error		True to fail on http error code, false to not.
+	 * @return			True on success, false otherwise.
+	 * @error			Invalid handle.
+	 */
+	public native bool SetErrorOnHttpError(bool error = true);
+	
+	/*
+	 * Should DownloadString() return the http header also?
```
KyleSanderson: Ditto.

KyleSanderson:
There's too much discussion here to have in a pull request. Please file a bug so we can help you more efficiently with implementing this.

extensions/curl/web_callback.cpp
WildCard65: Ok, I'll rework it on Monday.

PR #329, opened 2015-05-02
Exposed Webternet as natives.

WildCard65:
This is a remake of my last PR but uses SM's thread worker pool.


tools/buildbot/PackageScript
```diff
@@ -348,6 +348,7 @@ CopyFiles('plugins/include', 'addons/sourcemod/scripting/include',
     'usermessages.inc',
     'vector.inc',
     'version.inc',
+	'web.inc',
```
50Wliu: Tabs vs spaces

PR #331, opened 2015-05-07
Add support for Black Mesa Multiplayer.

psychonic:
For SM, this is largely just copying SDK 2013 support and compiling against hl2sdk-bms. (Only a single breaking difference at this time).

Radio menu support is disabled (not working), but ESC menus work. Nextmap is also disabled as they don't ship with a mapcycle.txt. I expect both of those to be rectified soon, at which point we can update our support.


psychonic:
Note: this won't compile until [MM:S PR 14](https://github.com/alliedmodders/metamod-source/pull/14) is merged into metamod-source/1.10-dev.

PR #332, opened 2015-05-07
Specifying more in details what GetClientLanguage does

TheMadSword:


plugins/include/lang.inc
```diff
@@ -60,6 +60,9 @@ native SetGlobalTransTarget(client);
 /**
  * Retrieves the language number of a client.
  *
+ * By default that language number is representing the language
+ * that the client's Steam application is using.
+ *
  * @param client	Client index.
```
psychonic: This is not necessarily true, however, it should be the language that the game is using.

By default, most Source games use the Steam client language, but at least a couple of them actually respect the game language set in Steam (rather than Steam UI language). By changing your text to "that the user's game client is using", it should be accurate for all.

asherkin:
Does this really belong in the function documentation at all? Is how SourceMod picks a language for a client relevant to the function manually retrieving that result?

We're somewhat lacking high-level documentation for the various systems, but it seems like this is either implementation detail or a note for the language/translation system as a whole (hardly anyone will be using this function directly).

TheMadSword:
@psychonic "By changing your text to "that the user's game client is using", it should be accurate for all."

That's not exactly true; try putting your CSS in German or Japanese for instance; GetClientLanguage will return the Steam language, since "cl_language english" is defined (if your Steam is in English) in CSS, even if the CSS game language (set in the properties of the game, accessible from the Steam game list) is set otherwise.

TheMadSword:
I changed it to not mention Steam nor the game client, as to avoid confusion.

PR #333, opened 2015-05-07
Increased PrintToChat, PrintCenter & PrintHint w/ "all" version, buffer sizes from 192 to 254

TheMadSword:
Increased PrintToChat & PrintToChatAll buffer size from 192 to 254; works in CSS/CSGO; dunno for the rest

### DIFF
--- a/core/HalfLife2.cpp
+++ b/core/HalfLife2.cpp
@@ -498,7 +498,7 @@ bool CHalfLife2::TextMsg(int client, int dest, const char *msg)
 		/* Use SayText user message instead */
 		if (chat_saytext != NULL && strcmp(chat_saytext, "yes") == 0)
 		{
-			char buffer[192];
+			char buffer[253];
 			UTIL_Format(buffer, sizeof(buffer), "%s\1\n", msg);
 
 #if SOURCE_ENGINE == SE_DOTA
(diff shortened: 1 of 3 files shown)


TheMadSword:
btw I tested 255 and it doesn't work; just if people were wondering

psychonic:
Going this close to the message limit is going to result in an overflow for games where the ChatSayText gamedata key is set.

In those cases, SayText is used instead of TextMsg in CHalfLife::TextMsg (see https://github.com/alliedmodders/sourcemod/blob/master/core/HalfLife2.cpp#L498)

SayText uses one more byte than a TextMsg.with one string (it supports up to four I believe).

I would recommend decreasing the size by at least 1 and/or testing with the "ChatSayText" key in core.games gamedata set to "yes".

TheMadSword:
It can't result in an overflow right now as the part you mention keeps a buffer of 192 bytes (I didn't change that part).

I'll test once more in CSS/CSGO with the ChatSayText key set to "yes" and see how it goes; I might need to close this PR & start another one.

asherkin:
Please don't open another PR, just push more commits to this one.

TheMadSword:
Done, tested once more in CSS/CSGO

TheMadSword:
PrintHint would have allowed 255, but -1 due to the PreByte thing @ Dark Messiah; both tested in CSS/CSGO (note that they often can't be completely seen, but this isn't new; they're still networked).

PR #334, opened 2015-05-07
Restore check_thunks error message.

asherkin:
This is mainly motivated to bring the 1.7 and 1.8+ versions back together, plus this message was probably here for a reason (and it doesn't look like things will do anything other than crash if this case is hit, hence the abort). I have no way of hitting the error case so that hasn't actually been tested, but it compiles and works fine normally. 1.7 changes will be separate because it still has the error using Msg.

@psychonic

### DIFF
--- a/public/asm/asm.c
+++ b/public/asm/asm.c
@@ -4,6 +4,8 @@
 #define _GNU_SOURCE
 #include <dlfcn.h>
 #include <string.h>
+#include <stdio.h>
+#include <stdlib.h>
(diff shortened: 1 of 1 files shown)


asherkin:
The diff for 1.7: https://github.com/alliedmodders/sourcemod/compare/alliedmodders:1.7-dev...asherkin:asm-1.7?expand=1

psychonic:
:goberserk:

PR #335, opened 2015-05-10
add include path to compile.sh

yedpodtrzitko:
SM 1.7 needs include path to be specified explicitly (afaik)


asherkin:
It does not. If it isn't correctly finding the include directory automatically, the problem is elsewhere (make sure binreloc is being linked into your build).

PR #336, opened 2015-05-12
Link libm in all bin, not just engine-specific ones.

psychonic:
This fixes the "undefined symbol: floorf" error that some people were getting in sourcemod.logic.so.


asherkin:
:+1:

PR #337, opened 2015-05-12
Populate MaxClients before OnPluginStart is called.

psychonic:
Now gets set in hook on LevelInit (and before plugins are initialized) rather than in hook on the later ServerActivate.

Also made players array initialized sooner for subsystems that relied on it being populated when max was set.


psychonic:
This was tested for functionality, but did have a memory leak the first time around as pointed out by @VoiDeD. Fixed now.

asherkin:
:+1:

PR #338, opened 2015-05-12
Remove Address_MinimumValid entry from sp incs

psychonic:
We have this defined in core for error checking, but it's useless in sp since unsigned comparisons are not supported.


asherkin:
:+1:

PR #339, opened 2015-05-13
Enable finding mapcycle in cfg dir on sdk2013 and bms.

psychonic:
Alternatively, we can just remove the engine checks as it probably will not hurt to enable the logic for all.

### DIFF
--- a/core/logic/smn_maplists.cpp
+++ b/core/logic/smn_maplists.cpp
@@ -91,7 +91,8 @@ public:
 		const char *pMapCycleFileName = m_pMapCycleFile ? smcore.GetCvarString(m_pMapCycleFile) : "mapcycle.txt";
 
 		if (strcmp(pEngineName, "tf2") == 0 || strcmp(pEngineName, "css") == 0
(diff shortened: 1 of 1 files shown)


powerlord:
Good to see you added in sdk2013 as well.

asherkin:
:+1: Enabling that logic for all seems sane though.

powerlord:
@asherkin 

I can see potential problems here for games that don't internally use cfg/mapcycle.txt style mapcyclefiles

1. We're now intentionally treating `mapcyclefile` differently than the game engine does.
2. `cfg/%s` is checked first.  Which means if someone had a copy of their `mapcyclefile` in cfg for some reason, that copy will be the version `nextmap`, `mapchooser`, `nominations`, `rockthevote`, and `randomcycle` use.
    * The reason this is OK for the 6 games previously mentioned is because their `mapcycle.txt` files are in cfg by default, with the implication that cfg is where mapcycles belongs.

asherkin:
That's fine. If there is a macycle in cfg for a game that doesn't normally put it there, then the server admin must have put it there explicitly and it makes sense to use it. SM takes over map changing so they wouldn't ever conflict. Unifying the logic across games makes sense from a support perspective.

PR #341, opened 2015-05-14
Add AMBuild2 script to Sample Extension

powerlord:
This adds an AMBuildScript, configure.py, and PackageScript to the Sample Extension.

This is adapted from SourceMod 1.8's AMBuildScript and PackageScript.

This AMBuildScript is also smart enough to notice if smsdk_ext.cpp is located in the project.  This was added for compatibility with SM 1.5 and below extensions.

NOTE:  This always requires MM:S to be present even for non-MMS extensions.  I should probably fix that, but afaict it only affects the include paths.

### DIFF
--- a/public/sample_ext/AMBuildScript
+++ b/public/sample_ext/AMBuildScript
@@ -1,0 +1,451 @@
+# vim: set sts=2 ts=8 sw=2 tw=99 et ft=python:
+import os, sys
+
+projectName = 'sample'
+
+#smsdk_ext.cpp will be automatically added later
+sourceFiles = [
+  'extension.cpp',
+]
+
+###############
+# Make sure to edit PackageScript, which copies your files to their appropriate locations
+# Simple extensions do not need to modify past this point.
+
+class SDK(object):
+  def __init__(self, sdk, ext, aDef, name, platform, dir):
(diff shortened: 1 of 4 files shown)


powerlord:
Note: I didn't touch the mms_sample_ext while I was working on this.  Does anyone even use that one?

asherkin:
Looks like you've merged the AMBuildScript and AMBuilder files, which makes it much harder for people to update in the future (as new SDKs get added).

powerlord:
@asherkin There, I've split them into separate files.

WildCard65:
You should probably also check for smsdk_ext.cpp in sdk folder

powerlord:
@WildCard65 Oh right, I don't have a copy of 1.5 handy.  I forgot it was in the sdk folder and not the base folder.  I'll fix that shortly.

Actually, I'm going to try to fix some of the comments as well as long as I'm doing another commit.

psychonic:
I won't have time to give this a real look until the weekend, but thanks for doing this! It's been needed for a while.

powerlord:
@psychonic So, did you ever get a chance to look at this?

psychonic:
Let's get this in there. We can always improve on it later if needed.

PR #343, opened 2015-06-04
Fix some buffer sizes in shipped plugins

psychonic:
See notes on commits.

### DIFF
--- a/plugins/admin-sql-threaded.sp
+++ b/plugins/admin-sql-threaded.sp
@@ -457,8 +457,8 @@ public void OnReceiveUser(Database db, DBResultSet rs, const char[] error, any d
 
 void FetchUser(Database db, int client)
 {
-	char name[65];
-	char safe_name[140];
+	char name[MAX_NAME_LENGTH];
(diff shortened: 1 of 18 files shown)


asherkin:
:ship:

plugins/admin-sql-threaded.sp
```diff
@@ -457,8 +457,8 @@ public void OnReceiveUser(Database db, DBResultSet rs, const char[] error, any d
 
 void FetchUser(Database db, int client)
 {
-	char name[65];
-	char safe_name[140];
+	char name[MAX_NAME_LENGTH];
+	char safe_name[(MAX_NAME_LENGTH * 2) - 1];
```
peace-maker: Shouldn't this be +1?

psychonic:
No, one of the MAX_NAME_LENGTH is a null, leaving MAX_NAME_LENGTH -1. each of those can take up to 2 chars after adding escape chars. So, ((MAX_NAME_LENGTH - 1) * 2 ) == (MAX_NAME_LENGTH - 2), plus one for a null, equals what is in the inc.

PR #346, opened 2015-06-10
Fix some ctype misuses (bug 6377).

psychonic:
isalpha, isdigit, isupper, and islower do not return 0/1. They return 0 or anything-else. Since the bool tag in pawn only supports exactly 0 and 1, we need to return 1 for all truthy returns in the natives that wrap these.


asherkin:
:shaved_ice:

PR #348, opened 2015-06-15
Add new ResolveFuzzyMapName native

powerlord:
This also modifies the internal ResolveFuzzyMapName to have multiple implementations, decided at compile time.

While I have tested to make sure I can [compile a plugin against it](https://github.com/powerlord/sourcemod-snippets/blob/master/scripting/resolvefuzzyname.sp), I have not yet had a chance to test it on a real server... and chances are I won't have a chance to until tomorrow.

### DIFF
--- a/public/IGameHelpers.h
+++ b/public/IGameHelpers.h
@@ -40,7 +40,7 @@
  */
 
 #define SMINTERFACE_GAMEHELPERS_NAME		"IGameHelpers"
-#define SMINTERFACE_GAMEHELPERS_VERSION		10
+#define SMINTERFACE_GAMEHELPERS_VERSION		11
 
 class CBaseEntity;
 class CBaseHandle;
@@ -314,6 +314,19 @@ namespace SourceMod
 		 * @return				Pointer to the string, or NULL if bad pointer.
 		 */
 		virtual const char *GetEntityClassname(CBaseEntity *pEntity) =0;
+
+		/**
+		 * @brief Resolves a fuzzy map name to its real map name that works
+		 * with the engine's Changelevel functionality.
+		 * Used internally for some games by IsMapValid.
+		 *
+		 * @param fuzzyName		Map entry name
+		 * @param outFullname	Full map path, excluding .bsp extension
+		 * @param size			Length of outFullname
+		 * @return				True if this map resolves to a real map, false if
+		 *						it doesn't exist.
+		 */
+		virtual bool ResolveFuzzyMapName(const char *fuzzyName, char *outFullname, int size) =0;
 
 		/**
 		 * @brief Returns whether or not a map name is valid to use with the
--- a/core/HalfLife2.cpp
+++ b/core/HalfLife2.cpp
@@ -1210,11 +1210,14 @@ const char *CHalfLife2::GetEntityClassname(CBaseEntity *pEntity)
 
 bool CHalfLife2::ResolveFuzzyMapName(const char *fuzzyName, char *outFullname, int size)
 {
+#if SOURCE_ENGINE != SE_TF2
+	// TF2's engine->IsMapValid doesn't seem to work at the moment, don't bother trying to call it
 	if (engine->IsMapValid(fuzzyName))
 	{
 		strncopy(outFullname, fuzzyName, size);
 		return true;
 	}
+#endif
 #if SOURCE_ENGINE >= SE_LEFT4DEAD
 	static ConCommand *pHelperCmd = g_pCVar->FindCommand("changelevel");
 	if (!pHelperCmd || !pHelperCmd->CanAutoComplete())
--- a/core/HalfLife2.cpp
+++ b/core/HalfLife2.cpp
@@ -1210,6 +1210,11 @@ const char *CHalfLife2::GetEntityClassname(CBaseEntity *pEntity)
 
 bool CHalfLife2::ResolveFuzzyMapName(const char *fuzzyName, char *outFullname, int size)
 {
+	if (engine->IsMapValid(fuzzyName))
+	{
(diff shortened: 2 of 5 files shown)


powerlord:
...figures, I noticed a typo in the comments for the native as soon as I finished the pull request.

powerlord:
Here are some results from my TF2 test server using my test plugin linked above:

   mappath glassworks
   glassworks is not a valid map
   mappath workshop/454118349
   workshop/454118349's real path is workshop/cp_glassworks_rc6.ugc454118349
   mappath dustbowl
   dustbowl's real path is cp_dustbowl
   mappath sawmill
   sawmill's real path is ph_sawmill_a2
   mappath nucleus
   nucleus's real path is arena_nucleus

KyleSanderson:
I don't like this returning false if the name matches, other then that this is a great addition as it should prevent casing mismatches from user input. Hopefully this makes it to other the Source engine forks.

KyleSanderson:
Just bringing this up as well, where the engine doesn't have support, we could roll something by hand in another PR. This isn't a complicated function, and we already have IDirectory.

powerlord:
@KyleSanderson Well, I couldn't tell if AutoCompleteSuggest would suggest things that were an identical match or not... the goal is obviously to make the two implementations identical.

As for adding our own implementations, it depends on if we JUST want it to work on things that work for ForceChangeLevel / changelevel / mapcycle entries or not.

psychonic:
The TF2 and L4D+ ones are already not identical, having different
matching logic.

powerlord:
@KyleSanderson After verifying on CSGO that it does return true for identical maps, I have changed the default behavior.

Having said that, it was also autocompleting de_dust to de_dust2 on CS:GO... so I modified the behavior to check the engine's IsMapValid command before it does any fuzzy matching.

powerlord:
Apparently TF2's engine->IsMapValid doesn't actually work anymore (it fails for cp_dustbowl for example)... should I leave it in for all versions, or just special-case it for the other two sections instead?

powerlord:
I know you guys aren't particularly fond of bumping the version numbers of SourceMod's interfaces, but I added ResolveFuzzyMapName to the IGameHelpers interface since its only core implementation also implements it in this pull request.

Now for a question:  When I started implementing changes to plugins to start using this, I ended up writing two stocks for things I kept using in multiple plugins.

The problem is that one of these stocks will change every time an additional game starts supporting workshop maps... requiring a recompile of the plugin calling it.  Should this function be moved to a native itself?

Basically, it's for getting the display name of a resolved workshop map.  At the moment, I've only implemented TF2 as I keep forgetting to check what GetCurrentMap returns for CSGO Workshop maps.

(As an example, In TF2, it turns `workshop/cp_basename_here.ugc123456789` into `cp_basename_here`)

public/IGameHelpers.h
```diff
@@ -316,6 +316,19 @@ namespace SourceMod
...
+		 * @param fuzzyName		Map entry name
+		 * @param outFullname	Full map path, excluding .bsp extension
+		 * @param size			Length of outFullname
+		 * @return				True if this map resolves to a real map, false if
+		 *						it doesn't exist.
+		 */
+		virtual bool ResolveFuzzyMapName(const char *fuzzyName, char *outFullname, int size) =0;
+
+		/**
 		 * @brief Returns whether or not a map name is valid to use with the
```
psychonic: I'd rather not expose this on the interface unless we have an immediate need or if it's been requested with a use case given.

The more we expose, the higher the maintenance burden.

psychonic:
I'm mildly concerned about not exposing the result of FindMap (due to the case of a map possibly existing but not being "ready"). However, if IsMapValid returns false for those (and it might or might not, I didn't look/remember), then that should be fine.

public/IGameHelpers.h
powerlord: Originally, I was planning on doing something with that in one of the extensions (SDKTools probably), but ended up just making it a stock instead.  (See pull request #350)

powerlord:
@psychonic 
IsMapValid is currently returning true for everything except eFindMap_NotFound.

In TF2, I'm not sure if there's a good way to resolve workshop maps if the user currently isn't using `tf_workshop_sync_map` on the map.  According to Valve's John S., FindMap will return eFindMap_PossiblyAvailable, but not rewrite what you passed into it... although for it to be picked up as a Workshop map, you have to pass in either `workshop/123456789` or `workshop/anythinghere.ugc123456789`.  If a Workshop map is currently being tracked, it will be resolved to the latter style name.

Incidentally, John S. also said that eventually, all workshop maps in the MapCycle will be synced automatically.

There's also a bug right now where doing a tf_workshop_map_sync in your server.cfg will likely crash the server as soon as a plugin attempts to read a map list in OnConfigsExecuted.  The [TF2 Workshop Map Loader plugin](https://forums.alliedmods.net/showthread.php?p=2309996#post2309996) exists to delay that by a few seconds to prevent this crash.  I was told last week Friday that this would be fixed in the next TF2 update.

One last note:  This is supposed to be cross-game compatible.  While we could alter it to return an enum of some sort, we'd have to be aware that only TF2 would have results that are really anything other than true or false.

public/IGameHelpers.h
powerlord: Anyway, and I forgot to outright say this, IGameHelpers can be reverted without breaking anything.

powerlord:
It occurred to me after my last post that yes, we could return our own enum based on whether it's a fuzzy match or not.  Still not sure how we'd handle TF2's `eFindMap_PossiblyAvailable` result though.

powerlord:
This is being closed as #351 was merged instead.

PR #349, opened 2015-06-17
sm_who reserved slot showing as admin fix

hlstriker:
Currently the sm_who command shows clients with the reservation flag as an admin. This causes a lot of confusion when a player needs an admin and wrongfully assumes another player with only the reservation flag is an admin.

This pull request will no longer show players with only the reservation flag as an admin. Technically admin commands can be mapped to the reservation flag but I feel this is bad plugin design and the reservation flag should be strictly used for a reserved slot.

### DIFF
--- a/plugins/basecommands/who.sp
+++ b/plugins/basecommands/who.sp
@@ -54,7 +54,15 @@ PerformWho(client, target, ReplySource:reply, bool:is_admin)
 	{
 		if (!is_admin)
 		{
-			ReplyToCommand(client, "[SM] %t", "Player is an admin", name);
+			new flags = GetUserFlagBits(target);
+			if (flags == ADMFLAG_RESERVATION)
+			{
+				ReplyToCommand(client, "[SM] %t", "Player is not an admin", name);
+			}
+			else
+			{
+				ReplyToCommand(client, "[SM] %t", "Player is an admin", name);
+			}
 		}
 		else
 		{
@@ -227,7 +235,7 @@ public Action:Command_Who(client, args)
 			}
 			else
 			{
-				if (flags == 0)
+				if (flags == 0 || flags == ADMFLAG_RESERVATION)
 				{
 					PrintToConsole(client, "%2d. %-24.23s %t", i, name, "No");
 				}


psychonic:
I have mixed feelings on this one. By default, only admins can use this command, and to them, it also displays the flags given. Additionally, any flag can technically be used for any purpose whether given actual administrative capabilities or otherwise. For example, it seems to be very common to give custom flag 1, 'o', to donors.

I don't disagree that there's an issue here, but this doesn't seem like the best solution. It may be overkill, but what about a config for basecommands to list which flags should (or should not) be considered "admins"?

KyleSanderson:
The name is very unfortunate for the subsystem, these users aren't necessarily admins, however they're using the admin system. As you're using this, I'm using this, and everyone else is using this, it's for identity mapping to assign permissions (alternatively, groups) to a user.

While fundamentally I agree with you, you're an "admin" when you have an identity assigned to you. It's way too late to change it now ;)

hlstriker:
Even though it is overkill I think there should be a way for a server operators to specify which flags appear as "admin" when non-admins use sm_who. I feel a client shouldn't be required to be deemed an administrator just because they have an identity assigned to them within the admin system.

It does seem rather silly to create a config purely for the who command but it sure would help prevent a few headaches in larger servers.

Perhaps a config that looks as simple as the following would work? Thinking of a name for said config file is the hard part.
```
Flags
{
	"a"	"0"
	"b"	"1"
	"c"	"1"
	"d"	"1"
	"e"	"1"
	"f"	"1"
	...
}
```

KyleSanderson:
So, how I use this is, each forum user gets an identity. If they belong to a usergroup they get flags... so something like
```
"8" // Level 2
{
        "Flags" "kp"
        "Immunity"      "2"
}

"7" // Level 1
{
        "Flags" "cdoj"
        "Immunity"      "1"
}

"28" // Level 0 (Dems)
{
        "Flags" "coj"
}

"14" // Base
{
        "Flags" "ab"
}

"6" // Donator
{
        "Flags" "t"
}
```

If who is causing you trouble, you could override it in admin_overrides.

DoctorMcKay:
I realize this PR is very old at this point, but I felt like throwing in my two cents anyway. By default, I think it should only say that a user is an admin if they have access to the "b" flag. The documentation specifies that all admins should have the generic flag, even though at the moment Core doesn't use it for anything besides access to the admin menu.

Therefore I think that since all "admins" should be getting the generic flag anyway, then basecommands should consider a user to be an admin if they have the generic flag.

PR #350, opened 2015-06-19
FindMap plugins

powerlord:
NOTE:  Since this was started, it switched from being based on pull request #348 to being based on the now-merged pull request #351.

This adds a new stock to sourcemod.inc: 

* GetFriendlyMapName changes a CSGO or TF2 workshop path to just the base mapname.  This is intended for display purposes in various plugins and/or plugins that need compatibility with existing names.

Note that in TF2, at this time you can't get the proper map path for a workshop map unless you `tf_workshop_map_sync` the map after the server start so that the server is actively tracking it.

The following plugins have all been updated to make calls to FindMap and GetFriendlyMapName as appropriate:

* MapChooser
* Nominations
* NextMap
* BaseVotes
* BaseTriggers
* RandomCycle


powerlord:
Based on input from John S. from Valve, all plugins that use the ProduceResolvedMapList function should be rewritten and the ProduceResolvedMapList stock removed entirely.

This is based on his comments about how tf_workshop_map_sync can be delayed.

powerlord:
That last commit is to convert a `ClearArray(g_NextMapList)` to the more modern `g_NextMapList.Clear()`

powerlord:
This is now based on PR #351, but both it and this have Linux issues at the moment.

I'll do some plugin testing on my test server once this is fixed.

powerlord:
Not sure when it happened, but I'm seeing bugs in nominations... need to check other plugins as well.

psychonic:
D:

-- 
Nicholas Hastings

powerlord:
@psychonic I thought I'd tested it since I switched it over to FindMap... maybe I didn't and just thought I did.

Then again, I'm testing it now.

FindMap itself is working as my mappath command is resolving these maps, they're just not being resolved in the current code for nominations/mapchooser/others

powerlord:
Oh wait, this may be because I'm not syncing workshop maps until AFTER plugins load due to a server crashing bug that should be fixed in tomorrow's update.  Let me test it again.

powerlord:
Yup, sure enough, that's exactly what was causing the issue I was seeing.

It'll be nice if they add the mapcyclefile auto-sync to tomorrow's TF2 update so I can remove the plugin I'm using to late-sync these maps.

powerlord:
I'm going to redo this pull request.

Not only is its history a mess because of the number of different methods I've done, but I think GetFriendlyMapName would be better served as a native.

In fact, I'm rewriting it as a native named GetWorkshopMapBaseName right now.  The name is a bit verbose, but it does explain what it's for.

So, expect to see a new pull request with that soon.  Not sure when as I'll need to do some testing on it after it's finished.

KyleSanderson:
@powerlord Just for reference, if you rebase your branch you can clean it up quite easily. This isn't recommended for a number of reasons, but you can recover this way.

powerlord:
@KyleSanderson You can, but they recommend you don't do it if it's already been pushed to a public repository.

PR #351, opened 2015-06-27
Expose FindMap/ResolveFuzzyMapName to plugins.

psychonic:
Similar goal as #348 - exposes the fuzzy map matching to plugins.

The biggest difference is that we attempt to expose as much as the FindMap functionality from TF2 (and eventually CS:S, DoD:S, etc.) as we can in other games.

@powerlord Can you give this a look and let me know if you have any concerns with the implementation?

Also, @VoiDeD

### DIFF
--- a/plugins/include/halflife.inc
+++ b/plugins/include/halflife.inc
@@ -97,12 +97,11 @@ enum FindMapResult
 	FindMap_Found,
 	// No match for this map name could be found.
 	FindMap_NotFound,
-	// A fuzzy match for this mapname was found and pMapName was updated to the full name.
-	// Ex: cp_dust -> cp_dustbowl
-	// Supported on many newer games.
+	// A fuzzy match for this map name was found.
+	// Ex: cp_dust -> cp_dustbowl, c1m1 -> c1m1_hotel
+	// Only supported for maps that the engine knows about. (This excludes workshop maps on Orangebox).
 	FindMap_FuzzyMatch,
-	// A match for this map name was found, and the map name was updated to the canonical version of the
-	// name.
+	// A non-canonical match for this map name was found.
 	// Ex: workshop/1234 -> workshop/cp_qualified_name.ugc1234
 	// Only supported on "Orangebox" games with workshop support.
 	FindMap_NonCanonical,
@@ -159,9 +158,9 @@ native bool:IsMapValid(const String:map[]);
 /**
  * Returns whether a full or partial map name is found or can be resolved
  * 
- * @param map 			Map path relative to maps/ folder, excluding .bsp extension.
- *						If result is FindMap_FuzzyMatch or FindMap_NonCanonical,
-						this will be updated to the full path.
+ * @param map 			Map name (usually same as map path relative to maps/ dir,
+ *						excluding .bsp extension). If result is FindMap_FuzzyMatch
+ *						or FindMap_NonCanonical, this will be updated to the full path.
  * @param maxlen		Maximum length to write to map var.
  * @return				Result of the find operation. Not all result types are supported on all games.
  */
--- a/plugins/include/halflife.inc
+++ b/plugins/include/halflife.inc
@@ -91,6 +91,26 @@ enum EngineVersion
 	Engine_BlackMesa,			/**< Black Mesa Multiplayer */
 };
 
+enum FindMapResult
+{
+	// A direct match for this name was found
+	FindMap_Found,
+	// No match for this map name could be found.
+	FindMap_NotFound,
+	// A fuzzy match for this mapname was found and pMapName was updated to the full name.
+	// Ex: cp_dust -> cp_dustbowl
+	// Supported on many newer games.
+	FindMap_FuzzyMatch,
+	// A match for this map name was found, and the map name was updated to the canonical version of the
+	// name.
+	// Ex: workshop/1234 -> workshop/cp_qualified_name.ugc1234
+	// Only supported on "Orangebox" games with workshop support.
+	FindMap_NonCanonical,
+	// No currently available match for this map name could be found, but it may be possible to load
+	// Only supported on "Orangebox" games with workshop support.
+	FindMap_PossiblyAvailable
+};
+
 #define INVALID_ENT_REFERENCE 0xFFFFFFFF
 
 /**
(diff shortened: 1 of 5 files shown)


plugins/include/halflife.inc
```diff
@@ -91,6 +91,26 @@ enum EngineVersion
 	Engine_BlackMesa,			/**< Black Mesa Multiplayer */
 };
 
+enum FindMapResult
+{
+	// A direct match for this name was found
```
VoiDeD: Does the api docgen understand regular comments?

psychonic: Yeah. Most of our newer api (largely methodmaps) have comments in this style, too.

plugins/include/halflife.inc
```diff
@@ -91,6 +91,26 @@ enum EngineVersion
 	Engine_BlackMesa,			/**< Black Mesa Multiplayer */
 };
 
+enum FindMapResult
+{
+	// A direct match for this name was found
+	FindMap_Found,
+	// No match for this map name could be found.
+	FindMap_NotFound,
+	// A fuzzy match for this mapname was found and pMapName was updated to the full name.
```
VoiDeD: Nit: the documentation references `pMapName` from the SDK side, god knows someone out there might get confused and _Make A Forum Thread_.

psychonic: Good catch! I'll remove that.

powerlord:
The only confusing thing I can see here is that auto-completing Workshop maps returns different results in L4D2+ than it does for OrangeBox games.

When TF2 auto-completes a workshop map, it will return eFindMap_NonCanonical on success or eFindMap_PossiblyAvailable if the map isn't being tracked and it can't get the "real" name ("real" is in quotes because the path doesn't actually exist on disc).

L4D2+ doesn't know if it's dealing with a workshop map or not, so just returns eFindMap_FuzzyMatch any time it does fuzzy matching.

Also, I'm assuming the FileExists is being done to avoid logging IsMapValid errors being logged to the server console.

On a side note, I've been told by Valve's John Schoenick that TF2's 

> IsMapValid() now requires a proper qualified filename, and no longer blindly wraps input in maps/%s.bsp due to engine changes. In a future SDK update these changes will be restricted to IVEngineServer v23 and requesting v22 will give you a legacy wrapper.

but since we use FindMap these days anyway, I guess this isn't really important.

plugins/include/halflife.inc
psychonic: I'm removing all mention of the var getting updates from the enum doc. It doesn't belong there anyway and is noted on the function doc.

psychonic:
> When TF2 auto-completes a workshop map, it will return eFindMap_NonCanonical or eFindMap_PossiblyAvailable on success depending on whether the map is being tracked and it can get the "real" name ("real" is in quotes because the path doesn't actually exist on disc).

> L4D2+ doesn't know if it's dealing with a workshop map or not, so just returns eFindMap_FuzzyMatch any time it does fuzzy matching.

Intended, for now. That's also documented in the enum. Does TF2 never give FuzzyMatch for workshop maps, even if they're fully downloaded and you use a partial, canonical name, ex. "workshop/<mapname>"?, rather than a non-canonical name like "workshop/<pubfileid>"

> Also, I'm assuming the FileExists is being done to avoid logging IsMapValid errors being logged to the server console.

Correct, as noted in the comment above that call.

> but since we use FindMap these days anyway, I guess this isn't really important.

Indeed. For that code path (just TF2 right now), we don't even call IsMapValid.

powerlord:
For `workshop/123456789` style
* If the map is being tracked (via `tf_workshop_map_syn`c or loading the map once), it will copy the `workshop/mapnamehere.ugc123456789` style name into the string variable and return `eFindMap_NonCanonical`
* If not tracked, but a copy of the map exists on your server, it will not change anything, but return `eFindMap_PossiblyAvailable`

I haven't tested what it will return for `workshop/mapnamehere.ugc123456789`

psychonic:
I see now. For TF2 workshop maps, they're not actually at the path displayed in console, nor even in the maps folder at all. That explains why FuzzyMatch doesn't/won't/can't work for them.

I still don't think that there is anything that we should (or can) do about it in code. I'll see about making the docs clearer.

psychonic:
It looks like there is also a compile error on Clang that I need to clear up. I only tested on MSVC/Windows.

powerlord:
I've updated Pull Request #350 to be based on this PR instead of #349.

Of course, that means it has the same Clang compile issue.  I'll have to make sure I merge it again after this one is fixed.

psychonic:
Hokay. Linux compile should be fixed now, and the native doc has been updated according to the findings above.

powerlord:
Hmm... Apparently TF2 is returning eFindMap_Found when it does non-Workshop fuzzy matching.

I should probably report this to Valve.

psychonic:
It returned FuzzyMatch for me when trying "cp_dust" and gave back "cp_dustbowl".

powerlord:
@psychonic Hmm... I was trying "dustbowl" and "dust" and it's executing the FindMap_Found case in my switch block.

powerlord:
@psychonic Were you testing that on Windows or Linux?

I'm testing this on TF2 Linux with [this plugin](https://github.com/powerlord/sourcemod-snippets/blob/ab92ebd0b3824f392f324bc638dd8f35f1795f1e/scripting/resolvefuzzyname.sp).

FindMap_NotFound, FindMap_NonCanonical, and FindMap_PossiblyAvailable are all working as expected (for not found, workshop resolved, and workshop unresolved).... I just happen to be getting FindMap_Found for both direct matches and fuzzy matches on TF2.

psychonic:
You're right, even on Windows. I must have not actually seen what I thought I saw. Using the exact case in Valve's own enum doc for eFindMap_FuzzyMatch, "Ex: cp_dust -> cp_dustbowl"), gives back eFindMap_Found (while still updating the buffer with the exact name).

I'll report it and just add a shim around it for now to return FuzzyMatch if the input doesn't match the output.

psychonic:
That is fixed now.

powerlord:
Thanks psychonic.  I can confirm this latest change appears to be working properly.

psychonic:
> I'll report it and just add a shim around it for now to return FuzzyMatch if the input doesn't match the output.

This has been acknowledged and will likely be fixed in a future game update.

PR #352, opened 2015-06-29
Lock database before setting character set

peace-maker:
SQL_SetCharset wasn't thread safe and could race with other threaded
queries causing a crash.


KyleSanderson:
Thanks for debugging this!

PR #353, opened 2015-07-02
Better string_t support for Get/Set EntPropString

psychonic:
This adds support for identifying Prop_Send string_t values for the DPT_String type (as opposed to char arrays), as well as getting values for Prop_Send string_t fields (Prop_Data was already supported) and setting values for both Prop_Send and Prop_Data string_t fields.

For setting, we have to allocate a string in the game's stringpool. The method here for it is a bit sketch, but it works and should not easily break.

This is not tested yet (other than the concept used for the AllocPooledString), but I plan to test in the near future.

### DIFF
--- a/core/HalfLife2.cpp
+++ b/core/HalfLife2.cpp
@@ -1278,6 +1278,8 @@ bool CHalfLife2::IsMapValid(const char *map)
 	return FindMap(szTmp, sizeof(szTmp)) != SMFindMapResult::NotFound;
 }
 
+// TODO: Add ep1 support for this. (No IServerTools available there)
+#if SOURCE_ENGINE >= SE_ORANGEBOX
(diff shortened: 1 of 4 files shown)


asherkin:
tumblr_n4bol3edlx1t4twpao1_400

ClaudiuHKS:
Hmmm, there is an issue related to SetEntPropString when the string is **m_szArmsModel**.
The arms are just removed.
Can you please check this out? Counter-Strike Global Offensive related.

psychonic:
In the future, please report bugs at https://bugs.alliedmods.net, rather than doing all of personally PMing me, personally emailing me, and commenting on a closed pull request.

This was filed and fixed on [bug 6391](https://bugs.alliedmods.net/show_bug.cgi?id=6391)

PR #354, opened 2015-07-02
GetMapDisplayName and associated core plugin changes

powerlord:
This replaces Pull Request #350.

This adds a new native named GetMapDisplayName.

Its purpose is to get the base name of a workshop map for display in various plugins.

At the present time, only CS:GO and TF2 actually do anything.

Its signature is 

    native bool GetMapDisplayName(const char[] map, char[] displayName, int maxlen);

This function returns false if a map can not be found or resolved, but true in all other circumstances (even for non-workshop maps).

No matter what it returns, displayName will always be populated, even if it is just by copying map to displayName.

The return type can be modified from its current behavior if needed.  Its current behavior was me being lazy.

The changes to core plugins are to implement both GetMapDisplayName and FindMap.

The plugins updated are:

* NextMap
* BaseTriggers
* BaseVotes (and basevotes/mapvote.sp)
* MapChooser
* Nominations
* RockTheVote
* RandomCycle

Note 1:  In TF2, the map must be tracked or it will not resolve.  This function will return false when a workshop map is asked for and FindMap returns that it is possibly available.  At the present moment, TF2 workshop maps can by tracked by using the `tf_workshop_map_sync` command or by loading the map once.

Valve's John S. has said

> In a future update we will automatically track maps found in the mapcycle, meaning you will only need tf_workshop_map_sync for maps you'd like to be prefetched but are not in the active cycle.

This change did not appear to land in the Gun Mettle Update.

### DIFF
--- a/core/HalfLife2.cpp
+++ b/core/HalfLife2.cpp
@@ -1273,6 +1273,38 @@ SMFindMapResult CHalfLife2::FindMap(const char *pMapName, char *pFoundMap, size_
 #endif
 }
 
+bool CHalfLife2::GetMapDisplayName(const char *pMapName, char *pDisplayname, size_t nMapNameMax)
+{
+	SMFindMapResult result = FindMap(pMapName, pDisplayname, nMapNameMax);
+
+	if (result == SMFindMapResult::NotFound)
+	{
+		return false;
+	}
+
+#if SOURCE_ENGINE == SE_CSGO
+	char *lastSlashPos;
+	// In CSGO, workshop maps show up as workshop/123456789/mapname
+	if (strncmp(pDisplayname, "workshop/", 9) == 0 && (lastSlashPos = strrchr(pDisplayname, '/')) != NULL)
+	{
+		ke::SafeSprintf(pDisplayname, nMapNameMax, "%s", &lastSlashPos[1]);
+		return true;
+	}
+#elif SOURCE_ENGINE == SE_TF2
+	char *ugcPos;
+	// In TF2, workshop maps show up as workshop/mapname.ugc123456789
+	if (strncmp(pDisplayname, "workshop/", 9) == 0 && (ugcPos = strstr(pDisplayname, ".ugc")) != NULL)
+	{
+		// Overwrite the . with a nul and SafeSprintf will handle the rest
+		ugcPos[0] = '\0';
+		ke::SafeSprintf(pDisplayname, nMapNameMax, "%s", &pDisplayname[9]);
+		return true;
+	}
+#endif
+
+	return true;
+}
+
 bool CHalfLife2::IsMapValid(const char *map)
 {
 	if (!map || !map[0])
--- a/core/HalfLife2.h
+++ b/core/HalfLife2.h
@@ -187,6 +187,7 @@ public: //IGameHelpers
 	bool IsMapValid(const char *map);
 	SMFindMapResult FindMap(char *pMapName, size_t nMapNameMax);
 	SMFindMapResult FindMap(const char *pMapName, char *pFoundMap = NULL, size_t nMapNameMax = 0);
+	bool GetMapDisplayName(const char *pMapName, char *pDisplayname, size_t nMapNameMax);
 #if SOURCE_ENGINE >= SE_ORANGEBOX
 	string_t AllocPooledString(const char *pszValue);
 #endif
--- a/plugins/basetriggers.sp
+++ b/plugins/basetriggers.sp
@@ -211,6 +211,7 @@ public Action:Command_Nextmap(client, args)
 	}
 	else
 	{
+		GetMapDisplayName(map, map, sizeof(map));
 		ReplyToCommand(client, "[SM] %t", "Next Map", map);
 	}
 	
@@ -291,7 +292,8 @@ public OnClientSayCommand_Post(client, const String:command[], const String:sArg
 	{
 		char map[PLATFORM_MAX_PATH];
 		GetNextMap(map, sizeof(map));
-			
+		GetMapDisplayName(map, map, sizeof(map));
+		
 		if (g_Cvar_TriggerShow.IntValue)
 		{
 			if (mapchooser && EndOfMapVoteEnabled() && !HasEndOfMapVoteFinished())
--- a/plugins/basevotes.sp
+++ b/plugins/basevotes.sp
@@ -116,7 +116,7 @@ public OnPluginStart()
 		OnAdminMenuReady(topmenu);
 	}
 	
-	g_SelectedMaps = CreateArray(33);
+	g_SelectedMaps = CreateArray(ByteCountToCells(PLATFORM_MAX_PATH));
 	
 	g_MapList = CreateMenu(MenuHandler_Map, MenuAction_DrawItem|MenuAction_Display);
 	g_MapList.SetTitle("%T", "Please select a map", LANG_SERVER);
@@ -311,8 +311,11 @@ public Handler_VoteCallback(Menu menu, MenuAction action, param1, param2)
 				
 				case (voteType:map):
 				{
+					// single-vote items don't use the display item
+					char displayName[PLATFORM_MAX_PATH];
+					GetMapDisplayName(item, displayName, sizeof(displayName));
 					LogAction(-1, -1, "Changing map to %s due to vote.", item);
-					PrintToChatAll("[SM] %t", "Changing map", item);
+					PrintToChatAll("[SM] %t", "Changing map", displayName);
 					new Handle:dp;
 					CreateDataTimer(5.0, Timer_ChangeMap, dp);
 					WritePackString(dp, item);		
--- a/plugins/basevotes/votemap.sp
+++ b/plugins/basevotes/votemap.sp
@@ -48,7 +48,7 @@ DisplayVoteMapMenu(client, mapCount, String:maps[5][])
 	
 	if (mapCount == 1)
 	{
-		strcopy(g_voteInfo[VOTE_NAME], sizeof(g_voteInfo[]), maps[0]);
+		GetMapDisplayName(maps[0], g_voteInfo[VOTE_NAME], sizeof(g_voteInfo[]));
 			
 		g_hVoteMenu.SetTitle("Change Map To");
 		g_hVoteMenu.AddItem(maps[0], "Yes");
@@ -61,7 +61,9 @@ DisplayVoteMapMenu(client, mapCount, String:maps[5][])
 		g_hVoteMenu.SetTitle("Map Vote");
 		for (new i = 0; i < mapCount; i++)
 		{
-			g_hVoteMenu.AddItem(maps[i], maps[i]);
+			decl String:displayName[PLATFORM_MAX_PATH];
+			GetMapDisplayName(maps[i], displayName, sizeof(displayName));
+			g_hVoteMenu.AddItem(maps[i], displayName);
 		}	
 	}
 	
@@ -288,8 +290,10 @@ int LoadMapList(Menu menu)
 	
 	for (new i = 0; i < map_count; i++)
 	{
+		decl String:displayName[PLATFORM_MAX_PATH];
 		GetArrayString(g_map_array, i, map_name, sizeof(map_name));
-		menu.AddItem(map_name, map_name);
+		GetMapDisplayName(map_name, displayName, sizeof(displayName));
+		menu.AddItem(map_name, displayName);
 	}
 	
 	return map_count;
--- a/plugins/mapchooser.sp
+++ b/plugins/mapchooser.sp
@@ -74,6 +74,8 @@ ConVar g_Cvar_RunOffPercent;
 Handle g_VoteTimer = null;
 Handle g_RetryTimer = null;
 
+// g_MapList stores unresolved names so we can resolve them after every map change in the workshop updates.
+// g_OldMapList and g_NextMapList are resolved. g_NominateList depends on the nominations implementation.
 /* Data Handles */
 ArrayList g_MapList;
 ArrayList g_NominateList;
@@ -288,15 +290,18 @@ public Action Command_SetNextmap(int client, int args)
 	}
 
 	char map[PLATFORM_MAX_PATH];
+	char displayName[PLATFORM_MAX_PATH];
 	GetCmdArg(1, map, sizeof(map));
 
-	if (!IsMapValid(map))
+	if (FindMap(map, displayName, sizeof(displayName)) == FindMap_NotFound)
 	{
 		ReplyToCommand(client, "[SM] %t", "Map was not found", map);
 		return Plugin_Handled;
 	}
-
-	ShowActivity(client, "%t", "Changed Next Map", map);
+	
+	GetMapDisplayName(displayName, displayName, sizeof(displayName));
+	
+	ShowActivity(client, "%t", "Changed Next Map", displayName);
 	LogAction(client, -1, "\"%L\" changed nextmap to \"%s\"", client, map);
 
 	SetNextMap(map);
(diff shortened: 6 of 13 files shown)


powerlord:
On a side note, I still need to do some testing to make sure all plugins are displaying the proper output.  

However, GetWorkshopMapBaseName has been tested on Linux TF2 and CS:GO servers by itself and it seems to be working.

These plugins were updated to switch calls to the GetFriendlyMapName stock to the GetWorkshopMapBaseName native and a variable name change here and there.

Having said that, RockTheVote's change is new and hasn't been tested at all.

powerlord:
I did some testing on it and things seem to be working as expected.  I merged the current master into it.  I did notice some minor naming issues I'm going to clean up, though (like *map instead of *pMapName)

core/HalfLife2.cpp
```diff
@@ -1258,6 +1258,50 @@ SMFindMapResult CHalfLife2::FindMap(char *pMapName, int nMapNameMax)
...
+#elif SOURCE_ENGINE == SE_TF2
+	char *ugcPos;
+	// In TF2, workshop maps show up as workshop/mapname.ugc123456789
+	if (strncmp(szTmp, "workshop/", 9) == 0 && (ugcPos = strstr(szTmp, ".ugc")) != NULL)
+	{
+		// Overwrite the . with a nul and snprintf will handle the rest
+		ugcPos[0] = '\0';
+		int newlength = strlen(szTmp);
+
+		snprintf(pBaseName, nMapNameMax < newlength ? nMapNameMax : newlength, "%s", szTmp + 9);
```
psychonic: Nit: &szTmp[9] syntax would be preferred over szTmp + 9.

psychonic:
The basics of this look okay to me, but I only skimmed through the mapchooser+friends edits.

If it's all tested, we should be good. My biggest concern there would be to make sure that we're still storing the full map "name" for votes, nominations, etc. rather than the new base name, as a base could potentially be ambiguous and should only be used for display.

core/HalfLife2.cpp
powerlord: I was using `szTmp[9]` before and Clang complained about it.  Didn't try it with a `&` though.

powerlord:
@psychonic As you said, the base name is for display only.  Generally, we create a variable for it, retrieve it, display it (or add it to a menu in the display name field), then let it go out of scope.

Having said that, the name returned from `FindMap` (**not** `GetWorkshopMapBaseName`) may be stored in adt_arrays.  Luckily, FindMap is safe to call on map names returned from FindMap, so that *should* be a non-issue.

asherkin:
Can we go for just GetMapBaseName (or GetMapDisplayName) instead?

powerlord:
@asherkin I'm not particularly attached to the function name, so we can change it.

Should we just go with `GetMapDisplayName` then?  It probably describes the function's purpose better than GetMapBaseName.

`GetMapDisplayName` might be the better of the two, not sure though.

asherkin:
GetMapDisplayName sounds good to me, @psychonic?

powerlord:
I'm going to push this for now so I don't forget.  If @psychonic doesn't like the name, we can always change it again later.

powerlord:
@asherkin @psychonic One question.

I've noticed that in some plugins I've updated LogAction to display the display name and some I didn't.

Since this behavior should be the same across plugins, should we be displaying the display name or the original name in LogAction?

I ask because I'm not sure which we should use due to OnLogAction's existence.

psychonic:
> GetMapDisplayName sounds good to me, @psychonic?

Yeah, that's fine.

psychonic:
@powerlord I think that logs should give the full map name or path, which display to users such as in chat or menus should use the shortened display name. @asherkin thoughts?

powerlord:
Hmm, should I update the documentation for GetMapDisplayName to note that it will return the result of FindMap even for non-workshop maps?

powerlord:
@psychonic @asherkin Right now, GetMapDisplayName returns true if FindMap returns FindMap_Found, FindMap_FuzzyMatch, or FindMap_NonCanonical.  Should I update it to only return true if the displayName is different than the map name passed in?

powerlord:
Yeah, I heard of an issue with `snprintf` on Windows, so I removed some of what I think is redundant logic (now that we overwrite the `.` with `\0`) that should lessen the chance of it happening.

Still, make sure you pass `PLATFORM_MAX_PATH` for the `nMapNameMax` variable and it won't ever happen.

powerlord:
The previous change has been tested as working on Linux, don't have a Windows server to test it, but it should work.

powerlord:
I ran into a new issue, though.

If you sync a lot of maps in server.cfg, the game server may not have them all cached by the time it generates the nominate list or voting list during the first map, which is done in OnConfigsExecuted.

Should we delay creating these menus/votes until the first time we use them in a map?

powerlord:
The previous changes dealing with late loading the nominations list and list of maps for the next vote appear to be working.

To clarify, the map **list** is not being late loaded, only the arrays and menus that use it are, in order to do late name resolution on them.

powerlord:
I'm kinda glad this hadn't been merged yet, as I just spotted a bug in the CS:GO code for GetMapDisplayName... when I fixed the pointer math to array syntax, I used the wrong pointer.

psychonic:
Sorry that I haven't had a chance to finish reviewing it. It's been a busy past few weeks,

powerlord:
I just force reset back a commit.  Turns out the sourcepawn commit referenced in sourcemod-git5533 is very buggy.

powerlord:
I'm not sure why, but TF2's FindMap seems to no longer resolve workshop map names like it used to.

For example, immediately after seeing this in the log (because I did tf_workshop_map_sync):

    [TF Workshop] Got updated information for map [ workshop/cp_glassworks_rc6a.ugc454118349 ]

this happens

    test_findmap workshop/454118349
    FindMap says PossiblyAvailab - "workshop/454118349"

It used to return NonCanonical and copy the longer map name into the map variable.

Short names are still resolved.

powerlord:
I emailed Valve's John Schoenick about this issue.  His response:

> FindMap does not return Found unless it has the map fully downloaded and available -- the message in question just means we have full information about the map, not that it is installed. I'll tweak it in the next update so that it still provides the full name, if known, in that situation, but otherwise this is working as intended.

I have replied to John pointing out the following:

1. The test I performed was 10 hours after the `tf_workshop_map_sync` command was initially issued, although it was issued again on map change.
2. The map in question is the same map I've had on my server since I initially started testing FindMap a week or so after Workshop support was added.  It was updated once since then.
3. FindMap used to return eFindMap_NonCanonical and update the map name for maps that were already on the server.

He hasn't responded to that yet, though.

powerlord:
The above behavior I'm seeing is a bug:

> There is a bug in the currently release build that could cause UGC files to never finish downloading. We hope to have a fix for that soon, but it would be good to rule that out.

(after doing more testing, I can't switch to workshop maps, even ones that are completely downloaded)

> That sounds like the aforementioned bug, which will be fixed by the next TF update. That would explain why FindMap() never considered it available as well :-/

powerlord:
Looks like today's TF2 update should fix the issues mentioned in the previous posts.  Needs testing still, though.

powerlord:
Did some testing after today's update.

All Map Workshop functionality appears to be working as intended.

core/HalfLife2.cpp
```diff
@@ -1258,6 +1258,49 @@ SMFindMapResult CHalfLife2::FindMap(char *pMapName, int nMapNameMax)
 #endif
 }
 
+bool CHalfLife2::GetMapDisplayName(const char *pMapName, char *pDisplayname, int nMapNameMax)
+{
+	char szTmp[PLATFORM_MAX_PATH];
```
KyleSanderson: I'm not sure you need this local array, it looks like you copy from it regardless to pDisplayname regardless of return.

core/HalfLife2.cpp
```diff
@@ -1258,6 +1258,49 @@ SMFindMapResult CHalfLife2::FindMap(char *pMapName, int nMapNameMax)
...
+		return true;
+	}
+#elif SOURCE_ENGINE == SE_TF2
+	char *ugcPos;
+	// In TF2, workshop maps show up as workshop/mapname.ugc123456789
+	if (strncmp(szTmp, "workshop/", 9) == 0 && (ugcPos = strstr(szTmp, ".ugc")) != NULL)
+	{
+		// Overwrite the . with a nul and snprintf will handle the rest
+		ugcPos[0] = '\0';
+
```
KyleSanderson: -1 style point for using two newlines everywhere in this function.

core/HalfLife2.cpp
powerlord: pMapName is const and FindMap can (and does) modify its first argument.

core/HalfLife2.h
```diff
@@ -183,6 +183,7 @@ class CHalfLife2 :
 	const char *GetEntityClassname(CBaseEntity *pEntity);
 	bool IsMapValid(const char *map);
 	SMFindMapResult FindMap(char *pMapName, int nMapNameMax);
+	bool GetMapDisplayName(const char *pMapName, char *pDisplayname, int nMapNameMax);
```
KyleSanderson: should probably be a size_t for the length of your string buffer, but the line above it also has an int so I won't fault you for it.

plugins/basetriggers.sp
```diff
@@ -290,8 +292,10 @@ public OnClientSayCommand_Post(client, const String:command[], const String:sArg
 	else if (strcmp(sArgs, "nextmap", false) == 0)
 	{
 		char map[PLATFORM_MAX_PATH];
+		char displayName[PLATFORM_MAX_PATH];
 		GetNextMap(map, sizeof(map));
-			
+		GetMapDisplayName(map, displayName, sizeof(displayName));
```
KyleSanderson: You're using this as a translation buffer to find the right map, you don't need to create a new buffer. If you do, we need to fix the native.

plugins/basevotes.sp
```diff
@@ -116,7 +116,7 @@ public OnPluginStart()
 		OnAdminMenuReady(topmenu);
 	}
 	
-	g_SelectedMaps = CreateArray(33);
+	g_SelectedMaps = CreateArray(ByteCountToCells(PLATFORM_MAX_PATH));
```
KyleSanderson: +++

plugins/basevotes.sp
```diff
@@ -311,8 +311,11 @@ public Handler_VoteCallback(Menu menu, MenuAction action, param1, param2)
 				
 				case (voteType:map):
 				{
+					// single-vote items don't use the display item
+					char displayName[PLATFORM_MAX_PATH];
+					GetMapDisplayName(item, displayName, sizeof(displayName));
```
KyleSanderson: This should be done before the vote happens, to ensure that users (and the vote initiator, if there is one) knows which map the vote counts for.

plugins/basevotes/votemap.sp
```diff
@@ -48,7 +48,9 @@ DisplayVoteMapMenu(client, mapCount, String:maps[5][])
 	
 	if (mapCount == 1)
 	{
-		strcopy(g_voteInfo[VOTE_NAME], sizeof(g_voteInfo[]), maps[0]);
+		decl String:displayName[PLATFORM_MAX_PATH];
+		GetMapDisplayName(maps[0], displayName, sizeof(displayName));
```
KyleSanderson: Write directly into g_voteInfo[VOTE_NAME], don't create a temporary buffer just to fill it, and copy into the next buffer immediately after.

plugins/mapchooser.sp
```diff
@@ -585,11 +591,19 @@ void InitiateVote(MapChange when, ArrayList inputlist=null)
...
 		
+		if (g_NextMapList.Length == 0)
+		{
+			CreateNextVote();
+		}
 		
+		char displayName[PLATFORM_MAX_PATH];
 		for (int i=0; i<nominationsToAdd; i++)
 		{
+
```
KyleSanderson: I think this snuck in?

plugins/nextmap.sp
```diff
@@ -153,7 +154,9 @@ void FindAndSetNextMap()
 		for (int i = 0; i < mapCount; i++)
 		{
 			g_MapList.GetString(i, mapName, sizeof(mapName));
-			if (strcmp(current, mapName, false) == 0)
+			strcopy(resolvedMap, sizeof(resolvedMap), mapName);
+			if (FindMap(resolvedMap, sizeof(resolvedMap)) != FindMap_NotFound &&
```
KyleSanderson: This pattern is brutal, sorry.

plugins/randomcycle.sp
```diff
@@ -91,11 +91,13 @@ public Action Timer_RandomizeNextmap(Handle timer)
 	
 	int b = GetRandomInt(0, g_MapList.Length - 1);
```
KyleSanderson: If you want to promote this to a do {} while(oldMaps && g_OldMapList.FindString(map) != -1); I think that would be best.

core/HalfLife2.cpp
KyleSanderson: That's unfortunately correct.

However, you have a nice array already present for you, via pDisplayname. If this is fed into FindMap you shouldn't have this issue. I can see the pain in pawn though.

plugins/basevotes.sp
powerlord: This is for a Yes/No map vote.  As such, the map name is part of the vote title and not the menu item, which is just the word "Yes" for its display.

I suppose I could pull it back out of `g_voteInfo[VOTE_NAME]` instead, but that brings up the question of why we weren't doing it that way in the first place.  I just assumed we had some sort of reason.

plugins/randomcycle.sp
powerlord: While that's not a bad idea, there's actually something more important to discuss here.

Pretty sure the decision earlier was to not pass results from FindMap or GetMapDisplayName to LogAction or SetNextMap.  Which means we need to make a copy of the string and send the copy to FindMap and PushString while keeping the original around for SetNextMap and LogAction.

KyleSanderson:
The engine at the moment in FindMap will write out 100 characters to an internal buffer then do the scan.

If you pass a length of 0, you can pass your const's in there and they should come out clean. This should avoid having to copy your buffer over and over. I'm really mixed on breaking this API in pawn and explicitly getting you guys to pass a const in instead of the following

```
char dispMap[24];
strncopy(dispMap, sizeof(dispMap), map);
new SMMapResult res = FindMap(dispMap, sizeof(dispMap);
```

I find this pattern to be really harmful in pawn, we have nothing else like it and it seems really bogus. In core at the moment we have IsMapValid creating a bogus temporary buffer to use the valve api.

powerlord:
@KyleSanderson Last time I checked, attempting to pass a `const char[]` to a method that takes a `char[]` will get you a compiler error in both SourcePawn and C++.

KyleSanderson:
https://github.com/alliedmodders/sourcemod/pull/396

powerlord:
Current build won't compile the plugins, but after fighting git for the last 20 minutes before finally convincing it that the submodule versions changed (it kept claiming they didn't), I'm going to do a "safety save."

KyleSanderson:
@powerlord you now have 21 commits in this branch and it's two months old, you should most likely rebase it once you have everything working.

powerlord:
@KyleSanderson Question about that... since I've merged master into this several times, do I just select all the commits and combine them?  I ask because all the commits made to master also show up in the commit log for this branch.

powerlord:
There, I think I rebased this correctly.  I had to watch it as it told me every time I changed HalfLife2.cpp that there was a conflict, but there was one instance where my change was the correct one (It was the one where the command name change).

It compiled using AMBuild on Windows at any rate.

powerlord:
I've amended the commit message as it was a mess after the rebase earlier.

psychonic:
Let's get this into the dev branch. We can iterate on it from there if necessary.

PR #355, opened 2015-07-02
Changing MAX_NAME_LENGTH from 32 to 128 due to CSGO having 128.

TheMadSword:
Changing MAX_NAME_LENGTH from 32 to 128 due to CSGO having 128 bytes buffer-size.

See related commit : https://github.com/alliedmodders/sourcemod/commit/9e0dbfcf6807c0782a6a8eea079bd14afe24c74d where a Duck says that CSGO uses 128 bytes name buffer.

### DIFF
--- a/plugins/include/clients.inc
+++ b/plugins/include/clients.inc
@@ -70,7 +70,7 @@ enum AuthIdType
  */
 
 #define MAXPLAYERS		65	/**< Maximum number of players SourceMod supports */
-#define MAX_NAME_LENGTH 32	/**< Maximum buffer required to store a client name */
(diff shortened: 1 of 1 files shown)


KyleSanderson:
Thanks for putting this in a separate PR.

psychonic:
As I already noted in IRC before this was created, we cannot raise this until everywhere in core is updated, else we could end up writing too much in some cases if core assumes the sp name size is 32.

KyleSanderson:
Missed that, sorry. With that being said however, the biggest overhead (delay) is getting users to update their plugins, not updating core.

PR #358, opened 2015-07-05
Removed case sensitive key checks

thekot:
Beam me up, Scotty!

### DIFF
--- a/plugins/admin-flatfile/admin-groups.sp
+++ b/plugins/admin-flatfile/admin-groups.sp
@@ -62,7 +62,7 @@ public SMCResult ReadGroups_NewSection(SMCParser smc, const char[] name, bool op
 	
 	if (g_GroupState == GroupState_None)
 	{
-		if (StrEqual(name, "Groups"))
+		if (StrEqual(name, "Groups", false))
 		{
 			g_GroupState = GroupState_Groups;
 		} else {
@@ -75,7 +75,7 @@ public SMCResult ReadGroups_NewSection(SMCParser smc, const char[] name, bool op
 		}
 		g_GroupState = GroupState_InGroup;
 	} else if (g_GroupState == GroupState_InGroup) {
-		if (StrEqual(name, "Overrides"))
+		if (StrEqual(name, "Overrides", false))
 		{
 			g_GroupState = GroupState_Overrides;
 		} else {
@@ -105,7 +105,7 @@ public SMCResult ReadGroups_KeyValue(SMCParser smc,
 	{
 		if (g_GroupState == GroupState_InGroup)
 		{
-			if (StrEqual(key, "flags"))
+			if (StrEqual(key, "flags", false))
 			{
 				int len = strlen(value);
 				for (int i=0; i<len; i++)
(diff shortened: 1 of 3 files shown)


psychonic:
Why is this change needed?

thekot:
I think it's really important to have those change while sourcemod have valve kv format support which has case related issues.

psychonic:
The file being read is not a kv file. What does it have to do with the kv format?

thekot:
Why does sourcemod core plugin reads files in smc format while it doesn't even have natives to write files in those format? Or I'm wrong and there is something like SMC_ToFile? But writing of kv files is supported. I'm a bit puzzled right now.

asherkin:
SMC has no in-memory representation, there is nothing to write from, it's all defined by the parser.

psychonic:
There is no in-memory SMC format to write to a file. It's an event-based parser, with the data storage being up to the plugin using it.

KeyValues uses a tree-based parser and keeps all of the data in memory. For more information on SMC vs KeyValues and why SMC exists, see http://www.sourcemod.net/devlog/?p=96

thekot:
So SMC is not a file format but a flexible parser, right? And it parses files in valve kv format. And those files may be obtained by simple call to KeyValuesToFile, which may produse case related issues.

psychonic:
> And it parses files in valve kv format

It does not. The format is visually similar, but there are multiple differences between them.
To name a few,
*KV does not requires keys to be quoted.
*KV does not support multiple comments.
*SMC doesn't support conditional evaluation.
*KV keys are case-insensitive. In SMC, it's left up to the consumer.

thekot:
> In SMC, it's left up to the consumer.
So, maybe there's a chance to merge those change. It's ain't gonna hurt to anyone.

I think there will always be some ambiguous about those *.cfg files format, which is look exactly like valve's kv one, but, yeah, you told me that it really another case-sensitive one.

thekot:
Those *.cfg files are for humans, right? Humans are case-insensitive beings! Don't force them to become robots.

_just kidding_

psychonic:
I would rather us not set a precedent of adding one-off fixes in core or base plugins to work around bugs in third-party plugins.

psychonic:
> Those *.cfg files are for humans, right? Humans are case-insensitive beings! Don't force them to become robots.
>
> *just kidding*

No, actually, those parts aren't meant to be edited by humans in this case. We ship example files with those key names already existing.

thekot:
> bugs in third-party plugins

As for me KeyValuesToFile is not a third-party "plugin".

> No, actually, those parts aren't meant to be edited by humans in this case.

What about "immunity" and other nested keys?

I think I'm done complaining here. You have a point and it seems I cant make you chage it.

One last question. Is it possible to show compiler warning on KeyValuesToFile usages to make it clear about possible drawbacks with those native?

psychonic:
> As for me KeyValuesToFile is not a third-party "plugin".

There is no inherent bug with that native. As the name suggests, it writes properly-formatted KV files. The issue here is trying to use a KV file as an SMC file. It's like renaming a .mp3 file to .wav and opening it in a media player application. It might detect the header and treat it as the right type, it might just error since it's not a .wav file. Either way, it can't be expected to Just Work as they're different formats.

> One last question. Is it possible to show compiler warning on KeyValuesToFile usages to make it clear about possible drawbacks with those native?

There was never anything anywhere to suggest that it would write a valid SMC file.

thekot:
Got it. Thanks for your time anyway

_Seems I'm going to use my very own version of admin-flatfile from now on. And It's not going to case any troubles, I think. Because you way so ~~lazy~~ inert to accept any changes._

PR #359, opened 2015-07-08
Map disabled plugins

KissLick:
Allows server masters to prevent some plugins load on certain maps.

This would be very useful for multi-mod servers like mg+course/climb+deathrun, zm+ze, skill surf+shoot surf etc. Map disabled plugins is a clean and easy way, how to manage plugins on maps.

**How does it work:**
- Is plugin disabled on this map?
  - Yes
    - Is plugin already running (from previous map)?
      - Yes -> Unload plugin before **OnMapStart** call
      - No -> Skip plugin loading and continue with next plugin
  - No -> Do nothing

**Advantages against:**
- Unloading plugins in map config
 - Some plugins change convars, precache files etc. in OnMapStart - might cause big problems
- Putting plugins in disabled folder and loading them in map config
 - This might cause big dependencies problems (eg. Timer with whole web of dependecies)
 - The need to unload plugins when the map ends


asherkin:
We generally advise large changes and new features be discussed before implementation to avoid wasting time.

That said, something like this is probably a reasonable addition, but I'm not sold on this specific implementation (the config file hell, and the fact all the logic is in core). I think a more reasonable path forward would be a global forward that is fired and returns an enum of [don't care, load, unload]. A simple boolean would probably also work, but would need to decide on the behaviour of conflicting results.

As for the implementation, STL is disallowed in SourceMod, and SourceHook::List is deprecated. Containers from AMTL should be used instead.

KissLick:
Oh, I got advised on SM forum to just pull request O:-)

I thought (and even tried) about global forward. I guess that would be the best case, but that requires the "plugins load controlling plugin" to load first before other plugins. It would work on map change, but not on server start. So I just ended up with configs... I wanted to be able to have a config for one map or a group of maps and this seemed the best way for it.

As for the STL, SH:List and AMTL, I am gonna have a look at that.

Thx

asherkin:
Special-casing a list of plugins that need to be loaded early still seems better (although it could be done automatically by doing another load pass and only loading plugins implementing that forward). I imagine that people will ask for map prefix support and stuff pretty quickly, although just supporting globbing might be enough.

PR #360, opened 2015-07-12
Add support for listening to, blocking, changing, and faking ClientCommandKeyValues

psychonic:
Adds forwards:

    // Pre-hook, blockable, can edit values on kv
    Action OnClientCommandKeyValues(int client, KeyValues kv)
    // Post-hook, only if not blocked
    void OnClientCommandKeyValues_Post(int client, KeyValues kv)

And native:

    // Bypasses hook if called within hook
    void FakeClientCommandKeyValues(int client, KeyValues kv)

Supported by all games that have CCKVs (all newer orangebox games and later, with the exception of Dota 2).

Editing the kv in OnClientCommandKeyValues will currently cause a crash on Windows, but I have a fix for that coming to tier1 in the SDK.

### DIFF
--- a/core/PlayerManager.cpp
+++ b/core/PlayerManager.cpp
@@ -2,7 +2,7 @@
  * vim: set ts=4 sw=4 tw=99 noet :
  * =============================================================================
  * SourceMod
- * Copyright (C) 2004-2009 AlliedModders LLC.  All rights reserved.
+ * Copyright (C) 2004-2015 AlliedModders LLC.  All rights reserved.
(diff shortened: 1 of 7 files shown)


psychonic:
> Editing the kv in OnClientCommandKeyValues will currently cause a crash on Windows, but I have a fix for that coming to tier1 in the SDK.

The tier1 code and Windows lib for each relevant SDK has now been updated.

asherkin:
giphy 1

PR #361, opened 2015-07-13
Corrected/updated Killtimer description.

InstantMuffin:
As you know Killtimer has two uses compared to CloseHandle.
-You can specify if you want to close the handle you passed
-You can kill repeating timers

The latter isn't obvious until you stumble upon the problem yourself and/or read up on what KillTimer does https://github.com/alliedmodders/sourcemod/blob/3291e3a38f8a458c7aebc233811e9514a2ec5f11/core/TimerSys.cpp#L388

Recent example showing that the first three people apparently don't know they should use KillTimer here:
https://forums.alliedmods.net/showthread.php?t=266180

Cheers and rip travis.

### DIFF
--- a/plugins/include/timers.inc
+++ b/plugins/include/timers.inc
@@ -90,7 +90,7 @@ typeset Timer
 native Handle:CreateTimer(Float:interval, Timer:func, any:data=INVALID_HANDLE, flags=0);
 
 /**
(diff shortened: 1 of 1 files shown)


asherkin:
This is incorrect, CloseHandle (or `delete` in the new syntax) works fine to kill a repeated timer, and should be preferred in all cases unless you need the additional options (as per the existing description).

PR #362, opened 2015-07-19
New forward - AskPluginAutoLoad

KissLick:
Ciao guys,

I added new forward function to allow/disallow (another) plugin to load.
```SourcePawn
forward APLRes:AskPluginAutoLoad(const String:plugin[], const String:map[]);
```
This function is called for every plugin that is about to auto load (plugins count times called in one plugin), **not** late load. If one plugin returns failure the given plugin is prevented from loading.

I also made little [test plugin](https://gist.github.com/KissLick/2bf6d04dfbf944aac7e1#file-test-askpluginautoload-sp) and [forwards order log](https://gist.github.com/KissLick/55978a1cc47ee005e5ab#file-askpluginautoload-forwards-order).

P.S.: This is releated to my previous [pull request](https://github.com/alliedmodders/sourcemod/pull/359) and thanks for reviewing my code.

EDIT: Changed to correspond with the latest commit.


plugins/include/sourcemod.inc
```diff
@@ -125,6 +125,8 @@ forward bool:AskPluginLoad(Handle:myself, bool:late, String:error[], err_max);
  */
 forward APLRes:AskPluginLoad2(Handle:myself, bool:late, String:error[], err_max);
 
+forward bool:AskPluginAutoLoad(const String:plugin[], const String:map[]);
```
powerlord: Shouldn't there be in-line documentation here as to what the forward does?  This is what shows up on the various SM documentation pages ([old](http://docs.sourcemod.net/api/) and [new](https://sm.alliedmods.net/new-api/)).

KissLick: Yep, added a doc for AskPluginAutoLoad

KissLick:
So what do you guys think?

PR #363, opened 2015-07-20
Introduce a datapack position tag.

VoiDeD:
Lots of users appear to be misusing the datapack position APIs and are making incorrect assumptions about what kind of values can be used as the position value.

This change aims to hide the value of a datapack's position behind a tag, thus requiring users to use the tagged value retrieved from a previous call to `GetPackPosition`, otherwise they'll encounter a compiler warning. This also updates the documentation to clarify how the functions should be used.

Hopefully using `enum` is still the usual way of declaring a tag in the new syntax.

@psychonic


psychonic:
Thanks.

PR #364, opened 2015-07-25
Fix maxlen in SetEntPropString not being initialized (bug 6391).

psychonic:
This var is only used when setting a direct string, rather than a string_t. Some flawed
logic was causing it to only be set if the sendprop didn't have a proxy function (although
all strings should; that's how we detect whether it's a string_t or not). The var only needs
to be set if it's not a string_t, but does not hurt anything if it is.


KyleSanderson:
r=me

PR #365, opened 2015-07-25
sampleext PackageScript: Add missing CopyFiles function.

powerlord:
PackageScript used a bunch of commented-out CopyFiles examples without actually including the CopyFiles function.


psychonic:
Thanks.

PR #366, opened 2015-08-08
Fix calling timer callback in paused plugins

peace-maker:
Don't try to call the timer callback, if it's not runnable.
Error wasn't reported before the exception refactoring. This reproduces the old behavior of just keeping the timer running and trying to call the callback until it's available again.

A better approach would be to stop all timers completely, if the plugin isn't going to change to running again from that paused state or just remove that pause state completely :+1:

### DIFF
--- a/core/logic/smn_timers.cpp
+++ b/core/logic/smn_timers.cpp
@@ -133,7 +133,10 @@ ResultType TimerNatives::OnTimer(ITimer *pTimer, void *pData)
 {
 	TimerInfo *pInfo = reinterpret_cast<TimerInfo *>(pData);
 	IPluginFunction *pFunc = pInfo->Hook;
-	cell_t res = static_cast<ResultType>(Pl_Continue);
+	if (!pFunc->IsRunnable())
+		return Pl_Continue;
+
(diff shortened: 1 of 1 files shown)


dvander:
I keep forgetting, why do we have paused plugins? Is there a reason we should have these checks everywhere just to keep unrunnable code loaded or should we just unload plugins that can't run?

KyleSanderson:
I think every time we discuss removing pausing support the requirements become too involved, the last IRC convo about it a couple months ago I remember was quite up there...

https://bugs.alliedmods.net/show_bug.cgi?id=5582 comes to mind ancient-bug wise, there may be others.

dvander:
That's a bug with pausing, which would adds support to my theory that pausing is always going to be buggy.

dvander:
Oh, I see - in that bug we did talk about ditching pausing too. I think it's time for us to just do this then.

dvander:
I'm merging this so we can backport it to 1.7. Pausing will get removed in 1.8, so future problems with it will have to go to the 1.7 branch directly.

PR #367, opened 2015-08-10
Add python syntax highlighting for AMBuildScripts on Github

ammaraskar:
Github allows per-file language overrides with the use of emacs mode lines. The global .gitattribute override only contributes to the language stats shown in the top bar.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+AMBuildScript linguist-language=Python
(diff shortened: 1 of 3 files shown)


ammaraskar:
Closing for now, since asherkin is looking into fixing this upstream to accommodate the current vim modeline

ammaraskar:
For anyone interested: https://github.com/github/linguist/pull/2546

PR #368, opened 2015-08-12
Update includes to remove binding syntax.

dvander:


core/logic/DebugReporter.cpp
```diff
@@ -177,6 +177,9 @@ void DebugReport::ReportError(const IErrorReport &report, IFrameIterator &iter)
 	g_Logger.LogError("[SM] Call stack trace:");
 
 	for (int index = 0; !iter.Done(); iter.Next(), index++) {
+		if (iter.IsInternalFrame())
+			continue;
+
```
psychonic: This looks like it slipped into this PR by accident.

plugins/include/datapack.inc
```diff
@@ -167,23 +167,68 @@ native void SetPackPosition(Handle pack, DataPackPos position);
  */
 native bool IsPackReadable(Handle pack, int bytes);
 
+// A DataPack allows serializing multiple variables into a single stream.
 methodmap DataPack < Handle
 {
```
psychonic: This whole methodmap can be moved nearer to the top of the file now for consistency and clarity. It was only at the bottom due to use the = syntax with existing natives above it.

plugins/include/regex.inc
```diff
@@ -132,11 +132,41 @@ native int MatchRegex(Handle regex, const char[] str, RegexError &ret = REGEX_ER
  */
 native bool GetRegexSubString(Handle regex, int str_id, char[] buffer, int maxlen);
 
+// Regular expression objects are used to match or decompose strings based on
+// patterns.
 methodmap Regex < Handle
 {
```
psychonic: This one can be moved to the top as well.

psychonic:
:+1: with notes.

PR #371, opened 2015-08-14
Fix build on Clang 3.6.

psychonic:
On Clang 3.6, a warning is thrown if you use the override keyword on some virtual members in a class but not all. While consistency is great, I don't think that having this break the build is worth it. (Alternatively, we can update the code in all cases were this matters).


dvander:
I agree, :bee:

PR #372, opened 2015-08-16
Cleanup some PluginSys idiosyncracies.

dvander:
This is a general cleanup, separated into a few commits for reviewability.

1. UnloadPlugin is now split up into an outer function and inner. The outer function just determines whether or not the unload can actually occur yet.
2. Dependency links were not formed correctly. The cache to optimize their creation was bogus in multiple ways, which has been fixed.
3. Dependency links were not created for library-only dependencies. That has been fixed.
4. The CPlugin class got some visual cleanup (organizing member variables, using AutoPtr, etc).
5. Plugin info is now stored in SH::String for safety. Requesting the info struct now taps into those strings instead of the runtime. This will let us persist info across unloads.

### DIFF
--- a/core/logic/PluginSys.cpp
+++ b/core/logic/PluginSys.cpp
@@ -71,6 +71,8 @@ CPlugin::CPlugin(const char *file)
 	m_errormsg[0] = '\0';
 	ke::SafeSprintf(m_filename, sizeof(m_filename), "%s", file);
 
+	memset(&m_info, 0, sizeof(m_info));
+
 	m_pPhrases = g_Translator.CreatePhraseCollection();
 }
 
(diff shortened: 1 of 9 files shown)


core/logic/PluginSys.cpp
```diff
@@ -1477,23 +1418,30 @@ void CPluginManager::TryRefreshDependencies(CPlugin *pPlugin)
...
+}
 
+bool CPluginManager::ScheduleUnload(CPlugin *pPlugin)
+{
 	/* This prevents removal during insertion or anything else weird */
-	if (m_plugins.find(pPlugin) == m_plugins.end())
-	{
-		return false;
-	}
+	assert(m_plugins.find(pPlugin) == m_plugins.end());
```
peace-maker: Shouldn't this be the other way around?

KyleSanderson:
If we have a plugin version of 3, m_DateTime is used uninitialized. Beyond that, lgtm.

dvander:
Great catch. Fixed.

PR #373, opened 2015-08-17
Don't use server commands for scheduling plugin unloads.

dvander:
If a plugin is currently active on the call stack, we can't unload immediately unload it since we would crash. The old code to deal with this used server commands to unload the plugin. That's pretty sketchy for a bunch of reasons.

The new method below adds the ability to attack one-off callbacks to IServerGameDLL::Think(), and then we use that to flush plugin unloads instead.

### DIFF
--- a/core/sourcemod.cpp
+++ b/core/sourcemod.cpp
@@ -46,11 +46,13 @@
 #include <amtl/os/am-path.h>
 #include <bridge/include/IExtensionBridge.h>
 #include <bridge/include/IScriptManager.h>
+#include <bridge/include/IProviderCallbacks.h>
 #include <bridge/include/ILogger.h>
 
 SH_DECL_HOOK6(IServerGameDLL, LevelInit, SH_NOATTRIB, false, bool, const char *, const char *, const char *, const char *, bool, bool);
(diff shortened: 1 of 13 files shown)


core/sourcemod.cpp
```diff
@@ -540,6 +543,7 @@ void SourceModBase::ShutdownServices()
 
 	SH_REMOVE_HOOK(IServerGameDLL, LevelShutdown, gamedll, SH_MEMBER(this, &SourceModBase::LevelShutdown), false);
 	SH_REMOVE_HOOK(IServerGameDLL, GameFrame, gamedll, SH_MEMBER(&g_Timers, &TimerSystem::GameFrame), false);
+	SH_REMOVE_HOOK(IServerGameDLL, Think, gamedll, SH_STATIC(*logicore.OnThink), false);
```
VoiDeD: Unintentional dereference?

dvander: Yup, thanks - though it shouldn't matter, it'll decay back into a pointer if it needs to.

asherkin:
tumblr_nsxrrobang1r0q799o1_500

dvander:
This image is terrifying.

dvander:
Pushed via command-line.

PR #374, opened 2015-08-17
Adding gamedata library support for vphysics

InstantMuffin:
Kyles: Had maple syrup before but not now because I'm not eating from a plate, like some sort of wild animal.


asherkin:
For the 3rd time, what's the use case for this? I don't see how this is useful without also exposing a helper to get and call functions on interfaces.

InstantMuffin:
I intended to call some CPhysicsCollisionData functions using an sdkcall, which without this change does not work. Psychonic said I could go ahead and pr, so here it is.

asherkin:
All the IPhysicsCollisionData functions are virtual, this is not necessary for calling them.

KyleSanderson:
@InstantMuffin do you have a use case for this?

PR #377, opened 2015-08-19
Add new method - Converts a bit string to a string of flag characters

s3bul:
I wrote this method a long time ago, maybe this will be helpful.


s3bul:
I added new syntax, can someone look for that?

peace-maker (changes requested):
Wow this is old. Sorry for the really long wait. The interchangeable bits vs AdminFlag vs character representation of admin permissions is ridiculous, but what we have I guess. Since we have natives for lots of the other conversions already this seems like a reasonable addition.

The function name could be a bit more descriptive. I'd suggest `FlagBitsToString`?

s3bul:
> `FlagBitsToString`

Yeah I add this very long time ago, but it's ok 😅 
Name `FlagBitsToString` is fine.

plugins/include/admin.inc
```diff
@@ -773,3 +773,34 @@ stock bool BitToFlag(int bit, AdminFlag &flag)
...
+	}
+	int i, c, numId = 0;
+	for (i = 0; i < numFlags; ++i)
+	{
+		if(FindFlagChar(array[i], c))
+		{
+			flags[numId++] = c;
+		}
+	}
+	for (i = numId; i < maxSize; ++i)
```
KyleSanderson, line 801: flags[numFlags] = '\0';

s3bul: Why `flags[numFlags] = '\0';`?
I think it should be `flags[numId] = '\0';` if `for` is too much

peace-maker (approved):
Thank you, this looks good now except for the usage of tabs for indentation in the doc comment, but this has waited for too long already. Sorry for taking this long and thank you for your patience!

PR #378, opened 2015-08-24
Fix ConVar .IntValue, .BoolValue, and .FloatValue assignments (Bug 6405)

powerlord:
This is a fix for [Bug 6405](https://bugs.alliedmods.net/show_bug.cgi?id=6405)

Simply put, this patch gets rid of `ConVar_BoolValue_set`, `ConVar_FloatValue_set`, and `ConVar_IntValue_set`.

Instead, it uses `sm_SetConVarNum` and `sm_SetConVarFloat` directly.

However, so we don't cause issues with pre-OrangeBox games,  the above two functions and `sm_SetConVarString` now check if `params[0]` (the parameter count) is >= the parameter number before using params 3 (replicate) and 4 (notify).

### DIFF
--- a/core/smn_console.cpp
+++ b/core/smn_console.cpp
@@ -361,13 +361,13 @@ static cell_t sm_SetConVarNum(IPluginContext *pContext, const cell_t *params)
 
 #if SOURCE_ENGINE < SE_ORANGEBOX
 	/* Should we replicate it? */
-	if (params[3] && IsFlagSet(pConVar, FCVAR_REPLICATED))
+	if (params[0] >= 3 && params[3] && IsFlagSet(pConVar, FCVAR_REPLICATED))
 	{
 		ReplicateConVar(pConVar);
 	}
 
 	/* Should we notify clients? */
-	if (params[4] && IsFlagSet(pConVar, FCVAR_NOTIFY))
+	if (params[0] >= 4 && params[4] && IsFlagSet(pConVar, FCVAR_NOTIFY))
 	{
 		NotifyConVar(pConVar);
 	}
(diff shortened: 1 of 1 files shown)


powerlord:
Using the test plugin from bug 6405, these are the results on my TF2 server:

```
incmaxrounds
L 08/24/2015 - 17:19:11: server_cvar: "mp_maxrounds" "6"
L 08/24/2015 - 17:19:11: [cvar_intvalue_bug.smx] maxrounds. old: 4, new: 6 (should be: 6)
incmaxrounds
L 08/24/2015 - 17:19:13: server_cvar: "mp_maxrounds" "8"
L 08/24/2015 - 17:19:13: [cvar_intvalue_bug.smx] maxrounds. old: 6, new: 8 (should be: 8)
incmaxrounds
L 08/24/2015 - 17:19:16: server_cvar: "mp_maxrounds" "10"
L 08/24/2015 - 17:19:16: [cvar_intvalue_bug.smx] maxrounds. old: 8, new: 10 (should be: 10)
```

powerlord:
I was hoping that, since this is a relatively simple patch that it would be patched in relatively quickly.

Particularly since the bug is showing up in core plugins that use the ConVar methodmap's properties.

psychonic:
This seems like it's avoiding whatever the real issue is. Unless I'm missing something (which is very possible), there is nothing wrong with the implementation or declaring of the getters and settings. If there's instead a parsing error in the compiler, that should still be fixed and would fix these too.

powerlord:
@psychonic The real issue seems to be some sort of caching in the setter.  You can tell it's in the setter because the getter is returning the correct value during testing.

Of course, there's also the question of why we were using a special method in the first place.  SourceMod already internally supports calling the same methods with different parameter counts... that's why params[0] exists in the first place.

dvander:
This change looks fine to me, but I don't understand what is actually changing. Before and after, the code looks semantically equivalent. What bug were you fixing by this change?

dvander:
Oh God, I see. The bug is really stupid. The replacement parameter vectors are "static" and they should not be static. Please just remove the static keyword and I'll merge that ASAP. That's better than changing the ABI of the old natives.

KyleSanderson:
It actually looks like replicate and notify are broken at the moment as well.

psychonic:
@KyleSanderson  can you elaborate?

What's broken other than the setting not giving the option to set them? (Keep in mind that those are only supported on ep1 anyway).

powerlord:
@KyleSanderson Speaking of elaborating... after @psychonic's considerations, do you mean broken it's broken in this code; the old code; or both?

@dvander Actually, I feel kinda dumb not catching that too, especially since I was explicitly looking for it.  Guess I'm too used to seeing static on global/class variables and methods rather than local variables.

dvander:
Fun fact, I made that mistake in two other places in smn_console.cpp as well. The difficulty of tracking down that sort of bug, makes me think that reusing older natives is actually better.

dvander:
ReplicateToClient is the last broken one, maybe that's what Kyle meant.

PR #379, opened 2015-08-26
Remove Format/FormatArgs from the core/logic bridge.

dvander:
This is in AMTL now so there's no point to pass it around via internal API.


TheDS:
LGTM otherwise.

PR #380, opened 2015-08-26
Remove strncopy from the core/logic bridge.

dvander:
I just uplifted strncopy to AMTL - so we can ditch its impl in core+logic and also remove it from the bridge (it was unneeded there to begin with).


TheDS:
:sheep:

dvander:
Merged via command-line.

PR #382, opened 2015-08-26
Move gnprintf and atcprintf to core/logic.

dvander:
These patches move the code over to a new file in core/logic and then add the logic bridge pieces to get player info for the `%N` format rule.


TheDS:
:shipit:

dvander:
Merged via command-line.

PR #385, opened 2015-08-28
Move RootConsoleMenu from core to core/logic.

dvander:
This occurs in a few steps:

1. Move unrelated junk out of the RootConsoleMenu class.
2. Break the IRootConsole API by no longer exposing internal data structures. This is ABI compatible but the old API calls for adding a root console menu item will now fail.
3. Move RootConsoleMenu into its own file.
4. Break ConsolePrint out into a new function.
5. Finally move the files to core/logic.


TheDS:
:sheep:

dvander:
Merged via command-line.

PR #386, opened 2015-08-30
Begin refactoring the core/logic bridge.

dvander:
Long term, the design of the core/logic bridge should be much more one-sided: what's currently called "core" should be a loader and a thin wrapper around ABI details of the SDK. Everything else should be under "logic". At that point, the bridge becomes one-way, and logic shouldn't need to provide anything back to core.

We've done a good job of moving things there so far, but to make the last stretch a refactoring will help a lot.

The first step is to rename `sm_core_t` to `CoreProvider`. This is the interface that a SourceMod loader must implement. Function pointers in this interface are now virtual members instead. (Except for the ones that will go away when the entire "move-stuff" project is complete.)

### DIFF
--- a/core/logic/AdminCache.cpp
+++ b/core/logic/AdminCache.cpp
@@ -328,7 +328,7 @@ void AdminCache::AddCommandOverride(const char *cmd, OverrideType type, FlagBits
 		return;
 
 	map->insert(cmd, flags);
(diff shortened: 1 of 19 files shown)


TheDS:
:8ball:

dvander:
Merged via command-line.

PR #387, opened 2015-08-30
Add a callback mechanism for the logic bridge.

dvander:
The new model for the core/logic bridge will be that logic provides a specific callback listener, rather than squirreling function pointers away inside the bridge structure that will eventually go away. The other two commits in this series are just cleanups/refactorings that came after.


TheDS:
:ship:

dvander:
Merged via command-line.

PR #388, opened 2015-08-31
Move ConVarManager hooks into a separate class.

dvander:
The next step of the bridge refactoring is to extract all the assorted game-specific hooks SourceMod has to do, so that it can push those notifications into logic rather than having each translation unit have to know everything about each game.

This PR shows how this is will be done for future patches. Here, ConVarManager's hooks are moved into a global class. Another advantage of this is that PlayerManager loses a dependency on ConVarManager.

I can see `GameHooks.cpp` getting super gross to read though, because the entire file is going to be an ifdef party. We'll see, I guess.

### DIFF
--- a/core/AMBuilder
+++ b/core/AMBuilder
@@ -41,6 +41,7 @@ project.sources += [
   'ConsoleDetours.cpp',
   'vprof_tool.cpp',
   'smn_commandline.cpp',
+  'GameHooks.cpp',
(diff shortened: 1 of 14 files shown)


TheDS:
:tada: ifdef party :tada:
:sheep:

dvander:
Merged via command-line.

PR #389, opened 2015-08-31
Split up intercom.h.

dvander:
I noticed that any change `intercom.h` ends up rebuilding most of the tree. Too much crud got added into it over time. This patch removes `intercom.h`, with each section getting moved into a new file under `bridge/include`. It also minimizes `#include` directives in shared headers.

Of course, this broke nearly every file under core/, so most of this patch is fixing up each .cpp file to only include exactly the components it needs.

This will matter less over time as most of our .cpp code lives in logic, but it helps minimal builds now and won't hurt it later. It also makes the headers easier to maintain, and many of them (like IExtensionSys/IScriptManager) will go away entirely once core is gone.


TheDS:
LGTM aside from a missing #include <string.h>

dvander:
Merged via command-line.

PR #390, opened 2015-09-01
Add translation files and updated GeoIP data to snapshots (bug 5728)

psychonic:
If we're to eventually change to a rolling release model, one of the first steps is to start having translations and updated GeoIP data included in our automatic builds. This addes that.

It's added at the BuildBot packaging step (after AMBuild has finished) both because it was easier, and to not affect time that manual builds take.

One thing that is not accounted for is versioning of translations. That will require further editing of the translator site and can be done at a later time since it won't currently cause an issue. (The only edit done to the site for this was adding the language id to the comment after each line in languages.cfg).

### DIFF
--- a/tools/buildbot/package.pl
+++ b/tools/buildbot/package.pl
@@ -3,7 +3,9 @@
 use strict;
 use Cwd;
 use File::Basename;
+use File::stat;
 use Net::FTP;
+use Time::localtime;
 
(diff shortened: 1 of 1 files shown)


asherkin:
tumblr_nsxrrobang1r0q799o3_500

PR #392, opened 2015-09-06
Move CDataPack from core to logic.

dvander:


core/logic/CDataPack.cpp
```diff
@@ -47,6 +51,24 @@ CDataPack::~CDataPack()
...
+static Vector<AutoPtr<CDataPack>> sDataPackCache;
+
+IDataPack * CDataPack::New()
+{
+  if (sDataPackCache.empty())
+    return new CDataPack();
+
+  CDataPack *pack = sDataPackCache.back().take();
+  sDataPackCache.pop();
+  return pack;
```
asherkin: Missing a call to `Initialize()` here?

dvander: Good catch, fixed.

PR #394, opened 2015-09-06
Remove ConCommand hook clutter.

dvander:
This series of patches isolates ConCommand dispatch hooking to GameHooks. Unfortunately it's a little tricky to do that, because there are three different signatures for a command dispatch callback.

What I did here was add a new reference-counted class, CommandHook, which encapsulates a hook id and a callback function. SH_ADD occurs against this class, which handles the different signatures, wrapping the arguments, and dispatching to the callback. When all references to a CommandHook are gone, the hook is removed. The reference is bumped during callbacks just in case some weird re-entrancy stuff goes on.

In addition a bunch more places that used CCommand have been converted to ICommandArgs.

All in all this was a very nice simplification to ChatTriggers and I think we can use the same mechanism to move CON_COMMAND usage over to logic. It should also simplify moving ConCmdManager and ChatTriggers.

### DIFF
--- a/core/ChatTriggers.cpp
+++ b/core/ChatTriggers.cpp
@@ -37,27 +37,13 @@
 #include "HalfLife2.h"
 #include "logic_bridge.h"
 #include "sourcemod.h"
+#include "provider.h"
 #include <amtl/am-string.h>
-
-#if SOURCE_ENGINE == SE_DOTA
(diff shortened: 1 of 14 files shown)


KyleSanderson:
:shipit: when m_bIsINS is resolved

dvander:
Filed PR #397 to deal with the horrible concmd cleaner code.

dvander:
MRES_IGNORE isn't required, but it's good practice. I'll fix that.

dvander:
Merged via command-line w/ nits fixed.

PR #395, opened 2015-09-06
Move the SetCommandClient hook into GameHooks.

dvander:
Also makes the client an argument into command callbacks, so consumers don't have to know about SetCommandClient.


KyleSanderson:
:shipit:

dvander:
Merged via command-line.

PR #397, opened 2015-09-09
Remove code that handles long-dead Metamod:Source versions.

dvander:
Our SourceMod builds require Metamod:Source 1.8 or higher.

Until now, it was still possible for users to build their own SourceMod against Metamod:Source 1.6 or 1.7. Quite a lot of code is dedicated to making this work, since the legacy API in those versions is missing a ton of stuff (notably, SourceHook VP hooks). Let's just kill all that off, since MM:S 1.8 was released in 2009.


psychonic:
:boat:

PR #398, opened 2015-09-11
Shim for TF2-branch games to find latest, non-shimmed engine iface.

psychonic:
This gross slab of hackeroni ensures that we get the latest (for now) engine iface available on TF2-branch games (including CS:S, DoD:S, HL2:DM, and SDK 2013 mods), so that we don't grab some older shim of an iface (causing our hooks to not fire).

This should allow seamless functionality now for TF2, both before and after the impending updates for CS:S, DoD:S, and HL2:DM, as well as both SDK 2013 mods that have updated to the newer iface and those that haven't.


dvander:
:ship:

PR #399, opened 2015-09-11
Fix IsMapValid call on non-TF2 TF2-branch games to use old behavior requiring only map name

psychonic:
This slimey hunk of hackeroni ensures that we get the "classic" IsMapValid behavior for games (other than TF2) that are on the TF2 branch, including CS:S, DoD:S, HL2:DM, and SDK 2013 mods.

The newer behavior is to have a relative (or maybe full?) map path as input, rather than just the map name. Since the rest of our logic only expects the map name, including on all other games, this normalizes that. It would also be necessary for TF2 if we weren't bypassing IsMapValid altogether and going through FindMap.

### DIFF
--- a/core/HalfLife2.cpp
+++ b/core/HalfLife2.cpp
@@ -1268,6 +1268,9 @@ SMFindMapResult CHalfLife2::FindMap(const char *pMapName, char *pFoundMap, size_
 	}
 
 	return static_cast<SMFindMapResult>(engine->FindMap(pFoundMap, static_cast<int>(nMapNameMax)));
+#elif SOURCE_ENGINE == SE_CSS || SOURCE_ENGINE == SE_DODS || SOURCE_ENGINE == SE_HL2DM || SOURCE_ENGINE == SE_SDK2013
+	static IVEngineServer *engine21 = (IVEngineServer *)(g_SMAPI->GetEngineFactory()("VEngineServer021", nullptr));
+	return engine21->IsMapValid(pMapName) == 0 ? SMFindMapResult::NotFound : SMFindMapResult::Found;
 #else
 	return engine->IsMapValid(pMapName) == 0 ? SMFindMapResult::NotFound : SMFindMapResult::Found;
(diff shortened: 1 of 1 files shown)


powerlord:
IsMapValid now takes a relative path starting with `map/` in TF2.  I haven't done experiments, but I think John Schoenick mentioned it in one of the many emails we had back and forth.

psychonic:
> IsMapValid now takes a relative path starting with map/ in TF2. 

That is what I was haphazardly mentioning. Since we don't call IsMapValid in TF2, it's not relevant. In other games on that branch (with this change), we call the old IsMapValid, so still not relevant.

core/HalfLife2.cpp
```diff
@@ -1268,6 +1268,9 @@ SMFindMapResult CHalfLife2::FindMap(const char *pMapName, char *pFoundMap, size_
 	}
 
 	return static_cast<SMFindMapResult>(engine->FindMap(pFoundMap, static_cast<int>(nMapNameMax)));
+#elif SOURCE_ENGINE == SE_CSS || SOURCE_ENGINE == SE_DODS || SOURCE_ENGINE == SE_HL2DM || SOURCE_ENGINE == SE_SDK2013
+	static IVEngineServer *engine22 = (IVEngineServer *)(g_SMAPI->GetEngineFactory()("VEngineServer022", nullptr));
+	return engine22->IsMapValid(pMapName) == 0 ? SMFindMapResult::NotFound : SMFindMapResult::Found;
```
KyleSanderson: It may be better to fallback here to engine.

KyleSanderson: :shipit:

psychonic:
I didn't realize that this didn't land yet, but that's good. It should have used engine v21 for the fallback instead of v22. Both are available but v21 is what CS:S, DoD:S, and HL2:DM are currently using. This is also the case for the oldest of SDK2013 mods.

PR #400, opened 2015-09-11
Remove sm_srvcmds.cpp.

dvander:
This series of patches adds a mechanism for logic to define CON_COMMANDs, then moves the commands from sm_srvcmds.cpp into various parts of logic. Although the caller code ends up looking quite nice, the actual implementation is a hack: we allocate a ConCommand, then add a hook for it.


KyleSanderson:
:shipit: after core/logic/RootConsoleMenu.cppL283

dvander:
Thanks, I'm surprised clang rolled with that one.

PR #402, opened 2015-09-11
Shim for TF2-branch games to find latest, non-shimmed engine iface. 2

psychonic:
This fixes some quirks from PR #398 being done a bit too haphazardly.


Drifter321:
:ship:

PR #403, opened 2015-09-11
Fix Makefile, and add info for byte checks.

Drifter321:
This fixes the makefile to work with  central (and the split of sourcepawn) I also added a bit more info to the printing of byte checks.

### DIFF
--- a/tools/gdc-psyfork/Makefile
+++ b/tools/gdc-psyfork/Makefile
@@ -22,12 +22,12 @@ BINARY = gdc
 
 LINK += -L. -lstdc++ -ldl -lm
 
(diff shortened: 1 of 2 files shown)


tools/gdc-psyfork/Makefile
```diff
@@ -22,12 +22,12 @@ BINARY = gdc
 
 LINK += -L. -lstdc++ -ldl -lm
 
-INCLUDE += -I. -I../../public -I../../public/sourcepawn -I../../core -I../../core/logic
+INCLUDE += -I. -I../../public -I../../public/sourcepawn -I../../core -I../../core/logic  -I../../sourcepawn/include
```
psychonic: You can probably ditch the -I../../public/sourcepawn now too. (not that I remember why this would even need those headers).

psychonic:
:goberserk:

PR #404, opened 2015-09-12
Fix a regression from the ConCmdManager refactoring that causes a crash using SourceMod commands in the server console.

dvander:
Can't use RETURN_META when no hooks are on the stack. No hooks will ever be on the stack here. I don't know why I added these.


KyleSanderson:
:shipit:

PR #396, opened 2015-09-09
Change FindMap to take a const char* for searching instead of char*.

KyleSanderson:
I dislike the FindMap API we have greatly because we're asking users to create a new buffer, fill it, and then pass it to FindMap to determine if the user input is valid. We don't have any other APIs like this, and it turns into a whole thing everytime this native is to be used. Internally we have the exact same pattern and we throw away the result...

The alternative I'm proposing is the top, bottom being present.
```native FindMapResult FindMap(const char[] map, char[] foundmap, int maxlen);```
```native FindMapResult FindMap(char[] map, int maxlen);```

This is only compile tested.

PR #406, opened 2015-09-12
Change FindMap to take a const char* for searching instead of char*.

KyleSanderson:
https://github.com/alliedmodders/sourcemod/pull/396


KyleSanderson:
r=Drifter.

PR #407, opened 2015-09-13
Added IsFree() edict check to FindEntityByNetClass

Thordin:
I had a null pointer crash in FindEntityByNetClass

https://github.com/Thordin/sourcemod/blob/3a59baab2c899a749d3d6d7f1d0906f75a8ad53d/extensions/tf2/extension.cpp#L429

I found the same function here except they added an IsFree() check on the edict, so maybe that was the problem.

https://github.com/alliedmodders/sourcemod/blob/master/extensions/sdktools/gamerulesnatives.cpp#L43


KyleSanderson:
Thanks Thordin

PR #409, opened 2015-09-16
Deprecate functions in non-threaded DBI API

KyleSanderson:
These natives can be very brutal on frame time and are commonly used inappropriately. Most recently we fixed a crash with SQL_SetCharset by locking the database https://github.com/alliedmodders/sourcemod/pull/352, however this impacted a plugin later down the road when <BCS|Ejziponken> reported his server was crashing (freezing) today from https://github.com/splewis/csgo-multi-1v1/commit/96edda2d43ce9cd102ac5089a84f02f10340dcb7#diff-c9ce3c3d5f758e48852dd0958ee38671. The problem here was the user is locking their database because they had threaded queries later on down the line.

This plugin seems more up to snuff then most, using a lot of API features; my worry being they're not using the DBI API correctly. I'm not saying there aren't plugins that don't need the natives, however the coarse rope burns more then it's worth.

```
    db = SQL_Connect(dbCfgName, true, error, sizeof(error));
    if (db == INVALID_HANDLE) {
        g_dbConnected = false;
        LogError("Could not connect: %s", error);
    } else {
        SQL_LockDatabase(db);
        SQL_SetCharset(db, "utf8");

        // create the table
        SQL_CreateTable(db, TABLE_NAME, g_TableFormat, sizeof(g_TableFormat));
```

Since they're all SQL_TQueries from what I can see going forward, there's no reason for the Lock to exist here. This isn't the only plugin that blocks for no reason.

The only real usecase I've seen for this is for getting InsertIDs. I'd much rather add an API to force a pawn callback before continuing down the queue. In the meanwhile; deprecate this API.

### DIFF
--- a/plugins/include/dbi.inc
+++ b/plugins/include/dbi.inc
@@ -105,6 +105,7 @@ enum DBPriority
  * @param maxlength		Maximum length of the error buffer.
  * @return				A database connection Handle, or INVALID_HANDLE on failure.
  */
+#pragma deprecated Use SQL_TConnect
 native Handle:SQL_Connect(const String:confname[], bool:persistent, String:error[], maxlength);
 
 /**
@@ -117,6 +118,7 @@ native Handle:SQL_Connect(const String:confname[], bool:persistent, String:error
  * @return				A database connection Handle, or INVALID_HANDLE on failure.
  * 						On failure the error buffer will be filled with a message.
  */
+#pragma deprecated Use SQL_TConnect with "default" as database name
 stock Handle:SQL_DefConnect(String:error[], maxlength, bool:persistent=true)
 {
 	return SQL_Connect("default", persistent, error, maxlength);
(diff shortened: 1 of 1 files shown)


dvander:
Potential path to nuking: make these natives not work outside of level change/startup callbacks without a pref set.

psychonic:
wtf

splewis:
For the record, I believe I do need the lock in that plugin since it does non-threaded SQL_FastQuery calls after connecting (unfortunately hidden in a function called SQL_AddColumn), and later uses SQL_TQuery calls for getting user data.

psychonic:
> wtf

I should have clarified here. This was a stale branch of mine from a year or so ago that doesn't even build, let alone being ready for the masses. Considerations for all first-party plugins aren't even handled. I was surprised to see a PR made for it.

asherkin:
We don't merge other developer's PRs by policy, lets not start creating them either.

KyleSanderson:
@splewis it's only if you're actively mixing them. If you go blocking and then switch to threaded afterwards you don't need it. You just can't go back to blocking.

PR #411, opened 2015-09-16
Add shim for TerminateRound for  CS:GO

Drifter321:
This simply adjusts the reason to match real value to the include one and vice versa for values coming from plugins.

### DIFF
--- a/plugins/include/cstrike.inc
+++ b/plugins/include/cstrike.inc
@@ -51,8 +51,8 @@
 enum CSRoundEndReason
 {
 	CSRoundEnd_TargetBombed = 0,           /**< Target Successfully Bombed! */
-	CSRoundEnd_VIPEscaped,                 /**< The VIP has escaped! */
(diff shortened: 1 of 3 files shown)


plugins/include/cstrike.inc
```diff
@@ -70,6 +70,8 @@ enum CSRoundEndReason
 	// The below only exist on CS:GO
 	CSRoundEnd_TerroristsSurrender,        /**< Terrorists Surrender */
 	CSRoundEnd_CTSurrender,                /**< CTs Surrender */
+	CSRoundEnd_TerroristsPlaned,		   /**< Terrorists Planted the bomb */
```
KyleSanderson: I believe you're missing a 't'.

KyleSanderson:
:shipit: after the terrorists stop getting planed.

psychonic:
:goberserk:

psychonic:
For 1.7-dev too, plz.

PR #412, opened 2015-09-17
Remove references to non-mapupdated plugins.

dvander:
This hasn't been supported in forever. I think we killed the feature in 1.5, so the only way it could be accessed is via an extension side-loading plugins. Extensions shouldn't really be doing that.

### DIFF
--- a/core/logic/PluginSys.cpp
+++ b/core/logic/PluginSys.cpp
@@ -1003,7 +1003,7 @@ IPlugin *CPluginManager::LoadPlugin(const char *path, bool debug, PluginType typ
 	LoadRes res;
 
 	*wasloaded = false;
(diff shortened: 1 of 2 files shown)


core/logic/PluginSys.cpp
```diff
@@ -2364,24 +2347,14 @@ void CPluginManager::RefreshAll()
 
 	List<CPlugin *>::iterator iter;
```
KyleSanderson: should be pulled, or the auto in the for loop on L2351.

KyleSanderson:
I remember using PluginType_MapOnly with a super buggy plugin when I first started using SM as per the author's direction. Definitely not the way to go about doing it :wink: 

:shipit:

dvander:
Merged via command-line.

PR #413, opened 2015-09-17
Clean up the first pass of plugin loading.

dvander:
This is some pretty straight forward shuffling of code around. "_LoadPlugin" gets split into a bunch of helpers and it gets a lot cleaner. You no longer need to understand the CPlugin::m_status state to follow the loading process.

The only part of note is that plugin timestamps are weird. I don't understand why (1) they were set so far into initialization or (2) right before being unloaded. I preserved the latter oddity for now, but all the details are hidden in CPlugin now, and timestamps are set in CPlugin::CreatePlugin.


KyleSanderson:
:shipit:

dvander:
Merged via command-line.

PR #414, opened 2015-09-17
Refactor LoadOrRequireExtensions.

dvander:
This nightmare of a function is weird. It does two totally separate things based on how you call it. Let's split it out into two functions and just share the iterator.

As per usual, this is split into a few separate commits because it has some precursor refactorings.


KyleSanderson:
:shipit:

dvander:
Merged via command-line.

PR #415, opened 2015-09-17
Simplify error propagation in the loading process.

dvander:
The loading process is unnecessarily complicated by having multiple error channels (the plugin state, the plugin error message, and the outparam error message). Sometimes we have to sync the two.

Instead, this makes CPlugin::m_errormsg canonical. At the very end, if the caller wants to capture the message, we copy it from the plugin.


KyleSanderson:
:shipit:

dvander:
Merged via command-line.

PR #416, opened 2015-09-18
Stop building Source 1 Dota 2 build.

psychonic:
The Source 1 version of Dota 2 is no longer downloadable or supported by Valve.

I don't know if we're ready to rip out the special-cased code for it or not yet, but we can at least stop spending build time on it.


KyleSanderson:
:shipit:

PR #417, opened 2015-09-19
Refactor ForwardSys.

dvander:
This is purely a syntactic refactoring. ForwardSys is possibly the oldest file in SourceMod. Having been introduced on Nov 11, 2006, it's almost 10 years old. It looks like early C++ using C89 syntax and has barely changed since it was introduiced.

Two main points of interest here. First, is I switched it to AMTL. Second, ForwardSys had an ad-hoc mechanism for protecting lists during iteration. I moved that out into its own helper class, and switched all of ForwardSys's lists to use it.

Mutation-during-iteration is a pervasive problem in SourceMod that we tend to just ignore until someone reports a specific test case. The new `ReentrantList<T>` inherits from `ke::LinkedList<T>` and is designed to handle this problem for future cases. It provides a new iterator and overrides remove/erase methods, making sure that when the list is mutated, any existing iterators are fixed up.

### DIFF
--- a/core/logic/ForwardSys.cpp
+++ b/core/logic/ForwardSys.cpp
@@ -47,11 +47,10 @@ void CForwardManager::OnSourceModAllInitialized()
 
 IForward *CForwardManager::CreateForward(const char *name, ExecType et, unsigned int num_params, const ParamType *types, ...)
 {
-	CForward *fwd;
 	va_list ap;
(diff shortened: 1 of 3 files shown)


KyleSanderson:
:shipit:

dvander:
Merged via command-line.

PR #418, opened 2015-09-20
Disentangle CPlugin from CPluginManager.

dvander:
These two classes have a pretty incestuous relationship. This set of patches removes direct field access one by one, and then removes their "friend" declarations.


KyleSanderson:
:shipit:

dvander:
Merged via command-line.

PR #419, opened 2015-09-20
Fix reentrancy issues in CPluginManager.

dvander:
This switches CPluginManager's lists to ReentrantList, fixing a bunch of potential reentrancy issues. It also fixes an issue where CPluginIterators could hold onto dangling pointers, if a plugin is somehow removed while the iterator is live.

In the interest of simplicity, I removed the CPluginIterator cache and it now copies the plugin list locally. It also registers itself as a plugin listener so it can watch for unloads.


KyleSanderson:
:shipit:

dvander:
Merged via command-line.

PR #420, opened 2015-09-20
Clean up plugin state.

dvander:
Plugin status is really confusing, and we currently try to assign all kinds of meaning toward Plugin_Loaded/Running/Failed/Error/BadLoad/etc. Everywhere we do this, there's some edge case where we're wrong. I've made an attempt to simplify when each status is used:
 - "Running" now means "all is normal" and implies that `OnPluginStart` succeeded.
 - "Loaded" now means "the first pass succeeded and we're waiting for the second".
 - "Error" now means "this plugin is no longer running, but we may re-activate it if a dependency comes online".
 - "Failed" now means "this plugin may or may not have been running, but now it's definitely not, and never will again".

This set of patches also introduces a new `PluginState`, which simply tells us which conceptual list a plugin is a member of. Currently, these lists are "Unregistered" (none), "Registered" (in PluginSys), and "WaitingToUnload" (meaning it is waiting on a frame hook callback that will unload the plugin).

We no longer use the status to decide when to invoke paired callbacks. For example, OnPluginUnloaded depended on the status, which was wrong. We now explicitly track whether or not the plugin went through OnPluginLoaded to ensure that they're paired correctly.


KyleSanderson:
:shipit:

dvander:
Merged via command-line.

PR #423, opened 2015-09-27
New forward - AskPluginAutoLoad

PremyslTalich:
Ciao guys,

finaly I had some time to rewrite and fix my `AskPluginAutoLoad` forward. For those who don't remember my attempt, read bellow.

---

I added new forward function to allow/disallow plugins to auto load on certain maps.

``` SourcePawn
forward APLRes:AskPluginAutoLoad(const String:plugin[], const String:map[]);
```

This function is called for every plugin that is about to auto load (plugins count times - 1 called in one plugin), **not** late load. If one plugin returns failure the given plugin is prevented from loading.
`AskPluginAutoLoad` is always called before `AskPluginLoad`, so I fixed the issue when blocked plugin called `AskPluginLoad`.

I also made little [test plugin](https://gist.github.com/KissLick/2bf6d04dfbf944aac7e1#file-test-askpluginautoload-sp).

**Advantages against:**
- Unloading plugins in map config
  - Some plugins change convars, precache files etc. in OnMapStart - might cause big problems
- Putting plugins in disabled folder and loading them in map config
  - This might cause big dependencies problems (eg. Timer with whole web of dependecies)
  - The need to unload plugins when the map ends

I tested this and everything seems to work like a charm! :-) So, what do you guys think about this?

### DIFF
--- a/core/logic/PluginSys.cpp
+++ b/core/logic/PluginSys.cpp
@@ -957,15 +957,6 @@ LoadRes CPluginManager::LoadPlugin(CPlugin **aResult, const char *path, bool deb
 
 			AskPluginAutoLoad(pPlugin, gamehelpers->GetCurrentMap());
 
-			// List<String>::iterator s_iter;
-			// for (s_iter = m_pluginsToLoad.begin(); s_iter != m_pluginsToLoad.end(); s_iter++)
-			// {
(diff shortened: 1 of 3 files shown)


asherkin:
Plugin loading is currently being rewritten (PRs 412 through 421, pending more in the near future), and this will almost certainly conflict.

PR #424, opened 2015-09-29
Promote git revision number to 4th version string component

asherkin:
With rolling releases it makes sense to promote this to being part of the version number proper (and drop the "-dev" tag).

Went with a 4th component rather than replacing the 3rd as it allows us to increment the 3rd component when people absolutely must update (primarily security issues I imagine) - I don't feel strongly either way. 99% of builds will be of the form 1.x.0.y.

Only the internal version number has changed, packaging has been left alone due to all the interaction with other tools.

For the current 1.8 release, this changes the version string from "1.8.0-dev+5699" to "1.8.0.5699".


KyleSanderson:
:shipit:

asherkin:
Pull requests that have a failing status can’t be merged from a phone.

God damn it GitHub 👿

Bara20:
Switch to Desktop Version?

asherkin:
It really wasn't that urgent.

PR #425, opened 2015-09-30
Fixed DBResultSet.IsFieldNull

WildCard65:
DBResultSet.IsFieldNull has an extra parameter that shouldn't be there(That parameter is the one that should be the automatically inserted first "this" param)
Also exists in SM 1.7


powerlord:
I have some code I'm writing now that may use this, so it'd be nice if this could be merged in, particularly since (as far as I can tell) the C++ is already expecting the post-change arguments.

PR #436, opened 2015-10-30
Fixed singles timer being deleted if plugin is paused (Bug 6338)

WildCard65:
This fixes plugin timers attempting to be called(If not repeating) if a plugin is being called, but get deleted when it fails.

### DIFF
--- a/core/TimerSys.cpp
+++ b/core/TimerSys.cpp
@@ -253,12 +253,12 @@ void TimerSystem::RunFrame()
 		pTimer = (*iter);
 		if (curtime >= pTimer->m_ToExec)
 		{
-            // Bug 6338: Timers being deleted if a plugin is paused.
-            if (pTimer->m_Listener->IsPaused(pTimer, pTimer->m_pData))
-            {
-                pTimer->m_ToExec = curtime + pTimer->m_Interval;
-                continue;
-            }
+			// Bug 6338: Timers being deleted if a plugin is paused.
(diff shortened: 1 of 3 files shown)


dvander:
Plugin pausing is going to be removed.

KyleSanderson:
> TimerInfo *pInfo = reinterpret_cast<TimerInfo *>(pData);

This is __completely__ illegal and will crash on invocation regardless of if the plugin is paused when an admin has a mismatched password. pTimer is also unused here, mistakes are fine, however these are completely incompatible structures. Extensions may use this interface as well.

Have you tested this?

WildCard65:
The reinterpret_cast line was copy + pasted from a few lines above.
pTimer is unused like it is in the rest of TimerNatives,
The extensions being able to use the new virtual method is unavoidable.

dvander:
Any third-party extension that uses ITimerSys will crash as soon as IsPaused() is called, since this does not have any version checks. Unfortunately adding a check here is pretty difficult since ITimedEvent is not versioned like our more modern callback classes. You would need to do something like what we just did for IPluginsListener: rename the old class, add a new one, and wrap the old one internally, all while preserving the ABI.

But it's moot, since ditching pausing has been in the works for a while and is coming to a finish, and extension API bumps on the stable branch are not a great idea. If you need a fix for 1.7, I would attempt to localize it to smn_timers.cpp by testing if the callback is runnable, and re-scheduling the timer there.

After pausing removal is finished, plugins that would have been paused will simply be unloaded, so any timers that would have executed will just never fire.

PR #437, opened 2015-11-01
Fix race between plugin unload and asynchronous query completion.

dvander:
IPluginsListener::OnPluginUnloaded API runs *after* plugins receive OnPluginEnd(). Since that's supposed to be the last callback in the plugin, it is an unspoken rule that extensions and listeners should not call into plugins at this stage.

Database.cpp of course breaks this rule: it can callback into plugins that think they're finished, which can cause unexpected state in the plugin and within SourceMod. This is particularly dangerous if a callback does something strange like adds a new ConVar: state that was removed in a different OnPluginUnloaded callback could now be revived, leaving leaks or danging pointers around SourceMod.

The easy fix for this, of course, is to just add a callback that fires *before* OnPluginEnd(). (Arguably, a partial fix is that OnPluginDestroyed is a better callback for releasing resources, but that ship has sailed because of extensions.)

Unfortunately we can't add new callbacks to IPluginsListener because it's not versioned. To work around this, I renamed IPluginsListener to IPluginsListener_V1 and Add/RemovePluginsListener to have a V1 as well. This should retain ABI compatibility, and whenever extensions recompile they'll seamlessly opt into the new listener.

I didn't take the final step of simply disallowing plugin callbacks within OnPluginUnloaded, but it's worth considering for the future.

### DIFF
--- a/core/logic/Database.cpp
+++ b/core/logic/Database.cpp
@@ -693,7 +693,7 @@ void DBManager::OnSourceModIdentityDropped(IdentityToken_t *pToken)
 	s_pAddBlock = NULL;
 }
 
-void DBManager::OnPluginUnloaded(IPlugin *plugin)
+void DBManager::OnPluginWillUnload(IPlugin *plugin)
 {
 	/* Kill the thread so we can flush everything into the think queue... */
 	KillWorkerThread();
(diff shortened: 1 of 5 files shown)


KyleSanderson:
:shipit:

PR #438, opened 2015-11-02
Remove plugin pausing.

dvander:
This is still a prototype, unfinished and untested. But it's nearing completion. This patch introduces a new "eviction" phase, which happens whenever we want a plugin to entered a paused state. Eviction works like this:
1. First, invoke callbacks related to dependency tracking in plugins. We do this for each dependent plugin, recursively, until the entire graph has been reached.
2. Pause each plugin in the dependency graph.
3. Iterate steps 1-2 until there are no changes in the dependency graph.
4. Break dependency links - mostly by removing and zapping fake natives, and clearing dependency lists.
5. Schedule an "eviction" event. This may happen immediately if we deem it safe to perform.
6. The eviction event will perform all plugin unload steps up to, but not including, removing it from the plugin list.

Pause change callbacks have been removed from both the C++ and SourcePawn API. ForwardSys was the only internal consumer, and it now checks the runtime pause status instead. It may still be possible to re-enter plugin code in between the dependency check and eviction event, but chances are low. Most likely is the caller will get a runtime error (as would be the case with Database.cpp, which will now check for pauses).

What's not done: automatic refreshing of plugins whose dependencies were removed and reloaded. I believe the correct way to approach this is different from how it works now: we should keep track of all the plugin filenames that it depended on, and only reload if all of them are present and operational.

### DIFF
--- a/core/logic/PluginGraph.cpp
+++ b/core/logic/PluginGraph.cpp
@@ -1,0 +1,153 @@
+// vim: set ts=4 sw=4 tw=99 noet :
+// =============================================================================
+// SourceMod
+// Copyright (C) 2004-2008 AlliedModders LLC.  All rights reserved.
+// =============================================================================
+//
+// This program is free software; you can redistribute it and/or modify it under
+// the terms of the GNU General Public License, version 3.0, as published by the
+// Free Software Foundation.
+//
+// This program is distributed in the hope that it will be useful, but WITHOUT
+// ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+// FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+// details.
+//
+// You should have received a copy of the GNU General Public License along with
+// this program.  If not, see <http://www.gnu.org/licenses/>.
+//
+// As a special exception, AlliedModders LLC gives you permission to link the
+// code of this program (as well as its derivative works) to "Half-Life 2," the
+// "Source Engine," the "SourcePawn JIT," and any Game MODs that run on software
+// by the Valve Corporation.  You must obey the GNU General Public License in
+// all respects for all other code used.  Additionally, AlliedModders LLC grants
+// this exception to all derivative works.  AlliedModders LLC defines further
+// exceptions, found in LICENSE.txt (as of this writing, version JULY-31-2007),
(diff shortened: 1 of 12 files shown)


core/logic/PluginGraph.cpp
```diff
@@ -0,0 +1,153 @@
...
+PluginGraph::~PluginGraph()
+{
+}
+
+bool
+PluginGraph::ComputeDependencies()
+{
+	// Reset the work queue state.
+	in_worklist_.clear();
+	assert(work_queue_.empty());
```
peace-maker: This assert seems to be in conflict with the loop condition below?

core/logic/PluginGraph.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+void
+PluginEvictionGraph::DoEviction()
+{
+	// Compute the dependency graph, then evict plugins. We iterate to a
+	// fixpoint since firing dependency callbacks could (in extremely rare
+	// cases) cause the dependency graph to change - for example via more
+	// eviction callbacks firing or by new plugins loading.
+	while (ComputeDependencies()) {
+		for (size_t i = change_list_.length() - 1; i < change_list_.length(); i--)
```
peace-maker: i >= 0?

core/logic/PluginGraph.cpp
```diff
@@ -0,0 +1,153 @@
...
+			FireDependencyCallbacks(change_list_[i]);
+	}
+
+	for (size_t i = full_list_.length() - 1; i < full_list_.length(); i--)
```
peace-maker: i >= 0?

dvander:
i >= 0 is always true for unsigned types.

Anyway, this patch probably needs to be rewritten, I'm not that confident in the approach yet.

KyleSanderson:
@dvander man this one is massively difficult. Jannik did a quick r? in 2016 but it looks like this all has gone stale. I know ultimately we all want this stuff to go away as it's a massive thorn in every ecosystem - is this still the right approach?

peace-maker:
I'll close this for now, since the implementation is stale and you mentioning the intention of a rewrite if you ever pick it up again @dvander. No need to pollute the pull requests.

PR #439, opened 2015-11-03
Deprecate FindSendPropOffs and FindDataMapOffs

asherkin:
These have had sane alternatives for ages now that still cover any use case they were actually useful for (via the optional local_offset param) and are consistently misused by people that really want the computed offset.

Also toss in a HasEntProp stock to cover the most common use case for all 4 of these functions.

Actually useful diff:
```
diff --git a/plugins/include/entity.inc b/plugins/include/entity.inc
index a1603b1..7592a96 100644
--- a/plugins/include/entity.inc
+++ b/plugins/include/entity.inc
@@ -403,17 +403,13 @@ native SetEntDataString(entity, offset, const String:buffer[], maxlen, bool:chan
  * @param prop			Property name.
  * @return				An offset, or -1 on failure.
  */
+#pragma deprecated Use FindSendPropInfo() instead. The local_offset if you actually want the local offset.
 native FindSendPropOffs(const String:cls[], const String:prop[]);
 
 /**
  * Given a ServerClass name, finds a networkable send property offset.
  * This information is cached for future calls.
  *
- * Note: This function will correctly compute nested offsets, unlike 
- * FindSendPropOffs().  YOU SHOULD NOT use this function to self-compute 
- * nested offsets.  For example, it is okay to add indexes for arrays, 
- * but not to add DT_LocalPlayer to m_nDisguiseClass.
- *
  * @param cls			Classname.
  * @param prop			Property name.
  * @param type			Optional parameter to store the type.
@@ -445,6 +441,7 @@ native FindSendPropInfo(const String:cls[],
  *						divisible by 8 (including 0 if unknown).
  * @return				An offset, or -1 on failure.
  */
+#pragma deprecated Use FindDataMapInfo() instead. The local_offset if you actually want the local offset.
 native FindDataMapOffs(entity,
 					   const String:prop[],
 					   &PropFieldType:type=PropFieldType:0,
@@ -482,21 +479,49 @@ native FindDataMapInfo(entity,
  */
 stock GetEntSendPropOffs(ent, const String:prop[], bool:actual=false)
 {
-	decl String:cls[64];
+	new String:cls[64];
 
 	if (!GetEntityNetClass(ent, cls, sizeof(cls)))
 	{
 		return -1;
 	}
 
+	new local;
+	new offset = FindSendPropInfo(cls, prop, _, _, local);
+
 	if (actual)
 	{
-		return FindSendPropInfo(cls, prop);
+		return offset;
+	} else {
+		return local;
 	}
-	else
+}
+
+/**
+ * Checks if an entity property exists on an entity.
+ *
+ * @param entity		Entity/edict index.
+ * @param type			Property type.
+ * @param prop			Property name.
+ * @return				Whether the property exists on the entity.
+ * @error				Invalid entity.
+ */
+stock bool:HasEntProp(entity, PropType:type, const String:prop[])
 {
-		return FindSendPropOffs(cls, prop);
+	if (type == Prop_Data) {
+		return (FindDataMapInfo(entity, prop) != -1);
+	}
+
+	if (type != Prop_Send) {
+		return false;
+	}
+
+	new String:cls[64];
+	if (!GetEntityNetClass(entity, cls, sizeof(cls))) {
+		return false;
 	}
+
+	return (FindSendPropInfo(cls, prop) != -1);
 }
 
 /**
```

### DIFF
--- a/plugins/include/entity.inc
+++ b/plugins/include/entity.inc
@@ -9,7 +9,7 @@
  * This program is free software; you can redistribute it and/or modify it under
  * the terms of the GNU General Public License, version 3.0, as published by the
  * Free Software Foundation.
- * 
+ *
  * This program is distributed in the hope that it will be useful, but WITHOUT
  * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
  * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
@@ -29,7 +29,7 @@
  *
  * Version: $Id$
  */
- 
+
 #if defined _entity_included
  #endinput
 #endif
@@ -67,9 +67,9 @@ enum PropFieldType
 	PropField_Entity,			/**< Valid for Data fields only (SendProp shows as int) */
 	PropField_Vector,			/**< Valid for SendProp and Data fields */
 	PropField_String,			/**< Valid for SendProp and Data fields */
-	PropField_String_T,			/**< Valid for Data fields.  Read only.  
-									 Note that the size of a string_t is dynamic, and 
-									 thus FindDataMapOffs() will return the constant size 
+	PropField_String_T,			/**< Valid for Data fields.  Read only.
+									 Note that the size of a string_t is dynamic, and
+									 thus FindDataMapOffs() will return the constant size
 									 of the string_t container (which is 32 bits right now).
 								 */
 };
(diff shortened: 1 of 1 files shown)


KyleSanderson:
>Also toss in a HasEntProp stock to cover the most common use case for all 4 of these functions.

This is kind of paper-esk, but it's still a good idea to have in-tree. I wonder if we should name this to be a little more similar to the rest of the GetEntProp* natives?

:shipit: based on the actual diff.

PR #440, opened 2015-11-05
Stop ignoring some warnings that we really shouldn't be.

asherkin:
alliedmodders/sourcepawn#50 needs to land before this.

@psychonic


psychonic:
I'm not crazy about -Wswitch. Could we change it to just warn?

Looks like there are some other things that need to be tweaked anyway to fix the build.

psychonic:
I just rebased this against master. Still not a fan of Wswitch, but if it compiles, 🚢

Headline:
Just waiting on alliedmodders/metamod-source#52 and then we should *hopefully* be good to go

core/logic/smn_core.cpp
```diff
@@ -951,7 +951,7 @@ static cell_t LogStackTrace(IPluginContext *pContext, const cell_t *params)
 
 	g_Logger.LogError("[SM] Stack trace requested: %s", buffer);
 	g_Logger.LogError("[SM] Called from: %s", pPlugin->GetFilename());
-	for (size_t i = 0; i < arr.length(); i)
+	for (size_t i = 0; i < arr.length(); ++i)
```
asherkin: Well that is a suspicious bug to be fixed here, might be best spinning that out to its own PR as it is a behaviour change?

extensions/sdktools/clientnatives.h
```diff
@@ -1,37 +0,0 @@
-/**
```
asherkin: What happened here?

pushbuild.txt
```diff
@@ -1,75 +1,76 @@
...
+Beeeeeeeees.
+Beads.
+snakesssssssss; maple
+R.I.P. Wade Boggs
+Your tier1 tower is under attack..
+Christmas!!!
+the mousse angles are smooth
+spoilers
+oh key you're so good to me
+well, ok rüd
```
KyleSanderson: god damn it... GH merge conflict rewrote all the line endings... :|

extensions/sqlite/AMBuilder
```diff
@@ -19,6 +19,7 @@ for arch in SM.archs:
   ]
   if builder.target.platform == 'linux':
     binary.compiler.postlink += ['-ldl', '-lpthread']
+    binary.compiler.cflags += ['-Wno-error=unused-const-variable']
```
KyleSanderson: This is not an option on gcc-4.9

KyleSanderson: Correction, GCC < 6 doesn't have this option.

Headline: Should just require a check for applicable versions I suppose, I presume this warning didn’t exist on gcc before then?

KyleSanderson: That's right, check out Travis-CI for compiler compliance.

peace-maker:
This requires permissive warnings due to hl2sdk code, so it's not as effective as it could be without touching all hl2sdk branches. Even though this is still desired, someone could take a new chance at this using newer compilers.

PR #441, opened 2015-11-06
Update GetMapDisplayName to use platform separator on CSGO

powerlord:
On CSGO, workshop maps use the path separator.  GetMapDisplayName didn't take this into account.

TF2 always uses `/`

While I was at it, I converted ke::SafeSprintf calls to ke::SafeStrcpy for items that just used the %s pattern.

This has been tested and appears to be working correctly in the following environments: TF2 Linux, CSGO Linux, TF2 Windows, and CSGO Windows.

I also fixed a few issues in the testsuite/findmap.sp as it was still using the old FindMap syntax and didn't quite have a long enough resultName.


asherkin:
Thanks!

PR #442, opened 2015-11-07
Use present netchannel instead of cached netchannel in CHookManager::SendFile.

KyleSanderson:
This corrects a potential oops if a file is being sent to another client when a packet is received from a different client. Last minute rejigging caused this unfortunately.


psychonic:
lgtm.

m_pActiveNetChannel should probably have a more specific name though if it's only relevant for some of the hooks.

PR #443, opened 2015-11-09
Fix crash on failing to load dependent extension.

asherkin:
This one is kind of nasty - it still leaves the failed ext in the libs, but is the sanest way to fix the crash without actually implementing `CExtensionManager::AddDependency` properly.

ShareSys has a couple of things that are designed to cleanup automatically on unload (rather than having a deregistration function that the extension is meant to call on unload) - this was not being done for autoloaded dependencies that failed to load, which could result in bad interface ptrs being left around. Instead, clean up immediately whenever any ext fails to load rather than waiting until unload time.

AFAICT, there is no downside to this change.


powerlord:
I think I've seen this kind of crash before when debugging DHooks.  I *think* you can work around it using QueryInterfaceDrop / NotifyInterfaceDrop, but a real fix is definitely a good thing.

psychonic:
lgtm

-- 
Nicholas Hastings

PR #444, opened 2015-11-10
Pre-fill buffer for GetClientAuth*

asherkin:
While we could use blank which will solve the worst of the problems without the perf hit, I think it is worth directly addressing this. Hopefully respecting the STEAM_ID_PENDING / STEAM_ID_BOT format will play nice with other code.

@psychonic


psychonic:
:boat:

PR #445, opened 2015-11-11
Expose experimental console debugger

peace-maker:
Exposes the experimental console debugger in sourcepawn. alliedmodders/sourcepawn#51 has to land and the submodule reference updated for this to compile.

Adds a `"EnableConsoleDebugger"` config option to core.cfg (defaulting to `"no"`) which enables a local debugging shell to step through plugins during execution. This isn't documented in the core.cfg for now, to keep non-developers from switching it on by accident ;) This should be considered alpha quality and we'll see how well it performs when developers start using it to debug their plugins.

It adds to the `sm` root console menu:

```
sm debug
SourceMod Debug Menu:
    start            - Start debugging a plugin
    next             - Start debugging the plugin which is loaded next
    bp               - Handle breakpoints in a plugin

sm debug start
[SM] Usage: sm debug start <#|file>

sm debug bp
[SM] Usage: sm debug bp <#|file> <option>
    list             - List breakpoints
    add              - Add a breakpoint
    clear            - Remove a breakpoint

sm debug bp plugin add
[SM] Usage: sm debug bp <#|file> add <file:line | file:function>
```

When `"EnableConsoleDebugger"` is set to `"yes"` the above menu is available to start debugging a plugin. Whenever it hits an error/exception or set breakpoint execution is stopped and a debugging shell is offered to examine and change plugin state, step through the code and list debugging information.

While the internal SourcePawn vm's watchdog is taken care of to handle debugging pauses, the watchdog in srcds, which monitors frame execution time, keeps running and terminates the process, if we spent too much time in the debugging shell. So `-nowatchdog` in the server startup line is required for games which include the watchdog. Maybe there's a way to disable it at runtime or at least warn if it's enabled?

Workflow demonstration

### DIFF
--- a/core/logic/DebugReporter.cpp
+++ b/core/logic/DebugReporter.cpp
@@ -36,8 +36,21 @@
 #include <am-string.h>
 #include <am-vector.h>
 #include "PluginSys.h"
+#include "common_logic.h"
+#include <bridge/include/CoreProvider.h>
 
 DebugReport g_DbgReporter;
+
+void DebugReport::OnSourceModStartup(bool late)
+{
+	if (!g_pConsoleDebugger)
+		return;
+
+	const char *enabled = bridge->GetCoreConfigValue("EnableConsoleDebugger");
+	if (enabled != nullptr && !strcasecmp(enabled, "yes")) {
+		g_pConsoleDebugger->SetEnabled(true);
+	}
+}
 
 void DebugReport::OnSourceModAllInitialized()
 {
@@ -238,6 +251,11 @@ void DebugReport::ReportError(const IErrorReport &report, IFrameIterator &iter)
 
 void DebugReport::OnRootConsoleCommand(const char *cmdname, const ICommandArgs *command)
 {
+	if (!g_pConsoleDebugger || !g_pConsoleDebugger->IsEnabled()) {
+		rootmenu->ConsolePrint("[SM] Console debugger not available.");
+		return;
+	}
+
 	int argcount = command->ArgC();
 	if (argcount >= 3) {
 		const char *cmd = command->Arg(2);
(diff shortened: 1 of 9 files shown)


50Wliu:
The picture/gif/whatever doesn't seem to have been uploaded to Github.

core/logic/DebugReporter.cpp
```diff
@@ -198,3 +220,145 @@ void DebugReport::ReportError(const IErrorReport &report, IFrameIterator &iter)
...
+			if (argcount < 4) {
+				rootmenu->ConsolePrint("[SM] Usage: sm debug start <#|file>");
+				return;
+			}
+
+			const char *arg = command->Arg(3);
+			CPlugin *pl = (CPlugin *)g_PluginSys.FindPluginByConsoleArg(arg);
+
+			if (!pl) {
+				rootmenu->ConsolePrint("[SM] Plugin %s is not loaded.", arg);
```
WildCard65: I think a better command error message should be along the lines of, "[SM] Plugin %s wasn't found."

core/logic/DebugReporter.cpp
```diff
@@ -198,3 +220,145 @@ void DebugReport::ReportError(const IErrorReport &report, IFrameIterator &iter)
...
+					IBreakpoint *bp;
+					for (unsigned int i = 0; i < breakpoints->length(); i++) {
+						bp = breakpoints->at(i);
+						ke::SafeSprintf(line, sizeof(line), "[SM] %2d  ", i + 1);
+
+						if (bp->line() > 0)
+							ke::SafeSprintf(line, sizeof(line), "%sline: %d", line, bp->line());
+
+						if (bp->temporary())
+							ke::SafeSprintf(line, sizeof(line), "%s  (TEMP)", line);
```
KyleSanderson: >>> http://pubs.opengroup.org/onlinepubs/9699919799/functions/vfprintf.html
>> http://pubs.opengroup.org/onlinepubs/9699919799/functions/snprintf.html
> If copying takes place between objects that overlap as a result of a call to sprintf() or snprintf(), the results are undefined.

peace-maker:
Instead of integrating this straight into core I'll try to get it working as a thirdparty extension.

PR #446, opened 2015-11-11
Fix custom menu sounds not working on newer games (bug 6433).

psychonic:
This hasn't been tested yet, but will be before landing.

While we support configuring menu sounds in core.cfg, it has always been up to the user to add the sounds to the downloads stringtable if there were custom sounds, so we arguably never "supported" these. However, even doing that, they do not work on newer games that have more aggressive precaching clientside.

The only way we've found to play custom sounds on any game since Left 4 Dead is to add the '*' sound char before the filename, forcing the game to stream it from disk directly instead of starting from the cache. This adds that for those games (whether or not the sound is custom). Note that people cannot just put it directly into the config with the asterisk as then it would attempt to call PrecacheSound with that added, which will fail due to that being treated as part of the filename.

### DIFF
--- a/core/MenuManager.cpp
+++ b/core/MenuManager.cpp
@@ -720,10 +720,8 @@ ConfigResult MenuManager::OnSourceModConfigChanged(const char *key,
 	return ConfigResult_Ignore;
 }
 
-const char *MenuManager::GetMenuSound(ItemSelection sel)
+bool MenuManager::GetMenuSound(ItemSelection sel, char *pszOut, size_t maxlen)
 {
-	const char *sound = NULL;
-
 	switch (sel)
 	{
 	case ItemSel_Back:
@@ -732,7 +730,12 @@ const char *MenuManager::GetMenuSound(ItemSelection sel)
 		{
 			if (m_SelectSound.size() > 0)
 			{
-				sound = m_SelectSound.c_str();
+#if SOURCE_ENGINE >= SE_LEFT4DEAD
+				ke::SafeSprintf(pszOut, maxlen, "*%s", m_SelectSound.c_str());
+#else
+				ke::SafeStrcpy(pszOut, maxlen, m_SelectSound.c_str());
+#endif
+				return true;
 			}
 			break;
 		}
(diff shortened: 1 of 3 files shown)


powerlord:
Last I heard, the * trick only works if the sounds are MP3s.

dlonagh:
Yeah. It's still better than not being able to use custom menu sounds entirely.

asherkin:
> Yeah. It's still better than not being able to use custom menu sounds entirely.

If that is indeed an issue, then this change will break the default sounds (which are WAVs) - so hardly an improvement :stuck_out_tongue:

psychonic:
I think it's unclear whether that breaks stock wavs, or just only doesn't work for custom ones.

Regardless, it just needs to be tested.

dlonagh:
Stock sounds won't be broken because they are from one of those game's .vpk files.
I've tested using wav files from hl2, css, and csgo and they work fine.
Custom .wav files are ones that don't work.

psychonic:
Unless you tested with this patch applied, it's irrelevant. We already know that they work without.

dlonagh:
I've just tested with a build I made from this branch (custom-menu-sound-fix).
I'm confirming that it works on both .wav (part of the game) and custom .mp3.

However, it no longer precaches them (for both default and custom files) so I had to create a plugin that adds them to the download table and precaches them.

test

psychonic:
Precaching was not removed, and functionality to add menu sounds to the downloads table never existed.

dlonagh:
I know it never did. I was just adding it to the facts.

As for precaching, I don't know why the console said the .wav wasn't precached earlier, but I just tested again multiple times it appears to be working fine. I think I must have made some typos on the sound path during my initial tests.

Everything is good!

dlonagh:
I know it never did. I was just adding it to the facts.

As for precaching, it's no longer precaching. Here, have a look:
test

dlonagh:
Any updates on this?

psychonic:
I no longer wish to maintain this PR. I will leave the branch in case someone else wants to run with it.

dlonagh:
Shame. All it needed was precaching though.
Everything else worked fine...

PR #448, opened 2015-11-18
Update SP and use BlamePluginError in sdkhooks.

Fyren:
I didn't find anywhere else that should use BPE instead of TNE, but maybe I missed them.


KyleSanderson:
Style seems to be off the rails. Code lgtm.

PR #449, opened 2015-11-18
Allow setting multiple chat trigger characters

asherkin:
Special characters only, minus a couple of potentially-problematic characters.

Technically this is a breaking change - but anyone using multi-char triggers is probably crazy anyway and has been driven off by now.

Not tested yet, here for feedback, etc, etc.

### DIFF
--- a/core/ChatTriggers.cpp
+++ b/core/ChatTriggers.cpp
@@ -62,26 +62,26 @@ ChatTriggers::~ChatTriggers()
 	m_ArgSBackup = NULL;
 }
 
-void ChatTriggers::SetChatTriggers(bool silent, const char *value)
+void ChatTriggers::SetChatTrigger(ChatTriggerType type, const char *value)
 {
-	ke::AutoArray<char> filtered(new char[strlen(value) + 1]);
+	ke::AutoPtr<char[]> filtered(new char[strlen(value) + 1]);
 
 	const char *src = value;
-	char *dest = filtered;
+	char *dest = filtered.get();
 	char c;
 	while ((c = *src++) != '\0') {
 		if (c <= ' ' || c == '"' || c == '\'' || (c >= '0' && c <= '9') || c == ';' || (c >= 'A' && c <= 'Z') || c == '\\' || (c >= 'a' && c <= 'z') || c >= 0x7F) {
-			logger->LogError("Ignoring %s chat trigger character '%c', not in valid set: %s", (silent ? "silent" : "public"), c, "!#$%&()*+,-./:<=>?@[]^_`{|}~");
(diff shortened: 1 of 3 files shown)


KyleSanderson:
I like it

dlonagh:
This is amazing! I love it.

iNilo:
Q: Would it be upto the plugin to detect what character was used, or should SM provide it?

This way we can make multiple commands.

/beacon @ct 
//would show up normally.
*beacon @ct 
//code checks what character used to change behavior, like hide the command from chat.

Just thinking out loud here.

pls dont kill me

KyleSanderson:
@iNilo that's exactly why ShowActivity exists.

iNilo:
Im aware, I ment more something like:
GetCmdTrigger(char[] buffer, int maxlength)

iNilo:
Any progress on this ?

asherkin:
No, we don't have some secret real tracker where actual work happens. 

If there was anything new to see, it would be here.

core/ChatTriggers.cpp
```diff
@@ -61,20 +62,43 @@ ChatTriggers::~ChatTriggers()
 	m_ArgSBackup = NULL;
 }
 
-ConfigResult ChatTriggers::OnSourceModConfigChanged(const char *key, 
-													const char *value, 
+void ChatTriggers::SetChatTriggers(bool silent, const char *value)
```
psychonic: This bool *offends* me. Can we get a Public/Silent enum?

asherkin: We certainly can. I've been trying to get home for 4 hours so maybe
tomorrow.

Sent from my iPhone

On 23 Jun 2016, at 21:04, Nicholas Hastings <notifications@github.com>
wrote:

In core/ChatTriggers.cpp
<https://github.com/alliedmodders/sourcemod/pull/449#discussion_r68304800>:

> @@ -61,20 +62,43 @@ ChatTriggers::~ChatTriggers()
>  	m_ArgSBackup = NULL;
>  }
>
> -ConfigResult ChatTriggers::OnSourceModConfigChanged(const char *key,
> -													const char *value,
> +void ChatTriggers::SetChatTriggers(bool silent, const char *value)

This bool *offends* me. Can we get a Public/Silent enum?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/alliedmodders/sourcemod/pull/449/files/ea29b11e85d1bee47a6b71585226b4091f1c4bb2#r68304800>,
or mute the thread
<https://github.com/notifications/unsubscribe/AA00UEj2w0vcl6Tp5U62IIVhictbNMxJks5qOubMgaJpZM4Gkq9C>
.

PR #451, opened 2015-11-25
Use GetTeamColor() Function Instead

stickz:


asherkin:
> No description given.

psychonic:
This seems like change for the sake of change. If touching it at all, a real improvement would be to move the team color to gamedata as red/blue shouldn't be assumed for every game.

stickz:
Yeah, it's pretty much just an improvement for the sake of improving things. If the team color were to be moved to gamedata, only one function would need changed instead of three repetitive calls.

PR #452, opened 2015-11-25
Set Variables at Initialization

stickz:


psychonic:
This is, at best, making the code less readable, and at most, accomplishing nothing.

An improvement to the original here here would be to replace those magic numbers with meaningful symbols, as at a glance, it's impossible to tell what the code is doing anyway.

stickz:
The binary size of the plugin is less when compiled. And the initialization stage of the variable isn't wasted by having no initializer. Plus it's consistent with other blocks of code; which, use this better method for setting the value of a variable.

asherkin:
> The binary size of the plugin is less when compiled. And the initialization stage of the variable isn't wasted by having no initializer.

A ternary expression can not be a constant initializer (due to not being constant), it'll compile to identical bytecode.

stickz:
I ran it through the compiler and the size was one byte less.

PR #453, opened 2015-11-25
Add Translation Support to Vote Menu

stickz:


asherkin:
This is not the correct way to translate menu items - [these menus are already correctly translated](https://github.com/alliedmodders/sourcemod/blob/master/plugins/funvotes.sp#L170-L193).

PR #455, opened 2015-11-25
Converted NormalSHook to typeset and added old prototype.

psychonic:
New prototype was added in PR #247 

This fixes compilation of plugins that are using pre-1.8 sound hook prototypes.

### DIFF
--- a/plugins/include/sdktools_sound.inc
+++ b/plugins/include/sdktools_sound.inc
@@ -336,37 +336,45 @@ typedef AmbientSHook = function Action (
   float &delay
 );
 
-/**
- * Called when a sound is going to be emitted to one or more clients.
- * NOTICE: all params can be overwritten to modify the default behavior.
- *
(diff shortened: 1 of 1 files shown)


50Wliu:
:green_heart: Thanks so much for this!

Also, shouldn't `clients[64]` be `clients[MAXPLAYERS]` instead?

psychonic:
Unfortunately, no. Much older versions of SM only using an array of size 64. If we increase to 65 here, users compiling against the newer include and (very) old SM will run into an issue where that last index is not valid.

psychonic:
As @VoiDeD pointed out to me yesterday, that's actually probably fine (if it can be added as a third typeset member), since reading past the numClients was never supported anyway, and that wouldn't have been higher than 64 before (and the plugin wouldn't be able to write more than 64 with the old-sized array).

KyleSanderson:
:shipit:

PR #456, opened 2015-11-26
Add support for hooking CanBeAutobalanced to SDKHooks.

psychonic:
Not many games support this, but it is in the SDK (and thus likely in many mods). This is also already an often used feature in TF2, provided by an unsupported extension that breaks often (using a different method to hook).

I don't know offhand of other games/mods that support this, but it's trivial to add the offset when finding one.

### DIFF
--- a/extensions/sdkhooks/extension.cpp
+++ b/extensions/sdkhooks/extension.cpp
@@ -916,10 +916,6 @@ bool SDKHooks::Hook_LevelInit(char const *pMapName, char const *pMapEntities, ch
 bool SDKHooks::Hook_CanBeAutobalanced()
 {
 	CBaseEntity *pPlayer = META_IFACEPTR(CBaseEntity);
-	int entity = gamehelpers->EntityToBCompatRef(pPlayer);
-
-	bool origRet = SH_MCALL(pPlayer, CanBeAutobalanced)();
-	bool newRet = origRet;
 
 	CVTableHook vhook(pPlayer);
 	ke::Vector<CVTableList *> &vtablehooklist = g_HookList[SDKHook_CanBeAutobalanced];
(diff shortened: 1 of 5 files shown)


extensions/sdkhooks/extension.cpp
```diff
@@ -905,6 +913,46 @@ bool SDKHooks::Hook_LevelInit(char const *pMapName, char const *pMapEntities, ch
 /**
  * CBaseEntity Hook Handlers
  */
+bool SDKHooks::Hook_CanBeAutobalanced()
+{
+	CBaseEntity *pPlayer = META_IFACEPTR(CBaseEntity);
+	int entity = gamehelpers->EntityToBCompatRef(pPlayer);
+
+	bool origRet = SH_MCALL(pPlayer, CanBeAutobalanced)();
```
KyleSanderson: should be inside the hook before the pawn calls.

extensions/sdkhooks/extension.cpp
```diff
@@ -905,6 +913,46 @@ bool SDKHooks::Hook_LevelInit(char const *pMapName, char const *pMapEntities, ch
...
+			// (so if multiple plugins returning different answers,
+			//  the one(s) that changed it win)
+			if ((res != 0) != origRet)
+				newRet = !origRet;
+		}
+
+		break;
+	}
+
+	RETURN_META_VALUE(MRES_SUPERCEDE, newRet);
```
KyleSanderson: use ignore here.

extensions/sdkhooks/extension.cpp
```diff
@@ -905,6 +913,46 @@ bool SDKHooks::Hook_LevelInit(char const *pMapName, char const *pMapEntities, ch
...
+			callback->Execute(&res);
+
+			// Only update our new ret if different from original
+			// (so if multiple plugins returning different answers,
+			//  the one(s) that changed it win)
+			if ((res != 0) != origRet)
+				newRet = !origRet;
+		}
+
+		break;
```
KyleSanderson: if override takes precedence, RETURN_META_VALUE MRES_SUPERCEDE here instead.

KyleSanderson:
No one else is probably hooking these, nits/hook citizen issues aside it seems absolutely fine.

extensions/sdkhooks/extension.cpp
```diff
@@ -905,6 +913,46 @@ bool SDKHooks::Hook_LevelInit(char const *pMapName, char const *pMapEntities, ch
...
+		if (vhook != vtablehooklist[entry]->vtablehook)
+		{
+			continue;
+		}
+
+		ke::Vector<IPluginFunction *> callbackList;
+		PopulateCallbackList(vtablehooklist[entry]->hooks, callbackList, entity);
+		for (entry = 0; entry < callbackList.length(); ++entry)
+		{
+			cell_t res;
```
KyleSanderson: initialize res incase there's a plugin error with forwardsys (pretty sure every other hook does this).

psychonic:
@KyleSanderson Thanks for fixing this up.

PR #457, opened 2015-11-26
Add x64 support to IDA vtable dump script

GoD-Tony:
This should fix the vtable dump script to work with dota's x64 bins.


psychonic:
Thanks. Been using your old standalone version of this for a while.

PR #459, opened 2015-11-29
Add BlamePluginError native

WildCard65:
Exposes recently added BlamePluginError method in IPluginContext to
allow SM plugins to utilize it. Ref:
https://forums.alliedmods.net/showthread.php?p=2367275

Note: Unable to test this as I have a VS 2015 installation(Can't remember where I found 2013 installer, will look for it).

### DIFF
--- a/core/logic/smn_core.cpp
+++ b/core/logic/smn_core.cpp
@@ -419,7 +419,7 @@ static cell_t BlamePluginError(IPluginContext *pContext, const cell_t *params)
 {
 	if (!params[1])
 	{
-		return 0;
+		return pContext->ThrowNativeError("No handle passed.");
 	}
 	HandleError err;
 	IPlugin *pPlugin = scripts->FindPluginByHandle(params[1], &err);
@@ -430,19 +430,19 @@ static cell_t BlamePluginError(IPluginContext *pContext, const cell_t *params)
 	IPluginContext *plContext = pPlugin->GetBaseContext();
 	if (!plContext)
 	{
-		return 0;
+		return pContext->ThrowNativeError("Unable to get plugin context for %x", params[1]);
 	}
 	IPluginFunction *plFunction = plContext->GetFunctionById(params[2]);
 	if (!plFunction)
 	{
-		return 0;
+		return pContext->ThrowNativeError("Unable to plugin function %d", params[2]);
 	}
 	char buffer[1024];
 	char *fmt;
 	int arg = 4;
 	pContext->LocalToString(params[3], &fmt);
 	size_t res = atcprintf(buffer, sizeof(buffer) - 1, fmt, pContext, params, &arg);
-	buffer[res++] = '\n';
+	buffer[res++] = '\0';
 	return plContext->BlamePluginError(plFunction, buffer);
 }
 
--- a/core/logic/smn_core.cpp
+++ b/core/logic/smn_core.cpp
@@ -48,6 +48,7 @@
 #include <unistd.h>
 #include <sys/times.h>
 #endif
+#include "sprintf.h"
 #include <IForwardSys.h>
 #include <ILibrarySys.h>
 #include <bridge/include/CoreProvider.h>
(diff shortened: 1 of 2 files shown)


core/logic/smn_core.cpp
```diff
@@ -414,6 +415,37 @@ static cell_t SetFailState(IPluginContext *pContext, const cell_t *params)
...
+	if (!plFunction)
+	{
+		return 0;
+	}
+	char buffer[1024];
+	char *fmt;
+	int arg = 4;
+	pContext->LocalToString(params[3], &fmt);
+	size_t res = atcprintf(buffer, sizeof(buffer) - 1, fmt, pContext, params, &arg);
+	buffer[res++] = '\n';
```
peace-maker: You need to readd the null terminator here!

peace-maker:
As this is used to report errors, it shouldn't just |return 0| if something is wrong, but still throw an error.

core/logic/smn_core.cpp
WildCard65: I meant to leave it in and remove the new line character, guess I reversed the line to delete.

dvander:
What is the point of this? The plugin is already implied, and the C++ API is intended for top-level code. Plugins shouldn't be blaming other plugins.

plugins/include/sourcemod.inc
```diff
@@ -314,6 +314,19 @@ native Handle:FindPluginByNumber(order_num);
 native SetFailState(const String:string[], any:...);
 
 /**
+ * This blames another plugin for a function error.
+ * @Note: Currently doesn't allow INVALID_HANDLE to be passed for plugin param.
+ *
+ * @param plugin        Plugin owning function your blaming error for.
```
50Wliu: At the very least, you should know the difference between `your` and `you're`...

plugins/include/sourcemod.inc
```diff
@@ -314,6 +314,19 @@ native Handle:FindPluginByNumber(order_num);
...
+ * @Note: Currently doesn't allow INVALID_HANDLE to be passed for plugin param.
+ *
+ * @param plugin        Plugin owning function your blaming error for.
+ * @param blame         Method your blaming error for.
+ * @param string		Format specifier string.
+ * @param ...			Formatting arguments.
+ * @noreturn
+ * @error               Throws if plugin handle couldn't be read.
+ */
+native void BlamePluginError(Handle plugin, Function blame, const char[] format, any ...);
```
50Wliu: The documentation reads `string`, but the argument here is `format`.

WildCard65:
@dvander Unless I misunderstood what @asherkin meant, I'm just going by what I believe asherkin meant by his reply to thread I linked.

plugins/include/sourcemod.inc
WildCard65: Above point not scratched out is also fixed in https://github.com/WildCard65/sourcemod/commit/90c85c57e35c44e5446faa77bbbf6e9aba51162f

plugins/include/sourcemod.inc
```diff
@@ -314,6 +314,19 @@ native Handle:FindPluginByNumber(order_num);
 native SetFailState(const String:string[], any:...);
 
 /**
+ * This blames another plugin for a function error.
+ * @Note: Currently doesn't allow INVALID_HANDLE to be passed for plugin param.
+ *
+ * @param plugin        Plugin owning function you're blaming error for.
+ * @param blame         Method your blaming error for.
```
50Wliu: `you're`

Fyren:
BPE is intended raise an error that's caused by a plugin that's not on the stack.  This is not what the forum user is dealing with.  He should be using TNE.

I suppose it's possible that a plugin might want to show an error that doesn't involve a SP stack, but it'd be due to something like manually managing callbacks and not using forwardsys.

asherkin:
> I suppose it's possible that a plugin might want to show an error that doesn't involve a SP stack, but it'd be due to something like manually managing callbacks and not using forwardsys.

I believe that is what he is doing (having read his previous threads for what he's working on), his plugin is acting as an intermediary between a plugin using his API and threaded SQL queries. I gather that he is in a threaded query callback and wishes to blame the callback of the plugin that was provided. This is the case BPE was added for, but the guy still hasn't replied so this PR is massively jumping the gun.

```
OtherPlugin ---> HisPlugin ---> TQuery ---> HisPlugin -X-> OtherPlugin
                                         Error is here ^ There is no native to blame.
```

PR #460, opened 2015-12-11
syntax 1.7

wyd3x:
update few files to syntax 1.7

### DIFF
--- a/plugins/include/clientprefs.inc
+++ b/plugins/include/clientprefs.inc
@@ -95,7 +95,7 @@ enum CookieMenuAction
  *                      exists, a handle to it will still be returned.
  * @error				Cookie name is blank.
  */
-native Handle:RegClientCookie(const String:name[], const String:description[], CookieAccess:access);
+native Handle RegClientCookie(char[] name, char[] description, CookieAccess access);
 
 /**
  * Searches for a Client preference cookie.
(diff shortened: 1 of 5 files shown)


plugins/include/clientprefs.inc
```diff
@@ -95,7 +95,7 @@ enum CookieMenuAction
  *                      exists, a handle to it will still be returned.
  * @error				Cookie name is blank.
  */
-native Handle:RegClientCookie(const String:name[], const String:description[], CookieAccess:access);
+native Handle RegClientCookie(char[] name, char[] description, CookieAccess access);
```
asherkin: Missing `const`.

plugins/include/clientprefs.inc
```diff
@@ -106,7 +106,7 @@ native Handle:RegClientCookie(const String:name[], const String:description[], C
  * @param name			Name of cookie to find.
  * @return				A handle to the cookie if it is found. INVALID_HANDLE otherwise.
  */
-native Handle:FindClientCookie(const String:name[]);
+native Handle FindClientCookie(char[] name);
```
asherkin: Missing `const`.

asherkin:
Rather than adding far too many line notes: You've dropped the `const`s on all string params.

asherkin:
Replaced by #461.

PR #462, opened 2015-12-12
Update IDA vtable dump script

GoD-Tony:
A small fix to work with multiple versions of IDA x86 & x64


asherkin:
Thank you good sir.

PR #463, opened 2015-12-13
Simply and Fix !Map Command

stickz:
This PR fixes a frequently occurring bug with the previous map code, where the map would not change when called. It also removes redundant code which isn't really needed to complete the map change. And provides a near instant response time when the client types the command.

### DIFF
--- a/plugins/basecommands/map.sp
+++ b/plugins/basecommands/map.sp
@@ -87,36 +87,27 @@ public Action:Command_Map(client, args)
 		return Plugin_Handled;
 	}
 
-	decl String:map[64];
+	char map[64];
 	GetCmdArg(1, map, sizeof(map));
 
-	if (!IsMapValid(map))
+	if (IsMapValid(map))
+	{
+		ChangeMap(client, map);
+	}
+	
+	else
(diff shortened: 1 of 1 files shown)


50Wliu:
I think the timer was there on purpose, to give players a warning...

stickz:
A three second warning is redundant. And there was a bug within the data timer code which prevented the map from changing when the client typed the command.

splewis:
What exactly was the bug that you fixed? As far as I can tell, you just removed the delay that was inserted by design and it was working fine before.

KyleSanderson:
The 3 second warning is not redundant, and is indeed a feature.

stickz:
The bug was something related to the data timer which was causing the server not to change level when the !map command was called. But I guess this bug fix won't be added to SM.

asherkin:
Any failures to change the map were not related to the code you blindly removed.

stickz:
Yes they were related, I debugged it with printouts and the IsValidMap() check passed. And the "!rcon changelevel mapname" command worked in every circumstance "!map mapname" did not.

PR #464, opened 2015-12-21
Compile SQLite so it supports DELETE/ UPDATE limit

ClaudiuHKS:
http://www.sqlite.org/compile.html#enable_update_delete_limit

We need this, to make the SQL drivers more friendly each-other.
I know that `unix_timestamp(now())` **MySQL** is `julianday('now')` **SQLite**, but for the basic update/ delete syntax... Updating/ deleting limitation must work on SQLite as well.

### DIFF
--- a/extensions/sqlite/AMBuilder
+++ b/extensions/sqlite/AMBuilder
@@ -12,7 +12,8 @@ elif binary.compiler.vendor == 'msvc':
 
 binary.compiler.defines += [
   'SQLITE_OMIT_LOAD_EXTENSION',
-  'SQLITE_THREADSAFE'
+  'SQLITE_THREADSAFE',
+  'SQLITE_ENABLE_UPDATE_DELETE_LIMIT'
 ]
 if builder.target_platform == 'linux':
   binary.compiler.postlink += ['-ldl', '-lpthread']


psychonic:
I'm not quite sure why this feature would be useful, and you haven't noted why this is desired yourself.

Regardless, the change made here will do absolutely nothing. From the link you gave, 
> Because of this, this option may only be used when the library is built from source, not from the amalgamation or from the collection of pre-packaged C files provided for non-Unix like platforms on the website. 

We do build from the amalgamation, which wasn't touched in the commit, and I believe would need to be regenerated if I'm reading correctly.

ClaudiuHKS:
Code from the amalgamation

**#if defined(SQLITE_ENABLE_UPDATE_DELETE_LIMIT) && !defined(SQLITE_OMIT_SUBQUERY)
SQLITE_PRIVATE Expr *sqlite3LimitWhere(Parse*,SrcList*,Expr*,ExprList*,Expr*,Expr*,char*);
#endif**

Fyren:
From the link that you provided yourself, the define is not sufficient by itself to make order by/limit work on update/delete.  You can see that the function does not get used.  parse.c needs to be regenerated and so the amalgamation itself would as well.

But, if it's not much work, I don't think making it happen needs to be justified beyond that where clauses are a pretty basic feature so parity there is good.

psychonic:
> But, if it's not much work, I don't think making it happen needs to be justified beyond that where clauses are a pretty basic feature so parity there is good.

ORDER and LIMIT on UPDATE/DELETE queries is not part of any SQL standard. While MySQL supports it, and SQLite doing so with the define, I don't think that we want to intentionally add support for non-standard SQL language features to what's meant to be an abstraction layer

KyleSanderson:
I really don't see the harm in adding the define in. Clientprefs for instance, which is native, already uses different queries for DBI. While the DBI layer is meant as an abstraction, it's to abstract data retrieval, not to limit the users ability to use raw SQL if a normal DB allows for it.

We already have SQL_SetCharset https://github.com/alliedmodders/sourcemod/blob/7fbcb81916476f442fe26f929aba37e1bfce5eeb/extensions/sqlite/driver/SqDatabase.cpp#L229 which is even more of a DBI specific action. I'm not advocating doubling down on a mistake, because that isn't one. I just cannot fathom the harm in enabling functionality if it gets a user further towards their goal with the product.

The code needs work and obvious testing as pointed out by Fyren and psychonic; which maybe means this should land in a bug. However, fundamentally, I don't see anything wrong with this.

:fried_shrimp:

ClaudiuHKS:
If you still wonder why this is very important, it is because for example you can get two entries into the database, which are cloned, let's say, by a programmer's mistake, and if you wanna delete the cloned entry, you could just do `delete from % where Steam = '%' limit totalEntries - 1;`.

psychonic:
Doing one-off queries to attempt to clean up poorly-implemented database schemas seems like a pretty poor use case.

PR #465, opened 2015-12-30
Allow plugins to suppress entity outputs (fix 5515).

klausenbusk:
https://bugs.alliedmods.net/show_bug.cgi?id=5515
It seems like this bug/patch has been forgotten, so I opened a pull request.
This is just Peace-Maker patch, where I updated the changes in the include file to the new syntax.

Kristian

### DIFF
--- a/plugins/include/sdktools_entoutput.inc
+++ b/plugins/include/sdktools_entoutput.inc
@@ -42,9 +42,15 @@
  * @param caller		Entity index of the caller.
  * @param activator		Entity index of the activator.
  * @param delay			Delay in seconds? before the event gets fired.
- * @noreturn
+ * @return					Anything other than Plugin_Continue will supress this event,
+ *								returning Plugin_Continue will allow it to propagate the results
+ *								of this output to any entity inputs.
  */
-typedef EntityOutput = function void (const char[] output, int caller, int activator, float delay);
+typeset EntityOutput
+{
+	function void (const char[] output, int caller, int activator, float delay);
+	function Action (const char[] output, int caller, int activator, float delay);
+};
 
 /**
  * Add an entity output hook on a entity classname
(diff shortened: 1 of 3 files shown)


klausenbusk:
Hmm. I did something wrong in the .inc file.
```
../../plugins/include/sdktools_entoutput.inc(52) : error 120: methodmap and class signatures must use new-style type declarations
```

plugins/include/sdktools_entoutput.inc
```diff
@@ -42,9 +42,15 @@
...
- * @noreturn
+ * @return					Anything other than Plugin_Continue will supress this event,
+ *								returning Plugin_Continue will allow it to propagate the results
+ *								of this output to any entity inputs.
  */
-typedef EntityOutput = function void (const char[] output, int caller, int activator, float delay);
+typeset EntityOutput
+{
+	function void (const char[] output, int caller, int activator, float delay);
+	function Action (const String:output[], int caller, int activator, float delay);
```
psychonic: This is the line causing your error. (String)

klausenbusk: Fixed (I hope).

klausenbusk:
What do you think? Do the code need any change?

psychonic:
Please don't bump Pull Requests. It's only been a few hours.

psychonic:
Thanks for helping push this forward.

klausenbusk:
Can I open a pr for 1.7?

psychonic:
PR's aren't necessary for cherry-picking changes to another branch unless large further changes are required to get it to merge due to structural changes between branches. I don't think that's the case here - we can do a normal cherry-pick when ready.

However, since it *just* landed, and since 1.8 is near stable anyway, I'm not sure that we'll opt to do so.

klausenbusk:
I see. I just wait for 1.8 then. I need it for a not mandatory plugin feature, I can code it some other way, but this is most correct way.

PR #467, opened 2016-01-05
Add missing |bool| return tag to RemoveFromTrie native

peace-maker:
Seems like it's been missing for years. The methodmap is fine.


psychonic:
Thanks!

PR #468, opened 2016-01-07
Update sourcepawn testing include to new syntax.

splewis:
Not sure if the args to AssertEq would be better off with a ``any:`` tag instead of as an int.

Also made them stocks so users don't get warnings about unused functions when then ``#include <testing>``.


dvander:
Looks good. AssertEq is supposed to only take ints, so that's fine.

PR #470, opened 2016-01-19
Update sample_ext's AMBuildScript

powerlord:
Basically, I've taken the recent changes to the AMBuildScript for SourceMod and applied them to the sample_ext AMBuildScript.

While I haven't tested everything completely on a complicated project, the sample_ext project builds just fine with it against all current SDKs available on Windows and Linux.

### DIFF
--- a/public/sample_ext/AMBuildScript
+++ b/public/sample_ext/AMBuildScript
@@ -12,7 +12,7 @@ class SDK(object):
     self.define = name
     self.platform = platform
     self.name = dir
-    self.path = None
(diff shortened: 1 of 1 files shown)


asherkin:
Thanks!

powerlord:
I've noticed that this doesn't work on SM 1.7.  Should I submit a new pull request to add

          os.path.join(self.sm_root, 'public', 'sourcepawn'),

back to it to make it work on 1.7 as well?

PR #473, opened 2016-02-03
Fix for crashes related to the GameRules_Set* natives.

akowald:
## The GameRules_Set* natives cause memory corruption when changeState is set to true.

This gave us a lot of grief with our stop that tank mod. Eventually a TF2 update made the crashes reliable enough that I was able to track down the root of the problem:

`GameRules_SetProp("m_bPlayingHybrid_CTF_CP", true, 4, 0, true);`
`*(bool *)((intptr_t)pProxy + offset) = (params[2] == 0) ? false : true;`

It makes no sense to directly set the value on CTFGameRulesProxy. Anytime the game makes a change to CTFGameRules, it calls CGameRulesProxy::NotifyNetworkStateChanged which then runs the equivalent of CBaseEdict::StateChanged(). It should be sufficient enough to mark the offset as changed so I left that portion unchanged. I've tested this and my client gets the change immediately.

I've looked at both the CTFGameRules and CTFGameRulesProxy pointers in memory and their class structure isn't the same. When running the `GameRules_SetProp`, it ends up corrupting memory that belongs to another entity.

### Recreate the crash
This is tricky due to the inherently random nature..but this process has been reliable on Windows (TF2):
1. Start the server on pl_badwater
2. Run the following command from a plugin: `GameRules_SetProp("m_bPlayingHybrid_CTF_CP", true, 4, 0, true);`
3. Change the map to ctf_turbine

It seems to frequently crash on round start as well. Last time I tested Linux was Tough Break and it crashed almost instantly. Since patching this, almost all our crashes have stopped. I think this would have been discovered sooner if the default value of changeState was not false. Thanks for reading!

### DIFF
--- a/extensions/sdktools/gamerulesnatives.cpp
+++ b/extensions/sdktools/gamerulesnatives.cpp
@@ -264,7 +264,6 @@ static cell_t GameRules_SetProp(IPluginContext *pContext, const cell_t *params)
 		*(int32_t *)((intptr_t)pGameRules + offset) = params[2];
 		if (sendChange)
 		{
-			*(int32_t *)((intptr_t)pProxy + offset) = params[2];
 			gamehelpers->SetEdictStateChanged(gamehelpers->EdictOfIndex(gamehelpers->EntityToBCompatRef(pProxy)), offset);
 		}
 	}
@@ -273,7 +272,6 @@ static cell_t GameRules_SetProp(IPluginContext *pContext, const cell_t *params)
 		*(int16_t *)((intptr_t)pGameRules + offset) = (int16_t)params[2];
 		if (sendChange)
 		{
-			*(int16_t *)((intptr_t)pProxy + offset) = (int16_t)params[2];
 			gamehelpers->SetEdictStateChanged(gamehelpers->EdictOfIndex(gamehelpers->EntityToBCompatRef(pProxy)), offset);
 		}
 	}
@@ -282,7 +280,6 @@ static cell_t GameRules_SetProp(IPluginContext *pContext, const cell_t *params)
 		*(int8_t *)((intptr_t)pGameRules + offset) = (int8_t)params[2];
 		if (sendChange)
 		{
-			*(int8_t *)((intptr_t)pProxy + offset) = (int8_t)params[2];
 			gamehelpers->SetEdictStateChanged(gamehelpers->EdictOfIndex(gamehelpers->EntityToBCompatRef(pProxy)), offset);
 		}
 	}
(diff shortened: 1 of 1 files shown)


psychonic:
I'm okay with this in concept, and that's some nice debugging.

I'm a little curious to find where and how we were getting the wrong entity, but that's not important. I believe that you're correct that setting the values on the proxy is redundant, but I'd like to verify and will do so soon.

If we're ripping out the setting of values on the proxy, we might as well also rip out the finding and storing of the ptr to it, including the then-unused FindEntityByNetClass func.

> I think this would have been discovered sooner if the default value of changeState was not false.

You're probably right. I have no idea why I made the default false as the functions were meant to be analogous to their SetEntProp* equivalents, which don't even make it an option to not network the change immediately. We should fix that too (but not in this PR).

extensions/sdktools/gamerulesnatives.cpp
```diff
@@ -264,7 +264,6 @@ static cell_t GameRules_SetProp(IPluginContext *pContext, const cell_t *params)
 		*(int32_t *)((intptr_t)pGameRules + offset) = params[2];
 		if (sendChange)
 		{
-			*(int32_t *)((intptr_t)pProxy + offset) = params[2];
```
psychonic: I have no idea why the diff view is showing some of these extraneous braces being removed. Maybe it's a Github bug since if they were there, it would have even compiled before.

asherkin: Yes, the actual [patch](https://patch-diff.githubusercontent.com/raw/alliedmodders/sourcemod/pull/473.diff) appears fine.

akowald:
> If we're ripping out the setting of values on the proxy, we might as well also rip out the finding and storing of the ptr to it, including the then-unused FindEntityByNetClass func.

You still need a pointer to the proxy for `gamehelpers->SetEdictStateChanged(gamehelpers->EdictOfIndex(gamehelpers->EntityToBCompatRef(pProxy)), offset);` for changeState, correct? Granted, it's a little round-a-bout but it still checks for validity. If you want to rip out FindEntityByNetClass what do you want to use instead, I don't think that servertools has a method for net classes.

The pointer to CTFGameRulesProxy was valid, but add 2408 (offset for m_bPlayingHybrid_CTF_CP) and you end up in another entity.

extensions/sdktools/gamerulesnatives.cpp
akowald: If we're ripping out the setting of values on the proxy, we might as well also rip out the finding and storing of the ptr to it, including the then-unused FindEntityByNetClass func.
You still need a pointer to the proxy for gamehelpers->SetEdictStateChanged(gamehelpers->EdictOfIndex(gamehelpers->EntityToBCompatRef(pProxy)), offset); for changeState, correct? Granted, it's a little round-a-bout but it still checks for validity. If you want to rip out FindEntityByNetClass what do you want to use instead, I don't think that servertools has a method for net classes.

The pointer to CTFGameRulesProxy was valid, but add 2408 (offset for m_bPlayingHybrid_CTF_CP) and you end up in another entity.

psychonic:
This is all making sense to me now. The proxy class does not actually contain the gamerules variables. The gamerules datatable is still fully contained in the actual gamerules class.

Disregard what I said about ripping more out.

psychonic:
Thanks for both the PR and explanation.

PR #475, opened 2016-02-10
TK Manager plugin for Insurgency 2014

cych:
Hello, could you make this plugin :  https://forums.alliedmods.net/showthread.php?p=714326
compatible with insurgency 2014.

thanks.


WildCard65:
1) This is not an issue section.
2) This has nothing to do with SM in general.
3) Post request in the thread you linked.

PR #476, opened 2016-02-15
Implement an auto-escaping Format native for SQL query construction

asherkin:
Probably best to look at this one commit-by-commit.

Will need some doc work on the wiki after landing.
The gist of it is that if formatting via the new native, "%s", "%N", "%L" etc are automatically escaped unless a new "!" flag is set (i.e. "%!s"), and an error is thrown if the string is going to be truncated (to avoid leaving a trailing escape character).

Old:
```
char name[MAX_NAME_LENGTH];
if (!GetClientName(client, name, sizeof(name))) {
  return false;
}

int safeNameLen = (strlen(name) * 2) + 1;
char[] safeName = new char[safeNameLen];
db.Escape(name, safeName, safeNameLen);

char steamId[32];
if (!GetClientAuthId(client, AuthId_Steam2, steamId, sizeof(steamId))) {
  return false;
}

int safeSteamIdLen = (strlen(steamId) * 2) + 1;
char[] safeSteamId = new char[safeSteamIdLen];
db.Escape(steamId, safeSteamId, safeSteamIdLen);

char buffer[512];
Format(buffer, sizeof(buffer), "UPDATE players SET name = '%s' WHERE steamid = '%s'", safeName, safeSteamId);
db.Query(OnQueryComplete, buffer);
```

New:
```
char steamId[32];
if (!GetClientAuthId(client, AuthId_Steam2, steamid, sizeof(steamid))) {
  return false;
}

char buffer[512];
db.Format(buffer, sizeof(buffer), "UPDATE players SET name = '%N' WHERE steamid = '%s'", client, steamId);
db.Query(OnQueryComplete, buffer);
```

### DIFF
--- a/core/logic/smn_database.cpp
+++ b/core/logic/smn_database.cpp
@@ -32,6 +32,7 @@
 #include "common_logic.h"
 #include "Database.h"
 #include "ExtensionSys.h"
+#include "sprintf.h"
 #include "stringutil.h"
 #include "ISourceMod.h"
 #include "AutoHandleRooter.h"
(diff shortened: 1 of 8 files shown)


psychonic:
Looks okay to me.

I'd also be fine with ditching the non-methodmap version of the function to encourage use of the newer syntax.

PR #477, opened 2016-02-20
Fix compilation of sample extension

2m:
I needed to do these changes in order to compile the sample extension.


powerlord:
Actually, I'm surprised we kept the Makefile in the `sample_ext` extension as we're trying to encourage the use of the AMBuild2 system for extensions.

Which reminds me, I should make a Wiki page about that.

dvander:
I think we're keeping them around until AMB2 can generate Makefiles. Just so authors have something to work with if they don't want to bother with AMBuild.

PR #480, opened 2016-02-25
GetClientSerial does not have @error documentation

iNilo:
I encountered a 

>  Native "GetClientSerial" reported: Client index 0 is invalid

With the clients.inc not saying it could throw errors.


asherkin:
Thanks!

PR #481, opened 2016-02-27
Fix use-after-free crash in SQLite extension

peace-maker:
When the server crashed and the process got terminated, the SqDriver
instance was killed first (e.g. by atexit). SqDatabase tries to access
SqDriver in its destructor.
This patch tells SqDatabase to not use anything from SqDriver anymore
after SqDriver got destroyed.

Next to that, the clientprefs extension relied on the IDatabase pointer
being valid to get the driver pointer. Cache the pointer, so the dbi
system still knows the IDBThreadOperation belonged to the now gone
driver, even after the database object is gone.

### DIFF
--- a/extensions/clientprefs/query.cpp
+++ b/extensions/clientprefs/query.cpp
@@ -92,7 +92,7 @@ void TQueryOp::RunThreadPart()
 
 IDBDriver *TQueryOp::GetDriver()
 {
-	return m_database->GetDriver();
+	return m_driver;
 }
 
 IdentityToken_t *TQueryOp::GetOwner()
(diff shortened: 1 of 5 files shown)


dvander:
One thing doesn't make sense to me, which is why we'd want to cache an invalid pointer. If the driver has been shut down, extensions should have observed that via the dependency drop callback and released any relevant resources. Further, if the driver extension itself is being unloaded, there's no hope that the SqDatabase accessing the driver will work anyway, so any extensions not obeying this protocol are in trouble.

Why is atexit being called in the first place? Is this crash reproducible?

peace-maker:
Yes, it's not really the root of the problem.
https://github.com/alliedmodders/sourcemod/blob/master/core/logic/Database.cpp#L371
This would crash, if there are still queries from clientprefs in the queue when calling op->GetDriver().
Maybe we need to kill the worker thread without it processing left operations when the driver got removed?

It's from any crash within srcds when e.g. dereferencing an invalid pointer. This just obfuscates the real crash reason.

asherkin:
Thanks! This one confuses many people.

PR #482, opened 2016-02-27
Fix printing dependent extension name on unload

peace-maker:
Typo which will print the name of the extension being unloaded instead
of the one which would be unloaded as well because of it.


KyleSanderson:
Thanks!

PR #483, opened 2016-02-28
Most of include files updated to new 1.7 syntax

MainlandHero:
I double checked to make sure I have included the newest changes in dev (since I was editing from stable, not originally from the latest dev)

Let me know if you would like me to update anything else.

### DIFF
--- a/plugins/include/helpers.inc
+++ b/plugins/include/helpers.inc
@@ -43,12 +43,12 @@
  * @param buffer		Buffer for text.
  * @param maxlength		Maximum length of text.
  */
-stock FormatUserLogText(client, String:buffer[], maxlength)
+stock void FormatUserLogText(int client, char[] buffer, int maxlength)
 {
-	decl String:auth[32];
-	decl String:name[MAX_NAME_LENGTH];
+	char auth[32];
+	char name[MAX_NAME_LENGTH];
 	
-	new userid = GetClientUserId(client);
+	int userid = GetClientUserId(client);
 	if (!GetClientAuthString(client, auth, sizeof(auth)))
 	{
 		strcopy(auth, sizeof(auth), "UNKNOWN");
@@ -69,12 +69,12 @@ stock FormatUserLogText(client, String:buffer[], maxlength)
  * @param filename		Filename of the plugin to search for.
  * @return				Handle to plugin if found, INVALID_HANDLE otherwise.
  */
-stock Handle:FindPluginByFile(const String:filename[])
+stock Handle FindPluginByFile(const char[] filename)
 {
-	decl String:buffer[256];
+	char buffer[256];
 	
-	new Handle:iter = GetPluginIterator();
-	new Handle:pl;
+	Handle iter = GetPluginIterator();
+	Handle pl;
 	
 	while (MorePlugins(iter))
 	{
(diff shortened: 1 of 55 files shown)


asherkin:
All the functions with `@noreturn` in their doc comment (and probably many more) should be `void` rather than `int`.

MainlandHero:
I was debating on whether it should be void or int, so I compared it to https://sm.alliedmods.net/new-api/ function by function to address the return type issue.

Which way should I do it? Does the API need to be fixed?

asherkin:
The API documentation is generated from the files, since pre-transitional-syntax did not have a `void` return type (hence the `@noreturn` annotation), all functions that haven't been updated yet show as `int`.

MainlandHero:
Went through and updated according to your comment. I also changed EmitGameSound (and related functions) in sdktools_sound.inc to have the correct return annotation (since it returns a bool based on the success of playing the sound)

Ne3tCode:
You missed
**`void`**`GameRules_SetProp[Float|Ent|Vector](...)`
**`void`**`GameRules_GetPropVector(...)`
**`bool`**`TR_PointOutsideWorld(...)`

Maybe something else.

MainlandHero:
Sorry about that, thanks for checking. I somehow skipped over gamerules. And I didn't get the other one because I was searching for @noreturn

Here are few more files I updated.
Things I am unclear with:

variable type info (any?) in `native void SetCookiePrefabMenu(...)` - clientprefs

MainlandHero:
Another thing I skipped past since I don't know how to fix is in the return statement of `stock float operator-(float oper)`. I left that as is.

Continue on, in profiler.inc, I guessed that void is the return type of the function `native void LeaveProfilingEvent();`

MainlandHero:
If there are still errors on the automatic builds or any mistakes I have missed, I will fix it tomorrow.

plugins/include/tf2_stocks.inc
```diff
@@ -338,7 +338,7 @@ stock TFTeam:TF2_GetClientTeam(client)
  * @noreturn
  * @error               Invalid client index.
  */
-stock TF2_ChangeClientTeam(client, TFTeam:team)
+stock void TF2_ChangeClientTeam(int client, TFTeam team)
 {
     ChangeClientTeam(client, _:team);
```
ErikMinekus: ```_:team``` should be ```view_as<int>(team)```

plugins/include/tf2_stocks.inc
```diff
@@ -325,7 +325,7 @@ static const String:TFResourceNames[TFResourceType][] =
  * @return				Current TFTeam of client.
  * @error				Invalid client index.
  */
-stock TFTeam:TF2_GetClientTeam(client)
+stock TFTeam TF2_GetClientTeam(int client)
 {
 	return TFTeam:GetClientTeam(client);
```
ErikMinekus: ```view_as<TFTeam>(GetClientTeam(client))```

plugins/include/sdktools_tempents_stocks.inc
```diff
@@ -273,7 +273,7 @@ stock TE_SetupBloodSprite(const Float:pos[3], const Float:dir[3], const color[4]
  * @param End_Radius		Final ring radius.
  * @param ModelIndex		Precached model index.
  * @param HaloIndex		Precached model index.
- * @param StartFrame		Initial frame to render.
+ * @param StartFrame		Initital frame to render.
```
ErikMinekus: Initital is not a word. Initial is correct.

plugins/include/helpers.inc
```diff
@@ -202,10 +204,10 @@ stock FindTarget(client, const String:target[], bool:nobots = false, bool:immuni
  * @return			Number of maps loaded or 0 if in error.
  */
 #pragma deprecated Use ReadMapList() instead. 
- stock LoadMaps(Handle:array, &fileTime = 0, Handle:fileCvar = INVALID_HANDLE)
+ stock int LoadMaps(Handle array, int &fileTime = 0, Handle fileCvar = INVALID_HANDLE)
  { 
- 	decl String:mapPath[256], String:mapFile[64];
- 	new bool:fileFound = false;
+ 	char mapPath[256], char mapFile[64];
```
ErikMinekus: Should be ```char mapPath[256], mapFile[64];```

plugins/include/tf2.inc
```diff
@@ -194,7 +194,7 @@ enum TFCond
 	TFCond_SpawnOutline, //114
 };
 
-const Float:TFCondDuration_Infinite = -1.0;
+const float TFCondDuration_Infinite = -1.0;
 
 enum TFHoliday
```
ErikMinekus: Below this enum are some public consts which still have a colon.

psychonic:
Thank you for the contribution!

We instead decided to go with PRs #509, #510, and #511 for this.

PR #484, opened 2016-02-28
Update basebans and SQL admins plugins for transitional syntax

ErikMinekus:


psychonic:
Thanks! Sorry for taking so long to look it over.

PR #485, opened 2016-02-29
Fix scheduled plugin unloading

peace-maker:
When unloading a plugin no plugin is allowed to execute code. There's
been some flawed logic in the scheduling, leading it to bail out if the
plugin was marked for scheduled unloading..

Also change the "error" message of `sm plugins unload` to something
nice.


peace-maker:
This fixes unloading plugins through other plugins like `sm_rcon sm plugins unload myplugin` for completion ;)

PR #486, opened 2016-03-02
Add initial version of safety checks for CS:GO to attempt to avoid user GSLT bans.

psychonic:
There's room to make this more robust, performant, stringent, etc., however I would like to get _something_ in place sooner than later as more users are receiving account-wide Game Server Login Token bans and having trouble figuring out why.

From testing, this appears to work as desired.

Please let me know if you see any glaring bugs or deal-breakers here.
@asherkin 
@Drifter321 

Further improvements are welcome after the initial version has landed.

### DIFF
--- a/core/smn_entities.cpp
+++ b/core/smn_entities.cpp
@@ -93,6 +93,15 @@ enum PropFieldType
 	PropField_String_T,			/**< Valid for Data fields.  Read only! */
 };
 
+inline bool CanSetPropName(const char *pszPropName)
+{
+#if SOURCE_ENGINE == SE_CSGO
+	return g_HL2.CanSetCSGOEntProp(pszPropName);
+#else
+	return true;
+#endif
+}
+
 inline edict_t *BaseEntityToEdict(CBaseEntity *pEntity)
 {
 	IServerUnknown *pUnk = (IServerUnknown *)pEntity;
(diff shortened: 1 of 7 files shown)


core/smn_entities.cpp
```diff
@@ -1995,6 +2012,11 @@ static cell_t GetEntPropString(IPluginContext *pContext, const cell_t *params)
 					element);
  			}
 
+			if (!CanSetPropName(prop))
```
KyleSanderson: I believe this is in the wrong function, this is GetEntPropString.

psychonic:
@KyleSanderson thanks. This should be fixed now.
@Drifter321 weapon_knife_t as mentioned in IRC should be allowed now as well.

extensions/sdktools/vnatives.cpp
```diff
@@ -144,6 +144,20 @@ static cell_t RemovePlayerItem(IPluginContext *pContext, const cell_t *params)
 class CEconItemView;
 static cell_t GiveNamedItem(IPluginContext *pContext, const cell_t *params)
 {
+	if (g_SdkTools.ShouldFollowCSGOServerGuidelines())
+	{
+		char *pWeaponName;
+		pContext->LocalToString(params[2], &pWeaponName);
+
+		// Don't allow knives other than weapon_knife,  weapon_knifegg, and wewapon_knife_t.
```
yedpodtrzitko: there's a tiny typo

PR #487, opened 2016-03-08
Add Nuclear Dawn Stock Functions

stickz:


plugins/include/nd_stocks.inc
```diff
@@ -0,0 +1,101 @@
...
+ *
+ * @param client		Client index.
+ * @pram PlayerManEnt		FindEntityByClassname(-1, "nd_player_manager") or -1
+ * @return			The level of the client or some value less than 2
+ * @error			Unloaded level or level less than 2.
+ */
+ 
+stock ND_RetreiveLevel(client, PlayerManEnt = -1)
+{
+	new PlayerManager = PlayerManEnt != -1 ? PlayerManEnt : FindEntityByClassname(-1, "nd_player_manager");
```
WildCard65: Wouldn't SDKTool's GetPlayerResourceEntity work here or does nd_player_manager not count as a resource entity?

stickz: Yup it works, I just tested it. Both outputted the same number.
```
#include <sourcemod>
#include <sdktools>

public OnPluginStart()
{
	RegAdminCmd("sm_TestPlayerMan", CMD_TestPlayerMan, ADMFLAG_KICK, "Tests player manager");
}

public Action:CMD_TestPlayerMan(client, args)
{
	new ResEntOne = GetPlayerResourceEntity();	
	new ResEntTwo = FindEntityByClassname(-1, "nd_player_manager");	
	PrintToChat(client, "GPRE: %d, FEBC: %d", ResEntOne, ResEntTwo);
}
```

stickz:
I'm adding more useful stocks currently.

plugins/include/nd_stocks.inc
```diff
@@ -0,0 +1,222 @@
...
+#if defined _nd_stocks_included_
+  #endinput
+#endif
+#define _nd_stocks_included_
+
+/* The team name constants */
+#define TEAM_UNASSIGNED		0
+#define TEAM_SPEC		1
+#define TEAM_CONSORT		2
+#define TEAM_EMPIRE		3
```
psychonic: The symbol names TEAM_UNASSIGNED and TEAM_SPEC can clash with defines people have for other games. This is probably better as an enum anyway and then have a stock to return/set a client team, retagged.

ex.
```
enum NDTeam {
    NDTeam_Unassigned, // etc.
}

stock NDTeam GetClientNDTeam(int client) { return view_as<NDTeam>(GetClientTeam(client)); }
```

plugins/include/nd_stocks.inc
```diff
@@ -0,0 +1,222 @@
...
+ * Retreives the level of a given client index
+ *
+ * @param client		Client index.
+ * @return			The level of the client or some value less than 2
+ * @error			Unloaded level or level less than 2.
+ */
+ 
+stock ND_RetreiveLevel(client)
+{
+	return GetEntProp(GetPlayerResourceEntity(), Prop_Send, "m_iPlayerRank", 1, client);
```
psychonic: Your wanting to use a cached player resource ent was fine, but the extra arg would have added confusion. Something like this would work:
stock int ND_GetPlayerRank(int client)
{
    int resourceEnt = INVALID_ENT_REFERENCE;
    if (!IsValidEntity(resourceEnt))
    {
        resourceEnt = EntIndexToEntRef(GetPlayerResourceEntity());
    }

    return GetEntProp(GetPlayerResourceEntity(), Prop_Send, "m_iPlayerRank", _, client);
}

plugins/include/nd_stocks.inc
```diff
@@ -0,0 +1,222 @@
...
+/** 
+ * Sets the class of a client index
+ *
+ * @param client		Client index.
+ * @parm2 MainClass		The main class to set to
+ * @parm3 SubClass		The sub class to set to
+ * @parm4 Gizmo			The gizmo to change to
+ */
+
+stock ND_SetPlayerClass(client, MainClass, SubClass, gizmo = GIZMO_NONE)
```
psychonic: If you're going to tag the enums (which is good), the args should be tagged as well (and then view_as int when calling SetEntProp).

plugins/include/nd_stocks.inc
```diff
@@ -0,0 +1,222 @@
...
+
+stock ND_GetPlayerSubClass(client)
+{
+	return GetEntProp(client, Prop_Send, "m_iPlayerSubclass");
+}
+
+
+/* Structure Functions */
+
+enum eNDStructFlags
```
psychonic: Just "NDStructFlags" would be better for consistency with the other enums in this inc and other first-party ones.

plugins/include/nd_stocks.inc
```diff
@@ -0,0 +1,222 @@
...
+};
+
+/** 
+ * Get the properties which belong to a structure
+ *
+ * @param entity		Entity index of the structure.
+ * @return			The flags that structure has
+ */
+
+stock ND_GetStructureFlags(entity)
```
psychonic: Again, using a named enum, but not tagging the return.

plugins/include/nd_stocks.inc
```diff
@@ -0,0 +1,222 @@
...
+/** 
+ * Remove specific property from a structure
+ *
+ * @param entity		Entity index of the structure.
+ * @parm2 flag			The flag to remove from the structure
+ */
+
+stock ND_RemoveStructureFlag(entity, flag)
+{
+	SetEntProp(entity, Prop_Send, "m_iStructFlags", ND_GetStructureFlags(entity) ~ flag);
```
psychonic: You want &~ here, not ~.

psychonic:
Any new code should use the new S[ourcePawn 1.7 Transitional syntax](https://wiki.alliedmods.net/SourcePawn_Transitional_Syntax).

plugins/include/nd_stocks.inc
```diff
@@ -0,0 +1,222 @@
...
+
+/**
+ * Gets the name of a given team index
+ *
+ * @param team			Team index.
+ * @return			The name of the team or "Unknown" on invalid team index
+ * @error			Invalid team index.
+ */
+
+stock String:ND_GetTeamName(team)
```
ErikMinekus: Functions can't return strings, pass a buffer instead like https://sm.alliedmods.net/new-api/sdktools_functions/GetTeamName.

plugins/include/nd_stocks.inc
```diff
@@ -0,0 +1,222 @@
...
+
+/** 
+ * Retreives the level of a given client index
+ *
+ * @param client		Client index.
+ * @return			The level of the client or some value less than 2
+ * @error			Unloaded level or level less than 2.
+ */
+ 
+stock ND_RetreiveLevel(client)
```
ErikMinekus: Did you mean: Retrieve

plugins/include/nd_stocks.inc
peace-maker: Non-public functions actually can return strings and arrays :+1:

stickz:
The compile checks are broken. It just passed an error through.

WildCard65:
@stickz char = String:

WildCard65:
Also when I said nameless enum, I mean something like:
```
enum
{
    ND_Assault_Infantry = 0,
    ND_Assault_Grendier,
    ND_Assault_Sniper
};
```
^ Example of nameless enum for assault sub-classes.

plugins/include/nd_stocks.inc
```diff
@@ -0,0 +1,267 @@
...
+
+/**
+ * Gets the name of a given team index
+ *
+ * @param team			Team index.
+ * @return			The name of the team or "Unknown" on invalid team index
+ * @error			Invalid team index.
+ */
+
+stock String:ND_GetTeamName(int team)
```
WildCard65: Probably be best to switch this to match SM's API when dealing with strings like:
stock int ND_GetTeamName(NDTeam team, char[] buffer, int maxlength) // Return means number of bytes written(returned by strcopy).
Note: I think this isn't needed because of SDKTool's GetTeamName native.

plugins/include/nd_stocks.inc
```diff
@@ -0,0 +1,267 @@
...
+/** 
+ * Get the main class of a client index
+ *
+ * @param client		Client index.
+ * @return			The main class of the client, see mainclass enum
+ */
+
+stock ND_MainClasses ND_GetPlayerMainClass(int client)
+{
+	return view_as<ND_MainClasses>GetEntProp(client, Prop_Send, "m_iPlayerClass");
```
WildCard65: ( and ) around call to GetEntProp

plugins/include/nd_stocks.inc
```diff
@@ -0,0 +1,267 @@
...
+};
+
+/** 
+ * Get the properties which belong to a structure
+ *
+ * @param entity		Entity index of the structure.
+ * @return			The flags that structure has
+ */
+
+stock ND_GetStructureFlags(int entity)
```
WildCard65: Tag as NDStructFlags

plugins/include/nd_stocks.inc
```diff
@@ -0,0 +1,267 @@
...
+/** 
+ * Give a structure a specific property
+ *
+ * @param entity		Entity index of the structure.
+ * @parm2 flag			Flag to give the structure, see enum above
+ */
+
+stock void ND_AddStructureFlag(int entity, NDStructFlags flag)
+{
+	SetEntProp(entity, Prop_Send, "m_iStructFlags", ND_GetStructureFlags(entity) | view_as<int>flag);
```
WildCard65: Better to do(once above stock is tagged):
```
view_as<int>(ND_GetStructureFlags(entity) | flags)
```

plugins/include/nd_stocks.inc
```diff
@@ -0,0 +1,267 @@
...
+/** 
+ * Remove specific property from a structure
+ *
+ * @param entity		Entity index of the structure.
+ * @parm2 flag			Flag to remove from structure, see enum above
+ */
+
+stock void ND_RemoveStructureFlag(int entity, NDStructFlags flag)
+{
+	SetEntProp(entity, Prop_Send, "m_iStructFlags", ND_GetStructureFlags(entity) &~ view_as<int>flag);
```
WildCard65: Ditto for above, but replace "|" with "&" and prepend "~" infront of "flags" in the statement.

plugins/include/nd_stocks.inc
stickz: The SM API doesn't make any sense in this case. The purpose of this function isn't to return a char array of characters in a team name; it's to return the full team name string. Regardless how the syntax handles things, it makes more sense for the datatype to be a string.

WildCard65: As already noted, "char" == "String" in SP

plugins/include/nd_stocks.inc
```diff
@@ -0,0 +1,267 @@
...
+ *
+ * @param team			NDTeam enum.
+ * @error			Invalid team or no commander
+ */
+
+stock void ND_DemoteTeamCommander(NDTeam team)
+{
+	int commander = ND_GetTeamCommander(team);
+	
+	FakeClientCommand(commander, "startmutiny");
```
yedpodtrzitko: would be wise to check if the value of `commander` is not -1 before doing this

plugins/include/nd_stocks.inc
```diff
@@ -0,0 +1,267 @@
...
+#define ND_ASSAULT_CLASS_INFANTRY 	0
+#define ND_ASSAULT_CLASS_GRENDIER 	1
+#define ND_ASSAULT_CLASS_SNIPER 	2
+
+#define ND_EXO_CLASS_SUPRESSION		0
+#define ND_EXO_CLASS_SEIGEKIT 		1
+
+#define ND_STEALTH_CLASS_ASSASSIN	0
+#define ND_STEALTH_CLASS_SNIPER 	1
+#define ND_STEALTH_CLASS_SABATUER 	2
```
yedpodtrzitko: saboteur

plugins/include/nd_stocks.inc
```diff
@@ -0,0 +1,267 @@
...
+	ND_MainClass_Support
+};
+
+/* Sub Classes */
+#define ND_ASSAULT_CLASS_INFANTRY 	0
+#define ND_ASSAULT_CLASS_GRENDIER 	1
+#define ND_ASSAULT_CLASS_SNIPER 	2
+
+#define ND_EXO_CLASS_SUPRESSION		0
+#define ND_EXO_CLASS_SEIGEKIT 		1
```
yedpodtrzitko: it's `siege`, not `seige`. Also why suffix `kit` here and not anywhere else?

plugins/include/nd_stocks.inc
```diff
@@ -0,0 +1,267 @@
...
+{
+	ND_MainClass_Assault = 0,
+	ND_MainClass_Exo,
+	ND_MainClass_Stealth,
+	ND_MainClass_Support
+};
+
+/* Sub Classes */
+#define ND_ASSAULT_CLASS_INFANTRY 	0
+#define ND_ASSAULT_CLASS_GRENDIER 	1
```
yedpodtrzitko: grenadier

plugins/include/nd_stocks.inc
```diff
@@ -0,0 +1,267 @@
...
+ * Gets the name of a given team index
+ *
+ * @param team			Team index.
+ * @return			The name of the team or "Unknown" on invalid team index
+ * @error			Invalid team index.
+ */
+
+stock char ND_GetTeamName(int team)
+{
+	char TeamName[12];
```
yedpodtrzitko: 10 would be enough

plugins/include/nd_stocks.inc
```diff
@@ -0,0 +1,267 @@
...
+stock ND_MainClasses ND_GetPlayerMainClass(int client)
+{
+	return view_as<ND_MainClasses>(GetEntProp(client, Prop_Send, "m_iPlayerClass"));
+}
+
+/** 
+ * Get the sub class of a client index
+ *
+ * @param client		Client index.
+ * @return			The sub class of the client, see subclass consants
```
yedpodtrzitko: - it's spelled cons**t**ants
- there are no such constants defined anywhere (?)

plugins/include/nd_stocks.inc
```diff
@@ -0,0 +1,267 @@
...
+
+/* Gizmo(s) */
+#define ND_GIZMO_NONE 			0
+
+/** 
+ * Sets the class of a client index
+ *
+ * @param client		Client index.
+ * @parm2 MainClass		The main class to set, see mainclass enum
+ * @parm3 SubClass		The sub class to set, see subclass consants
```
yedpodtrzitko: - it's spelled cons**t**ants
- there are no such constants defined anywhere (?)

plugins/include/nd_stocks.inc
yedpodtrzitko: What is this function good for? Imo would more useful to have something like `bool ND_HasStructureFlash(int entity, NDStructFlags flag)` instead.

plugins/include/nd_stocks.inc
WildCard65: The constants are the defines.

yedpodtrzitko: no - those are classes, and each of them its subclasses (which is this function/comment related to)

WildCard65: The enum MainClass is the main classes, the defines like ND_ASSAULT_CLASS_INFANTRY is a subclass.  
In SP, defines ARE constants like they are in C++

yedpodtrzitko: oh yes, right.

plugins/include/nd_stocks.inc
stickz: An IsValidCommander() function would be better. Raising an error is a good thing in this case instead of just having it silently fail, so, the client knows they've done something wrong.

plugins/include/nd_stocks.inc
stickz: The kit suffix is part of the actual subclass name in the selection panel, unlike, other classes.

KyleSanderson:
Returning Arrays is not permitted/first-party supported in SP at the time of this writing.

As the feedback from Yed for the myriad of typos hasn't been resolved since March, I'm closing this.

PR #489, opened 2016-03-14
Don't try to call functions in paused plugins

peace-maker:
This avoids spam of "Plugin not runnable" exceptions on shutdown or
plugin unload.

When re/unloading a plugin which has other ones depending on it, like
the adminmenu, It pauses the depending plugins putting them in an
"Depends on plugin: %s" error state. ForwardSys doesn't remove them from
the forward lists on pause, specially the global forwards, and still
tries to call all the global forwards like OnPlayerRunCmd and
OnLibraryAdded etc. on the paused plugins. Executing functions in paused
runtimes has been ignored in the VM before introducing the "Exception"
mechanism, but now they're all logged.

This adds checks to make sure the plugin is runnable before calling a
function. (Stolen from #438)

### DIFF
--- a/core/logic/ForwardSys.cpp
+++ b/core/logic/ForwardSys.cpp
@@ -252,6 +252,9 @@ int CForward::Execute(cell_t *result, IForwardFilter *filter)
 		if (filter)
 			filter->Preprocess(func, temp_info);
 
+		if (func->GetParentRuntime()->IsPaused())
+			continue;
+
 		for (unsigned int i=0; i<num_params; i++)
 		{
 			int err = SP_ERROR_PARAM;
--- a/core/logic/smn_database.cpp
+++ b/core/logic/smn_database.cpp
@@ -236,6 +236,9 @@ public:
 	}
 	void CancelThinkPart()
 	{
+		if (!m_pFunction->IsRunnable())
+			return;
+
 		m_pFunction->PushCell(BAD_HANDLE);
 		m_pFunction->PushCell(BAD_HANDLE);
 		m_pFunction->PushString("Driver is unloading");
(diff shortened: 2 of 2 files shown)


KyleSanderson:
This may be unpopular, however if a plugin is paused they should be cleaned up, similar to an unload.

Queries are finished, in a blocking manner as-is with unload.

Forwards are unregistered. If they're still registered, we already have IForward::GetFunctionCount, checking this and then doing nothing because the single plugin defeats that. We're only hurting ourselves here.

Sprinkling these checks don't seem "legit" to me.

@dvander ?

dvander:
Like everything scary relating to plugin unloading I'd rather just see where the removal goes. My current thinking is that we can eliminate pausing down to the window of time in between scheduling the unload and the actual unload, and during this time it is okay for errors to be spammy. The duration should be short, and if we fire callbacks correctly it should minimize weird cases, assuming authors interact with SetFailState correctly.

dvander:
Failing that, or if the spam is still too high, there is specific machinery we'd want to use to solve this kind of problem: new callbacks that instruct listeners to cancel pending operations, or more likely, just set a flag in the VM to silence console error reporting on that runtime.

peace-maker:
Posts about this stack up on the forums. This could be checked in and backed out when pausing is removed?

Concerning the forward function count being off, one could work around that by caching the `m_functions.length()` and adjust it `OnPluginPauseChange` or run through the list and only count runnable functions in `CForward::GetFunctionCount()`.

asherkin:
> This could be checked in and backed out when pausing is removed?

This seems like the right thing to do here, it's been many months and we're no closer to finishing the colossal undertaking that is removing pausing - in the mean time this is seriously harming user faith in SM and the utility of error logs (that have previously been very high quality and low noise). @psychonic?

psychonic:
Make it so.

PR #490, opened 2016-03-16
Format instead of BreakString

iNilo:
BreakString cuts off the following arguments:
sm_psay @me 1 2 3 //1
using format, or whatever is better it now prints:
sm_psay @me 1 2 3 //1 2 3


asherkin:
This doesn't work in the general case, GetCmdArgString could still have quotes around it if it was from a chat trigger.

Adding a call to StripQuotes is probably enough, but this clearly hasn't been tested for the various edge cases :stuck_out_tongue:.

PR #492, opened 2016-03-18
GetEntityRenderColor to get the entity colors

iNilo:
there is an SetEntityRenderColor, but no GetEntityRenderColor.
imo this can be a default stock.


iNilo:
nvm someone made it already.
My bad

PR #493, opened 2016-03-20
Only hook IBaseFileSystem::FileExists if we're also hooking INetChannel

KyleSanderson:
This works around a TF Replay SourceHook crash that Johns was seeing when using 1.8.
SourceHook is not thread-safe at the time of this writing.

As TF is the only game that has fully implemented replay (the changes from the CS:S beta branch (appid 260) were thrown away during the Steam| conversion, other games never had the feature), this should minimize the footprint significantly.

### DIFF
--- a/extensions/sdktools/hooks.cpp
+++ b/extensions/sdktools/hooks.cpp
@@ -78,8 +78,6 @@ void CHookManager::Initialize()
 		g_pSM->LogError(myself, "Failed to find PlayerRunCmd offset - OnPlayerRunCmd forward disabled.");
 		PRCH_enabled = false;
 	}
-	
-	SH_ADD_HOOK(IBaseFileSystem, FileExists, basefilesystem, SH_MEMBER(this, &CHookManager::FileExists), false);
 
 	basefilesystemPatch = SH_GET_CALLCLASS(basefilesystem);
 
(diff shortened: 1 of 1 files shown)


psychonic:
Is there a way to change this around to actually fix the issue?

If nothing else, can we disable this logic on TF2 and throw some sort of error if we detect that replay is enabled? (rather than consumers of this still getting crashes in that case).

asherkin:
Might as well just test and merge the patch that makes SH per-thread and safe.

KyleSanderson:
@psychonic hard to say. We could lock this in a mutex, however depending on what Valve decides to do, this can introduce skipping or hard locks. Asher has stats on the footprint for 1.8, however considering it took this long for anyone to notice (nothing in Accelerator to my knowledge), it's tough to say how many real consumers there are of Replay (and less so for these forwards).

We can disable it if Replay is active, however if someone is relying on this for server security as Valve outright refuses to actually address malicious file transfers from clients as an issue we would be doing them a disservice.

Ryan Kistner many moons ago changed SourceHook to utilize TLS, that may be the better path at this point as code is trending more and more towards MP.

asherkin:
The problem is not that the hook gets called from the other threads - the problem is that random SH callbacks are invoked if ANY hooked function is called from multiple threads. This can only be fixed in SH or by not hooking the function at all.

KyleSanderson:
That's much worse then I thought it was, thanks for the clarification asher.

So yeah, without going down the path towards https://en.wikipedia.org/wiki/Giant_lock we'd need to utilize TLS or another strategy in SourceHook.

psychonic:
Let's get this in as it's an improvement regardless.

Afterward, one of us can add disabling of this on servers with replay, at least until the required MM:S fixes are in.

PR #494, opened 2016-03-25
Always send changes when using the GameRules_SetProp* natives.

akowald:
This is a follow up to https://github.com/alliedmodders/sourcemod/pull/473#issuecomment-179506496.

Essentially, it forces changeState to true for all the GameRules_SetProp* natives. This brings the behavior in line with that of the SetEntProp* natives. I can't find a use case for not marking the offsets as changed. That just means that clients won't see the change until the game logic happens to change another game rules property, calling StateChanged().


psychonic:
Thanks!

PR #495, opened 2016-03-30
Add support for retrieving game server Steam Id (bug 6404).

psychonic:
This has been requested a few times and isn't too bad to add.

It could have been done simpler, but this leaves a segue to move more natives to the logic bin.

### DIFF
--- a/core/logic/smn_halflife.cpp
+++ b/core/logic/smn_halflife.cpp
@@ -1,0 +1,69 @@
+/**
+* vim: set ts=4 sw=4 tw=99 noet :
+* =============================================================================
+* SourceMod
+* Copyright (C) 2004-2016 AlliedModders LLC.  All rights reserved.
+* =============================================================================
+*
+* This program is free software; you can redistribute it and/or modify it under
+* the terms of the GNU General Public License, version 3.0, as published by the
+* Free Software Foundation.
+*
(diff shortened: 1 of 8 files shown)


core/logic/smn_halflife.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+static cell_t GetServerAccountId(IPluginContext *pContext, const cell_t *params)
+{
+	return (cell_t)(gamehelpers->GetServerSteamId64() & 0xFFFFFFFF);
+}
+
+REGISTER_NATIVES(halflifeNatives)
+{
+	{ "GetServerAuthId",			GetServerAuthId },
+	{ "GetServerSteamAccountId",	GetServerAccountId },
```
KyleSanderson: Should be similar to GetSteamAccountID so it doesn't drive people crazy.

psychonic: The existing GetPlayerAccountID is the exception, not the rule. We use "Id" everywhere else in the API. (It should actually be GetClientAccountId - it's just a bit late for that now).

psychonic:
As @KyleSanderson (?) pointed out, this didn't actually work at runtime on Linux (or Mac) for the engine on which we needed to call the steam_api function directly to get the id.

This has now been fixed.

KyleSanderson:
:shipit:

PR #500, opened 2016-04-21
Actual support for milliseconds since epoch

jbzdarkid:
While it is also valid to use addr[1] = (time_t)(millis % (1 << 31)), it's about 3% slower.
I use long long rather than long because some systems don't require long to be 64 bit. Long long is always at least 64 bits.

### DIFF
--- a/core/logic/smn_core.cpp
+++ b/core/logic/smn_core.cpp
@@ -152,13 +152,17 @@ void LogAction(Handle_t hndl, int type, int client, int target, const char *mess
 
 static cell_t GetTime(IPluginContext *pContext, const cell_t *params)
 {
-	time_t t = g_pSM->GetAdjustedTime();
+	time_t seconds = g_pSM->GetAdjustedTime();
+	struct timeval timev;
+	gettimeofday(&timev, NULL);
+	long long millis = (long long)seconds * 1000 + timev.tv_usec / 1000;
+
 	cell_t *addr;
 	pContext->LocalToPhysAddr(params[1], &addr);
+	addr[0] = (time_t)(millis >> 31); // Shift left by the length of an int to get the first half.
+	addr[1] = (time_t)(millis - addr[0]); // Subtract the first half to get the second half.
 
-	*(time_t *)addr = t;
-
-	return static_cast<cell_t>(t);
+	return static_cast<cell_t>(seconds);
 }
 
 #if defined SUBPLATFORM_SECURECRT
--- a/core/logic/smn_core.cpp
+++ b/core/logic/smn_core.cpp
@@ -47,6 +47,7 @@
 #include <limits.h>
 #include <unistd.h>
 #include <sys/times.h>
+#include <sys/time.h>
 #endif
 #include <IForwardSys.h>
 #include <ILibrarySys.h>


jbzdarkid:
The internet suggests:

    #include <windows.h>
    SYSTEMTIME time;
    GetSystemTime(&time);
    WORD millis = (time.wSecond * 1000) + time.wMilliseconds;

So likely this could be written to use `time.wMilliseconds` instead of `timev.tv_usec / 1000`. I don't have frequent access to windows, if someone could test this that would be wonderful ^_^

asherkin:
Plugins using GetTime are expecting seconds, you can't just change to giving them milliseconds.

This will need an additional native, what is your use case?

jbzdarkid:
GetTime still return seconds, this only sets milliseconds to a passed-in array (as stated in the comment). Use case is for timing anything more accurately than 1 tick. For me, that's finding the tf2 weapon function times.

asherkin:
The array param is already used to return seconds, it's behaviour can't be changed. 

If you're just timing something, use the high-resolution profiling natives, that's what they're there for.

jbzdarkid:
GetEngineTime() isn't higher profile, it also returns accuracy to the frame count only. If the array was used to return seconds, than the comment should be changed -- "64 bit timestamp" usually means milliseconds since epoch.

asherkin:
https://sm.alliedmods.net/new-api/profiler

jbzdarkid:
Unless I'm misunderstanding the API, that's only reports to the second, not millisecond.

asherkin:
`GetProfilerTime` returns a floating point number, it has microsecond precision (granted, you wont get that if you measure over a long time).

jbzdarkid:
GetProfilerTime	Returns the amount of high-precision time in seconds that passed during the profiler's last start/stop cycle.

This does not imply (to me) millisecond accuracy, but I'll try it tonight, when I can.

jbzdarkid:
Ok, GetProfilerTime() seems to do what I need. However, I still have a few concerns:
The spec says that GetTime(bigArray[2]) will set bigArray to a "64 bit timestamp". As far as the code goes, this will never happen, as it always simply sets it to the 32 bit seconds since epoch.
Now as far as I can tell, [nobody is using this feature](https://github.com/search?q=GetTime+filename%3A.sp&type=Code), though this isn't the most complete search.

It's also worth noting that there's no easy way (and I wouldn't say Profiler is easy) to get millisecond accuracy. While the uses are pretty small, using seconds will break eventually (2038), and there are applications where frame precision isn't sufficient.

If you don't decide to implement this PR, please at least change the documentation for GetTime to clarify that it returns seconds to the array, and the documentation for GetProfilerTime() to indicate that it returns microsecond accuracy.

psychonic:
> If you don't decide to implement this PR, please at least change the documentation for GetTime to clarify that it returns seconds to the array

From anything I can find when searching, "64-bit timestamp" already implies that it's in seconds, just allowing for higher values - nevermind that the PR doesn't even compile on all supported platforms.

jbzdarkid:
I'm not expecting you to accept the PR without a successful test pass, I'm just not hugely invested in fixing it if there's a conceptual concern.

My results for '64 bit timestamp' turn up NTP time, Windows time storage, and using 64 bits for seconds. Regardless, some clarification would be nice.

asherkin:
> The spec says that GetTime(bigArray[2]) will set bigArray to a "64 bit timestamp". As far as the code goes, this will never happen, as it always simply sets it to the 32 bit seconds since epoch.

No, time_t is 64-bit seconds since epoch. The array param gets the full 64-bits in two halfs (with the high half always being 0 until 2038, as you mention - and that's why it exists, and why few bother to deal with it), and then it's truncated to just the low half to be returned as a cell_t (since SourcePawn does not support 64-bit integers). 

`GetTime` is documented as returning a UNIX timestamp, that is crystal clear as being seconds since epoch. I agree that the clarification on that should probably move from the `@return` anotation to the description though.

jbzdarkid:
I guess the answer is a 'Not broken, won't fix'. Alright, I'll remember to use Profiler next time I want higher precision.

PR #502, opened 2016-04-26
Allow support for running on filesystems that use 64-bit inodes on Linux.

psychonic:
Most supported games don't even support this case, but at least CS:GO does. WIthout
this fix, some filesystem calls can fail, or in the case of readdir, fail to return all/any files.
This was first observed when using an XFS-formatted volume on CentOS 7 x64.

@asherkin


asherkin:
I'm afraid I am out of Adventure Time gifs. I shall top up when I get around to watching the latest season.

🚢

PR #503, opened 2016-04-26
Allow Plugin_Changed return on event PreHooks to not block.

psychonic:
This avoids issues from confusion that users have, returning Plugin_Changed
after altering event data, when Plugin_Changed isn't required and wasn't supported
in that scenario, and instead acted like Plugin_Handled.


KyleSanderson:
r+, this has screwed people up (including myself) for close to a decade.

PR #505, opened 2016-05-08
Add support for sending gameevent to specific client.

psychonic:
I also initially planned on adding support for blocking existing events, per client on-the-fly, but API was getting complex and potentially inefficient. (This can still be added at some point). The net effect of that can still be achieved by blocking existing events with a prehook, saving event data, and then resending to the desired clients (not ideal, but doable).

Only tested on TF2 in Windows

### DIFF
--- a/core/EventManager.cpp
+++ b/core/EventManager.cpp
@@ -31,6 +31,8 @@
 
 #include "EventManager.h"
 #include "sm_stringutil.h"
+#include "PlayerManager.h"
+
 #include "logic_bridge.h"
 #include <bridge/include/IScriptManager.h>
 
@@ -358,6 +360,13 @@ void EventManager::FireEvent(EventInfo *pInfo, bool bDontBroadcast)
 
 	/* Add EventInfo struct to free event stack */
 	m_FreeEvents.push(pInfo);
+}
+
+void EventManager::FireEventToClient(EventInfo *pInfo, IClient *pClient)
+{
+	// The IClient vtable is +4 from the IGameEventListener2 (CBaseClient) vtable due to multiple inheritance.
+	IGameEventListener2 *pGameClient = (IGameEventListener2 *)((intptr_t)pClient - 4);
+	pGameClient->FireGameEvent(pInfo->pEvent);
 }
 
 void EventManager::CancelCreatedEvent(EventInfo *pInfo)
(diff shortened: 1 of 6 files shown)


core/EventManager.cpp
```diff
@@ -31,6 +31,8 @@
 
 #include "EventManager.h"
 #include "sm_stringutil.h"
+#include "PlayerManager.h"
+
```
WildCard65: So is this blank line needed?

psychonic: ...

psychonic:
Thinking about it more, it shouldn't be necessary to add any further hooking. Even copying the event data in the "workaround" can be avoided, since you can use Event.FireToClient in an event prehook.

Pattern would be something like:

```
public Action Ev_SomeEventPre(Event event, const char[] name, bool dontBroadcast)
{
    for (int i = 1; i <= MaxClients; ++i)
    {
        if (IsClientBlah(client))
            event.FireToClient(i);
    }

    return Plugin_Handled;
}
```

Benoist3012:
So event.BroadcastDisabled = true; is no longer required?

psychonic:
I don't fully understand the question, but this is a better example of the above, if you don't want to actually block the event serverside.

```
public Action Ev_SomeEventPre(Event event, const char[] name, bool dontBroadcast)
{
    for (int i = 1; i <= MaxClients; ++i)
    {
        if (IsClientBlah(client))
            event.FireToClient(i);
    }

    event.BroadcastDisabled = true;

    return Plugin_Continue;
}
```

When you only block broadcast, it stops the event from going to clients, but the server will still do any processing that it may have for the event (such as the log line for player_death). If you block the event itself, that gets blocked, too.

core/smn_events.cpp
```diff
@@ -146,6 +147,54 @@ static cell_t sm_FireEvent(IPluginContext *pContext, const cell_t *params)
...
+
+	if (!pPlayer->IsConnected())
+	{
+		return pContext->ThrowNativeError("Client %d is not connected", client);
+	}
+
+	IClient *pClient = pPlayer->GetIClient();
+	if (!pClient)
+	{
+		return pContext->ThrowNativeError("Sending events to fakeclients is not supported on this game (client %d)", client);
```
KyleSanderson: There are super rare cases where a client can be connected but still have an invalid net channel (KAC/SMAC used to cause crashes in convar queries). Despite this, everything looks fine.

:shipit:

psychonic: Pics or it didn't happen.

KyleSanderson: Ironically it was part of the code you refactored :smile: 

https://hg.alliedmods.net/releases/sourcemod-1.3/rev/82cfdef347cc

PR #461, opened 2015-12-11
New syntax

wyd3x:
Update to new syntax (now const saved, my bad last time)

PR #509, opened 2016-05-12
Update small includes for transitional syntax

ErikMinekus:
The remaining 16 includes have a lot more changes, so they'll be in a separate PR. Tested with newdecls required and including all includes, all default plugins and a bunch of my own compile without warnings. Obsoletes #483 and #461.


psychonic:
The "@noreturn" lines can be removed, as they're made redundant with actually having a void return type.

psychonic:
Thanks!

PR #512, opened 2016-05-19
Don't enable FileExists hook (SendFile forward) on TF2 servers with Replay

psychonic:
It will crash. See PR #493 for details.

@KyleSanderson


KyleSanderson:
Merge when build succeeds.

:shipit:

PR #513, opened 2016-05-21
Fix late/incorrect calls to TR_GetEntityIndex potentially causing crash.

psychonic:
When the global trace handle is used for this, the containted entity pointer can be stable. This adds separate storage of an SM entity reference. Also fixed it to return -1 as documented when there is no entity and using trace handle, rather than returning 0/world.

### DIFF
--- a/extensions/sdktools/trnatives.cpp
+++ b/extensions/sdktools/trnatives.cpp
@@ -58,6 +58,7 @@ private:
 /* Used for the global trace version */
 Ray_t g_Ray;
 trace_t g_Trace;
+cell_t g_TraceEntRef = INVALID_EHANDLE_INDEX;
 Vector g_StartVec;
 Vector g_EndVec;
 Vector g_HullMins;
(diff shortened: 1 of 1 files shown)


KyleSanderson:
Ancient crash with an equally popular ancient spray plugin. Nice job.

As an aside, can't this happen with the tr handle too; with the entity pointer eventually going away if someone holds onto it? I can't recall.

It's an improvement regardless of the above comment. :shipit:

psychonic:
That's a good point. If people hold onto the trace handle for even a frame, that can indeed become invalid. I'm going to rework this a bit.

psychonic:
@KyleSanderson should be covered now.

KyleSanderson:
I'm liking it a lot better. If we really want to cover our bases before invoking any natives using the entity from the tr pointer we can see if the reference, and thus the pointer is still valid.

Same as before though, :shipit:

psychonic:
I believe we're okay. We own the trace_t and never pass it back into the engine after having it initially populate it for us.

PR #517, opened 2016-06-04
Fix erroneous () in DebugReport:: printouts.

KyleSanderson:
This fixes a hard blocker for 1.8 from https://github.com/alliedmodders/sourcemod/pull/448.


dvander:
What exactly is going on here? I'm guessing the parens on the script name is a mistake, but what about the function name?

KyleSanderson:
The parens on the function name was always there. However without a prototype you can easily deduce that plugin::function exists. The () reads as func(void), which is invalid because there's no guarantee that the error in function takes no params. That's why I'm considering both of these cases to be invalid usages of ().

PR #518, opened 2016-06-06
Fix UTIL_ReplaceAll not properly tracking length (6472)

Fyren:
UTIL_ReplaceEx returns a pointer to the character after the newly-replaced stuff.  While maxlength is the size of the original buffer, the remaining space left needs to get reduced by how far forward we jumped after a replace happens.


KyleSanderson:
:shipit:

PR #519, opened 2016-06-06
fix CloseHandle signature

yedpodtrzitko:
[`CloseHandle` returns `bool`](https://github.com/alliedmodders/sourcemod/blob/0c8e6e29184bf58851954019a2060d84f0c556f9/core/logic/smn_handles.cpp#L58-L71). This PR updates its signature and docstring

### DIFF
--- a/plugins/include/handles.inc
+++ b/plugins/include/handles.inc
@@ -52,9 +52,10 @@ enum Handle // Tag disables introducing "Handle" as a symbol.
  *       sure you read the documentation on whatever provided the Handle.
  *
  * @param hndl		Handle to close.
+ * @return			True on success, false if Handle does not exists or on Handle access violation.
  * @error			Invalid handles will cause a run time error.
  */
-native void CloseHandle(Handle hndl);
+native bool CloseHandle(Handle hndl);
 
 /**
  * Clones a Handle.  When passing handles in between plugins, caching handles


asherkin:
That doccomment is not correct: https://mxr.alliedmods.net/sourcemod/source/core/logic/smn_handles.cpp#67

Also, please take 5 minutes to bother writing a description, or no one is going to take 10 minutes to review and merge your PR.

yedpodtrzitko:
I was about to add description afterward, sorry for the slowness.

I was not sure about that, so I looked at  `IsValidHandle` in the same file, which is marked as `bool`:
https://github.com/alliedmodders/sourcemod/blob/master/plugins/include/handles.inc#L97

...and its actual implementation returns also `0` / `1`
https://github.com/alliedmodders/sourcemod/blob/0c8e6e29184bf58851954019a2060d84f0c556f9/core/logic/smn_handles.cpp#L38-L50

Should both these functions have `int` return type then?

KyleSanderson:
IsValidHandle is considered technical debt and no longer exists in the API for this reason.

It's true on success, and false on a Handle Access violation, as you can see from the code.

dvander:
It's not that useful to have a return value on `CloseHandle` - there isn't anything sensible to do on a false return, and it's almost certainly an oversight that it doesn't just error in the access violation case anyway. I would suggest leaving it documented as "void", even though the implementation has a quirky return case.

PR #520, opened 2016-06-11
Increase topmenu display buffer size

peace-maker:
The buffer for menu options was 64 chars. That limit can easily be
reached by some utf8 string which would get truncated. Double the buffer
size to 128 chars.

### DIFF
--- a/extensions/topmenus/TopMenu.cpp
+++ b/extensions/topmenus/TopMenu.cpp
@@ -36,7 +36,7 @@
 struct obj_by_name_t
 {
 	unsigned int obj_index;
-	char name[64];
+	char name[TOPMENU_DISPLAY_BUFFER_SIZE];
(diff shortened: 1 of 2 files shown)


asherkin:
At some point this presumably hits an engine/game limitation, do we know the max length for a decent variety of games?

psychonic:
IIRC, max for hudmenu is just under 512. (The buffer is 512, but the data gets sent across in up to two usermessages, with a couple of bytes of overhead on each).

peace-maker:
That limit isn't handled in the normal menu api too iirc. If we were to check for the limit, we shouldn't do that in the topmenu extension but the radiomenu display style. I wouldn't prematurely limit the display text of menu items, but just show less menu items per page, if they don't fit in the buffer?

PR #522, opened 2016-06-20
Fix .travis.yml

galeksandrp:


asherkin:
LLVM disabled the APT repo because it was a significant drain on their bandwidth and disk IO, I'm not comfortable with us downloading a large archive from them on every build instead - and think it's terrible that Travis CI are pushing people to do that (deep-linking to an issue comment bypassing both the reason the APT repo is down, and how to cache the archive).

Cache usage needs to be added before we merge this, but that seems like a snake pit with cache invalidation or the cache servers being slow - we also only build PRs with Travis CI, and it seems they only update the cache from pushes. (Or maybe we could just host the archive on AM? It's ~100M. @psychonic @dvander)

psychonic:
> Or maybe we could just host the archive on AM? It's ~100M

I don't foresee any issues with that.

asherkin:
Thanks for this, we're going with #523 instead.

PR #525, opened 2016-07-02
Add nmethod CloseTimerfor timers

s3bul:


asherkin:
You should use the `delete` keyword from the new syntax in normal use, we don't need an additional stock for the 0.001% of cases where the `autoClose` param is required.

s3bul:
Example of use this method
```sourcepawn
if(timerHandle  != INVALID_HANDLE) CloseTimer(timerHandle);
timerHandle = CreateTimer(0.8, ActionMethod, GetClientSerial(client));
```
so i dont need delete the variable.

asherkin:
delete timerHandle;

will do the same job as

    if (timerHandle != INVALID_HANDLE) {
      KillTimer(timerHandle);
      timerHandle = INVALID_HANDLE;
    }

s3bul:
Ok, i have more learn this sm 1.7 :D

PR #526, opened 2016-07-02
Feature/vectors new method

s3bul:


plugins/include/vector.inc
```diff
@@ -177,3 +177,29 @@ stock void MakeVectorFromPoints(const float pt1[3], const float pt2[3], float ou
 	output[1] = pt2[1] - pt1[1];
 	output[2] = pt2[2] - pt1[2];
 }
+
+/**
+ * Checking whether the vectors are different
+ *
+ * @param vec1		First vector.
+ * @param vec2		Second vector.
+ * @param mistake	Max mistake.
```
yedpodtrzitko: Isn't it called `deviation` in this context usually?
&lt;/bikeshedding&gt;

s3bul: You mean the `mistake` is the wrong name?

yedpodtrzitko: I don't say it's a mistake (lol), but the more common name is `deviation`.

asherkin:
I don't think this is something we're interested in taking upstream - generally feature additions should be discussed before development either on the bug tracker or the forum.

As a note, the FloatCompare native should not be used directly, it is an internal helper for the relational operators.

There is some good information here on how to implement this function properly when the concern is rounding errors: http://floating-point-gui.de/errors/comparison/

PR #528, opened 2016-07-10
Add clearer TFCond: TFCond_HealingDebuff

akowald:
This condition is activated whenever a pyro does direct fire damage to another player. **TFCond_NoTaunting** was probably repurposed by Valve leading up to the last major update. I purpose **TFCond_HealingDebuff**, straight from the server's strings.


asherkin:
See #497, which does a lot more of these and has pending discussion on how to handle it in the long run.

akowald:
I'm fine with that. But what's happening with #497, seems it's stalled.

PR #531, opened 2016-07-27
Added missing information

luki1412:
There is no information about sm_reserve_type 2 in the cfg file, even though reserve type 2 exists in the plugin and on the wikia.

[https://wiki.alliedmods.net/Reserved_Slots_(SourceMod)](https://wiki.alliedmods.net/Reserved_Slots_(SourceMod))


KyleSanderson:
Thanks for the contribution!

PR #532, opened 2016-07-31
Fix /nominations returning Plugin_Continue on success

powerlord:
sm_nominate (/nominate and !nominate) was returning Plugin_Continue when a map was successfully nominated.

This should be Plugin_Handled.


psychonic:
Thanks.

PR #533, opened 2016-08-01
Minor comma removal

404UserNotFound:
One of the latest updates to tf2_stocks.inc added new TF_CUSTOM and TF_WEAPON enum entries. The last entry in each enum ended with a comma where there should have been no comma. Highly insignificant but it bugs the hell out of me.


KyleSanderson:
This is very common to avoid a line edit when a new member is added. If there's no warning or compilation error, this is far from legitimate.

PR #534, opened 2016-08-13
Updated basevotes

luki1412:
Basevotes convars are missing from sourcemod.cfg and there is no config file for this plugin, so I added AutoExecConfig.


KyleSanderson:
Thanks!

PR #536, opened 2016-08-13
Updated sourcemod.cfg

luki1412:
Added missing convars and their description.

### DIFF
--- a/configs/cfg/sourcemod.cfg
+++ b/configs/cfg/sourcemod.cfg
@@ -90,12 +90,39 @@ sm_reserved_slots 0
 // Default: 0
 sm_hide_slots 0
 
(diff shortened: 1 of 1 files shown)


KyleSanderson:
Can't do it, sorry. You're totally allowed to remove those plugins, we shouldn't start throwing errors by default due to a missing convar.

Thanks though man, it was a nice thought.

luki1412:
I wasnt sure whether to add autoexeccfg or an entry in sourcemod.cfg, but there are entries in sourcemod.cfg that require plugins and throw errors due to missing convars already. Currently antiflood.smx, reservedslots.smx, basechat.smx, basecomm.smx, basetriggers.smx have entries in sourcemod.cfg.

PR #537, opened 2016-08-23
Add ICellArray creation/deletion to logic bridge

peace-maker:
Add an ICellArray interface to expose the adt_array CellArray class from logic to core.
~~Add methods to ISourceMod to create and delete ICellArray instances in logic.~~

Now core ~~and extensions~~ can use CellArrays in their forwards.

### DIFF
--- a/public/ISourceMod.h
+++ b/public/ISourceMod.h
@@ -40,10 +40,11 @@
 #include <IHandleSys.h>
 #include <sp_vm_api.h>
 #include <IDataPack.h>
+#include <ICellArray.h>
 #include <time.h>
 
 #define SMINTERFACE_SOURCEMOD_NAME		"ISourceMod"
-#define SMINTERFACE_SOURCEMOD_VERSION	13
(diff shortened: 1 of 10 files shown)


WildCard65:
See Bailo's comment on #92

public/ISourceMod.h
```diff
@@ -318,6 +319,23 @@ namespace SourceMod
...
+		* @return			A new ICellArray object.
+		*/
+		virtual ICellArray *CreateCellArray(size_t blocksize) = 0;
+
+		/**
+		* @brief Releases a cell array's resources so it can be re-used.
+		*
+		* @param pack		An ICellArray object to release.
+		*/
+		virtual void FreeCellArray(ICellArray *arr) = 0;
```
dvander: This looks good for the most part. What extensions need access to cell arrays? If none are known or planned at this time, please drop the ISourceMod.h changes.

peace-maker:
The methods to create and delete CellArrays have been removed from ISourceMod.

PR #539, opened 2016-09-01
Expose SourceMods existing Think hook to Pawn

KyleSanderson:
SourceMod utilizes Think instead of GameFrame to enable plugins to load/unload while the server is hibernating. As an unfortunate side effect, these plugins can't do much if anything while the server is hibernating. Exposing a forward is a simple, requested for years, feature that helps server operators keep hibernation enabled while utilizing SourceMod.

PR #540, opened 2016-09-02
Add ``RequestThink`` and ``OnServerThink``

fakuivan:
Added AddThinkAction to the sm interface and a flag to RequestFrame on pawn to specify a Think hook instead of a GameFrame one. This allows plguins to schedule actions to be executed on the next think call, even if the server is hibernating.

Status: the branch builds no problem but it generates a runtime error while trying to load the sdktools and sdkhooks extensions ("Unable to load extension "sdkhooks.ext": undefined symbol: Warning" and "Unable to load extension "sdktools.ext": undefined symbol: Warning") any help would be appreciated.

### DIFF
--- a/core/sourcemod.cpp
+++ b/core/sourcemod.cpp
@@ -699,6 +699,36 @@ void SourceModBase::ProcessGameFrameHooks(bool simulating)
 	}
 }
 
+void SourceModBase::AddServerThinkHook(GAME_FRAME_HOOK hook)
+{
+	m_think_hooks.push_back(hook);
+}
+
+void SourceModBase::RemoveServerThinkHook(GAME_FRAME_HOOK hook)
+{
+	for (size_t i = 0; i < m_think_hooks.size(); i++)
+	{
+		if (m_think_hooks[i] == hook)
+		{
+			m_think_hooks.erase(m_think_hooks.iterAt(i));
+			return;
+		}
+	}
+}
+
+void SourceModBase::ProcessServerThinkHooks(bool finalTick)
+{
+	if (m_think_hooks.size() == 0)
+	{
+		return;
+	}
+
+	for (size_t i = 0; i < m_think_hooks.size(); i++)
+	{
+		m_think_hooks[i](finalTick);
+	}
+}
+
 size_t SourceModBase::Format(char *buffer, size_t maxlength, const char *fmt, ...)
 {
 	size_t len;
--- a/core/sourcemod.cpp
+++ b/core/sourcemod.cpp
@@ -254,6 +254,7 @@ void SourceModBase::StartSourceMod(bool late)
 {
 	SH_ADD_HOOK(IServerGameDLL, LevelShutdown, gamedll, SH_MEMBER(this, &SourceModBase::LevelShutdown), false);
 	SH_ADD_HOOK(IServerGameDLL, GameFrame, gamedll, SH_MEMBER(&g_Timers, &TimerSystem::GameFrame), false);
+	SH_ADD_HOOK(IServerGameDLL, Think, gamedll, SH_MEMBER(&g_Timers, &TimerSystem::Think), false);
 
 	enginePatch = SH_GET_CALLCLASS(engine);
 	gamedllPatch = SH_GET_CALLCLASS(gamedll);
@@ -543,6 +544,7 @@ void SourceModBase::ShutdownServices()
 
 	SH_REMOVE_HOOK(IServerGameDLL, LevelShutdown, gamedll, SH_MEMBER(this, &SourceModBase::LevelShutdown), false);
 	SH_REMOVE_HOOK(IServerGameDLL, GameFrame, gamedll, SH_MEMBER(&g_Timers, &TimerSystem::GameFrame), false);
+	SH_REMOVE_HOOK(IServerGameDLL, Think, gamedll, SH_MEMBER(&g_Timers, &TimerSystem::Think), false);
 	SH_REMOVE_HOOK(IServerGameDLL, Think, gamedll, SH_MEMBER(logicore.callbacks, &IProviderCallbacks::OnThink), false);
 }
 
@@ -720,6 +722,11 @@ size_t SourceModBase::FormatArgs(char *buffer,
 void SourceModBase::AddFrameAction(FRAMEACTION fn, void *data)
 {
 	::AddFrameAction(FrameAction(fn, data));
+}
+
+void SourceModBase::AddThinkAction(FRAMEACTION fn, void *data)
+{
+	::AddThinkAction(FrameAction(fn, data));
 }
 
 const char *SourceModBase::GetCoreConfigValue(const char *key)
--- a/plugins/include/functions.inc
+++ b/plugins/include/functions.inc
@@ -498,10 +498,20 @@ typedef RequestFrameCallback = function void (any data);
 
 /**
  * Creates a single use Next Frame hook.
+ * @note If you need to queue an action even when the server is hibernating,
+ * Use RequestFrame instead
  *
  * @param Function	Function to call on the next frame.
  * @param data		Value to be passed on the invocation of the Function.
- * @param HookThink	Enable this flag if you want your function to be called 
- *					even if the server is hibernating.
  */
-native void RequestFrame(RequestFrameCallback Function, any data=0, bool HookThink=false);
+native void RequestFrame(RequestFrameCallback Function, any data=0);
+
+/**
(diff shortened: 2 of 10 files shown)


fakuivan:
It seems to be a bug on ambuild with gcc, on it has no runtime errors (as far as loading sm goes) when compiled with clang. [Discussion thread](https://forums.alliedmods.net/showthread.php?t=287145)

fakuivan:
A new update for tf2 servers came out allowing them to enter hibernation mode, making this pr more relevant now. Any reason why this might not be merged yet? I've tested it on Linux and it works fine.

http://www.teamfortress.com/post.php?id=25364

psychonic:
TF2 does not hibernate in the same sense that L4D and CS:GO do due to the engine being tickless. Frames are still processed.

fakuivan:
Well every day you learn something new, this can still be used on other engines though, It's a small tweak, and it works fine. I don't see any reason why you wouldn't merge it.

KyleSanderson:
As mentioned in the thread before you put any time into this, if the rest of SM isn't firing any callbacks: how do you invoke this native.

KyleSanderson:
RequestFrame is a native.
FrameAction is the private function call that happens once, on the next GameFrame.

How do you RequestFrame, when there are no GameFrames, and almost the entire event queue in SM is paused?

Are you actually using this in production? I don't see how this is possible using a proper implementation: which is why this PR has not been merged.

fakuivan:
https://forums.alliedmods.net/showthread.php?t=287015#post2449450

Well, I wouldn't ask for it if it wasn't because I have a plugin running on my servers that require this functionality to work properly... you just needed to read the thread that you furiously commented on. Is that too much to ask?

asherkin:
Kyle isn't asking if you require this, that is obvious - he is asking if you are using *this patch* on your servers currently.

fakuivan:
Yes, I currently have a "custom" build of sourcemod installed on them, running the plugin that I just mentioned.

public/ISourceMod.h
```diff
@@ -289,6 +289,16 @@ namespace SourceMod
...
 
 		/**
+		* @brief Adds an action to be executed on the next available think call.
+		*
+		* This function is thread safe.
+		*
+		* @param fn			Function to execute.
+		* @param data		Data to pass to function.
+		*/
+		virtual void AddThinkAction(FRAMEACTION fn, void *data) = 0;
```
peace-maker: When changing a public interface, you should always remember to bump the interface's version number.
`SMINTERFACE_SOURCEMOD_VERSION` at the top in this case.

dvander:
Two comments, ignoring the idea itself:
 - Adding bool flags is discouraged. Either the name should be less mechanical (maybe "runWhenIdle"), or it should be a new native.
 - This breaks the layout of the ISourceMod vtable. New functions must be added to the end.

core/sourcemod.cpp
```diff
@@ -254,6 +254,7 @@ void SourceModBase::StartSourceMod(bool late)
 {
 	SH_ADD_HOOK(IServerGameDLL, LevelShutdown, gamedll, SH_MEMBER(this, &SourceModBase::LevelShutdown), false);
 	SH_ADD_HOOK(IServerGameDLL, GameFrame, gamedll, SH_MEMBER(&g_Timers, &TimerSystem::GameFrame), false);
+	SH_ADD_HOOK(IServerGameDLL, Think, gamedll, SH_MEMBER(&g_Timers, &TimerSystem::Think), false);
```
peace-maker: You should consider moving this to the logic binary like in #539 so we don't hook Think twice.

peace-maker:
I think this would go in addition to a global `OnServerThink` forward from the previous incomplete PR. I don't know if switching SM's entire internal event dispatching to a Think hook instead of a Frame one is what people want. Still this moves some subtle details to the plugin authors which I feel shouldn't have to bother with hibernation in their code explicitly.

KyleSanderson:
Speaking for myself:

Changing SM to support hibernation now is tricky. The arguably vocal minority say there's a previous API guarantee saying Timers, SQL, and other callbacks are bound to a GameFrame occuring. Personally, I disagree that this was a guarantee from SM but is rather an implementation detail.

The flip side of the coin is we've had so much churn recently (past couple years) that people are afraid to update SM to master due to sporadic issues, like the runnable plague that was recently quelled. If changing this behaviour is going to further upset the balance: we shouldn't do it. Honestly we don't have enough information (at the moment), to confidently determine what the path forward is.

Thoughts? DBI already queries on Think, just doesn't run any callbacks (iirc, might be wrong now).

core/sourcemod.cpp
fakuivan: or calling ``&IProviderCallbacks::OnThink`` from ``&IProviderCallbacks::OnThink`` ?

fakuivan:
* [Move ``AddThinkAction`` down because vtables](https://github.com/alliedmodders/sourcemod/pull/540/commits/64c4fecb088f0d4d07b995815a42568c157769ce)
* [Ditch the ``HookThink`` flag for a new native](https://github.com/alliedmodders/sourcemod/pull/540/commits/6d94d7ba5dbda9ed4d18529d74f1d10d24122d0f)

@dvander How about now?

plugins/include/functions.inc
```diff
@@ -498,8 +498,20 @@ typedef RequestFrameCallback = function void (any data);
 
 /**
  * Creates a single use Next Frame hook.
+ * @note If you need to queue an action even when the server is hibernating,
+ * Use RequestFrame instead
```
fakuivan: RequestThink*

public/ISourceMod.h
```diff
@@ -318,6 +318,30 @@ namespace SourceMod
...
+
+		/**
+		* @brief Adds a function to be called each server think.
+		*
+		* @param hook		Hook function.
+		*/
+		virtual void AddServerThinkHook(GAME_FRAME_HOOK hook) = 0;
```
GoD-Tony: Will you need a new typedef here or is reusing `GAME_FRAME_HOOK` fine?

fakuivan: Both ``GameFrame`` and ``Think`` take and return the same number of parameters so it should be ok, I could add one but I've reused most of the symbols that frame hooks uses so I think It would be more confusing to use a different symbol here and reuse others elsewhere.

KyleSanderson: They take the same params, but GAME_FRAME_HOOK is a define which stipulates "simulating" as the parameter which is undesirable because the definitions are different.

Please use another define, then _I Think_ it should be fine.

ramonberrutti:
Hi, this will be added? I'm needing this feature for public release :)

asherkin:
> Hi, this will be added? I'm needing this feature for public release :)

It is still under discussion, I think the current leaning is to change GameFrame globally to Think, we have to be cautious because going either way this is a hard one to go back on — which actually points to making the global change first anyway, as if that doesn’t work it is easier to revert and merge this new native than go the other way.

fakuivan:
@asherkin Wouldn't it be easier and less risky to just implement this? After all the current feature is called ``OnGameFrame``.

fakuivan:
@ramonberrutti You can use [this](https://github.com/fakuivan/Think-Hooks) in the meantime.

ramonberrutti:
Thanks @fakuivan I don't need for me, I need for public release, I build a (Redis & Mongo Module) for Sourcemod, and I need to have a loop for the Pub/Sub of Redis. (Yes now I'm hooking the Think method, like you do in that module) but the idea is to no include Sourcehook in the library. Also if a player want to set task with the Sub of redis will can't.

KyleSanderson:
@asherkin how do you feel about this one now?

KyleSanderson:
@asherkin @Fyren lets ice this PR and give @fakuivan some closure after 4 years if we can. What I'm thinking is ultimately this (not this PR specifically - but feature) should have indeed been promoted nearly half a decade ago and frames are not tied to when the server is running production frames, but when think occurs we invoke OnGameFrame / SM logic alike. This is an ancient bug, and while intentions were correct future Valve games changed hibernation definitions and we should have adapted more agile. I'm going to give this some more time but ultimately I'm going to finally change the backend behaviour if no one's objecting as this is still a major pain point for our consumers.

asherkin:
Thanks for getting the (very slow) ball rolling with this, it was definitely the start of the modern discussion around changes here. As suggested back in 2017, SM will now run things as normal even if hibernating and that'll roll out with 1.11 becoming the stable release.

PR #541, opened 2016-09-16
Add cstrike_stocks.inc

Bara20:
Add CS_GetEntitySpeed + CS_SetEntitySpeed

I don't which games has m_flLaggedMovementValue.


WildCard65:
Probably should have it match tf2_stocks.inc in that cstrike_stocks should include cstrike

plugins/include/cstrike_stocks.inc
```diff
@@ -0,0 +1,96 @@
...
+}
+
+/**
+ * Sets an entity's speed.
+ *
+ * @param entity	Entity index.
+ * @param amount	Speed to set (default = 1.0, half = 0.5, double = 2.0).
+ * @error			Invalid entity index, or lack of mod compliance.
+ */
+stock void SetEntitySpeed(int entity, float amount)
```
WildCard65: Missing CS_ prefix?

KyleSanderson:
We appreciate the thought, however this has serious repercussions such as impacting not only movement; but fall speed. If you're able to bypass m_flMaxspeed's restriction (or similar) and have it stick like this; that would absolutely be a welcomed addition. Changing this prop only results in quite unfortunate UX, so I'm closing this.

Drifter321:
Alright, lets start with why are stocks required? AFAIK the prop exists on both cs:go and cs:s, so why is this needed? Also this isnt a great way to set a players speed.

PR #545, opened 2016-10-02
Strip invalid UTF-8 char from end of player names.

psychonic:
Steam supports player names up to 32 UTF-8 characters, excluding the NULL terminator. The Source Engine supports either 32 or 128 bytes for a name (depending on version), including the NULL terminator.

In all cases, this can lead to the engine copy of the name ending in the middle of a character due to the truncation. When this occurs, rather than passing this copy of the name to the GetClientName native and the %N format specifier, let's fix our stored copy.

### DIFF
--- a/core/PlayerManager.cpp
+++ b/core/PlayerManager.cpp
@@ -1963,7 +1963,6 @@ CPlayer::CPlayer()
 void CPlayer::Initialize(const char *name, const char *ip, edict_t *pEntity)
 {
 	m_IsConnected = true;
-	m_Name.assign(name);
 	m_Ip.assign(ip);
 	m_pEdict = pEntity;
 	m_iIndex = IndexOfEdict(pEntity);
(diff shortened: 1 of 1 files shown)


core/PlayerManager.cpp
```diff
@@ -2135,7 +2136,30 @@ void CPlayer::Disconnect()
...
+	while (i < sizeof(szNewName))
+	{
+		if (szNewName[i] == 0)
+			break;
+
+		unsigned int cCharBytes = _GetUTF8CharBytes(&szNewName[i]);
+		size_t newPos = i + cCharBytes;
+		if (newPos > (sizeof(szNewName) - 1))
+		{
+			szNewName[i] = 0;
```
KyleSanderson: You should do this after the loop regardless, because if you break normally this will not be written to the array.

psychonic: The array is already NULL terminated. This is just for adding it sooner.

KyleSanderson: Right; sorry. lgtm

I kind of wonder about targeting though. If you copy the entire string we won't find the name anymore because the names would vary between the engine and SM. This is if the last, invalid character is actually sent to the client.

psychonic: The last character wouldn't be printable, so it shouldn't exist in display anywhere to copy from.

PR #548, opened 2016-10-08
Use newer MM:S api for Ep1/Original engine.

psychonic:
These are the only engine versions that are using the legacy MM:S API. Now all use the same.

See also: https://github.com/alliedmodders/metamod-source/pull/29

### DIFF
--- a/loader/loader.cpp
+++ b/loader/loader.cpp
@@ -58,7 +58,6 @@
 #endif
 
 #define METAMOD_API_MAJOR			2
-#define FILENAME_1_6_EP1			"sourcemod.2.ep1" PLATFORM_EXT
 #define FILENAME_1_6_EP2			"sourcemod.2.ep2" PLATFORM_EXT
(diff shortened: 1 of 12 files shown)


loader/loader.cpp
```diff
@@ -58,15 +54,11 @@
...
-	inline bool IsPathSepChar(char c) 
-	{
-		return (c == '/');
-	}
 	#include <dlfcn.h>
 #endif
 
 #define METAMOD_API_MAJOR			2
-#define FILENAME_1_4_EP1			"sourcemod.1.ep1" PLATFORM_EXT
+#define FILENAME_1_6_EP1			"sourcemod.2.ep1" PLATFORM_EXT
```
peace-maker: That's doubled up now.

psychonic: Good catch.

KyleSanderson:
:shipit:

PR #549, opened 2016-10-10
Add command argument autocompletion for plugins

peace-maker:
Add an additional callback argument to Reg*Cmd natives to provide suggestions for possible arguments.
Currently only the SRCDS windows GUI supports the autocompletion feature properly. You need a [seperate extension](https://forums.alliedmods.net/showthread.php?t=286937) to enable it on the command line versions. This only works in the server console.

In the callback

``` sourcepawn
typedef CommandCompletionCallback = function void (const char[] cmd, int args, ArrayList suggestions);
```

plugin authors can use the `GetCmdArg*` natives to get the (partial) arguments just like in the normal command callback and add possible arguments to the `suggestions` array.

There is a new native to get the argument suggestions for arbitary commands too.

``` sourcepawn
native ArrayList GetCmdAutoCompleteSuggestions(const char[] partial);
```

The autocompletion capability is exposed to logic too through the bridge. A future PR might add argument autocompletion for the rootmenu.

~~At the moment `episode1` and `darkm` engine branches aren't supported when trying to hook autocompletion for existing commands due to a limitation in SourceHook and fixed-size array parameters.~~

### DIFF
--- a/bridge/include/BridgeAPI.h
+++ b/bridge/include/BridgeAPI.h
@@ -35,7 +35,7 @@ namespace SourceMod {
 
 // Add 1 to the RHS of this expression to bump the intercom file
 // This is to prevent mismatching core/logic binaries
-static const uint32_t SM_LOGIC_MAGIC = 0x0F47C0DE - 55;
+static const uint32_t SM_LOGIC_MAGIC = 0x0F47C0DE - 56;
 
 } // namespace SourceMod
 
--- a/bridge/include/CoreProvider.h
+++ b/bridge/include/CoreProvider.h
@@ -31,6 +31,8 @@
 #include <stdint.h>
 #include <IAdminSystem.h>
 #include <amtl/am-function.h>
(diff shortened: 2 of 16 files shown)


peace-maker:
I've just remembered that this requires changes in the OBV branch games' SDKs. The CUtlVector and/or CUtlString classes are out of date and need to be updated from the SDK2013. In my case, replacing the whole tier1 library with the one from SDK2013 did the job for these games.

peace-maker:
Closing this since it's only adding value to the windows srcds gui console by default, which doesn't appear to be used too often. Giving plugins the possibility to provide autocompletion suggestions for the console commands they register is only useful, if the server is running a [third party extension](https://forums.alliedmods.net/showthread.php?t=286937) to be able to access those suggestions from the server console.

While this has proven to be really useful and a timesaver after adding autocompletion to the `sm` command root menu, it appears to be too much to be in the core by the lack of interest in this PR.

PR #550, opened 2016-10-10
Add option for persistent previous maps

stickz:
This commit adds an option to the mapchooser allowing for previous maps to be stored persistently. In the event of plug-in termination, it will remember what the previous maps played were. And recall them like nothing happened, if the user has the corresponding ConVar enabled. 

It solves the annoyance of having to start from scratch with previous excluded maps, when something like a server reboot or crash happens; or even if the plug-in is updated/reloaded for whatever reason.

### DIFF
--- a/plugins/mapchooser.sp
+++ b/plugins/mapchooser.sp
@@ -1276,7 +1276,11 @@ void WritePreviousMapsToText()
 	
 	if (file == null)
 	{
-		LogError("[SM] Error writing to %s. Check file permissions.", MAPCHOOSER_TXT);
+		if (g_Cvar_PersistentMaps.BoolValue)
+		{
+			LogError("[SM] Error writing to %s. Check file permissions.", MAPCHOOSER_TXT);
+		}
+		
 		return;
 	}
     


plugins/mapchooser.sp
```diff
@@ -1223,3 +1242,63 @@ public int Native_GetNominatedMapList(Handle plugin, int numParams)
...
+    	g_OldMapList.Clear();
+            
+    	char path[PLATFORM_MAX_PATH];
+    	BuildPath(Path_SM, path, PLATFORM_MAX_PATH, MAPCHOOSER_TXT);
+        
+    	File file = OpenFile(path, "r");
+    
+    	char map[64];
+    
+  	while (!file.EndOfFile() && file.ReadLine(map, sizeof(map)))
```
KyleSanderson: EndOfFile is undefined if you haven't read a line.

plugins/mapchooser.sp
```diff
@@ -1223,3 +1242,63 @@ public int Native_GetNominatedMapList(Handle plugin, int numParams)
...
+    	file.Close();
+}
+
+// Write the previous maps to a text file
+void WritePreviousMapsToText()
+{    
+    	char path[PLATFORM_MAX_PATH];
+    	BuildPath(Path_SM, path, PLATFORM_MAX_PATH, MAPCHOOSER_TXT);
+        
+   	DeleteFile(path);
```
KyleSanderson: If you're opening in write mode, as you're doing below, you don't need to delete the existing file.

plugins/mapchooser.sp
```diff
@@ -1223,3 +1242,63 @@ public int Native_GetNominatedMapList(Handle plugin, int numParams)
...
+        
+    	File file = OpenFile(path, "w");
+    
+    	char lastMap[64];
+    
+    	for (int idx = 0; idx < g_OldMapList.Length; idx++)
+    	{
+        	g_OldMapList.GetString(idx, lastMap, sizeof(lastMap));
+
```
KyleSanderson: Why is there so much spacing?

plugins/mapchooser.sp
```diff
@@ -1223,3 +1242,63 @@ public int Native_GetNominatedMapList(Handle plugin, int numParams)
...
+// Create a text file (if not present)
+void CreatePreviousMapsTextFile()
+{
+	char path[PLATFORM_MAX_PATH];
+   	BuildPath(Path_SM, path, PLATFORM_MAX_PATH, MAPCHOOSER_TXT);
+        
+    	if (!FileExists(path))
+    	{      
+    	    	File file = OpenFile(path, "w");
+		file.Close();
```
KyleSanderson: What is going on here? FileExist should be used instead of creating empty files, that you later remove and re-add.

plugins/mapchooser.sp
```diff
@@ -232,8 +239,7 @@ public void OnConfigsExecuted()
 	for (int i=0; i<MAXTEAMS; i++)
 	{
 		g_winCount[i] = 0;	
-	}
-	
+	}
```
KyleSanderson: What happened here?

stickz: I think an unneeded space got removed.

plugins/mapchooser.sp
stickz: This function creates the text file, if it's not existent OnPluginStart(), to avoid no read/write errors.

plugins/mapchooser.sp
stickz: I just put in there as a safety feature, to ensure the loop terminates when the last file line is reached. There isn't any errors while running; but that can be removed, if it's not required.

plugins/mapchooser.sp
stickz: Alright I'll remove that.

plugins/mapchooser.sp
stickz: Fixed. The code is bunched closer together, so it's harder to read now.

plugins/mapchooser.sp
```diff
@@ -105,6 +107,8 @@ int g_winCount[MAXTEAMS];
 #define VOTE_EXTEND "##extend##"
 #define VOTE_DONTCHANGE "##dontchange##"
 
+#define MAPCHOOSER_TXT "data/mapchooser.txt"
```
peace-maker: I think `maphistory.txt` would be a better name than just `mapchooser.txt`

plugins/mapchooser.sp
```diff
@@ -126,6 +130,7 @@ public void OnPluginStart()
 	g_Cvar_ExtendRoundStep = CreateConVar("sm_extendmap_roundstep", "5", "Specifies how many more rounds each extension makes", _, true, 1.0);
 	g_Cvar_ExtendFragStep = CreateConVar("sm_extendmap_fragstep", "10", "Specifies how many more frags are allowed when map is extended.", _, true, 5.0);	
 	g_Cvar_ExcludeMaps = CreateConVar("sm_mapvote_exclude", "5", "Specifies how many past maps to exclude from the vote.", _, true, 0.0);
+	g_Cvar_PersistentMaps = CreateConVar("sm_mapvote_persisentmaps", "0", "Specifies  whether or not previous maps should be persistently stored.", _, true, 0.0, true, 1.0);
```
peace-maker: Nit: two spaces after `Specifies`

plugins/mapchooser.sp
peace-maker: You removed the new line but kept the tab after the closing brace. Could just leave it as it was 😉

plugins/mapchooser.sp
```diff
@@ -1223,3 +1242,59 @@ public int Native_GetNominatedMapList(Handle plugin, int numParams)
...
+
+// Read previous maps from text file
+void ReadPreviousMapsFromText()
+{
+    	g_OldMapList.Clear();
+            
+    	char path[PLATFORM_MAX_PATH];
+    	BuildPath(Path_SM, path, PLATFORM_MAX_PATH, MAPCHOOSER_TXT);
+        
+    	File file = OpenFile(path, "r");
```
peace-maker: You should add a null check here, if the file can't be opened and return. That lets you get rid of that `CreatePreviousMapsTextFile` function too!

plugins/mapchooser.sp
```diff
@@ -1223,3 +1242,59 @@ public int Native_GetNominatedMapList(Handle plugin, int numParams)
...
+    	file.Close();
+}
+
+// Write the previous maps to a text file
+void WritePreviousMapsToText()
+{    
+    	char path[PLATFORM_MAX_PATH];
+    	BuildPath(Path_SM, path, PLATFORM_MAX_PATH, MAPCHOOSER_TXT);
+  
+    	File file = OpenFile(path, "w");
```
peace-maker: Add null check here too. User could always be lacking write permissions in the data folder or for that file specifically.

plugins/mapchooser.sp
stickz: I chose `mapchooser.txt` because it's less likely to conflict with other plugins.

plugins/mapchooser.sp
stickz: Done. I told the plugin to log an error as well, so the user knows something is wrong.

plugins/mapchooser.sp
```diff
@@ -1223,3 +1240,58 @@ public int Native_GetNominatedMapList(Handle plugin, int numParams)
...
+void ReadPreviousMapsFromText()
+{
+	char path[PLATFORM_MAX_PATH];
+	BuildPath(Path_SM, path, PLATFORM_MAX_PATH, MAPCHOOSER_TXT);
+        
+	File file = OpenFile(path, "r");
+	
+	if (file == null)
+	{
+		file.Close();
```
KyleSanderson: If the file is null you don't need to close it...

plugins/mapchooser.sp
```diff
@@ -1223,3 +1240,58 @@ public int Native_GetNominatedMapList(Handle plugin, int numParams)
...
+		return;
+	}
+
+	g_OldMapList.Clear();
+	char map[64];
+
+	while (!file.EndOfFile() && file.ReadLine(map, sizeof(map)))
```
KyleSanderson: This needs to be a do while. You cannot call EndOfFile before reading a line, this is present in almost every libc.

plugins/mapchooser.sp
```diff
@@ -1223,3 +1240,58 @@ public int Native_GetNominatedMapList(Handle plugin, int numParams)
...
+{    
+	char path[PLATFORM_MAX_PATH];
+	BuildPath(Path_SM, path, PLATFORM_MAX_PATH, MAPCHOOSER_TXT);
+  
+	File file = OpenFile(path, "w");
+	
+	if (file == null)
+	{
+		LogError("[SM] Error writing to %s. Check file permissions.", MAPCHOOSER_TXT);
+		file.Close();
```
KyleSanderson: Ditto on calling CloseHandle against an INVALID_HANDLE: why?

plugins/mapchooser.sp
```diff
@@ -243,6 +247,17 @@ public void OnConfigsExecuted()
 			LogError("Warning - Bonus Round Time shorter than Vote Time. Votes during bonus round may not have time to complete");
 		}
 	}
+	
+	/* Recall previous maps from a text file, if persistentcy is enabled */
+	if (g_FirstConfigExec)
```
KyleSanderson: can be a static.

plugins/mapchooser.sp
```diff
@@ -126,6 +129,7 @@ public void OnPluginStart()
 	g_Cvar_ExtendRoundStep = CreateConVar("sm_extendmap_roundstep", "5", "Specifies how many more rounds each extension makes", _, true, 1.0);
 	g_Cvar_ExtendFragStep = CreateConVar("sm_extendmap_fragstep", "10", "Specifies how many more frags are allowed when map is extended.", _, true, 5.0);	
 	g_Cvar_ExcludeMaps = CreateConVar("sm_mapvote_exclude", "5", "Specifies how many past maps to exclude from the vote.", _, true, 0.0);
+	g_Cvar_PersistentMaps = CreateConVar("sm_mapvote_persisentmaps", "0", "Specifies whether or not previous maps should be stored persistently.", _, true, 0.0, true, 1.0);
```
schrodingerballs: "You could learn how to spell things"
- stickz, 2016

plugins/mapchooser.sp
```diff
@@ -1223,3 +1240,62 @@ public int Native_GetNominatedMapList(Handle plugin, int numParams)
 
 	return;
 }
+
+/* Persisently store previous maps incase user wants to recall them */
```
schrodingerballs: > You could learn how to spell things

stickz, 2016

plugins/mapchooser.sp
```diff
@@ -263,6 +278,8 @@ public void OnMapEnd()
 	{
 		g_OldMapList.Erase(0);
 	}	
+	
+	WritePreviousMapsToText();
```
peace-maker: You might only want to write/create the file if the convar is set to 1. Otherwise people might get error spam from a disabled feature.

stickz: I'll compensate for that by only logging an error if the feature is enabled. It's best to write to text regardless, because of something happens between ConVar change and map end, it won't save.

peace-maker: If the feature is disabled, it shouldn't change the file. Maybe some admin doesn't want a map to appear in the history and disables that feature for that map?

stickz: In that instance, the map would still be in history because it's read from the ArrayList. Preventing it from writing to text would have absolutely no effect, unless an abnormal event happened. The proper way to accomplish such a task, would be to pop it then update the text file afterwards.

plugins/mapchooser.sp
Headline: `mapchooser_history.txt` ?

PR #553, opened 2016-10-16
Try to call even less functions in paused plugins

peace-maker:
See #489
There are even more callbacks and hooks in sourcemod that just keep
calling the callbacks in paused plugins.
This should add a `IsRunnable()` check to all of them.


peace-maker:
Some thirdparty extensions like dhooks keep calling functions of paused plugins too. Maybe it's better to just silence the `SP_ERROR_NOT_RUNNABLE` error in the vm instead of working around it in the code calling into it.

dvander:
Yes.

peace-maker:
In the VM or in SM's ReportError? @dvander

PR #554, opened 2016-10-17
Added AutoExecConfig to each plugin that has its own ConVars

luki1412:
Removed each ConVar that requires a plugin from sourcemod.cfg, so the console doesnt print Unknown command for ConVars from plugins that aren't loaded.
Added AutoExecConfig to each plugin that has its own ConVars.
Expanded description for 2 reservedslots ConVars.

### DIFF
--- a/configs/cfg/sourcemod.cfg
+++ b/configs/cfg/sourcemod.cfg
@@ -50,68 +50,6 @@ sm_immunity_mode 1
 // Default: 0
 sm_time_adjustment 0
 
-// Specifies the amount of time that is allowed between chat messages.  This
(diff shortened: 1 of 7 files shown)


peace-maker:
Thank you for your contribution! Sorry for the insanely long delay.

While this would certainly be the correct way to handle convars of the stock plugins, the possibilty to split the convars into multiple files is long gone. This will cause too much confusion with convars being set twice in two config files to potentially different values due to admins not updating their sourcemod.cfg. If there's a sane way to introduce this, I'd be all in, but I don't see any path for a smooth transition. So we'll have to live with those unbound convar warnings 😞

PR #555, opened 2016-10-19
Update sm_psay (better compatibility for csgo)

Bara:
Because " doesn't work in chat and will replaced with '

### DIFF
--- a/plugins/basechat.sp
+++ b/plugins/basechat.sp
@@ -290,10 +290,12 @@ public Action Command_SmPsay(int client, int args)
 	}
 	
 	char buffer[100];
+	buffer[0] = ' ';
+	
 	for (int i = 2; i <= args; i++)
 	{
-		GetCmdArg(i, buffer, sizeof(buffer));
-		Format(message, sizeof(message), "%s %s", message, buffer);
+		GetCmdArg(i, buffer, sizeof(buffer) - 1);
+		StrCat(message, sizeof(message), buffer);
 	}
 	
 	SendPrivateChat(client, target, message);
--- a/plugins/basechat.sp
+++ b/plugins/basechat.sp
@@ -277,20 +277,28 @@ public Action Command_SmPsay(int client, int args)
 		return Plugin_Handled;	
 	}	
 	
-	char text[192], arg[64], message[192];
+	char text[192], sTarget[64], message[192];
 	GetCmdArgString(text, sizeof(text));
 
-	int len = BreakString(text, arg, sizeof(arg));
-	BreakString(text[len], message, sizeof(message));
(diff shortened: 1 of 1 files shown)


plugins/basechat.sp
```diff
@@ -277,20 +277,28 @@ public Action Command_SmPsay(int client, int args)
...
-		return Plugin_Handled;	
+	{
+		return Plugin_Handled;
+	}
+	
+	char buffer[100];
+	for (int i = 2; i <= args; i++)
+	{
+		GetCmdArg(i, buffer, sizeof(buffer));
+		Format(message, sizeof(message), "%s %s", message, buffer);
```
WildCard65: Better to do the following (with above change):
```sourcepawn
StrCat(message, sizeof(message), buffer)
```

Note: Didn't want to click "Start a review" button so I deleted last one and reposting it.

Bara20: If i add StrCat....
/psay Player Text1 Text2
Output (without a space):
(Private to Player) Bara: Text1Text2

plugins/basechat.sp
```diff
@@ -277,20 +277,28 @@ public Action Command_SmPsay(int client, int args)
...
 	if (target == -1)
-		return Plugin_Handled;	
+	{
+		return Plugin_Handled;
+	}
+	
+	char buffer[100];
+	for (int i = 2; i <= args; i++)
+	{
+		GetCmdArg(i, buffer, sizeof(buffer));
```
Bara20: You mean "buffer[0] = '\0';"?

WildCard65: no:
```sourcepawn
char buffer[100];
buffer[0] = ' '; // Very first character is a whitespace character.
for (int i = 2; i <= args; i++)
{
    GetCmdArg(i, buffer[1], sizeof(buffer) - 1); // This will put arg string into buffer starting at index of 1, this will leave character in buffer[0] what ever it was before call.
    // Rest of code here.
}
```

plugins/basechat.sp
```diff
@@ -290,10 +290,12 @@ public Action Command_SmPsay(int client, int args)
 	}
 	
 	char buffer[100];
+	buffer[0] = ' ';
+	
 	for (int i = 2; i <= args; i++)
 	{
-		GetCmdArg(i, buffer, sizeof(buffer));
-		Format(message, sizeof(message), "%s %s", message, buffer);
+		GetCmdArg(i, buffer, sizeof(buffer) - 1);
```
WildCard65:
```sourcepawn
GetCmdArg(i, buffer[1], sizeof(buffer) - 1); //Pass buffer starting from index 1 instead of 0.
```

PR #558, opened 2016-10-27
Suppress logging of "Plugin not runnable" errors

peace-maker:
SourceMod itself and thirdparty extensions call functions in plugins
without making sure it is runnable. The plugin pausing feature isn't
supported anymore and is set to get nuked #438 which will reduce the
amount of not-runnable error reports.

We can't fix every third party extension, so suppressing the error from
the logs will stop the spam until pausing is removed.

#489 can be reverted after this.

### DIFF
--- a/core/logic/DebugReporter.cpp
+++ b/core/logic/DebugReporter.cpp
@@ -157,6 +157,12 @@ int DebugReport::_GetPluginIndex(IPluginContext *ctx)
 
 void DebugReport::ReportError(const IErrorReport &report, IFrameIterator &iter)
 {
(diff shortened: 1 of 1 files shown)


dvander:
Looks good, but do you need to pull a new version of SP into the subrepo?

peace-maker:
Yes, I thought that would be done outside of this PR. Updated the submodule.

peace-maker:
In SM 1.8 one could work around this by comparing the error message to "Plugin not runnable" and checking if there's no plugin to blame at all?

PR #560, opened 2016-11-17
Suppress logging of "Plugin not runnable" errors in SM 1.8

peace-maker:
This is a backport of #558 to the 1.8 branch.

Since we can't really easily change the VM in 1.8 anymore (right?), this does a string comparison against the error message to know what to suppress. Also tries to not suppress errors manually generated by plugins which happen to start with the same phrase.

### DIFF
--- a/core/logic/DebugReporter.cpp
+++ b/core/logic/DebugReporter.cpp
@@ -180,6 +180,13 @@ void DebugReport::ReportError(const IErrorReport &report, IFrameIterator &iter)
 	}
 
 	iter.Reset();
+	
+	// Don't log an error if a function wasn't runnable.	
+	// This is necassary due to the way SM is handling and exposing
+	// scripted functions. It's too late to change that now.
+	// Make sure this error wasn't manually thrown by a plugin.
+	if (!blame && !strcmp(report.Message(), "Plugin not runnable"))
+		return;
 
 	g_Logger.LogError("[SM] Exception reported: %s", report.Message());
 


core/logic/DebugReporter.cpp
```diff
@@ -180,6 +180,13 @@ void DebugReport::ReportError(const IErrorReport &report, IFrameIterator &iter)
 	}
 
 	iter.Reset();
+	
+	// Don't log an error if a function wasn't runnable.	
+	// This is necassary due to the way SM is handling and exposing
+	// scripted functions. It's too late to change that now.
+	// Make sure this error wasn't manually thrown by a plugin.
+	if (!strcmp(report.Message(), "Plugin not runnable") && !blame)
```
KyleSanderson: You almost always want to do the cheaper null check before doing an array comparison.

psychonic: If error reporting becomes a hot path, the game server has bigger issues.

dvander:
Regardless of performance it doesn't seem like a great idea. On 1.9 I hope we landed some change to IErrorReport and not this.

core/logic/DebugReporter.cpp
KyleSanderson: Believe it or not, but this is a hot path and people have reported missed frames. Yes, problematic plugins, never the less...

peace-maker:
Reversed the check order @KyleSanderson 

> Regardless of performance it doesn't seem like a great idea. On 1.9 I hope we landed some change to IErrorReport and not this.

Yes, like you suggested #558

peace-maker:
People keep contacting me about this one.

PR #561, opened 2016-11-29
Add DETOUR_DECL_* marcos for more parameter counts

peace-maker:
Add more macros for up to 8 parameters, because DETOUR_DECL_MEMBER8 is
the highest one yet.


KyleSanderson:
Absolutely nothing wrong with adding this; but do you have a present consumer?

peace-maker:
Meee
https://github.com/peace-maker/srcds-autocomplete/blob/master/extension.cpp#L75

PR #562, opened 2016-12-01
Update IDA makesig script

GoD-Tony:
This fixes the error: `makesig.idc:47: No such attribute: object.Op0`

Apparently this scripting language doesn't support short-circuit evaluation, so the not-null check in this (`pInfo.n == 1`) case has to be nested in `if` statements.


asherkin:
Quality programming language!!

PR #563, opened 2016-12-02
Fix case of not calling OnMapTimeLeftChanged in CS:GO

peace-maker:
Round end reasons in CS:GO are +1 from the ones in CS:S. The
CSRoundEnd_GameStart reason for when the OnMapTimeLeftChanged forward
should be called should be one higher than in CS:S.


KyleSanderson:
Thanks!

PR #564, opened 2016-12-04
Fixed description for CharToLower and CharToUpper

ErikMinekus:
Follow-up for 098a693.


psychonic:
I'm not crazy about the phrasing of these, and it was probably the cause for the initial mix-up. Regardless, at least making it accurate is definitely an improvement. Thanks!.

PR #565, opened 2016-12-04
Add support for more SQLite database open options.

psychonic:
As requested here, https://forums.alliedmods.net/showthread.php?t=265261

Tested with:
```
public OnPluginStart()
{
	RegServerCmd("sqlite_memtest", sqlite_memtest);
}

public Action sqlite_memtest(int argc)
{
	char error[1024];
	Database hi = SQLite_UseDatabase(":memory:", error, sizeof(error));  
	if (hi == INVALID_HANDLE)
	{
		PrintToServer("Failed: %s", error);
	}
	else
	{
		PrintToServer("Success!");
	}
	
	return Plugin_Handled;
}
```

### DIFF
--- a/extensions/sqlite/AMBuilder
+++ b/extensions/sqlite/AMBuilder
@@ -12,7 +12,9 @@ elif binary.compiler.vendor == 'msvc':
 
 binary.compiler.defines += [
   'SQLITE_OMIT_LOAD_EXTENSION',
-  'SQLITE_THREADSAFE'
+  'SQLITE_THREADSAFE',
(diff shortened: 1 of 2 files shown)


asherkin:
🚢 

I wonder if we should pass through [URI filenames](https://www.sqlite.org/uri.html) as well? (Looks like just checking if it starts with "file:" should do it.)

psychonic:
@asherkin Good idea.

Added that now, too. I retested :memory: and name, as well as testing a file URI. All seemed to work well.

PR #566, opened 2016-12-09
Fix crash in CloneArray for too large arrays

peace-maker:
If there is not enough memory to clone an array, throw an error instead
of crashing the server.


KyleSanderson:
Thanks!

PR #568, opened 2016-12-12
Change sm_msay behavior to use GetMaxPageItems instead of hardcoded values

iNilo:
Fix for CS:GO not showing the sm_msay exit button

mSayPanel.CurrentKey = 10:
mSayPanel.CurrentKey = 10

mSayPanel.CurrentKey = 9:
mSayPanel.CurrentKey = 9


psychonic:
Good catch on the issue. Instead of hardcoding a number, we should use the MaxMenuItems gamedata key that already exists. (That may not be the exact name)

iNilo:
As in https://sm.alliedmods.net/new-api/menus/GetMaxPageItems ?
It is not exposed in the Panel methodmaps

iNilo:
Implemented.

iNilo:
Renamed PR

psychonic:
Looks good. Thanks!

PR #570, opened 2016-12-22
Fix Offsets for CommitSuicide TF

Jonhasacat:


Jonhasacat:
Not sure why it's showing all those files as edited, the previous edits were already merged into the main repo.

The only file that should actually be different is game.tf.txt

KyleSanderson:
@Jonhasacat Right now, no files are different. This probably isn't intentional?

Are you sure CommitSuicide is broken?

psychonic:
There's no change here, as seen on the Files Changed tab.

The TF2 offset for CommitSuicie hasn't changed.

Jonhasacat:
AFK Manager isn't properly suiciding players when they are switched to spec.

After testing this further, it doesn't appear to be working on my end either

PR #571, opened 2016-12-26
Fix incomplete TextMsg packet

thexa4:
The source 2013 handler for TextMsg expects five strings in the message here: https://github.com/ValveSoftware/source-sdk-2013/blob/master/mp/src/game/client/hud_chat.cpp#L124

Because sourcemod doesn't send the last four strings along each message sent from sourcemod triggers an assert in tier1/bitbuf.cpp if the game is not built with RELEASE.

### DIFF
--- a/core/HalfLife2.cpp
+++ b/core/HalfLife2.cpp
@@ -574,6 +574,10 @@ bool CHalfLife2::TextMsg(int client, int dest, const char *msg)
 
 	pBitBuf->WriteByte(dest);
 	pBitBuf->WriteString(msg);
+	pBitBuf->WriteString("");
+	pBitBuf->WriteString("");
+	pBitBuf->WriteString("");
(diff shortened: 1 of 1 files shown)


KyleSanderson:
Thanks @thexa4. I'm not certain we support running SourceMod on non-release builds of the SDK.

Is there a shipped game where you're encountering this?

thexa4:
I encountered this when accidentally releasing a non-release build of tier1 for Empires.

You're probably also losing speed on release branches due to unoptimized code paths when recovering from this.

It's not an essential fix, but makes running the game in debug mode a lot easier.

KyleSanderson:
Nope, just after data for historical purposes. Thanks for the background @thexa4.

thexa4:
Ah, no problem. Thanks for the quick response :)

PR #572, opened 2016-12-31
Fix memory leak in Regex extension.

WildCard65:
"x" is never deleted when compile fails.

### DIFF
--- a/extensions/regex/extension.cpp
+++ b/extensions/regex/extension.cpp
@@ -91,7 +91,16 @@ static cell_t CompileRegex(IPluginContext *pCtx, const cell_t *params)
 		return 0;
 	}
 
-	return g_pHandleSys->CreateHandle(g_RegexHandle, (void*)x, pCtx->GetIdentity(), myself->GetIdentity(), NULL);
+	HandleError error = HandleError_None;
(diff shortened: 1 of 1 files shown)


KyleSanderson:
Thanks. If you could also handle handle allocation failing like we do in the rest of SM this seems like an improvement.

extensions/regex/extension.cpp
```diff
@@ -87,10 +87,20 @@ static cell_t CompileRegex(IPluginContext *pCtx, const cell_t *params)
...
 		return 0;
 	}
 
-	return g_pHandleSys->CreateHandle(g_RegexHandle, (void*)x, pCtx->GetIdentity(), myself->GetIdentity(), NULL);
+	HandleError error = HandleError_None;
+	Handle_t regexHandle = g_pHandleSys->CreateHandle(g_RegexHandle, (void*)x, pCtx->GetIdentity(), myself->GetIdentity(), &error);
+	if (!regexHandle || error != HandleError_None)
+	{
+		delete x;
+		pCtx->ReportError("Allocation of regex handkle failed, error code #%d", error);
```
KyleSanderson: > handkle

psychonic:
Thanks.

PR #573, opened 2017-01-03
Event.FireToClient fixups

psychonic:
This removes a couple of checks that are IMO unnecessary. See commit messages for details and reasoning.

### DIFF
--- a/core/smn_events.cpp
+++ b/core/smn_events.cpp
@@ -160,12 +160,6 @@ static cell_t sm_FireEventToClient(IPluginContext *pContext, const cell_t *param
 		return pContext->ThrowNativeError("Invalid game event handle %x (error %d)", hndl, err);
 	}
 
-	/* If identities do not match, don't fire event */
-	if (pContext->GetIdentity() != pInfo->pOwner)
-	{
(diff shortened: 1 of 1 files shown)


dvander:
🚢

Benoist3012:
Funny thing is, dontbroadcast bugged me on the previous PR #505.
I got a question regarding that function, can we now hook an event with pre flag, and set the dontbroadcast boolean on true and fire to one client. Or we need to copy all the event?

psychonic:
@Benoist3012 Yes, that was the original intent. The example code in that PR showed it working, when it really didn't. I had either missed that case in my testing, or tested before adding the handle ownership check.

KyleSanderson:
https://github.com/alliedmodders/sourcemod/blob/0aa11374d8bdf305fae4acc4c251d7eff61a3050/core/smn_events.cpp#L135 should probably fix this up as well.

psychonic:
@KyleSanderson I think we want to keep that check since the handle gets freed at the end of that path.

peace-maker:
This should land in 1.8 too!

psychonic:
What all is left holding us up from just making 1.9 the stable branch?

PR #577, opened 2017-01-23
Fix crash when database transaction fails (bug 6531)

peace-maker:
If one of the queries fails in a transaction, the DBI system would only
allocate an array of the size of the amount of all the successful
queries before the failed one.
It writes data for all the queries though effectively writing past the
array bounds leading to heap corruption.

Create the right sized array!


KyleSanderson:
Great catch, thanks!

PR #579, opened 2017-01-28
[HashMap] Adds new IntHashMap and IntMap natives.

gnif:
This patch makes the following changes
  * Refactors StringHashMap to a generic HashMap template
  * Adds the IntHashMap class
  * Adds new IntMap natives
  * Adds IntMap tests to the tries test suite

This does not change the existing Trie API in any way so as to retain backwards compatibility, and all existing and new test suite tests pass.

### DIFF
--- a/core/logic/smn_adt_trie.cpp
+++ b/core/logic/smn_adt_trie.cpp
@@ -33,12 +33,14 @@
 #include "common_logic.h"
 #include <am-moveable.h>
 #include <am-refcounting.h>
-#include <sm_stringhashmap.h>
+#include <sm_hashmap.h>
 #include "sm_memtable.h"
 #include <IHandleSys.h>
 
-HandleType_t htCellTrie;
-HandleType_t htSnapshot;
+HandleType_t htStringCellTrie;
+HandleType_t htStringSnapshot;
+HandleType_t htIntCellTrie;
+HandleType_t htIntSnapshot;
 
 enum EntryType
 {
@@ -164,14 +166,19 @@ private:
 	cell_t data_;
 };
 
-struct CellTrie
+struct StringCellTrie
 {
 	StringHashMap<Entry> map;
 };
 
-struct TrieSnapshot
+struct IntCellTrie
 {
-	TrieSnapshot()
+	IntHashMap<Entry> map;
+};
+
+struct StringTrieSnapshot
+{
+	StringTrieSnapshot()
 		: strings(128)
 	{ }
 
@@ -185,6 +192,22 @@ struct TrieSnapshot
 	BaseStringTable strings;
 };
 
+struct IntTrieSnapshot
+{
+	IntTrieSnapshot()
+		: ints(128)
+	{ }
+
+	size_t mem_usage()
+	{
+		return length * sizeof(int) + ints.GetMemTable()->GetMemUsage();
+	}
+
+	size_t length;
+	ke::AutoPtr<int[]> keys;
+	BaseIntTable ints;
+};
+
 class TrieHelpers : 
 	public SMGlobalClass,
 	public IHandleTypeDispatch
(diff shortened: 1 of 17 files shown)


public/sm_hashmap.h
```diff
@@ -236,6 +273,12 @@ class StringHashMap
 	size_t memory_used_;
 };
 
+template <typename T>
+using StringHashMap = HashMap<T, ke::AString, detail::StringHashMapPolicy, detail::CharHash, const char *>;
```
WildCard65: This feature makes SM depend on VS 2013 minimum for Windows

psychonic: No, all of the C++11 usage already in SM makes it depend on VS 2013 minimum for Windows.

dvander:
Is the goal here to avoid having to wrap integer keys in strings? Generally just writing stocks is preferable to having new natives, in cases like this, especially when we wind up with a lot of duplication.

public/sm_hashmap.h
```diff
@@ -88,24 +88,61 @@ namespace detail
...
 			return lookup.length() == key.length() &&
 				   memcmp(lookup.chars(), key.chars(), key.length()) == 0;
 		}
-		static inline uint32_t hash(const CharsAndLength &key) {
+		static inline uint32_t hash(const CharHash &key) {
+			return key.hash();
+		}
+	};
+
+	class IntHash
```
dvander: The hash and key are stored in HashTable entries, so you actually don't need this class. The reason for CharsAndLength is to avoid always recomputing the key length as we scan the table for collisions. When the key is plain-old-data, you can just make KeyStoreType == KeyContainerType. Then use ke::HashInt32 as your hash function.

plugins/include/adt_trie.inc
```diff
@@ -169,6 +273,22 @@ methodmap StringMapSnapshot < Handle
...
+ * Insertion, deletion, and lookup in a hash map are all considered to be fast
+ * operations, amortized to O(1), or constant time.
+ *
+ * The word "Trie" in this API is historical. As of SourceMod 1.6, tries have
+ * been internally replaced with hash tables, which have O(1) insertion time
+ * instead of O(n).
+ *
+ * @return 			New Map Handle, which must be freed via CloseHandle().
+ */
+native IntMap CreateIntTrie();
```
gnif: Great, I felt the same but retained them for completeness.

public/sm_hashmap.h
```diff
@@ -88,24 +88,61 @@ namespace detail
...
+		}
+		static inline uint32_t hash(const IntHash &key) {
 			return key.hash();
 		}
 	};
 }
 
-template <typename T>
-class StringHashMap
+template <typename T, typename K, typename P, typename C, typename KT>
```
gnif: Agreed, I will rename these to more descriptive names.

public/sm_hashmap.h
gnif: I was not sure here, I initially planned to not have a dedicated class for it but I was time constrained to look into what was actually going on here so in the rush to finish it was just cloned.

public/sm_hashmap.h
```diff
@@ -223,7 +232,7 @@ class StringHashMap
 	}
 
 	// Only value needs to be set after.
-	bool add(Insert &i, const char *aKey)
+	bool add(Insert &i, KeyLookupType aKey)
```
dvander: I think all instances where `KeyLookupType` is used as a parameter, it should be `const KeyLookupType&`?

public/sm_hashmap.h
```diff
@@ -236,6 +245,12 @@ class StringHashMap
 	size_t memory_used_;
 };
 
+template <typename T>
+using StringHashMap = HashMap<T, ke::AString, detail::StringHashMapPolicy, detail::CharsAndLength, const char *>;
+
+template <typename T>
+using IntHashMap = HashMap<T, int32_t, detail::IntHashMapPolicy, const int32_t, const int32_t>;
```
dvander: And then s/`const int32_t`/`int32_t`/

core/logic/smn_adt_trie.cpp
```diff
@@ -185,52 +192,123 @@ struct TrieSnapshot
...
+	{ }
+
+	size_t mem_usage()
+	{
+		return length * sizeof(int) + ints.GetMemTable()->GetMemUsage();
+	}
+
+	size_t length;
+	ke::AutoPtr<int[]> keys;
+	BaseIntTable ints;
```
dvander: BaseMemTable is crazy. Ten years ago it seemed like a good idea...

Just use `ke::Vector<int32_t>` here it should amount to the same thing. Then you don't need the `BaseIntTable` class.

dvander:
Actually you can probably merge `keys` and `ints` in the snapshot together...

H1nk:
@gnif, would be nice if you/someone could fix up #579. IntHashMap would be nice to avoid wrapping ints to strings for mapping clients <--> properties.

gnif:
I am sorry but I no longer have time to contribute to this, someone else will need to pursue this.

Headline22:
@gnif I'm interested in pursuing this further. Can you give me push access to your fork that way I don't have to create a whole new pull request?

KyleSanderson:
@gnif so this hasn't aged well (which isn't on you). @Headline was looking to pursue this additionally in 2017. Are you open to any changes on this one? At the very least a rebase?

PR #580, opened 2017-01-31
Add convenience logic for "offset" to Address definitions in GameConf

FlaminSarge:
Allows for the last "read" offset for an Address to instead be "offset", which doesn't deref the addr pointer after applying the offset
Shortens the necessary coding to get a particular Address inside a function
(old method: store offset in a separate GameConf Offsets entry, apply the offset to the Address in SP)

It builds on OSX, but I don't have srcds on OSX to test with, nor any other environments to build on.
Any suggestions?

### DIFF
--- a/core/logic/GameConfigs.cpp
+++ b/core/logic/GameConfigs.cpp
@@ -323,6 +323,7 @@ SMCResult CGameConfig::ReadSMC_NewSection(const SMCStates *states, const char *n
 			m_Address[0] = '\0';
 			m_AddressSignature[0] = '\0';
 			m_AddressReadCount = 0;
+			m_AddressLastIsOffset = false;
 
 			strncopy(m_Address, name, sizeof(m_Address));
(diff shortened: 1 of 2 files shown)


FlaminSarge:
When I started writing this up, that's the first thing I tried. However, having a flag on each index (either as a bitfield or bool array, doesn't matter) tended to be more complicated than simply having a single boolean saying that the last specified one is an offset.

FlaminSarge:
By the way, is sigLength used at all in AddressConf? If not, was there a reason for keeping it around?

FlaminSarge:
I looked through the history for sigLength and it seems like it was only added because the first param for AddressConf was a char*, even though that pointer is only read from, never written to (and therefore sigLength never gets used).

Do any further changes need to be made to this PR? I tried various cleanups for the conditionals, but in the end this was still the most readable format.

PR #581, opened 2017-02-03
Fix typo in menu manager not drawing a spacer

peace-maker:
The "Back" slot was hidden even if the page had a "Next" button. Draw the spacer correctly!


psychonic:
Thanks!

PR #584, opened 2017-02-09
Fix regression with GetEntPropEnt always returning -1

peace-maker:
Regression in #583

The `CBaseHandle *` was compared to a `IHandleEntity **` always failing
instead of comparing the actual base handles.

Fixes a small typo in MatchTypeDescAsInteger too, where a FIELD_CUSTOM
could fall through badly.

### DIFF
--- a/core/smn_entities.cpp
+++ b/core/smn_entities.cpp
@@ -1190,6 +1190,7 @@ inline int MatchTypeDescAsInteger(_fieldtypes type, int flags)
 			// Variant, read as int32.
 			return 32;
 		}
+		break;
 	case FIELD_SHORT:
 		return 16;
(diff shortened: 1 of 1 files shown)


KyleSanderson:
This looks good to me, if someone could test @psychonic's original test case so we don't regress that, we should be good to go.

:shipit:

peace-maker:
Since we only changed GetEntPropEnt and not GetEntPropFloat it wouldn't be affected.

KyleSanderson:
Sorry, not the original case; but the case he's added support for. *

It's a variant_t holding an edict (unless if I'm terribly mistaken), I can't immediately see a prop on an ent for it.

psychonic:
Good catches. Thank you much.

PR #586, opened 2017-02-09
improved CharToUpper and CharToLower

assyrianic:
for CharToUpper, using bitwise AND with an underscore always returns an upper case letter, even if it's already upper case! `'c' & '_' == 'C'` likewise, `'C' & '_' == 'C'`

for CharToLower, using bitwise OR with space literal always yields a lower case letter, even if the letter is already lower case. `'c' | ' ' == 'c'` likewise, `'C' | ' ' == 'c'`

simple bitwise operation that does the same thing more efficiently rather than having to require an if statement to catch exceptional chars and convert them.

### DIFF
--- a/plugins/include/string.inc
+++ b/plugins/include/string.inc
@@ -400,7 +400,9 @@ native bool StripQuotes(char[] text);
  */
 stock int CharToUpper(int chr)
 {
-	return chr & '_';
+	if (IsCharLower(chr))	
+		return chr & '_';
+	return chr;
 }
 
 /**
@@ -412,7 +414,9 @@ stock int CharToUpper(int chr)
  */
 stock int CharToLower(int chr)
 {
-	return chr | ' ';
+	if (IsCharUpper(chr))	
+		return chr | ' ';
+	return chr;
 }
 
 /**
--- a/plugins/include/string.inc
+++ b/plugins/include/string.inc
@@ -400,11 +400,7 @@ native bool StripQuotes(char[] text);
  */
 stock int CharToUpper(int chr)
 {
-	if (IsCharLower(chr))
-	{
-		return (chr & ~(1<<5));
-	}
-	return chr;
+	return chr & '_';
 }
 
 /**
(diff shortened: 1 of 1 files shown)


asherkin:
The bitwise math in the current functions is already doing exactly the same thing. The if statements are there to guard non-alpha characters. Please update your existing pull request rather than creating a new one.

assyrianic:
guard non-alpha characters? How many people realistically use CharToUpper and CharToLower on non-ascii chars?

Secondly, the bitwise only does it on the opposite case for a given char.
Thirdly, the bitwise math uses bit shifts with bitwise. Why bitshift instead of giving the direct constant?

I've done a more improved commit to come to an agreement.

dvander:
Is there a particular reason you want this change? The compiler output will be literally identical before/after. `~(1<<5)` and `'_'` are the same constant.

(And, it's not so much whether or not people realistically use them on non-ascii chars - it's that we have to preserve the behavior that already exists. It's very hard to verify that no one will be affected.)

assyrianic:
'_' is 95 as a constant... ~(1<<5) is -33 as a constant so no they're not the same.
Also, why not keep it simple?

peace-maker:
Well, `_` is `0101 1111` in binary, `~(1<<5)` is `1101 1111`, so your code would strip the MSB while the old one doesn't touch it. Regarding preserving behavior.
Since the only difference between uppercase and lowercase characters in ascii is the 6th bit, it'd say the current code is more readable.

dvander:
Well, the is-ascii check guarantees no MSB. ("the same constant for what the operation needs to do"). But I agree, it's not as readable since the bit flip is a magic char rather than an explicit mask. And in terms of performance it's no change:

```
  9     movzbl  8(%ebp), %eax
 10     andl    $95, %eax

 20     movsbl  8(%ebp), %eax
 21     andl    $-33, %eax
```

I'm sure there are *tons* of places in SourceMod that could use tidying up :) this one seems fine as-is though.

PR #587, opened 2017-02-12
Add FireEntityOutput native.

Benoist3012:
Allows plugins to force fire an Output from an entity.
The native can be used in combo with SetVariantX, similarly like how you would use it with AcceptEntityInput native.
So it does support every kind of output
```
COutputInt;
COutputFloat;
COutputString;
COutputEHANDLE;
COutputVector;
COutputPositionVector;
COutputColor32;
COutputEvent
```

Example code:
```
int iEnt = Entity_FindEntityByName("clock_death_counter", "math_counter");
SetVariantInt(12);
FireEntityOutput(iEnt,"OnGetValue");
```


psychonic:
If keeping sdktools_variant_t.inc as a separate file, there should be an `#include` for it in sdktools.inc, like there are for the other split sdktools incs.

You'll also want to add it to the `CopyFiles('plugins/include', 'addons/sourcemod/scripting/include'` section in PackageScript for it to be included in the package.

Benoist3012:
For the first one, I think I forgot to upload the file.
I'll fix the rest tomorrow.

tools/buildbot/PackageScript
```diff
@@ -331,6 +331,7 @@ CopyFiles('plugins/include', 'addons/sourcemod/scripting/include',
     'sdktools.inc',
     'sdktools_client.inc',
     'sdktools_engine.inc',
+	'sdktools_variant_t.inc',
```
WildCard65: Python is strict on tabbing/spacing

Benoist3012: I didn't notice it, I guess notepad++ did it automatically http://i.imgur.com/mFvKkQL.png

WildCard65: Depending on your tab settings in NP++, you may never notice an indentation issue in a script until something it discovers it (like python throwing an indentation exception/git diff)

Benoist3012: I'll pay more attention to this now, thanks.

psychonic:
Thanks!

PR #588, opened 2017-02-17
Update sm_map behavior to better reflect FindMap/GetMapDisplayName results

FlaminSarge:
sm_map now makes use of GetMapDisplayName in the same way that sm_setnextmap did in #354 

sm_map and its menu now use PLATFORM_MAX_PATH for map name buffer size instead of 64

~~nominations has a new message for when a map is found, but not available in the nominations pool (to differentiate it from the map not being found at all)~~

sm_setnextmap now uses ShowActivity2

### DIFF
--- a/plugins/basecommands/map.sp
+++ b/plugins/basecommands/map.sp
@@ -42,7 +42,7 @@ public int MenuHandler_ChangeMap(Menu menu, MenuAction action, int param1, int p
 	}
 	else if (action == MenuAction_Select)
 	{
-		char map[64];
+		char map[PLATFORM_MAX_PATH];
 		
 		menu.GetItem(param2, map, sizeof(map));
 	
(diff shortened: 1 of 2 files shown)


psychonic:
Looks okay to me. My only hesitation is adding a new translation phrase. We've not had any new ones since switching to a rolling release model, and I'm not sure that we have a great way at the moment to ask for updated phrases.

Regardless, I suppose an accurate English phrase is better than a incorrect localized one here, so not a huge deal as-is.

@asherkin ?

FlaminSarge:
Any concerns with the actual language of the phrase?

psychonic:
Nope.

FlaminSarge:
Do any further changes need to be made to this?

KyleSanderson:
It's just the translation phrase that has a hang-up as we haven't added one to my knowledge since switching to the rolling model (and atleast a year prior to that, iirc). If these were split into two separate PRs, only one would be pending.

FlaminSarge:
I can pull out the changes to nominations.sp and its i18n, then, but basically the issue is that there's no good way set up for telling people 'hey we need translations for this new thing'? What was the old setup, again?

FlaminSarge:
Pulled out the nominations commit pending the discussion on how to get translations done.

KyleSanderson:
Sorry, I thought I pulled this two days ago. Thanks FS.

PR #590, opened 2017-02-28
rockthevote.sp - RoundToFloor() -> RoundToCeil()

theSaint92:
The reason of update is described here https://forums.alliedmods.net/showthread.php?p=2499471


asherkin:
Thanks!

PR #591, opened 2017-03-02
Add OnPlayerRunCmdPost forward

peace-maker:
Allow plugins to catch the values that are actually used by the game
after some eventual changes by other plugins.

This is useful for a plugin like BotMimic 👍

### DIFF
--- a/plugins/include/sdktools_hooks.inc
+++ b/plugins/include/sdktools_hooks.inc
@@ -59,6 +59,23 @@
 forward Action OnPlayerRunCmd(int client, int &buttons, int &impulse, float vel[3], float angles[3], int &weapon, int &subtype, int &cmdnum, int &tickcount, int &seed, int mouse[2]);
 
 /**
+ * @brief Called after a clients movement buttons were processed.
+ *
+ * @param client	Index of the client.
+ * @param buttons	The current commands (as bitflags - see entity_prop_stocks.inc).
+ * @param impulse	The current impulse command.
+ * @param vel		Players desired velocity.
+ * @param angles	Players desired view angles.
+ * @param weapon	Entity index of the new weapon if player switches weapon, 0 otherwise.
+ * @param subtype	Weapon subtype when selected from a menu.
+ * @param cmdnum	Command number. Increments from the first command sent.
+ * @param tickcount	Tick count. A client's prediction based on the server's GetGameTickCount value.
+ * @param seed		Random seed. Used to determine weapon recoil, spread, and other predicted elements.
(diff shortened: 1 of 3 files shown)


Benoist3012:
We need that, I see many uses for it :1st_place_medal:

meev:
Lol, use postthink instead if theres something sourcemod needs it is plugin priority so you can prioritize certain plugins to come first in all the hooks.

KyleSanderson:
My only aversion to merging something like this is this is our most expensive hook. With something like SDKHooks style hooking we can atleast limit who we're watching, and avoid running the function for other players we don't care about.

For performance reasons, and performance reasons alone, I'm against merging this.

peace-maker:
So you'd want an extra `HookRunPlayerCmdPost` native? That'd be awkward with the old "pre" forward being global.
We cannot just change the current one too due to backwards compatibility.

psychonic:
I'll echo here what I said in a related discussion with @peace-maker and @KyleSanderson on IRC.

Doing a benchmark should be able to prove the in/significance of any potential perf issues, but I'm not sure that it would even matter, since even if it is still a global forward, the hooking is already opt-in, only hooking if a plugin has that forward.

Deprecating the prehook, moving it (and new post hook) to SDKHooks with a compat shim would not be difficult, but it sounds like that doesn't really solve anything, adding a further perf hit per @asherkin 

TL;DR, neither approach is the end of the world. I'm okay with it if value is added and with it staying opt-in, one way or the other (as the pre-hook is now, or via SDKHooks). If you want to benchmark to figure out which method is the most performant in practice, have at it.

plugins/include/sdktools_hooks.inc
```diff
@@ -59,6 +59,23 @@
...
+ * @param vel		Players desired velocity.
+ * @param angles	Players desired view angles.
+ * @param weapon	Entity index of the new weapon if player switches weapon, 0 otherwise.
+ * @param subtype	Weapon subtype when selected from a menu.
+ * @param cmdnum	Command number. Increments from the first command sent.
+ * @param tickcount	Tick count. A client's prediction based on the server's GetGameTickCount value.
+ * @param seed		Random seed. Used to determine weapon recoil, spread, and other predicted elements.
+ * @param mouse		Mouse direction (x, y).
+ */
+forward void OnPlayerRunCmdPost(int client, int buttons, int impulse, float vel[3], float angles[3], int weapon, int subtype, int cmdnum, int tickcount, int seed, int mouse[2]);
```
asherkin: The arrays should be `const`, other than that I think this is all good.

PR #593, opened 2017-03-09
Fix crash when accessing gamerules during mapchange

peace-maker:
If the gamerules proxy entity doesn't exist, don't try to generate the reference for it.

https://forums.alliedmods.net/showthread.php?t=294762


KyleSanderson:
Thanks!

PR #594, opened 2017-03-09
Fix workshop/ and workshop\ not always working with GetMapDisplayName

powerlord:
The previous patch had a bug where the first slash was always required to be the same as the OS separator path.

This time, the fix has been tested using plugins/testsuite/mapdisplayname.sp on CS:GO Windows:

    test_mapdisplayname workshop\125488374\de_dust2_se
    GetMapDisplayName says "de_dust2_se" for "workshop\125488374\de_dust2_se"
    test_mapdisplayname workshop/125488374/de_dust2_se
    GetMapDisplayName says "de_dust2_se" for "workshop/125488374/de_dust2_se"

There is one caveat with this patch:  
It does not check for the presence of a slash immediately after workshop.  So, if you pass it `workshopblahblabh/de_dust` it will return `de_dust`

Then again, it isn't really checking the entire format anyway, only that it starts with `workshop` and has a `/` or `\` somewhere in it.

### DIFF
--- a/core/HalfLife2.cpp
+++ b/core/HalfLife2.cpp
@@ -1253,14 +1253,10 @@ bool CHalfLife2::GetMapDisplayName(const char *pMapName, char *pDisplayname, siz
 	}
 
 #if SOURCE_ENGINE == SE_CSGO
-	// In CSGO, the path separator is used in workshop maps.
-	char workshop[10];
-	ke::SafeSprintf(workshop, SM_ARRAYSIZE(workshop), "%s%c", "workshop", PLATFORM_SEP_CHAR);
-
 	char *lastSlashPos;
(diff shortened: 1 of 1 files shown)


KyleSanderson:
This isn't your fault, but I don't like this function at all the way it is.

Based on the fact we're stripping / or \ anyways, why not just always check to see if they're there instead of requiring workshop. Same goes for stripping .ugc and kin, it's a super cheap check that doesn't need to be guarded in engine ifdefs.

powerlord:
@KyleSanderson 

I'll be honest, I don't host servers or write plugins any more, I was just fixing logic for people who noticed that it's acting inconsistently with the way CS:GO currently works.

By all means, feel free to rewrite it any way you please.

PR #595, opened 2017-03-11
Ignore replay and sourcetv in @spec targeting.

Drifter321:
This makes it so sm_kick @spec and other commands don't target replay and sourcetv. This is a side effect from a change to calling disconnect for bots instead of kickid.


Benoist3012:
I got that issue on SF2 with admins accidentally forcing the source tv bot by using `sm_sf2_setplaystate @all` fixed on my side, but that would be cool if it's directly fixed from sourcemod.

ambaca:
This fix work nice,
How about rest multitarget strings ?  "@dead" "@all" "@bots", etc.
What kind of drawback we could get to not kick SourceTV+Replay with multitargeting ?

PR #596, opened 2017-03-14
Change CHalfLife2::GetMapDisplayName to always check for string substitutions.

KyleSanderson:
This is a reworking of the issue identified in https://github.com/alliedmodders/sourcemod/pull/594/

Instead of doing specific character and string substitutions based on the game, we should just be defensive and do it for every game.


powerlord:
Was the branch name *really* necessary?

PR #597, opened 2017-03-15
Change the message displayed when a map isn’t in the nominations maplist

FlaminSarge:
New English text is “Map ‘%s’ is not in the nominations pool.”

Split off from #588 due to the pending discussion on how i18n requests will work under the new setup.

### DIFF
--- a/plugins/nominations.sp
+++ b/plugins/nominations.sp
@@ -145,7 +145,7 @@ public Action Command_Addmap(int client, int args)
 	int status;
 	if (!g_mapTrie.GetValue(resolvedMap, status))
 	{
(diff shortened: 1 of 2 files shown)


FlaminSarge:
How was https://sm.alliedmods.net/translator set up? What complications arise from the new model that prevent its use?

FlaminSarge:
Either an I18N label for issues or use of the Github Projects feature might be appropriate for handling and tracking I18N requests.

FlaminSarge:
Rebased this PR.

KyleSanderson:
@psychonic @asherkin there's no technical reason for this to stall for 3+ years. Are you okay if we land this new translatable field?

peace-maker (approved):

PR #598, opened 2017-03-20
HUD is now supported on CSGO

kossolax:
https://forums.alliedmods.net/showthread.php?t=295215

### DIFF
--- a/gamedata/core.games/common.games.txt
+++ b/gamedata/core.games/common.games.txt
@@ -100,6 +100,7 @@
 			"engine"			"sdk2013"
 			"engine"			"contagion"
 			"engine"			"bms"
+			"engine"			"csgo"
 		}
 		
 		"Offsets"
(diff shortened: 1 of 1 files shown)


peace-maker:
Wasn't there some gotcha about mixing "game" and "engine" keys for csco?

kossolax:
oops, you are right. It's "game" not "engine".

psychonic:
While "engine" isn't technically correct for the functionality, it may not cause any issues, depending on whether or not CS:CO overrides the files that changed.

kossolax:
Sadly it isn't working ingame if you use "engine", but it works with "game".
I went wrong with the online github editor, sorry about that.

psychonic:
If that were true, there would be a ton of broken gamedata right now for csgo, and even more before we added support for csco, not to mention a ton also broken for other games.

Accelerator74:
I confirm. With the "engine" there is no effect. Changed to "game" - everything worked.

Bara20:
But works "games" for (cs)co servers, but I can test it later with "game" under (cs)co.

Bara20:
Doesn't work with "csgo" as game, just with "csco". @psychonic

PR #599, opened 2017-03-20
Generated new offsets for Empires

thexa4:
Generated using vtablescanner: https://forums.empiresmod.com/index.php?threads/calculating-updated-sourcemod-vtable-offsets.20571/


KyleSanderson:
Appreciate the effort man. We usually only maintain offsets for gamedata we use in our extensions/core. It looks like you were able to port a previous signature to it's corresponding symbol? Are there any others like this in the offset list you've given?

thexa4:
It seemed like the empires offsets missed a lot of entries other games did have so it was easier for me to add them all. If you have an overview of what offsets are used I can make a filtered list.

We converted the linux server build scripts to no longer strip symbol information so I think most linux signatures should be able to be converted into symbols.

KyleSanderson:
https://github.com/alliedmodders/sourcemod/blob/aee82c9be4cc023130a1e8c298fcc23ba40b861d/gamedata/sdkhooks.games/game.empires.txt

Comparing other mods, or games running the same SDK may reveal other keys that could become available to empires. We don't really define any keys, it's more of a "hey, I need the address for CBaseCombatClam", then our gamedata system grabs the value for what we're looking for, using the operating system as the key. From there the gamedata iface will look in a defined library (server, or engine), and will attempt to lookup the symbol or sigscan until a match is hopefully found.

Great news on the stripping front. :-)

psychonic:
I've updated the gamedata for Empires now separately.

PR #600, opened 2017-03-21
Fixed typo on previous PR

kossolax:
"engine" doesn't work here. It should be "game". #598


psychonic:
It would have worked if it was first, but then would have broken the "game" entries.

There are some non-obvious oddities and inconsistencies when mixing "engine" and "game" in a section, and as I hinted in the other PR, this does mean that some of our other gamedata is broken as well. I will either fix that separately, or just _fix_ the gamedata reader.

When reading, it works as expected when seeing just "engine" or just "game" entries. As soon as it encounters the other type, it requires a match of each type (but any matches before that point are okay with just the one).

PR #601, opened 2017-04-18
Add PrintToConsoleAll

Headline22:
We have PrintToChatAll, PrintToCenterAll, but not a PrintToConsoleAll. Might as well be consistent...


psychonic:
LGTM. Thanks!

PR #603, opened 2017-04-30
Add missing parameter to DataPack.IsReadable

peace-maker:
The `int bytes` parameter is mentioned in the docs but was missing from the prototype.
Copy&Paste heros!


psychonic:
Thanks!

PR #604, opened 2017-05-01
Fix for sm_msay (CS:GO don't have 0 button)

DjCoraxa:


Bara20:
How about a csgo check, because other games has button 0?

DjCoraxa:
Done

psychonic:
GetMaxPageItems should already be returning 9 on CS:GO (and whatever correct value on other games). It defaults to 10, but gets overridden with gamedata.

At a glance, I'm not sure why it would be using 0 at the exit key with sm_msay, but adding a special case in the plugin for one game isn't the right fix.

It also looks like another set of unrelated changes slipped in.

KyleSanderson:
Alongside psychonics comments, this is also broken as you're changing the behaviour for every other platform besides CS:GO which maintains the existing behaviour. 

Are you sure this is broken?

plugins/basechat.sp
```diff
@@ -407,7 +407,10 @@ void SendPanelToAll(int from, char[] message)
 	mSayPanel.DrawItem("", ITEMDRAW_SPACER);
 	mSayPanel.DrawText(message);
 	mSayPanel.DrawItem("", ITEMDRAW_SPACER);
-	mSayPanel.CurrentKey = GetMaxPageItems(mSayPanel.Style);
+	
+	if (g_GameEngine != Engine_CSGO)	mSayPanel.CurrentKey = (GetMaxPageItems(mSayPanel.Style) - 1);
```
KyleSanderson: This should probably be ==...

Bara20:
https://wiki.alliedmods.net/Contributing_to_SourceMod#Coding_Style

PR #606, opened 2017-05-07
Add natives to check for NULL_VECTOR and NULL_STRING

peace-maker:
Until now the only use for the `NULL_VECTOR` and `NULL_STRING` pubvars was to pass them to C++ natives for them to be able to skip some work if the callee isn't interested in the results. Plugins themselves couldn't use them for that purpose or anything else at all.
These patches add support for plugins to use them in the same manner as an extension could.

`IsNullVector` and `IsNullString` can be used to check if the `NULL_VECTOR` or `NULL_STRING` was passed as a function/forward argument.

`IsNativeParamNullVector` and `IsNativeParamNullString` can be used to see if someone passed the NULL pubvars when calling your native.

`Call_PushNullVector` and `Call_PushNullString` allow you to call any function or forward and pass the NULL pubvars, which can be checked with the above natives.
Because of the need for the local address instead of the physical one we can't use `IPluginContext::GetNullRef` for these two natives, so the pubvars had to be looked up manually again without caching. If this is a concern we could expose the cached local address too in SourcePawn, but I tried not to change the VM here.

### DIFF
--- a/core/logic/ForwardSys.cpp
+++ b/core/logic/ForwardSys.cpp
@@ -273,7 +273,26 @@ int CForward::Execute(cell_t *result, IForwardFilter *filter)
 				 */
 				if (type == Param_String)
 				{
-					err = func->PushStringEx((char *)param->byref.orig_addr, param->byref.cells, param->byref.sz_flags, param->byref.flags);
+					// If NULL_STRING was pushed, push the reference to the pubvar of the callee instead.
+					if (param->isnull)
+					{
+						IPluginRuntime *runtime = func->GetParentRuntime();
+						uint32_t null_string_idx;
+						err = runtime->FindPubvarByName("NULL_STRING", &null_string_idx);
+
+						if (!err)
+						{
+							cell_t null_string;
+							err = runtime->GetPubvarAddrs(null_string_idx, &null_string, nullptr);
+
+							if (!err)
+								err = func->PushCell(null_string);
+						}
+					}
+					else
+					{
+						err = func->PushStringEx((char *)param->byref.orig_addr, param->byref.cells, param->byref.sz_flags, param->byref.flags);
+					}
 				}
 				else if (type == Param_Float || type == Param_Cell)
 				{
@@ -281,7 +300,26 @@ int CForward::Execute(cell_t *result, IForwardFilter *filter)
 				}
 				else
 				{
-					err = func->PushArray(param->byref.orig_addr, param->byref.cells, param->byref.flags);
+					// If NULL_VECTOR was pushed, push the reference to the pubvar of the callee instead.
+					if (param->isnull && type == Param_Array)
+					{
+						IPluginRuntime *runtime = func->GetParentRuntime();
+						uint32_t null_vector_idx;
+						err = runtime->FindPubvarByName("NULL_VECTOR", &null_vector_idx);
+
+						if (!err)
+						{
+							cell_t null_vector;
+							err = runtime->GetPubvarAddrs(null_vector_idx, &null_vector, nullptr);
+
+							if (!err)
+								err = func->PushCell(null_vector);
+						}
+					}
+					else
+					{
+						err = func->PushArray(param->byref.orig_addr, param->byref.cells, param->byref.flags);
+					}
 					assert(type == Param_Array || type == Param_FloatByRef || type == Param_CellByRef);
 				}
 			}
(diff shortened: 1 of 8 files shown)


core/logic/ForwardSys.cpp
```diff
@@ -273,15 +273,53 @@ int CForward::Execute(cell_t *result, IForwardFilter *filter)
...
+					if (param->isnull && type == Param_Array)
+					{
+						IPluginRuntime *runtime = func->GetParentRuntime();
+						uint32_t null_vector_idx;
+						err = runtime->FindPubvarByName("NULL_VECTOR", &null_vector_idx);
+
+						if (!err)
+						{
+							cell_t null_vector;
+							err = runtime->GetPubvarAddrs(null_vector_idx, &null_vector, nullptr);
```
KyleSanderson: Do these need to be different? Can't we have a global null here?

core/logic/ForwardSys.cpp
```diff
@@ -273,15 +273,53 @@ int CForward::Execute(cell_t *result, IForwardFilter *filter)
 				 */
 				if (type == Param_String)
 				{
-					err = func->PushStringEx((char *)param->byref.orig_addr, param->byref.cells, param->byref.sz_flags, param->byref.flags);
+					// If NULL_STRING was pushed, push the reference to the pubvar of the callee instead.
+					if (param->isnull)
```
KyleSanderson: Can we eval the pointer to the array and see if it's null or not instead of maintaining/adding this extra layer?

core/logic/smn_fakenatives.cpp
```diff
@@ -439,5 +491,7 @@ REGISTER_NATIVES(nativeNatives)
 	{"SetNativeArray",			SetNativeArray},
 	{"SetNativeCellRef",		SetNativeCellRef},
 	{"SetNativeString",			SetNativeString},
+	{"IsNativeParamNullVector",	IsNativeParamNullVector},
```
KyleSanderson: I'm not certain we should add these, operators are much more natural here in other languages; can we get anywhere close here? Do these need to be different if we maintain a single null?

core/logic/ForwardSys.cpp
peace-maker: That's probably a better idea. Would you leave the `CForward::PushNullVector` methods or just allow passing null to PushArray(Ex) and PushString(Ex) (while ignoring the flags)? PushString already allows for NULL, but I don't think that's sane.

core/logic/smn_fakenatives.cpp
peace-maker: What do you mean by operators?
We could add a single native for this one, since it only checks the parameter number, but the other natives have to be different to be able to take a `const char[]` and a `const float vec[3]` for the pubvars. And for consistency having two seperate natives for `NULL_STRING` and `NULL_VECTOR` respectively like the `IsNullVector` natives seems more intuitive for the user?

core/logic/ForwardSys.cpp
peace-maker: They have to be different because of the IsNull[String|Vector] signature. The arguments have be different to check for the pubvar, which could be passed around in local calls too.

InstantMuffin:
Finally this is happening

core/logic/ForwardSys.cpp
dvander: This seems fine to me as-is, though the nested |ifs| all over this function are gross, you could factor out a helper if you wanted.

dvander:
Looks good, if you want to do that minor clean-up let me know, otherwise I will merge as-is.

peace-maker:
Cleanup done 🐤

InstantMuffin:
https://img.devrant.io/devrant/rant/r_535939_oThZB.gif please

core/logic/ForwardSys.cpp
```diff
@@ -400,6 +440,7 @@ int CForward::PushCell(cell_t cell)
 		m_params[m_curparam].pushedas = Param_Cell;
 	}
 
+	m_params[m_curparam].isnull = false;
 	m_params[m_curparam++].val = cell;
```
asherkin: Drive-by comment, I find this pattern absolutely terrifying.

peace-maker: You could clean it up in another PR ♿️

PR #613, opened 2017-05-27
[1.8] Add support for compiling with VS2015.

ErikMinekus:
Cherry picked commit 416abd8 from master, which never landed on 1.8-dev. Fixes building SM 1.8 on VS2015 and VS2017.


asherkin:
Thanks!

PR #614, opened 2017-05-27
Add A2S_Rules fix for CS:GO to CStrike extension.

psychonic:
This has been discussed a bit back and forth on IRC. I'm still not 100% sold on the idea, but here's what an implementation would look like. It's not too bad.

This would also fix our plugin stats for CS:GO.


Drifter321:
Looks good to me. My only concern is the additional gamedata but its minimal and very easy to find.

KyleSanderson:
This changes the max MTU size and when we talked about this when the situation was more "emotional", lets say, this is changing limits the game has defined. For whatever reason Valve still believes limiting the maximum MTU is a good idea, we should respect that (our positions have essentially inversed ;) ).

The solution at that time was to scrap querying servers, and instead do a periodic upload of server data back to an AM WebAPI to do stats. This is a lot more work as there's backend changes, but it was the path forward however many years ago we discussed this.

I'm flexible, I see some SW.ext code in there and that's totally fine, but we're not only changing their MTU on rules inquiries, but on all network traffic.

psychonic:
> but we're not only changing their MTU on rules inquiries, but on all network traffic.

That is not true.

This isn't changing it on all network traffic. The check only affects outbound, connectionless packets. No netchannel traffic is affected. Also, CS:GO is the only engine branch with this check, despite most of the other logic around it being identical to others.

peace-maker:
You could add a hook on `host_rules_show` and only change the maxMTU and set the keyvalues on steam if it's enabled.

psychonic:
I had considered that. There's just minimal gain if it isn't "fixed" to match all other games by default, still requiring manual action similar to how it does as-is.

psychonic:
I just realized that I may have misunderstood what you were saying. You didn't mention changing the default, just not making the changes - doing anything, if it is disabled. That's a good idea!

KyleSanderson:
Can we merge this?

psychonic:
I want to do what @peace-maker suggested first, only having it enabled when host_rules_show is true, but haven't had the time yet.

Drifter321:
I had a look at this today, the original value should probably be compared to make sure its the value it should be before we start patching incorrect bytes. This way we can identify if an offset changed easier besides a random crash.

Drifter321:
Not tested but implemented the change. Compiles locally but fails with ambuild atm needs to be looked at. Regarding my previous crash i totally forgot the signature contained the actual patched bytes so we should be fine without checking. The gamedata also still seems to be good.

gamedata/sm-cstrike.games/game.csgo.txt
```diff
@@ -210,4 +210,28 @@
...
+			// This isn't the start of the function, but closer to where we want to patch, including the exact bytes we will replace
+			"NET_SendPacket"
+			{
+				"library"	"engine"
+				"windows"	"\x81\xFB\xB0\x04\x00\x00\x7E\x31\xFF\x75\x08"
+				"linux"		"\x84\xDB\x0F\x85\x2A\x2A\x2A\x2A\x81\x7D\x18\xB0\x04\x00\x00"
+			}
+		}
+		
+		"Offsets"
```
peace-maker: We could use the new "offset" key in the "Addresses" section here now!

extensions/cstrike/AMBuilder
```diff
@@ -25,6 +25,30 @@ for sdk_name in ['css', 'csgo']:
   if sdk_name not in SM.sdks:
     continue
   sdk = SM.sdks[sdk_name]
+  
+  if sdk_name == 'csgo':
+    project.sources += ['rulesfix.cpp']
+  
+  binary = SM.HL2Config(project, 'game.cstrike.ext.' + sdk.ext, sdk)
```
peace-maker, line 32: This change needs to be rebased against the `arch` changes in the build scripts.

KyleSanderson:
@psychonic @Drifter321 what's the skinny on this one?

peace-maker (approved):
The branch is up to date again and ready to be merged. I've tested on windows and the game rule query is nice and useful.

PR #619, opened 2017-05-28
Updating reload offsets

RussianRonin:
Updating reload offsets so that the reload function can work correctly in plugins.

Let me know if I am missing any information.


RussianRonin:
Disregard. I see they were already fixed.

PR #622, opened 2017-06-11
Fix sm_msay (Correct setting of the exit button depending on the game)

komashchenko:
Correct setting of the exit button depending on the game


psychonic:
Looks like this is already upstream in master and just being ported to the 1.8-dev branch. Looks safe to me. Thanks.

PR #623, opened 2017-06-12
Use the same weapon drop function in CS:GO as in CS:S

Drifter321:
In a previous update it appeared that the CSWeaponDrop function was replaced from taking 2 bools to a bool and a vector. This does not appear to be the case anymore or never was so this reverts to using the overload function with 2 bools while keeping backcompat for older SM versions.


KyleSanderson:
r+

PR #628, opened 2017-06-23
Kick fixes

psychonic:
Cache IClient pointer on clients, since it's also used for per-client events, not just when kicking.

Support bot IClient lookups for SDK 2013 mods that are on latest engine.

Remove unsafe kick queue bypass for bots.

### DIFF
--- a/core/PlayerManager.cpp
+++ b/core/PlayerManager.cpp
@@ -1967,6 +1967,13 @@ void CPlayer::Initialize(const char *name, const char *ip, edict_t *pEntity)
 	|| SOURCE_ENGINE == SE_BMS   \
 	|| SOURCE_ENGINE == SE_INSURGENCY
 	m_pIClient = engine->GetIServer()->GetClient(m_iIndex - 1);
+#elif SOURCE_ENGINE == SE_SDK2013
+	// Source SDK 2013 mods that ship on Steam can be using older engine binaries
+	static IVEngineServer *engine22 = (IVEngineServer *)(g_SMAPI->GetEngineFactory()("VEngineServer022", nullptr));
+	if (engine22)
+	{
+		m_pIClient = engine22->GetIServer()->GetClient(m_iIndex - 1);
+	}
 #else
 	INetChannel *pNetChan = static_cast<INetChannel *>(engine->GetPlayerNetInfo(m_iIndex));
 	if (pNetChan)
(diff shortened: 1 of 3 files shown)


KyleSanderson:
Usual drive-by, is it safe to cache m_pIClient on all mods? I know there's a lot of sensitivity with the NetChannel. I definitely like the clean-up.

psychonic:
@KyleSanderson Yes, it persists for the life of the connection.

KyleSanderson:
@psychonic we have issues with the engine not triggering our disconnection callbacks, so we can have an invalid NetChannel while still being connected (ConvarQuery comes to mind). I think if we check for a valid connection we should be okay if they're a player? I'd almost just fix this and not change the caching as terrible as that sounds.

psychonic:
If that's still a problem, not catching disconnect in some cases, it affects other things as well. It would not only cause a problem with this. We cache other pointers tied to the lifetime of the player.

core/PlayerManager.cpp
```diff
@@ -1978,6 +1960,31 @@ void CPlayer::Initialize(const char *name, const char *ip, edict_t *pEntity)
...
+	|| SOURCE_ENGINE == SE_INSURGENCY
+	m_pIClient = engine->GetIServer()->GetClient(m_iIndex - 1);
+#elif SOURCE_ENGINE == SE_SDK2013
+	// Source SDK 2013 mods that ship on Steam can be using older engine binaries
+	static IVEngineServer *engine22 = (IVEngineServer *)(g_SMAPI->GetEngineFactory()("VEngineServer022", nullptr));
+	if (engine22)
+	{
+		m_pIClient = engine22->GetIServer()->GetClient(m_iIndex - 1);
+	}
+	else
```
Ne3tCode: `else` what? There is something wrong. Maybe you wanted to write something like:
```C++
#if SOURCE_ENGINE == SE_TF2
   ...
#else
#if SOURCE_ENGINE == SE_SDK2013
	if ()
	{
	}
	else
#endif
	{
	}
#endif
```
Because your current code looks like this:
```C++
if (engine22)
{
	m_pIClient = ...
}
else
	UpdateAuthIds(); // <--
```

psychonic: I was actually just trying to demonstrate why you shouldn't toss in another commit in an extremely rushed fashion, right after creating the PR and right before running off to lunch.

Joking aside, good catch. Thank you!

Ne3tCode: Blame asherkin for approving the changes :rofl:

PR #647, opened 2017-08-03
Revert "Add natives to check for NULL_VECTOR and NULL_STRING"

asherkin:
Reverts alliedmodders/sourcemod#606

This broke extension API due to `IChangeableForward` inheriting from `IForward`.

`IForwardManager::IsVersionCompatible` needs to be changed to mark bcompat as broken here to re-merge.


peace-maker:
I've removed the new `PushNullString` and `PushNullVector` functions from `IForward` and allowed passing NULL to `PushArray` and `PushString` instead.
https://github.com/peace-maker/sourcemod/commit/c5d71dd3025766504ea06e439243bd8c50ee4a7c

This should retain backwards compatibillity while still allowing the new natives to operate.

As a side note:
I didn't touch `PushStringEx` because it already handled NULL on its own. This was explicitly disallowed in `PushArray` before. Why is NULL not allowed when passing arrays, but OK when passing strings?

PR #649, opened 2017-08-09
[basevotes / funvotes] Ignore vote actions if target is no longer available

### LINKED ISSUE #641: [CS:GO] Basevotes error
```
L 07/27/2017 - 22:39:23: [SM] Exception reported: Client index 5 is invalid
L 07/27/2017 - 22:39:23: [SM] Blaming: basevotes.smx
L 07/27/2017 - 22:39:23: [SM] Call stack trace:
L 07/27/2017 - 22:39:23: [SM]   [0] BanClient
L 07/27/2017 - 22:39:23: [SM]   [1] Line 349, /home/builds/sourcemod/linux-1.8/build/plugins/basevotes.sp::Handler_VoteCallback
```

Timocop:
Checks the target userid at the end of the vote and ignores the actions if userid is invalid instead of throwing exceptions.
Fixes #641.

### DIFF
--- a/plugins/funvotes.sp
+++ b/plugins/funvotes.sp
@@ -247,19 +247,35 @@ public int Handler_VoteCallback(Menu menu, MenuAction action, int param1, int pa
 				
 				case (burn):
 				{
-					PrintToChatAll("[SM] %t", "Set target on fire", "_s", g_voteInfo[VOTE_NAME]);					
-					LogAction(-1, g_voteClient[VOTE_CLIENTID], "Vote burn successful, igniting \"%L\"", g_voteClient[VOTE_CLIENTID]);
-					
-					IgniteEntity(g_voteClient[VOTE_CLIENTID], 19.8);	
+					if(GetClientOfUserId(g_voteClient[VOTE_USERID]) == 0)
+					{
+						PrintToChatAll("[SM] %t", "Player no longer available");
+						LogAction(-1, -1, "Vote burn failed, unable to burn \"%s\" (%d) (reason \"%s\")", g_voteInfo[VOTE_NAME], g_voteClient[VOTE_CLIENTID], "Player no longer available");
+					}
+					else
+					{
+						PrintToChatAll("[SM] %t", "Set target on fire", "_s", g_voteInfo[VOTE_NAME]);					
+						LogAction(-1, g_voteClient[VOTE_CLIENTID], "Vote burn successful, igniting \"%L\"", g_voteClient[VOTE_CLIENTID]);
+						
+						IgniteEntity(g_voteClient[VOTE_CLIENTID], 19.8);	
+					}
 				}
 				
 				case (slay):
 				{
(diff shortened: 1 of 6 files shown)


plugins/funvotes/voteburn.sp
```diff
@@ -40,6 +40,8 @@ void DisplayVoteBurnMenu(int client, int target, char[] name)
 	}
 	
 	g_voteClient[VOTE_CLIENTID] = target;
```
asherkin: `VOTE_CLIENTID` should probably go away completely.

plugins/funvotes.sp
```diff
@@ -247,19 +247,35 @@ public int Handler_VoteCallback(Menu menu, MenuAction action, int param1, int pa
...
 				case (slay):
 				{
-					PrintToChatAll("[SM] %t", "Slayed player", g_voteInfo[VOTE_NAME]);					
-					LogAction(-1, g_voteClient[VOTE_CLIENTID], "Vote slay successful, slaying \"%L\"", g_voteClient[VOTE_CLIENTID]);
-					
-					ExtinguishEntity(g_voteClient[VOTE_CLIENTID]);
-					ForcePlayerSuicide(g_voteClient[VOTE_CLIENTID]);
+					if(GetClientOfUserId(g_voteClient[VOTE_USERID]) == 0)
+					{
+						PrintToChatAll("[SM] %t", "Player no longer available");
```
asherkin: I don't think we need prints on failure (the logging can stay) - people know the client has gone.

One day these should all be changed to `ShowActivity`, but not in this PR.

Timocop: Isnt it better to let the clients know the vote has actually ended?
People may know the client has gone, but they dont know if the vote actually ended unless they start a new vote.

Timocop:
Everything ok with this?

asherkin:
I think so, and I've dismissed my reject, but I'd like someone else to take a pass over it before merging.

ErikMinekus:
Instead of putting the player name in the vote failed messages, I would use ```%L``` with ```voteTarget```, to be consistent with the vote successful messages.

And instead of literally putting "Player no longer available" as the reason for failing, I would use ```%T``` and pass ```LANG_SERVER``` as an additional argument, so that the English phrase is used.

If you want me to nitpick about the code style:
```sourcepawn
int voteTarget;
if((voteTarget = GetClientOfUserId(g_voteTarget)) == 0)
```
I would write this as:
```sourcepawn
int voteTarget = GetClientOfUserId(g_voteTarget);
if (voteTarget == 0)
```
And instead of using else and having another indentation level, you could ```return 0``` after the vote fails.

Timocop:
Logging should always be in english, or do you want to translate logs on AlliedModders all the time if someone has a problem? You can look in other AlliedModders plugins, they do the same. Same as the code style you mentioned.

ErikMinekus:
I forgot that the server language can be changed in the configs, in that case the code is fine.

Headline22:
Looks good to me 👍

asherkin:
Thanks for this!

PR #653, opened 2017-08-14
Removed gender from nominations.phrases.txt

shavitush:
I honestly didn't mind it at all, but a bunch of butt-hurt girl gamers kept complaining in my servers as if I'm the one responsible for it.


e-slut:
SEXIST!!!!!!!!!!!

PR #654, opened 2017-08-15
Remove excess assignment to null after delete

Kailo97:
Remove excess assignment.


ErikMinekus:
Can you remove these lines as well:

[funvotes.sp](https://github.com/alliedmodders/sourcemod/blob/master/plugins/funvotes.sp#L301)
[mapchooser.sp](https://github.com/alliedmodders/sourcemod/blob/master/plugins/mapchooser.sp#L685)

PR #659, opened 2017-08-18
Temp fix for weaponID related functions

Drifter321:
This is a temp fix to fix the following natives

AliasToWeaponID
GetTranslatedWeaponAlias
WeaponIDToAlias

Afaict these were either inlined or removed in windows and in linux were either removed or an additional param added. The additional param is i believe the player inventory. Im not sure the right way to approach fully fixing but i believe deprecating them on cs:go or adding an additional param to them for passing the player may be the best way to approach fixing it. Parsing the item.games.txt alone isnt actually enough (for this at least) since it appears to actually ignore the weapon name in functions.

For example doing buy ak47 on ct will give you an m4 while the reverse on t. So it appears to be based on  item_sub_position from items.games.txt which both have different id's and since these functions are commonly used within HandleBuyCommand forward they should be fixed to return the id of the weapon they are ACTUALLY going to get instead of the one based on name. Which would require the client to get the team.

As for the GetWeaponPrice, WeaponInfo is completly gone. There is no function to get the weapon price anymore this appears to be gotten from the CEconItem. This one is a bit harder to approach cause again it requires either parsing item.games.txt and again checking against the ACTUAL weapon the player is going to get since prices differ for example between AK and M4 or somehow getting the CEconItem of the item they are going to get and grabbing the price from there.


asherkin:
I'm not going to look at the code tonight, but 👍 from me on the concept.

Kailo97:
About GetTranslatedWeaponAlias.
This functions exists and have 15 alias (last "" to "").
Only one (0x1047DCC0 .dll, 0xB1DFB0 .so in latest bins, CCSBot::GiveWeapon) functing use this function.
For linux GetTranslatedWeaponAlias is 0xAC24D0 with signature: "\x55\x89\xE5\x56\x53\x31\xDB\x83\xEC\x10\x8B\x75\x08\x8D\x76\x00\x8B\x04\xDD\x2A\x2A\x2A\x2A\x89\x74\x24\x04\x89\x04\x24\xE8\x2A\x2A\x2A\x2A\x85\xC0\x74\x2A\x83\xC3\x01\x83\xFB\x0F"
But windows compiler decided for optimazing (only one call) to inline this function into CCSBot::GiveWeapon =(

KyleSanderson:
Besides SM_ARRAYSIZE which is a complete nit, this seems okay to me based on the circumstances.

I don't think passing the player would be relevant here because CS_AliasToWeaponID / CS_WeaponIDToAlias / CS_GetTranslatedWeaponAlias are agnostic to clients and are bound to weaponids (maybe these are going to become name-tags?)? Can we pass NULL or similar in?

Are the weaponids completely wrong in CS_OnBuyCommand / CS_OnGetWeaponPrice for usage with probably CS_GetWeaponPrice?

Spitballing:
Can we figure out how the game is fixing this up? Can we do similar with a setjmp/longjmp?

Drifter321:
More fixes. HandleCommand_Buy_Internal is no longer passed the weapon so the weapon is now gotten from the CEconItemDefinition (i believe that is what the function returns). OnGetWeaponPrice has a hacky fix as well the price on the CEconItemView is first grabbed the forward for OnGetWeaponPrice called and the var on the view set. Then after calling the original function the var is set back to the original price. I tested it lightly everything seems to work. The only thing that does not work is GetWeaponPrice as there is no clear enough way to grab the price of the weapon. Maybe a loadoutslot to id map to get the loadoutslot of each weapon since thats kinda required now.

gamedata/sm-cstrike.games/game.csgo.txt
```diff
@@ -13,26 +13,22 @@
...
 	{
-		"Keys"
-		{
-			"GetWeaponPriceByteCheck"	"E9"
-			"GetWeaponPriceByteCheck_Linux"	"E8"
-		}
 		"Offsets"
 		{
-			//Offset of szClassName in CCSWeaponInfo 
+			// Offset of szClassName in CEconItemDefinition(?)
```
komashchenko: This is not CEconItemDefinition, it has the name of the weapon you can get through * (char **) ((intptr_t) this + 448)

Drifter321:
I already tried that offset it does not work (unless that isn't the one for windows which is possible). But you are correct it's not a definition.

gamedata/sm-cstrike.games/game.csgo.txt
Drifter321: You are correct I realized that already.

Drifter321:
So what i thought was and item definition is actually a CCSWeaponData. The class also contains an itemdefintion ptr which is located at +16. At +140 the default name for the weapon is stored (ignores inventory basically).

Kailo97:
if just dump data of object
20: clip1 size
36: clip2 size

according sub_A7E790 (linux) members of CCSWeaponData
204: "ingame price" attribute
208: "kill award" attribute
212: char* to alias
216: "cycletime" attribute
220: "cycletime alt" attribute
224: "time to idle" attribute
228: "idle interval" attribute
232: "is full auto" attribute
236: "damage" attribute
240: "armor ratio" attribute
244: "bullets" attribute
248: "penetration" attribute
252: "flinch velocity modifier large" attribute
256: "flinch velocity modifier small" attribute
260: "range" attribute
264: "range modifier" attribute
268: "throw velocity" attribute
// think 272-280 is array[3]
272: ?
276: ?
280: ?
284: "has silencer" attribute
288: "silencer model" attribute
292: "crosshair min distance" attribute
296: "crosshair delta distance" attribute
300: "max player speed" attribute
304: "max player speed alt" attribute
308: "spread" attribute
312: "spread alt" attribute
316: "inaccuracy crouch" attribute
320: "inaccuracy crouch alt" attribute
324: "inaccuracy stand" attribute
328: "inaccuracy stand alt" attribute
332: "inaccuracy jump initial" attribute
336: "inaccuracy jump" attribute
340: "inaccuracy jump alt" attribute
344: "inaccuracy land" attribute
348: "inaccuracy land alt" attribute
352: "inaccuracy ladder" attribute
356: "inaccuracy ladder alt" attribute
360: "inaccuracy fire" attribute
364: "inaccuracy fire alt" attribute
368: "inaccuracy move" attribute
372: "inaccuracy move alt" attribute
376: "inaccuracy reload" attribute
380: "recoil seed" attribute
384: "recoil angle" attribute
388: "recoil angle alt" attribute
392: "recoil angle variance" attribute
396: "recoil angle variance alt" attribute
400: "recoil magnitude" attribute
404: "recoil magnitude alt" attribute
408: "recoil magnitude variance" attribute
412: "recoil magnitude variance alt" attribute
416: "recovery time crouch" attribute
420: "recovery time stand" attribute
424: "recovery time crouch final" attribute
428: "recovery time stand final" attribute
432: "recovery transition start bullet" attribute
436: "recovery transition end bullet" attribute
440: "unzoom after shot" attribute //  byte
441: "hide view model zoomed" attribute // byte
444: "zoom levels" attribute
448: "zoom fov 1" attribute
452: "zoom fov 2" attribute
456: "zoom time 0" attribute
460: "zoom time 1" attribute
464: "zoom time 2" attribute
468: ?
472: ?
476: "addon scale" attribute
480: ? // pointer
484: ? // pointer
488: "tracer frequency" attribute
492: "tracer frequency alt" attribute
496: ? // pointer
500: ? // pointer
508: ? // pointer
504: ? // pointer
512: ? // pointer
516: "heat per shot" attribute
520: ? // pointer
524: ? // pointer
528: "inaccuracy pitch shift" attribute
532: "inaccuracy alt sound threshold" attribute
536: "bot audible range" attribute
540: ?
544: ?
548: "wrong team msg" attribute
552: "has burst mode" attribute // byte
553: "is revolver" attribute // byte
554: "cannot shoot underwater" attribute // byte

Total size of object 556 bytes (not a guess, found mem allocate).

Drifter321:
This should fix everything now (at least for now). It isnt very sustainable and in the future a new solution should be found. This was lightly tested everything seems to work correctly. In the future the new functions should probably also be open to plugins (GetLoadoutSlotFromName atleast?).

extensions/cstrike/forwards.cpp
```diff
@@ -22,10 +24,77 @@ CDetour *DCSWeaponDrop = NULL;
...
+	else
+	{
+		const char *underscore = strstr(szClassname, "_");
+		if (underscore)
+		{
+			Q_strncpy(weaponName, (const char *)((intptr_t)underscore + 1), sizeof(weaponName));
+		}
+		else
+		{
+			Q_strcpy(weaponName, szClassname);
```
psychonic: See other note about strcpy.

Majkovsky:
I have tested the latest version of **Dev - 6142** unfortunately not working anymore
Linux - Centos

> L 08/21/2017 - 17:40:25: [SM] Exception reported: Stack leak detected: sp:80628 should be 80660!
> L 08/21/2017 - 17:40:26: [SM] Exception reported: Property "ȊR��卸CDЊRH��" not found on the gamerules proxy 
> L 08/21/2017 - 17:43:04: [CSTRIKE] Failed to lookup GetCCSWeaponData signature.

Majkovsky:
I have tested version Dev - 6143 I have Linux - Centos problem is still the same, server does not want to start I do not have any logs here.

asherkin:
@Drifter321 has all this been merged now? can the PR be closed?

Drifter321:
Merged into master yes. Was gonna close it when I pushed the changes into 1.8 but the pr can be closed unless anyone has any other comments.

Majkovsky:
Linux **SourceMod 1.9 - 6148** it works ;)
There is only one problem with the spawn grants and medishot.

Majkovsky:
And you can move the dev version to stable - 6148

PR #660, opened 2017-08-19
Fix some CS:GO gamedata

kossolax:
I've just tested some of them.

 - Weapon_Drop
 - Reload

 I'm testing Weapon_CanUse in few minutes


kossolax:
Well, Weapon_CanUse edition makes crash.

the vtable was altered between reload and this. I'm checking.

PR #662, opened 2017-08-20
Added repeating param to end of OnPlayerRunCmd and OnPlayerRunCmdPost

SM91337:
This will allow people to determine if the UserCmd is truly being sent from the client or if its previous command being repeated by the server.

You can read more information about this here: https://list.valvesoftware.com/pipermail/csgo_servers/2015-July/011465.html

The parameter was added for more flexibility rather than just not firing the forward at all.
This is a none breaking change as the parameter has been added to the end, you can optionally adjust your plugins otherwise they will continue working as before.

### DIFF
--- a/core/sourcemm_api.cpp
+++ b/core/sourcemm_api.cpp
@@ -63,7 +63,7 @@ int vsp_version = 0;
 
 PLUGIN_EXPOSE(SourceMod, g_SourceMod_Core);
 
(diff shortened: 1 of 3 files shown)


SM91337:
The correct way to do this would be to check CBasePlayer::ResendUserCmdForTimingOutPlayer but I don't know if such thing exists in games other than CSGO or where it would go.

peace-maker:
Did you actually test your changes now? You should take your time before opening a pull request..
Maybe we can use the same condition the server code uses to detect if there is no new message instead of doing that heavy comparison.

PR #668, opened 2017-08-22
Un deprecate FCVAR_PLUGIN

SM91337:
Can be useful to know if a plugin actually added the ConVar or not.


SM91337:
Fucked up PR.

psychonic:
If there's a use case, we could add a way to check that, but we can't use the old FCVAR_PLUGIN value, it has a different meaning on newer engine versions.

PR #673, opened 2017-08-26
Fix core config values not being cached

peace-maker:
Valid core config options aren't cached to be retrieved using `GetCoreConfigValue` after they've been loaded from core.cfg or set through the `sm config` root console menu.

E.g. `sm config ServerLang` would return `[SM] No such config option "ServerLang" exists.` all the time.

Stop notifying other listeners if the config key was consumed, but don't skip adding it to the cache.

Also fix `FollowCSGOServerGuidelines` always showing as unhandled command when being changed through `sm config FollowCSGOServerGuidelines yes`.

### DIFF
--- a/core/CoreConfig.cpp
+++ b/core/CoreConfig.cpp
@@ -277,7 +277,7 @@ SMCResult CoreConfig::ReadSMC_KeyValue(const SMCStates *states, const char *key,
 
 ConfigResult CoreConfig::SetConfigOption(const char *option, const char *value, ConfigSource source, char *error, size_t maxlength)
 {
-	ConfigResult result;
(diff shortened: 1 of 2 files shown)


KyleSanderson:
I don't have a client but can you track down the cset or bug that accidentally broke this?

peace-maker:
I think it never worked.
https://github.com/alliedmodders/sourcemod/commit/9e445a5adfa8948f6ddcb46b1c62e5fe569783c4

You just couldn't really notice it until https://github.com/alliedmodders/sourcemod/commit/c19712dccc188791c96dd89f0b43b7973f410ba1

PR #674, opened 2017-08-26
Added more translation to basechat

TiBarification:
Also fixed colors, because it didn't work without `\x01` at begin of string.

### DIFF
--- a/plugins/basechat.sp
+++ b/plugins/basechat.sp
@@ -334,7 +334,7 @@ void SendChatToAll(int client, const char[] message)
 		}
 		FormatActivitySource(client, i, nameBuf, sizeof(nameBuf));
 		
-		PrintToChat(i, "\x01\x04(%t) %s: \x01%s", "Say all", nameBuf, message);
+		PrintToChat(i, "\x01\x0B\x04%t: \x01%s", "Say all", nameBuf, message);
 	}
 }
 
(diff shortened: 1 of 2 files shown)


Bara20:
\x01 doesn't fix csgo colors. You need a space before \x01. [Take a look at this.](https://wiki.alliedmods.net/Csgo_quirks#Denoting_Text-Colors)

psychonic:
Do other prints from SourceMod base plugins have this extra space? I would think that it should only be added for CS:GO, if at all since it will cause these specific messages to be offset from the rest of chat.

TiBarification:
I think now it won't affect on non-csgo games.

omar-damaj:
Has this fix been pushed yet? I am still seeing this issue on CS:GO servers. Seems like a fair fix that wont affect other games and would fix it until a better fix comes out.

psychonic:
> Has this fix been pushed yet?

I haven't looked though the entire commit history, but judging by the PR still being "Open", no, I would assume that it hasn't been.

omar-damaj:
I do apologize I did not check the status... Would this be possibly pushed soon? Just wondering if it would be or if there is a fix yet.

TiBarification:
Tested on both games, tf2 and cs:go, all works fine. Ready to merge.

GabeHirakawa:
@TiBarification I'm curious, is there a reason you changed the original prefixes from ADMINS/To ADMINS?

psychonic:
@TiBarification did you test that the coloring works on CS:GO? I don't think that the 0xB trick has worked for a couple of years.

As @Bara noted, a space is commonly used now instead.

Bara:
@psychonic he added a space like here: https://github.com/alliedmodders/sourcemod/pull/674/files#diff-d6b22c2d8f239d0172a347a30daf28bcR404 ?

TiBarification:
>@TiBarification did you test that the coloring works on CS:GO?

Yes, it works fine.

>I don't think that the 0xB trick has worked for a couple of years.

Maybe it is really minor `symbol/color code`. I was guided by wiki.

psychonic:
Wow, I totally missed that, only still seeing the 0B (which isn't needed or doing anything, but barely harming anything). Good catch. Thanks.

PR #678, opened 2017-09-09
add VS Preview 2017 as valid compiler (msvc-1911)

thewavelength:
When someone uses Visual Studio 2017 Preview, the compiler version increased from 1910 to 1911.
Compiling SourceMod with 2017 Preview works fine. The build script has to be adjusted.


core/AMBuilder
```diff
@@ -77,6 +77,8 @@ for sdk_name in SM.sdks:
         vs_year = '2015'
       elif msvc_ver == 1910:
         vs_year = '2017'
+      elif msvc_ver == 1911:
```
WildCard65: Instead of duplicating the if statement, add an or case to the previous iteration for VS 2017 check.

PR #681, opened 2017-09-14
add support for SourceMod libraries in GameConfig files

thewavelength:
Adds support for further SourceMod libraries in GameConfig files.
Supported libraries included:
- `sourcemod` (SourceMod core)
- `sourcemod.logic` (SourceMod core/logic)
- `sourcepawn.jit` (SourcePawn JIT)

I need this for a current project and would be happy if this makes it into the master.


Drifter321:
No.

PR #688, opened 2017-09-19
Migrate CDataPack from a Compact Cassette tape.

KyleSanderson:
This lift and shifts CDataPack from a linear datatype to something using a vector with type mixed in. Iterators will change (again), so plugins manually using SetPackPosition without ReadPackPosition (against the API) will experience breakage; if this even builds.

### DIFF
--- a/core/logic/CDataPack.cpp
+++ b/core/logic/CDataPack.cpp
@@ -198,10 +198,20 @@ bool CDataPack::IsReadable(size_t bytes) const
 const char *CDataPack::ReadString(size_t *len) const
 {
 	if (!this->IsReadable())
+	{
+		if (len)
+			*len = 0;
+
 		return "";
+	}
 
 	if (this->elements[this->position].type != CDataPackType::String)
+	{
+		if (len)
+			*len = 0;
+
 		return "";
+	}
 
 	const ke::AString &val = *static_cast<ke::AString *>(this->elements[this->position++].pData.vval);
 	if (len)
--- a/core/logic/CDataPack.cpp
+++ b/core/logic/CDataPack.cpp
@@ -46,7 +46,7 @@ CDataPack::~CDataPack()
 
 static ke::Vector<ke::AutoPtr<CDataPack>> sDataPackCache;
 
-IDataPack * CDataPack::New()
+IDataPack *CDataPack::New()
 {
   if (sDataPackCache.empty())
     return new CDataPack();
@@ -205,20 +205,21 @@ const char *CDataPack::ReadString(size_t *len) const
 
 void *CDataPack::GetMemory() const
 {
-	if (!IsReadable() || elements[position].type != CDataPackType::Raw)
-		return nullptr;
-
-	return &(reinterpret_cast<size_t *>(elements[position].pData.vval)[1]);
+	return nullptr;
 }
 
 void *CDataPack::ReadMemory(size_t *size) const
 {
-	void *ptr = GetMemory();
-	if (ptr != nullptr)
-		++position;
-	
+	void *ptr = nullptr;
+	if (!IsReadable() || elements[position].type != CDataPackType::Raw)
+		return ptr;
+
+	size_t *val = reinterpret_cast<size_t *>(elements[position].pData.vval);
+	ptr = &(val[1]);
+	++position;
+
 	if (size)
-		*size = (reinterpret_cast<size_t *>(ptr)[-1]); /* Egor!!!! */
+		*size = val[0]; /* Egor!!!! */
 
 	return ptr;
 }
--- a/core/logic/CDataPack.cpp
+++ b/core/logic/CDataPack.cpp
@@ -33,16 +33,15 @@
 #include <string.h>
 #include "CDataPack.h"
 #include <amtl/am-autoptr.h>
-#include <amtl/am-string.h>
 
 CDataPack::CDataPack()
 {
-	this->Initialize();
+	Initialize();
 }
 
 CDataPack::~CDataPack()
 {
-	this->Initialize();
+	Initialize();
 }
 
 static ke::Vector<ke::AutoPtr<CDataPack>> sDataPackCache;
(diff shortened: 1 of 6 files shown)


Headline22:
Floats seem to show 0.0 on ReadFloat.

```cpp
public void OnPluginStart()
{
	int integer = 32;
	float floatingPoint = 32.0;
	char[] string = "Hello.";
	
	DataPack pack = new DataPack();
	
	pack.WriteCell(integer);
	pack.WriteFloat(floatingPoint);
	pack.WriteString(string);
	
	DataPackPos pos = pack.Position;
	
	pack.Reset();
	
	int integer2;
	float floatingPoint2;
	char string2[8];
	
	integer2 = pack.ReadCell();
	floatingPoint2 = pack.ReadFloat();
	pack.ReadString(string2, sizeof(string2));
	
	PrintToServer("int: %i, float: %.2f, string: %s, pos: %i",
					integer2,
					floatingPoint2,
					string2,
					pos);
	
	

	delete pack;
}
```

**Output:**
```
int: 32, float: 0.00, string: Hello., pos: 1
[SM] Plugin testcase.smx reloaded successfully.
```

Headline22:
So the test case for DataPack operations seems to work as expected now.

👍 From me 

<details>
<summary>Test case </summary>

```cpp
public void OnPluginStart()
{
	int integer = 32;
	float floatingPoint = 32.0;
	char[] string = "Hello.";
	
	DataPack pack = new DataPack();
	
	pack.WriteCell(integer);
	pack.WriteFloat(floatingPoint);
	pack.WriteString(string);
		
	pack.Reset();
	
	int integer2;
	float floatingPoint2;
	char string2[8];
	
	integer2 = pack.ReadCell();
	floatingPoint2 = pack.ReadFloat();
	
	pack.ReadString(string2, sizeof(string2));
	
	PrintToServer("int: %i, float: %.2f, string: %s",
					integer2,
					floatingPoint2,
					string2);
	
	char string3[8];
	PrintToServer("%i", pack.Position);
	pack.Position--; // go back to string
	PrintToServer("%i", pack.Position);

	pack.ReadString(string3, sizeof(string3));
	PrintToServer("%s", string3);
	
	delete pack;
}
```
</details>


<details>
<summary>Actual output </summary>

```
int: 32, float: 32.00, string: Hello.
3
2
Hello.
[SM] Plugin testcase.smx reloaded successfully.
```
</details>

core/logic/smn_datapacks.cpp
```diff
@@ -185,11 +185,16 @@ static cell_t smn_ReadPackCell(IPluginContext *pContext, const cell_t *params)
...
 
-	if (!pDataPack->IsReadable(sizeof(char) + sizeof(size_t) + sizeof(cell_t)))
+	if (!pDataPack->IsReadable())
 	{
 		return pContext->ThrowNativeError("DataPack operation is out of bounds.");
 	}
 
+	if (pDataPack->GetCurrentType() != CDataPackType::Cell)
+	{
+		return pContext->ThrowNativeError("Invalid Datapack Type (got %d / expected %d).", pDataPack->GetCurrentType(), CDataPackType::Cell);
```
Fyren: Please check all the error messages for consistency, including the ones you didn't touch.  The docs use "data pack" and errors shouldn't be in title case but normal sentence case.

I'd also recommend prettifying the enum types for the user so they see names and not the integer value of the enum.  Lots (most?  all?) of our other errors don't do anything like that, so it's definitely not required.

core/logic/CDataPack.cpp
```diff
@@ -72,286 +66,171 @@ CDataPack::Free(IDataPack *pack)
...
+	this->position = 0;
 }
 
 void CDataPack::ResetSize()
 {
-	m_size = 0;
+	this->Initialize();
 }
 
 size_t CDataPack::CreateMemory(size_t size, void **addr)
```
Fyren: Do we even know if a single person is using the generic/raw memory functions?  This functionality is not exposed to plugins, so it'd have to be an extension.  I'd be in favor of axing these completely.  (Since they're in the interface, just make them no-ops/return null.)

If they're kept, the docs in IDataPack.h about pointers being invalidated (in the case of realloc) are wrong and should be updated.

You also alloc a byte more than necessary?  Did you mean to store the size here (originally stored as 32-bits)?  Related to the behavior change I mention for ReadMemory?

core/logic/CDataPack.cpp
```diff
@@ -72,286 +66,171 @@ CDataPack::Free(IDataPack *pack)
...
-	}
-	if (*reinterpret_cast<char *>(m_curptr) != String)
-	{
-		return NULL;
-	}
-	m_curptr += sizeof(char);
-
-	size_t real_len = *(size_t *)m_curptr;
+	if (!this->IsReadable())
+		return "";
```
Fyren: The API says "pointer to the string, or NULL if out of bounds," so the old behavior seems correct to me.  We should actually use `nullptr` now, though (and switch any other uses in the file as well).

core/logic/CDataPack.cpp
```diff
@@ -72,286 +66,171 @@ CDataPack::Free(IDataPack *pack)
...
 
-	m_curptr += sizeof(size_t);
-	char *str = (char *)m_curptr;
-
-	if ((strlen(str) != real_len) || !(IsReadable(real_len+1)))
-	{
-		return NULL;
-	}
+	if (this->elements[this->position].type != CDataPackType::String)
+		return "";
```
Fyren: The API doesn't say what would happen in this case, but the old behavior was to return `NULL`, so I'd also make it `nullptr`.

core/logic/CDataPack.cpp
```diff
@@ -72,286 +66,171 @@ CDataPack::Free(IDataPack *pack)
...
+	if (!this->IsReadable())
+		return nullptr;
+	
+	if (this->elements[this->position].type != CDataPackType::Raw)
+		return nullptr;
+
+	return this->elements[this->position].pData.vval;
 }
 
 void *CDataPack::ReadMemory(size_t *size) const
```
Fyren: Behavior here appears to have changed?  The old behavior checked the stored size was <= the requested size and returned null if it wasn't.  It also returned the real size of the memory chunk, which you haven't stored.

core/logic/CDataPack.cpp
```diff
@@ -33,25 +33,19 @@
...
-
-#define DATAPACK_INITIAL_SIZE 64
+#include <amtl/am-string.h>
 
 CDataPack::CDataPack()
 {
-	m_pBase = (char *)malloc(DATAPACK_INITIAL_SIZE);
-	m_capacity = DATAPACK_INITIAL_SIZE;
-	Initialize();
+	this->Initialize();
```
Fyren: Style: Why are you using `this->` everywhere?  SM code in general doesn't appear to use this style.

core/logic/CDataPack.cpp
```diff
@@ -72,286 +66,171 @@ CDataPack::Free(IDataPack *pack)
...
 void CDataPack::Initialize()
 {
-	m_curptr = m_pBase;
-	m_size = 0;
-}
-
-void CDataPack::CheckSize(size_t typesize)
-{
-	if (m_curptr - m_pBase + typesize <= m_capacity)
+	for (size_t iter = 0; iter < this->elements.length(); ++iter)
```
Fyren: Style: this isn't an iterator, it's an index.

core/logic/CDataPack.cpp
```diff
@@ -72,286 +66,171 @@ CDataPack::Free(IDataPack *pack)
...
+		*len = val.length();
 	}
 
-	m_curptr += real_len + 1;
-
-	return str;
+	return val.chars();
 }
 
 void *CDataPack::GetMemory() const
```
Fyren: This is not the same behavior.  The analogue in the new implementation would be returning the address of the InternalPack.  However, I don't see how/why anyone would ever use this function because the pointer is opaque to the user.

core/logic/CDataPack.cpp
```diff
@@ -72,286 +66,171 @@ CDataPack::Free(IDataPack *pack)
...
-	m_curptr += sizeof(size_t);
-
-	float val = *reinterpret_cast<float *>(m_curptr);
-	m_curptr += sizeof(float);
-	return val;
+	
+	return this->elements[this->position++].pData.fval;
 }
 
 bool CDataPack::IsReadable(size_t bytes) const
```
Fyren: The docs here utterly suck.  The imply the old implementation and so this is not really satisfiable with the new one.

PR #692, opened 2017-09-24
Update LogAction to include map param.

RussianRonin:


KyleSanderson:
Appreciate the thought, but this is fundamentally wrong unfortunately. You can't pass an array into a native that takes a single cell. The native is also expecting a client, not a map filename.

RussianRonin:
Oh ok. Thanks for the info!

Issue #645, opened 2017-08-01
Forwards still push disconnected client indexes


Timocop:
**MetaMod:Source v1.11.0-dev+1096V
Stripper v1.2.2
SourceMod v1.9.0.6119
Left 4 Dead (Linux)**

Everytime SMAC or other plugins kicks a player (not bot) some sdkhook forwards still fire even when the client has been kicked. Which also floods my sourcemod errors logs for ~1sec.
This only happend after the sdkhook fix (fd87cb6b239fc7a7cf3703c30c53396752efb18b, 29d118743446c2116d36ddb543f8673b8836944d).
Or this this intended after the fix?
Look at the timespan in the logs below.

## Connection log:
> 0/32 Players - 21:45:56 - <[DE]tom_the_teacher> <STEAM_1:0:***> <***> DISCONNECTED after 31 minutes. \<All players idle\>

_("All players idle" is a plugin which kicks all afk players when there are no non-afk players left)_

## Sourcemod error logs:
> L 07/30/2017 - 21:45:57: [SM] Exception reported: Entity 3 (3) is invalid
> L 07/30/2017 - 21:45:57: [SM] Blaming: fo_fatman.smx
> L 07/30/2017 - 21:45:57: [SM] Call stack trace:
> L 07/30/2017 - 21:45:57: [SM]   [0] GetEntPropEnt
> L 07/30/2017 - 21:45:57: [SM]   [1] Line 976, D:\Left 4 Dead Modding\L4D SERVER\L4D SERVER\scripting\include\smlib/clients.inc::Client_GetActiveWeapon
> L 07/30/2017 - 21:45:57: [SM]   [2] Line 657, fo_fatman.sp::OnPlayerRunCmd
> L 07/30/2017 - 21:45:57: [SM] Exception reported: Client 3 is not in game
> L 07/30/2017 - 21:45:57: [SM] Blaming: smac_speedhack.smx
> L 07/30/2017 - 21:45:57: [SM] Call stack trace:
> L 07/30/2017 - 21:45:57: [SM]   [0] IsPlayerAlive
> L 07/30/2017 - 21:45:57: [SM]   [1] Line 145, smac_speedhack.sp::OnPlayerRunCmd
> L 07/30/2017 - 21:45:57: [SM] Exception reported: Client 3 is not in game
> L 07/30/2017 - 21:45:57: [SM] Blaming: l4d_teammate_light.smx
[...]

PR #693, opened 2017-09-24
Fix OnPlayerRunCmd crashing with invalid CUserCmd ptr (Issue #645).

KyleSanderson:
Issue reported by Peace-Maker in issue #645.


asherkin:
If I'm reading the minidump that P-M linked correctly, the CUserCmd wasn't `0x00000000`, it was `0x00000003`, so this won't help.

Generally "1 owner / 1 ip" crash reports are not worth any time, especially not in the presence of a number of related non-standard extensions (`tangotimer_doubletracking.ext.2.csgo.so`, `cmdfix.ext.2.csgo.so` in that reports case).

peace-maker:
The first extension just exposes 2 natives to do float math on a double in c++ since sourcepawn is 32bit floats only, the second one is GoD-Tony's.

PR #697, opened 2017-09-26
Making blind's LogAction message consistent...

hlstriker:
with the other plugin's messages.

### DIFF
--- a/plugins/funcommands/blind.sp
+++ b/plugins/funcommands/blind.sp
@@ -76,7 +76,7 @@ void PerformBlind(int client, int target, int amount)
 	
 	EndMessage();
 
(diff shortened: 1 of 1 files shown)


KyleSanderson:
@hlstriker merge it all into one PR man.

hlstriker:
@KyleSanderson I used github's edit file feature and couldn't figure out how to merge them together. Knew I would get yelled at for that, sorry :)

KyleSanderson:
@hlstriker totally would have just msg'd on steam if I had it on here; all's well ;)

If there's anything else like this though where it's one-liners; lets fix it en-mass.

PR #698, opened 2017-10-01
Quick helper TraceRayClientAimEndPoint

ambaca:
More stocks? Helpers?
I was bored. Testing.
This bad example ?

Simple function to get quick the end position where player is looking at. Player index itself have filter out.

### DIFF
--- a/plugins/include/sdktools_stocks.inc
+++ b/plugins/include/sdktools_stocks.inc
@@ -73,3 +73,29 @@ stock int FindTeamByName(const char[] name)
 
 	return found_team;
 }
+
+/**
+ * Get the end point coordinates where client aim at.
+ *
+ * @param client			Player Index.
+ * @param pos				Destination vector to store the end point coordinates.
+ * @param flags				Trace flags. MASK_SHOT by default.
+ */
+stock void TraceRayClientAimEndPoint(int client, float pos[3], int flags = MASK_SHOT)
+{
+	float client_pos[3], client_ang[3];
+
+	GetClientEyePosition(client, client_pos);
+	GetClientEyeAngles(client, client_ang);
+
+	Handle trace = TR_TraceRayFilterEx(client_pos, client_ang, flags, RayType_Infinite, TraceRayClientAimEndPointCB, client);
+	TR_GetEndPosition(pos, trace);
+	delete trace;
+}
+public bool TraceRayClientAimEndPointCB(int entity, int contentsMask, any data)
+{
+	if(data == entity)
+		return false;
+
+	return true;
+}
--- a/plugins/include/sdktools_stocks.inc
+++ b/plugins/include/sdktools_stocks.inc
@@ -77,29 +77,39 @@ stock int FindTeamByName(const char[] name)
 /**
  * Get the end point coordinates where client aim at.
  *
- * @param client			Player Index.
- * @param pos				Destination vector to store the end point coordinates.
- * @param flags				Trace flags. MASK_SHOT by default.
+ * @param client		Player Index.
+ * @param pos			Destination vector to store the end point coordinates.
+ * @param flags			Trace flags. MASK_SHOT by default.
+ * @return				True when aim end point hit inside skybox, playable area. "pos" argument return new coordinates.
+						False when hit unplayable area or out of space. "pos" argument return unchanged.
+ * @error				Invalid client index or client is not in game.
  */
-stock void TraceRayClientAimEndPoint(int client, float pos[3], int flags = MASK_SHOT)
+stock bool ClientAimEndPos(int client, float pos[3], int flags = MASK_SHOT)
 {
-
-	if( !(0 < client <= MaxClients) || !IsClientInGame(client))
-		return;
-
 	float client_pos[3], client_ang[3];
 
 	GetClientEyePosition(client, client_pos);
 	GetClientEyeAngles(client, client_ang);
 
-	Handle trace = TR_TraceRayFilterEx(client_pos, client_ang, flags, RayType_Infinite, TraceRayClientAimEndPointCB, client);
-	TR_GetEndPosition(pos, trace);
-	delete trace;
-}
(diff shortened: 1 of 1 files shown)


plugins/include/sdktools_stocks.inc
```diff
@@ -73,3 +73,33 @@ stock int FindTeamByName(const char[] name)
...
+	if( !(0 < client <= MaxClients) || !IsClientInGame(client))
+		return;
+
+	float client_pos[3], client_ang[3];
+
+	GetClientEyePosition(client, client_pos);
+	GetClientEyeAngles(client, client_ang);
+
+	Handle trace = TR_TraceRayFilterEx(client_pos, client_ang, flags, RayType_Infinite, TraceRayClientAimEndPointCB, client);
+	TR_GetEndPosition(pos, trace);
```
Headline22: You should probably be checking if TR_DidHit here. Maybe make the function return true if pos was populated and the trace ray actually hit.

plugins/include/sdktools_stocks.inc
```diff
@@ -73,3 +73,33 @@ stock int FindTeamByName(const char[] name)
...
+	delete trace;
+}
+public bool TraceRayClientAimEndPointCB(int entity, int contentsMask, any data)
```
Headline22: public should be 'stock'

plugins/include/sdktools_stocks.inc
```diff
@@ -73,3 +73,33 @@ stock int FindTeamByName(const char[] name)
...
 }
+
+/**
+ * Get the end point coordinates where client aim at.
+ *
+ * @param client			Player Index.
+ * @param pos				Destination vector to store the end point coordinates.
+ * @param flags				Trace flags. MASK_SHOT by default.
+ */
+stock void TraceRayClientAimEndPoint(int client, float pos[3], int flags = MASK_SHOT)
```
Headline22: Maybe "GetClientAimPos" is a better name

Headline22: Maybe "GetClientAimPos" is a better name

plugins/include/sdktools_stocks.inc
MitchDizzle: I'd be curious to see if 'private' would work also.

Headline22: @MitchDizzle private is a reserved keyword which is currently unused

MitchDizzle: Oh right, i was using statics to imitate private variables.

plugins/include/sdktools_stocks.inc
```diff
@@ -73,3 +73,33 @@ stock int FindTeamByName(const char[] name)
...
+	GetClientEyePosition(client, client_pos);
+	GetClientEyeAngles(client, client_ang);
+
+	Handle trace = TR_TraceRayFilterEx(client_pos, client_ang, flags, RayType_Infinite, TraceRayClientAimEndPointCB, client);
+	TR_GetEndPosition(pos, trace);
+	delete trace;
+}
+public bool TraceRayClientAimEndPointCB(int entity, int contentsMask, any data)
+{
+	if(data == entity)
```
MitchDizzle: Only two paths, this could be the return value
`return (data != entity)`

plugins/include/sdktools_stocks.inc
MitchDizzle: With what Headline suggested, maybe this shouldn't return void, and should return a boolean. `true` if the pos argument is changed, `false` if something went wrong in getting the end point.

plugins/include/sdktools_stocks.inc
```diff
@@ -73,3 +73,33 @@ stock int FindTeamByName(const char[] name)
...
+ *
+ * @param client			Player Index.
+ * @param pos				Destination vector to store the end point coordinates.
+ * @param flags				Trace flags. MASK_SHOT by default.
+ */
+stock void TraceRayClientAimEndPoint(int client, float pos[3], int flags = MASK_SHOT)
+{
+
+	if( !(0 < client <= MaxClients) || !IsClientInGame(client))
+		return;
```
MitchDizzle: This is just a minor nit but instead of just returning because the client isn't valid we should throw an error or remove these checks. The developer using this stock should add these checks before even thinking about using this stock.

Headline22: If the author of the pr change the function to return a bool, it'd be ok to just return false here, or not check it at all.

ambaca:
hmmm, does this look good now? Or is this whole stock superfluous ?
I was thinking somekind fast-ready function in SM API, so you not need make whole TraceRay code separatelly in your plugin work 😛

Headline22:
Looks good to me, nice work.

KyleSanderson:
I'm not certain we have public callbacks hidden in stocks like this anywhere else. I'm also not certain we want to start introducing semi-hidden code paths like this where the filter isn't a param.

Timocop:
And the callback name is not realy unique enough there could be some collisions happening. Maybe add a prefix infront of it like smlib does?

Headline22:
@KyleSanderson If we had the filter as a parameter it would kill this abstraction, making it less usable for those new to sourcemod. Perhaps if we make the convention for internal sourcemod callbacks for stocks like this '__<calbackname>', we'd be able to get away with it. Almost no scripts would break and we could advise against registering callbacks with the '__' prefix

Headline:
While this does have some utility, it’s not really fit to belong as a part of sourcemod. Thanks for your patience and for considering a contribution!

MitchDizzle:
Speaking of not really fitting, why is FindTeamByName a thing? It seems to be a little too restrictive of a stock to be useful for anyone. Almost like it exists because GetTeamName exists.

asherkin:
It was originally expected that team indexes (and lots of other things) would be a lot more dynamic across games, so there is a lot of redundancy in the older SM features to allow control via gamedata.

PR #699, opened 2017-10-03
Remove spacers from MenuManager when we have control options.

KyleSanderson:
This has been a long standing gripe with SM menus that I forgot about until today.

When I migrated my environment to SM from Mani Admin Plugin (MAP) in 2k9 the one thing that always irk'd me was that SM drew a ton of whitespace if there were control options at the bottom. I meant to get this in for SM 1.5, and then again for 1.6 but always forgot about it...

Turns out this was a feature in SM and might be a "personal change". Screenshots attached.

Before: https://i.imgur.com/vxGvLH5.jpg
After: https://i.imgur.com/TvuCdKt.jpg


dvander:
Make it a pref

PR #701, opened 2017-10-04
CStrike extension changes for CS:GO

Drifter321:
Here are hopefully the final changes related to the August CS:GO update. This is mostly a cleanup of all the hack fixes, moving away (sort of) from the CSWeaponID enum and optimizing code slightly. I have tested this on both CS:GO (win/lin) and CS:S (win) albeit lightly.

The enum is still used however, new weapons will be added to the include enum (but dont have to be technically) and their values will be their item definition index's. Previously weapons that replaced another in loadout were hardcoded to return the replaced items id, this is no longer the case. This i believe is the best way forward, it breaks very little plugins. Mostly it should only break weapon restrict (which I already made fixes for). 

Two new natives were also added, one to go from weaponid to itemdefindex and vice versa. Additionally slight code changes were made a long the way.


plugins/include/cstrike.inc
```diff
@@ -358,7 +380,7 @@ native int CS_WeaponIDToAlias(CSWeaponID weaponID, char[] destination, int len);
  * @param weaponID		WeaponID to check
  * @return				Returns true if its a valid WeaponID false otherwise.
  *
- * @note This will return false always for CSWeapon_NONE
+ * @note This will return false always for CSWeapon_NONE. Should only be called after OnMapStart since weapon info isnt intialized before.
```
KyleSanderson: We should make this a native error on CS:GO if the map hasn't started.

https://github.com/alliedmodders/sourcemod/blob/65a8f3cbf8223a387a765bbcbbcfc312d27dce93/extensions/sdktools/vnatives.cpp#L925

extensions/cstrike/natives.cpp
```diff
@@ -898,6 +862,35 @@ static cell_t CS_UpdateClientModel(IPluginContext *pContext, const cell_t *param
...
+{
+#if SOURCE_ENGINE == SE_CSGO
+	ItemIndexMap::Result res = g_mapDefIdxToClass.find((uint16_t)params[1]);
+
+	if (!res.found())
+		return  pContext->ThrowNativeError("Invalid item definition passed.");
+
+	return res->value.m_iWeaponID;
+#else
+	return pContext->ThrowNativeError("CS_ItemDefIndexToID is not supported on this game");
```
KyleSanderson: We can't use TestFeature if we do it like this, we're better off (imo) not binding these natives to begin with on unsupported platforms.

extensions/cstrike/itemdef-hash.h
```diff
@@ -0,0 +1,95 @@
...
+	{
+		m_iLoadoutSlot = iLoadoutSlot;
+		m_iPrice = iPrice;
+		m_iDefIdx = iDefIdx;
+		m_iWeaponID = iWeaponID;
+
+		Q_strncpy(m_szClassname, szClassname, sizeof(m_szClassname));
+
+		//Create the item name
+		Q_strncpy(m_szItemName, GetWeaponNameFromClassname(szClassname), sizeof(m_szItemName));
```
KyleSanderson: I know we have a couple of calls using Q_strncpy from the SDK; I think we're rolling with ke::SafeStrcpy as the supported replacement.

plugins/include/cstrike.inc
Drifter321: In some cases it will be fine, for example in CS:GO or after the first map. The note is more a caution so people know they shouldn't rely on it returning true before OnMapStart.

extensions/cstrike/natives.cpp
Drifter321: This follows the other natives that dont work on CS:S

psychonic:
I actually agree with @KyleSanderson on this one, but changing them all
is outside the scope of this PR.

Ruben Gonzalez wrote:

plugins/include/cstrike.inc
```diff
@@ -130,7 +130,29 @@ enum CSWeaponID
...
+	CSWeapon_KNIFE_FLIP = 505,
+	CSWeapon_KNIFE_GUT = 506,
+	CSWeapon_KNIFE_KARAMBIT = 507,
+	CSWeapon_KNIFE_M9_BAYONET = 508,
+	CSWeapon_KNIFE_TATICAL = 509,
+	CSWeapon_KNIFE_FALCHION = 512,
+	CSWeapon_KNIFE_SURVIVAL_BOWIE = 514,
+	CSWeapon_KNIFE_BUTTERFLY = 515,
+	CSWeapon_KNIFE_PUSH = 516,
+	CSWeapon_MAX_WEAPONS //THIS MUST BE LAST, EASY WAY TO CREATE LOOPS. When looping, do CS_IsValidWeaponID(i), to check.
```
asherkin: Looping over this seems like a terrible idea with a 400-wide gap?

extensions/cstrike/util_cstrike.h
```diff
@@ -31,158 +31,9 @@
...
-	CSGOWeapon_C4, //49
-	CSGOWeapon_KEVLAR = 50,
-	CSGOWeapon_ASSAULTSUIT,
-	CSGOWeapon_NVG,
-	CSGOWeapon_DEFUSER
-};
-#endif
+ //THIS IS THE INCLUDE ENUM DO NOT CHANGE ONLY UPDATE THE INCLUDE
+ //This is used to match to old weaponid's to their correct enum value
+ //Anything after defuser will pass the itemdef as they will be the id set in include
```
asherkin: Is `defuser` here still correct?

plugins/include/cstrike.inc
Drifter321: Probably true, that is why i added one in the middle (CSWeapon_MAX_WEAPONS_NO_KNIFES). This one will probably be used more often, and if you want the knifes an additional loop could be used from CSWeapon_BAYONET to CSWeapon_MAX_WEAPONS.

extensions/cstrike/util_cstrike.h
Drifter321: Good catch, it should be heavy assault suit, for some reason they added it to game later in time but used a one in the middle instead of appending to the end. I didnt know that until after I wrote the comment will fix.

plugins/include/cstrike.inc
peace-maker: You could add something like CSWeapon_FIRST_KNIFE if there is ever going to be a new knife inserted before the bayonet?

Drifter321: The seem to be determined to add knifes in the 500's so i think its fine.

PR #702, opened 2017-10-08
Ditch C header/Fix build warnings

Headline22:
Fixing some warnings that were spit out due to unsafe double -> float comparisons, and some other easily-fixable warnings were corrected, as well.

~~Someone please double check that the change in MernsenneTwister.h is safe. The warning was~~ 
~~```\sourcemod\core\logic\MersenneTwister.h(128): warning C4146: unary minus operator applied to unsigned type, result still unsigned```~~

^Redacted from pr

### DIFF
--- a/extensions/sdkhooks/extension.cpp
+++ b/extensions/sdkhooks/extension.cpp
@@ -957,7 +957,7 @@ bool SDKHooks::Hook_CanBeAutobalanced()
 			// Only update our new ret if different from original
 			// (so if multiple plugins returning different answers,
 			//  the one(s) that changed it win)
-			if (res != origRet)
+			if ((bool)res != origRet)
 				newRet = !origRet;
 		}
 
--- a/core/logic/smn_float.cpp
+++ b/core/logic/smn_float.cpp
@@ -29,7 +29,7 @@
  * Version: $Id$
  */
 
-#include <math.h>
+#include <cmath>
 #include <string.h>
 #include <stdlib.h>
 #include "common_logic.h"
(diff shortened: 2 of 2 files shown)


TomyLobo:
All those ``(float)`` casts would be unnecessary if C++ headers had been used instead of C headers.
wrong: ``#include "math.h"``
still wrong: ``#include <math.h>``
correct: ``#include <cmath>``

<sub><sup>Keep in mind that the C++ stdlib resides in the ``std`` namespace.</sup></sub>

In the correct C++ header, there's an overload for floats for all the math functions.

Btw, the same goes for all the other C headers.
In addition to ``<cmath>``, there are ``<climits>``, ``<cstdio>``, ``<ctime>``, ``<cstdlib>`` and others, so the recommendation is to switch over all of the headers to the correct C++ headers.
If the authors are not willing to make a fundamental change like that, the C headers also have float variants of the math functions.
Their names have a suffix of "f" (like ``ceilf``, ``sinf``) to work around C's lack of overloading.

extensions/sdkhooks/extension.cpp
```diff
@@ -957,7 +957,7 @@ bool SDKHooks::Hook_CanBeAutobalanced()
 			// Only update our new ret if different from original
 			// (so if multiple plugins returning different answers,
 			//  the one(s) that changed it win)
-			if (res != origRet)
+			if ((bool)res != origRet)
```
TomyLobo: Why would that cast be necessary?

Issue #700, opened 2017-10-03
Issue with the DBI driver


Pukata:
After the csgo update around a week ago, i started getting crashes on my server.
I have many plugins which makes use of the DBI extension, so hard to pinpoint the exact behaviour that causes the crash, especially since they appear randomly, and need quite a few players on the server (15+ players) to trigger a crash. But they all tend to happen at the start of the map, when everyone is connecting, and a lot of database calls needs to be made.

Stack trace:
https://crash.limetech.org/sg4jaud5vyu3

Server info:
Linux version 4.9.30 (gcc version 6.3.0 20170516 (Debian 6.3.0-18) ) 
Metamod:Source version 1.10.7-dev
SourceMod Version: 1.8.0.6025

Also worth noting that im running the server in an lxc container.

Had a look at some crash statistics, and seems like im not the only one with the driver issue:
https://crash.limetech.org/stats/dbi.mysql.ext.so

peace-maker:
Looks like you're mixing threaded and non-threaded queries in a plugin.

Pukata:
But that hasn't been a problem up until now?

PR #704, opened 2017-10-12
Disable SQL_(Un)LockDatabase and always lock/unlock on blocking queries.

KyleSanderson:
This is too hard for people to use and is a ghost from the past https://forums.alliedmods.net/showpost.php?p=1406603&postcount=16 .

I've seen weird things in accelerator, there's issues like https://github.com/alliedmodders/sourcemod/issues/700 every couple months. Doubling down on this, if someone locks the DB and encounters a plugin error; they get into the weird state of having the database locked without being able to check if it's actually locked. Instead of this, every blocking query we should just lock/unlock for them. If there's no contention, they get the lock pretty darn quickly, and if there is contention, they should have taken the mutex anyways.

### DIFF
--- a/core/logic/smn_database.cpp
+++ b/core/logic/smn_database.cpp
@@ -800,7 +800,9 @@ static cell_t SQL_Query(IPluginContext *pContext, const cell_t *params)
 	pContext->LocalToString(params[2], &query);
 
 	IQuery *qr;
-	
+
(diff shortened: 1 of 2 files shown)


peace-maker:
You should add locking to other natives that require that too if this is wanted.
https://wiki.alliedmods.net/SQL_(SourceMod_Scripting)#Locking

PR #706, opened 2017-10-19
Add KeyValues.ExportToString

Headline22:
Before this pull request, in order for a plugin to get a string version of the KeyValues tree they'd have to export to a file and read it back in. That is the reason this was created. General use of this functionality looks something like this:

```cpp
int length = kv.ExportLength;
char[] buffer = new char[length];
kv.ExportToString(buffer, length);
```

### DIFF
--- a/core/smn_keyvalues.cpp
+++ b/core/smn_keyvalues.cpp
@@ -1113,7 +1113,7 @@ static cell_t KeyValues_Import(IPluginContext *pContext, const cell_t *params)
 	return smn_CopySubkeys(pContext, new_params);
 }
 
-// KeyValues.ExportToString(char[] buffer, int maxlen);
+// int KeyValues.ExportToString(char[] buffer, int maxlen);
 static cell_t smn_KeyValuesToString(IPluginContext *pContext, const cell_t *params)
 {
 	Handle_t hndl = static_cast<Handle_t>(params[1]);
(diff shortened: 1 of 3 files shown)


core/smn_keyvalues.cpp
```diff
@@ -1113,6 +1113,67 @@ static cell_t KeyValues_Import(IPluginContext *pContext, const cell_t *params)
 	return smn_CopySubkeys(pContext, new_params);
 }
 
+// int KeyValues.ExportToString(char[] buffer, int maxlen);
```
KyleSanderson: we don't have comments like this anywhere else in the file from what I can see. Atleast not immediately around it.

SM91337:
:thumbsup: Very useful ty.

core/smn_keyvalues.cpp
```diff
@@ -1113,6 +1113,65 @@ static cell_t KeyValues_Import(IPluginContext *pContext, const cell_t *params)
...
+		!= HandleError_None)
+	{
+		return pContext->ThrowNativeError("Invalid key value handle %x (error %d)", hndl, herr);
+	}
+	KeyValues *kv;
+	CUtlBuffer buffer;
+	
+	kv = pStk->pCurRoot.front();
+
+	kv->RecursiveSaveToFile(buffer, 0);
```
asherkin: Can this fail?

core/smn_keyvalues.cpp
```diff
@@ -1113,6 +1113,65 @@ static cell_t KeyValues_Import(IPluginContext *pContext, const cell_t *params)
...
+	
+	char* outStr;
+	pContext->LocalToString(params[2], &outStr);
+	size_t maxlen = static_cast<size_t>(params[3]);
+	
+	buffer.GetString(outStr, maxlen);
+	return buffer.TellGet();
```
asherkin: I think both calls to TellGet should possibly be TellPut, which would remove the need for the extra GetString, which could also open up caching the buffer for the common use case, but test it.

Headline22: TellPut is for writing to the CUtilBuffer, no? 

inside of UtlBuffer.h
```cpp
	// Where am I writing (put)/reading (get)?
	int TellPut( ) const;
	int TellGet( ) const;
```

It doesn't seem very obvious from their comment, but I'll test your recommendation

core/smn_keyvalues.cpp
Headline22: The way it's currently done is KeyValues.ExportToFile calls KeyValues::SaveToFile which just does some checks around opening the file & then calls RecursiveSaveToFile anyway. Calling it how we are here is good as-is

core/smn_keyvalues.cpp
asherkin: The useful comment directing towards it is somewhere in the `CUtlBuffer` impl iirc, but the gist is that TellPut is the number of bytes written, which is actually what we're after.

PR #709, opened 2017-10-20
 Lowercase windows/mac paths to avoid plugin duplicate loadings (Bug #6491)

Headline22:
Prior to this pull request, plugins could be loaded in twice & have two separate instances of the same plugin running simultaneously. This adds a check for each plugin loaded in, and ensures that the code hash of the new plugin doesn't equal a code hash of a plugin that's already loaded.

Here's the original bug report: https://bugs.alliedmods.net/show_bug.cgi?id=6491

### DIFF
--- a/core/logic/PluginSys.cpp
+++ b/core/logic/PluginSys.cpp
@@ -46,6 +46,7 @@
 #include "frame_tasks.h"
 #include <amtl/am-string.h>
 #include <amtl/am-linkedlist.h>
+#include <amtl/am-uniqueptr.h>
 #include <bridge/include/IVEngineServerBridge.h>
 #include <bridge/include/CoreProvider.h>
 
@@ -953,9 +954,9 @@ LoadRes CPluginManager::LoadPlugin(CPlugin **aResult, const char *path, bool deb
 
 /* For windows, we convert the path to lower-case in order to avoid duplicate plugin loading */
 #if defined PLATFORM_WINDOWS || defined PLATFORM_APPLE
-	char *finalPath = strdup_tolower(path);
+	ke::UniquePtr<char> finalPath = ke::UniquePtr<char>(strdup_tolower(path));
 #else 
-	char *finalPath = strdup(path);
+	ke::UniquePtr<char> finalPath = ke::UniquePtr<char>(strdup(path));
 #endif
 
 
@@ -963,7 +964,7 @@ LoadRes CPluginManager::LoadPlugin(CPlugin **aResult, const char *path, bool deb
 	 * Does this plugin already exist?
 	 */
 	CPlugin *pPlugin;
-	if (m_LoadLookup.retrieve(finalPath, &pPlugin))
+	if (m_LoadLookup.retrieve(finalPath.get(), &pPlugin))
 	{
 		/* Check to see if we should try reloading it */
 		if (pPlugin->GetStatus() == Plugin_BadLoad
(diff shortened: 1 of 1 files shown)


core/logic/PluginSys.cpp
```diff
@@ -932,6 +932,31 @@ void CPluginManager::LoadPluginsFromDir(const char *basedir, const char *localpa
 	libsys->CloseDirectory(dir);
 }
 
+inline bool HashesAreEqual(const unsigned char* a, const unsigned char* b)
+{
+	return strlen((const char*)a) == strlen((const char*)b) && memcmp(a, b, strlen((const char*)a)) == 0;
```
asherkin: This should just be `return memcmp(a, b, 16) == 0;`, CodeHashes are 16 byte uchar arrays, not strings, they are not NUL terminated and the data can contain 0s, calling `strlen` on binary data is not valid.

core/logic/PluginSys.cpp
```diff
@@ -932,6 +932,31 @@ void CPluginManager::LoadPluginsFromDir(const char *basedir, const char *localpa
 	libsys->CloseDirectory(dir);
 }
 
+inline bool HashesAreEqual(const unsigned char* a, const unsigned char* b)
+{
+	return strlen((const char*)a) == strlen((const char*)b) && memcmp(a, b, strlen((const char*)a)) == 0;
+}
+
+CPlugin* PluginExistsInHashSet(CPlugin *plugin, NameHashSet<CPlugin *> &hashset)
```
asherkin: We should probably have a proper secondary hashset keyed on the CodeHash rather than iterating all loaded plugins.

core/logic/PluginSys.cpp
```diff
@@ -960,6 +985,16 @@ LoadRes CPluginManager::LoadPlugin(CPlugin **aResult, const char *path, bool deb
 
 	CPlugin *plugin = CompileAndPrep(path);
 
+	// Now that we've got CPlugin: check hashes to see if plugin is already loaded. Checking names is not enough
```
asherkin: I don't really like that we have to load the plugin completely to get to this point, and the CodeHash is a pretty weird construct to use for this, but I can't really think of any saner options that don't make plugin loading more expensive globally rather than just in an edge case.

Fyren:
I would assume there's some way to see if two file handles represent the actual same file.  Then we'd only have to fopen (and possibly make the file handle available in CPlugin or IPluginRuntime or somewhere, if it isn't already).

Headline22:
Instead of using code hash and a loop through NameHashSet on every plugin load, the latest commit just converts the path of the key to lower case if the platform is windows. This is better for performance, and linux is completely unaffected.

TheDS:
I think it makes sense to do this on macOS as well. So adding a PLATFORM_APPLE check should suffice.

While technically macOS has support for case sensitive file systems, the default is case insensitive like Windows. Steam doesn’t run on the case sensitive versions anyways: https://support.steampowered.com/kb_article.php?ref=8601-RYPX-5789

core/logic/PluginSys.cpp
```diff
@@ -932,16 +932,38 @@ void CPluginManager::LoadPluginsFromDir(const char *basedir, const char *localpa
...
 {
 	if (m_LoadingLocked)
 		return LoadRes_NeverLoad;
 
+/* For windows, we convert the path to lower-case in order to avoid duplicate plugin loading */
+#if defined PLATFORM_WINDOWS || defined PLATFORM_APPLE
+	char *finalPath = strdup_tolower(path);
+#else 
+	char *finalPath = strdup(path);
+#endif
```
asherkin: Would prefer a `ke::UniquePtr` here to avoid the explicit `free()` calls later, also the comment is outdated.

PR #710, opened 2017-10-21
TF2 Makebleed function param fix

Benoist3012:
Fix the error:
`L 10/21/2017 - 19:05:10: [SM] Exception reported: Failed to locate function
L 10/21/2017 - 19:05:10: [SM] Blaming: sf2.smx
L 10/21/2017 - 19:05:10: [SM] Call stack trace:
L 10/21/2017 - 19:05:10: [SM]   [0] TF2_MakeBleed
`
I don't know if the recent two fix patches changed the signature or you've missed it, but here a fixed linux sig for MakeBleed

### DIFF
--- a/extensions/tf2/natives.cpp
+++ b/extensions/tf2/natives.cpp
@@ -45,7 +45,7 @@ cell_t TF2_MakeBleed(IPluginContext *pContext, const cell_t *params)
 	if(!pWrapper)
 	{
 		REGISTER_NATIVE_ADDR("MakeBleed",
-			PassInfo pass[5]; \
+			PassInfo pass[6]; \
 			pass[0].flags = PASSFLAG_BYVAL; \
 			pass[0].size = sizeof(CBaseEntity *); \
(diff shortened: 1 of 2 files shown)


Drifter321:
It has a new param, you cant simply change the symbol without modifying what uses it.

Benoist3012:
I didn't know about that, okay then, I'll edit the tf2 extension add the changes.

Benoist3012:
Rigth so new param seems to be an in using some kind of bit flags, it's 34 everywhere but at one place where it's 3 and is used on dragon fury the new flamethrower added by valve, no idea what are the effects.

psychonic:
Sounds like it might be damage bits.

asherkin:
Does it make sense to expose all these new params in the native now?

psychonic:
I considered doing that with the newest one, and then saw the others that we don't expose.

I think it's worth doing, just outside of the scope of the fix.

PR #712, opened 2017-10-23
Update TF2_IsPlayerInCondition

Pelipoika:
Update TF2_IsPlayerInCondition to support the new conditions


psychonic:
Good catch. I didn't even realize we were past the threshold for another var.

PR #716, opened 2017-10-31
Create FrameIterator type

Headline22:
I decided to redact these changes from the LogStackTrace pull request since these are unrelated.

If we decide to add a stack frame iterator for sourcemod plugins, this is what that implementation would look like. Comments & criticisms are welcome.

### DIFF
--- a/core/logic/FrameIterator.cpp
+++ b/core/logic/FrameIterator.cpp
@@ -1,0 +1,90 @@
+/**
+ * vim: set ts=4 sw=4 tw=99 noet :
+ * =============================================================================
+ * SourceMod
+ * Copyright (C) 2004-2017 AlliedModders LLC.  All rights reserved.
+ * =============================================================================
+ *
+ * This program is free software; you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License, version 3.0, as published by the
+ * Free Software Foundation.
+ * 
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * As a special exception, AlliedModders LLC gives you permission to link the
+ * code of this program (as well as its derivative works) to "Half-Life 2," the
+ * "Source Engine," the "SourcePawn JIT," and any Game MODs that run on software
+ * by the Valve Corporation.  You must obey the GNU General Public License in
+ * all respects for all other code used.  Additionally, AlliedModders LLC grants
+ * this exception to all derivative works.  AlliedModders LLC defines further
+ * exceptions, found in LICENSE.txt (as of this writing, version JULY-31-2007),
+ * or <http://www.sourcemod.net/license.php>.
+ *
+ * Version: $Id$
+ */
+ 
+#include "FrameIterator.h"
+
+SafeFrameIterator::SafeFrameIterator(IFrameIterator *it)
+{	
+	while (!it->Done())
+	{
+		FrameInfo info = FrameInfo(it);
+		frames.append(info);
+		it->Next(); 
+	}
+	
+	it->Reset();
+	current = 0;
+}
+
+bool SafeFrameIterator::Done() const
+{
+	return current >= (frames.length() - 1);
+}
+
+void SafeFrameIterator::Next()
+{
+	current++;
+}
+
+void SafeFrameIterator::Reset()
+{
+	current = 0;
+}
+
+int SafeFrameIterator::LineNumber() const
+{
+	if (this->Done())
+	{
+		return -1;
+	}
+
+	return (int)frames[current].LineNumber;
+}
+
+const char *SafeFrameIterator::FunctionName() const
+{
+	if (this->Done())
+	{
+		return NULL;
+	}
+
+	return frames[current].FunctionName.chars();
+}
+
+const char *SafeFrameIterator::FilePath() const
+{
+	if (this->Done())
+	{
+		return NULL;
+	}
+
+	return frames[current].FilePath.chars();
+}
--- a/core/logic/FrameIterator.cpp
+++ b/core/logic/FrameIterator.cpp
@@ -51,7 +51,7 @@ bool SafeFrameIterator::Done() const
 
 bool SafeFrameIterator::Next()
 {
-	if(!this->Done())
+	if (!this->Done())
 	{
 		current++;
 		return true;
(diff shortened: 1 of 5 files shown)


plugins/include/sourcemod.inc
```diff
@@ -645,6 +645,40 @@ native int LoadFromAddress(Address addr, NumberType size);
...
 native void StoreToAddress(Address addr, int data, NumberType size); 
 
+methodmap FrameIterator < Handle {
+	// Creates a stack frame iterator to build your own stack traces.
+	// @return New handle to a FrameIterator.
+	public native FrameIterator();
+	
+	// Determines if there are any more stack frames to iterate.
+	// @return True if there are no more stack frames left.
+	public native bool Done();
```
KyleSanderson: I don't think this is natural, Next should return true or false. Done can be done away with completely (I think), or you can keep it to match SQL_MoreRows.

https://github.com/alliedmodders/sourcemod/blob/750767289558236da6c3554c734745aef35f58dd/core/logic/smn_database.cpp#L1062

core/logic/smn_core.cpp
```diff
@@ -780,6 +792,163 @@ static cell_t IsNullString(IPluginContext *pContext, const cell_t *params)
...
+	if (!filePath)
+	{
+		return pContext->ThrowNativeError("Iterator out of bounds!");
+	}
+
+	char* buffer;
+	pContext->LocalToString(params[2], &buffer);
+	size_t size = static_cast<size_t>(params[3]);
+	
+	snprintf(buffer, size, "%s", filePath);
```
KyleSanderson: ke::SafeStrcpy

core/logic/FrameIterator.h
```diff
@@ -0,0 +1,78 @@
...
+class SafeFrameIterator
+{
+public:
+
+	/**
+	 * Struct which holds all of the cached values for each individual frame.
+	 */
+	struct FrameInfo
+	{
+		unsigned LineNumber;
```
KyleSanderson: Typically you want larger constructs to be at the top for packing. I'm not sure if these are going to be promoted or not but I would half expect them to be.

core/logic/smn_core.cpp
```diff
@@ -780,6 +792,163 @@ static cell_t IsNullString(IPluginContext *pContext, const cell_t *params)
...
+
+	if ((err=handlesys->ReadHandle(hndl, g_FrameIter, &sec, (void **)&iterator)) != HandleError_None)
+	{
+		return pContext->ThrowNativeError("Could not read Handle %x (error %d)", hndl, err);
+	}
+	
+	const char* filePath = iterator->FilePath();
+	if (!filePath)
+	{
+		return pContext->ThrowNativeError("Iterator out of bounds!");
```
KyleSanderson: It's good you're checking these, but I would try and give a little more information if possible. I'm not sure the '!' is present anywhere else in SM, so matching style would be an improvement.

core/logic/FrameIterator.cpp
```diff
@@ -0,0 +1,90 @@
...
+		it->Next(); 
+	}
+	
+	it->Reset();
+	current = 0;
+}
+
+bool SafeFrameIterator::Done() const
+{
+	return current >= (frames.length() - 1);
```
KyleSanderson: current < frames.length()

core/logic/smn_core.cpp
```diff
@@ -91,13 +95,21 @@ class CoreNativeHelpers :
 	}
 	void OnHandleDestroy(HandleType_t type, void *object)
 	{
-		IPluginIterator *iter = (IPluginIterator *)object;
-		iter->Release();
+		if (type == g_FrameIter)
```
KyleSanderson: maybe use switch here, there's a mixed style throughout SM.

core/logic/smn_core.cpp
```diff
@@ -780,6 +792,163 @@ static cell_t IsNullString(IPluginContext *pContext, const cell_t *params)
...
+	sec.pIdentity = g_pCoreIdent;
+	sec.pOwner = pContext->GetIdentity();
+
+	if ((err=handlesys->ReadHandle(hndl, g_FrameIter, &sec, (void **)&iterator)) != HandleError_None)
+	{
+		return pContext->ThrowNativeError("Could not read Handle %x (error %d)", hndl, err);
+	}
+	
+	const char* functionName = iterator->FunctionName();
+
```
KyleSanderson: crunch here, you're doing it below.

core/logic/FrameIterator.cpp
```diff
@@ -0,0 +1,90 @@
...
+}
+
+void SafeFrameIterator::Next()
+{
+	current++;
```
KyleSanderson: This should not increment endlessly, after you hit Done this should stop.

core/logic/smn_core.cpp
Headline22: Compiler demands that g_FrameIter and the likes must be constant variables. Errors specifically are
```
core\logic\smn_core.cpp(100): error C2131: expression did not evaluate to a constant
core\logic\smn_core.cpp(100): note: failure was caused by non-constant arguments or reference to a non-constant symbol
core\logic\smn_core.cpp(100): note: see usage of 'g_FrameIter'
core\logic\smn_core.cpp(105): error C2131: expression did not evaluate to a constant
core\logic\smn_core.cpp(105): note: failure was caused by non-constant arguments or reference to a non-constant symbol
core\logic\smn_core.cpp(105): note: see usage of 'g_PlIter'
core\logic\smn_core.cpp(100): error C2051: case expression not constant
core\logic\smn_core.cpp(105): error C2051: case expression not constant
```

KyleSanderson: Ah, yes you're (well, msvc is) absolutely right.

core/logic/FrameIterator.cpp
```diff
@@ -0,0 +1,96 @@
...
+}
+
+bool SafeFrameIterator::Done() const
+{
+	return current == frames.length();
+}
+
+bool SafeFrameIterator::Next()
+{
+	if(!this->Done())
```
KyleSanderson: style

core/logic/FrameIterator.cpp
```diff
@@ -0,0 +1,96 @@
...
+ * or <http://www.sourcemod.net/license.php>.
+ *
+ * Version: $Id$
+ */
+ 
+#include "FrameIterator.h"
+
+SafeFrameIterator::SafeFrameIterator(IFrameIterator *it)
+{	
+	while (!it->Done())
```
KyleSanderson: style, but do {} while(it->Next());

KyleSanderson: actually if you do this you'd have to check it->Done() first; so it's all good.

asherkin: There could be 0 frames to iterate.

KyleSanderson: That's not possible for the Pawn API.

PR #719, opened 2017-11-03
Add note for CS:GO regarding CreateItemByName.

Drifter321:
This adds a simple note to CreateEntityByName to make it obvious that it calls CreateItemByName internally on CS:GO first. Just want to make sure the wording is clear.


asherkin:
I don’t really see the point of this - it is literally just for lemons that read the patch notes but don’t understand what they’re reading (those that do read the code.)

Drifter321:
Fair point. I guess in the future people will completely forget it was added anyway. Plus it wont even matter since the API search wont return it since its a comment (I believe). Anyway, closed!

KyleSanderson:
This is an implementation detail that's good to know... This is worthwhile to have in the tree.

PR #721, opened 2017-11-09
Expose ``KeyValueStack`` to extensions to interact with plugin kv handles

fakuivan:
Exposes methods via the sm interface for creating and reading kv handles and adds a new interface for ``KeyValueStack``. This should allow extensions to provide kv handles to plugins and read them when necessary.


psychonic:
Do we really need a way to expose the whole KeyValueStack, rather than just a way to create the handle from the KeyValues*? @asherkin

fakuivan:
Any updates on this?

PR #722, opened 2017-11-10
Add Timer Datatype

Headline:
Timers don't have very many functions attributed to the Handle, so this is quite a small methodmap. This is one step towards methodmap completion, though.

This pull request allows for sourcemod plugin creators to make timers like so

```cpp
Timer timer = new Timer(10.0, Timer_Callback);
...
timer.Kill();
```

### DIFF
--- a/plugins/include/timers.inc
+++ b/plugins/include/timers.inc
@@ -45,7 +45,7 @@
 /**
  * Any of the following prototypes will work for a timed function.
  */
-typeset Timer
+typeset TimerCallback
 {
 	/**
  	 * Called when the timer interval has elapsed.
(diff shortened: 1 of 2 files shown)


powerlord:
The only catch I can see is that it will break any plugins that pass timer callbacks around.

Headline22:
This is true, but I do not believe any plugins do this.

plugins/include/timers.inc
```diff
@@ -45,7 +45,7 @@
 /**
  * Any of the following prototypes will work for a timed function.
  */
-typeset Timer
```
peace-maker: You should change the tag of the `Handle timer` in the callback signatures to `Timer timer` too.

Headline: Absolutely right. Thanks pm

plugins/include/timers.inc
```diff
@@ -201,9 +233,9 @@ native bool IsServerProcessing();
  * @param flags				Timer flags.
  * @return					Handle to the timer object.  You do not need to call CloseHandle().
  */
-stock Handle CreateDataTimer(float interval, Timer func, Handle &datapack, int flags=0)
+stock Timer CreateDataTimer(float interval, TimerCallback func, Handle &datapack, int flags=0)
```
ZeronTheXeon: Handle &datapack -> DataPack &datapack

Headline: 👍

Deathreus:
While you're refactoring the timer include:
```/**
 * Returns whether or not the server is processing frames or not.  
 *
 * The server does not process frames until at least one client joins the game.  
 * Once the first player has in, even if that player, leaves, the server's 
 * timers and entities will work.
 *
 * @return					True if the server is ticking, false otherwise.
 */
native bool IsServerProcessing();
```
That brief bugs my inner grammar nazi

JoyKoy:
Is this gonna be merged?

KyleSanderson:
We can't land this, it's going to break close to a decades worth of plugins.

asherkin:
That is no reason to close it - it can just remain open until the next worthwhile widespread breaking change.

Headline:
@asherkin i'll remember this so when that time comes we can bring this back

PR #723, opened 2017-11-19
Fix "Exception reported: Client 12 is not in game"

komashchenko:
L 11/19/2017 - 16:37:02: [SM] Exception reported: Client 12 is not in game
L 11/19/2017 - 16:37:02: [SM] Blaming: adminhelp.smx
L 11/19/2017 - 16:37:02: [SM] Call stack trace:
L 11/19/2017 - 16:37:02: [SM]   [0] PrintToConsole
L 11/19/2017 - 16:37:02: [SM]   [1] Line 105, /home/builds/sourcemod/linux-1.9/build/plugins/adminhelp.sp::HelpCmd

### DIFF
--- a/plugins/adminhelp.sp
+++ b/plugins/adminhelp.sp
@@ -58,7 +58,7 @@ public void OnPluginStart()
 
 public Action HelpCmd(int client, int args)
 {
-	if(client != 0 && !IsClientInGame(client))
+	if (client && !IsClientInGame(client))
 	{
 		return Plugin_Handled;
 	}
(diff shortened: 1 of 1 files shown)


plugins/adminhelp.sp
```diff
@@ -58,6 +58,8 @@ public void OnPluginStart()
 
 public Action HelpCmd(int client, int args)
 {
+	if(!IsClientInGame(client)) return Plugin_Handled;
```
WildCard65: Nit pick: Please make style match the rest of the file.

KyleSanderson: This is so close to being right, but breaks console support and moves the error (reliably) there as rcon/console is 0.

PR #727, opened 2017-11-27
Revert "Make CS:GO specific natives exist only on CS:GO"

Drifter321:
This reverts commit 687db96b38112ba455d21fab5515e94600236942.

Creating a pr for this just in case I am missing something. When a plugin is made to work for both CS:S and CS:GO having natives not exist on both cause issues since the native has to either be marked as optional in the plugin or REQUIRE_EXTENSIONS undefined. The change possibly broke other plugins besides mine as well.


psychonic:
🚀

PR #728, opened 2017-11-30
Only show extensions that failed to load if the extension is required.

Drifter321:
I might have missed something since first time really touching any of this. Also not sure if we should instead print a different message for optional instead of omitting it completely.

### DIFF
--- a/core/logic/ExtensionSys.cpp
+++ b/core/logic/ExtensionSys.cpp
@@ -43,8 +43,9 @@
 CExtensionManager g_Extensions;
 IdentityType_t g_ExtType;
 
-void CExtension::Initialize(const char *filename, const char *path)
+void CExtension::Initialize(const char *filename, const char *path, bool bRequired)
 {
+	m_bRequired = bRequired;
 	m_pAPI = NULL;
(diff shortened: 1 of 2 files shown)


Drifter321:
After discussing with psychonic. I made some changes to display them separately instead.

Example output
`sm exts list
[SM] Displaying 9 extensions:
[01] CS:GO GNI Econ (1.0.2): Creates EconItemView for items.
[02] BinTools (1.8.0-manual): Low-level C/C++ Calling API
[03] CS Tools (1.8.0-manual): CS extended functionality
[04] SDK Tools (1.8.0-manual): Source SDK Tools
[05] Top Menus (1.8.0-manual): Creates sorted nested menus
[06] Client Preferences (1.8.0-manual): Saves client preference settings
[07] SQLite (1.8.0-manual): SQLite Driver
[08] CS:S DM (2.1.6-dev): Deathmatch for CS:S
[09] SDK Hooks (1.8.0-manual): Source SDK Hooks

[SM] Displaying 1 optional extension not found:
[01] "SteamWorks.ext.dll"`

PR #729, opened 2017-12-01
Remove goto from ExtensionSys.cpp

Drifter321:
Since I touched the function, i decided to also clean up the goto stuff.

Im not sure what

g_pSM->BuildPath(Path_SM, path, PLATFORM_MAX_PATH, "extensions/auto.%s/%s." PLATFORM_LIB_EXT, filename, bridge->gamesuffix);

is from (i have a vague remembrance of something with auto.) I assume it was how old engine specific extensions were loaded idk if we should still keep it or not.. but i left it.


core/logic/ExtensionSys.cpp
```diff
@@ -72,74 +72,69 @@ CLocalExtension::CLocalExtension(const char *filename, bool bRequired)
...
-	g_pSM->BuildPath(Path_SM, 
-		path, 
-		PLATFORM_MAX_PATH,
-		"extensions/auto.%s/%s." PLATFORM_LIB_EXT,
-		filename,
-		bridge->gamesuffix);
+			/* COMPAT HACK: One-halfth, if ep2v, see if there is an engine specific build in the new place with old naming */
+			if (strcmp(bridge->gamesuffix, "2.tf2") == 0
+				|| strcmp(bridge->gamesuffix, "2.dods") == 0
+				|| strcmp(bridge->gamesuffix, "2.hl2dm") == 0
```
KyleSanderson: Not your fault: I think I missed css here whenever this generalization stuff landed?

core/logic/ExtensionSys.cpp
```diff
@@ -72,74 +72,69 @@ CLocalExtension::CLocalExtension(const char *filename, bool bRequired)
...
-	/* Try the "normal" version */
-	if (!libsys->IsPathFile(path))
-	{
-		normal:
-		g_pSM->BuildPath(Path_SM,
-			path,
-			PLATFORM_MAX_PATH,
-			"extensions/%s." PLATFORM_LIB_EXT,
-			filename);
+			//Try further
```
KyleSanderson: Inconsistent comment style.

core/logic/ExtensionSys.cpp
```diff
@@ -72,74 +72,69 @@ CLocalExtension::CLocalExtension(const char *filename, bool bRequired)
...
-	/* COMPAT HACK: One-halfth, if ep2v, see if there is an engine specific build in the new place with old naming */
-	if (strcmp(bridge->gamesuffix, "2.tf2") == 0
-		|| strcmp(bridge->gamesuffix, "2.dods") == 0
-		|| strcmp(bridge->gamesuffix, "2.hl2dm") == 0
-		)
 	{
 		g_pSM->BuildPath(Path_SM,
 			path,
 			PLATFORM_MAX_PATH,
-			"extensions/%s.2.ep2v." PLATFORM_LIB_EXT,
```
KyleSanderson: Behaviour change, assuming this is a fix?

core/logic/ExtensionSys.cpp
Drifter321: CS:S never was compat hacked. This was added after since CS:S for a long while was its own branch.

core/logic/ExtensionSys.cpp
Drifter321: Not a behavior change, this is strictly for bintools. which previously used goto to the same buildpath.

asherkin:
@Drifter321 @KyleSanderson This change appears to have caused a crash loading certain extensions (PTAH, might be related to number of exts or loading an ext after another fails to load), https://crash.limetech.org/xmglmbqdoqnz, it has been bisected to this commit.

Headline:
@asherkin's comment is relevant to #789

asherkin:
This commit wasn’t in master at that point, might indeed be PTAH’s issue.

Headline:
Oh, yeah you're right. These crashes looked identical, my bad.

PR #732, opened 2017-12-01
Make sure map is running before allowing stuff to get Precached.

SM91337:
Seemed simple enough to add.

### DIFF
--- a/core/smn_halflife.cpp
+++ b/core/smn_halflife.cpp
@@ -214,7 +214,7 @@ static cell_t PrecacheModel(IPluginContext *pContext, const cell_t *params)
 {
 	if (!g_pSM->IsMapRunning())
 	{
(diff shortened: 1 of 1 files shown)


core/smn_halflife.cpp
```diff
@@ -211,6 +211,11 @@ static cell_t GetCurrentMap(IPluginContext *pContext, const cell_t *params)
 
 static cell_t PrecacheModel(IPluginContext *pContext, const cell_t *params)
 {
+	if (!g_pSM->IsMapRunning())
+	{
+		return pContext->ThrowNativeError("Cannot Precache model when no map is running");
```
Headline: nit: Can you make P in "Precache" lowercase in all of your warnings

KyleSanderson:
If the intention is to have this prevent the (alleged) late load crashing, the map is already running in this case. What are you trying to solve?

PR #733, opened 2017-12-03
Block reserved device names on windows

Headline:
This should fix some issues windows users may experience should a reserved windows device name sneaks up


KyleSanderson:
hmmmmmmmmmmmmm... I think if there's a space given to this, all security is bypassed. Same with special characters... Why not just use QueryDosDevice and if it's mapped to /?? discard, then you know it's some legacy trash?

PR #735, opened 2017-12-14
Add forgotten `const` to parameters

Headline:
If anything here is wrong let me know

Thanks to Kruzya for bringing this up in the sm discord

### DIFF
--- a/plugins/include/usermessages.inc
+++ b/plugins/include/usermessages.inc
@@ -109,7 +109,7 @@ native UserMsg GetUserMessageId(const char[] msg);
  * @param maxlength		Maximum length of string buffer.
  * @return				True if message index is valid, false otherwise.
  */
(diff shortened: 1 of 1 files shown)


plugins/include/usermessages.inc
```diff
@@ -109,7 +109,7 @@ native UserMsg GetUserMessageId(const char[] msg);
  * @param maxlength		Maximum length of string buffer.
  * @return				True if message index is valid, false otherwise.
  */
-native bool GetUserMessageName(UserMsg msg_id, char[] msg, int maxlength);
+native bool GetUserMessageName(UserMsg msg_id, const char[] msg, int maxlength);
```
asherkin: This native is now broken, `msg` is an outparam.

Headline: ty

PR #738, opened 2017-12-19
Fix typo and update CSGOBadList

InvexByte:
`m_iCompetitiveRankType` is a new netprop which appeared a few updates ago.
It should be blocked alongside the other netprops.


shavitush:
Is it used for Wingman ranks?

InvexByte:
@shavitush Yes.

PR #740, opened 2017-12-23
Individualize NameHashSet Hashing & Revisit #709

### LINKED ISSUE #831: Path file case detection
# Description
SM 1.9.0.6237
Issue with handling of path files - not sure scope of bug, however it was noticed during sm_rcon sm plugins unload PluginFolder/plugin.smx

Folder exists in /plugins dir with capital letters. Inside is plugin. (eq plugins/PluginFolder/plugin.smx)
run sm_rcon sm plugins unload PluginFolder/plugin
error that plugin is not loaded
run sm_rcon sm plugins unload pluginfolder/plugin
success

# Example:
```
sm_rcon sm plugins unload SourceIRC/sourceirc-relayall
[SM] Plugin SourceIRC\sourceirc-relayall.smx is not loaded.
sm_rcon sm plugins unload sourceirc/sourceirc-relayall
[SM] Plugin SourceIRC -> Relay All will be unloaded on the next frame.
```

Headline:
On #709, we forgot to convert the `NameHashSet<CPlugin *>`'s lookups to be lowercase. This will fix some casing issues people were experiencing trying to unload their (case-edited) plugins.

EDIT (12/27/17): 
Upon asherkin’s message below, I reverted the first commit of this PR & the #709 commit. The new solution to this entire problem, including the one #709 tried to solve, is to change the way `NameHashSet` hashes the key by forcing policies to also implement a `hash` method which will be used instead of the blanket implementation of `hash` that existed before.

Edit (6/17/2018):
Fixes #831

### DIFF
--- a/core/ConVarManager.h
+++ b/core/ConVarManager.h
@@ -43,6 +43,7 @@
 #include <compat_wrappers.h>
 #include "concmd_cleaner.h"
 #include "PlayerManager.h"
+#include <sm_stringhashmap.h>
 
 using namespace SourceHook;
 
@@ -66,6 +67,10 @@ struct ConVarInfo
 	static inline bool matches(const char *name, const ConVarInfo *info)
 	{
 		return strcmp(name, info->pVar->GetName()) == 0;
+	}
+	static inline uint32_t hash(const detail::CharsAndLength &key)
+	{
+		return key.hash();
 	}
 };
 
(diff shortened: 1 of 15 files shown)


asherkin:
I wonder if we should just change the hash strategy for the `NameHashSet`, and revisit the original change.

Headline:
That’s probably a cleaner way of accomplishing this. I’ll submit a patch in the near future

Headline:
@asherkin can you add this to your list for a review?

duck says he's cool to take this into 1.10 and if all is well after a while it's cool to cherrypick into 1.9

PR #741, opened 2017-12-26
Default last_command_client_ to 0 instead of -1.

KyleSanderson:
Over two years ago (!) we changed CommandClient() to return -1 as opposed to 0 if it was uninitialized. This is reproducible by simply going:

```
#pragma semicolon 1
public OnPluginStart()
{
	AddCommandListener(Listener);
}

public Action:Listener(client, const String:sCommand[], argc)
{
	PrintToServer("%N: %s", client, sCommand);
	return Plugin_Continue;
}
```
This may have uncovered a bug, that we never set CommandClient to 0 when commands are invoked by the server. This PR just reverts the behaviour to prior to when this landed for 1.8 as I'm only updating my (ancient) gear now.

Fix from: https://github.com/alliedmodders/sourcemod/commit/fe16e8e47cd854c482ba04f292eaa1e2d4e2e34b
Docs: https://github.com/alliedmodders/hl2sdk/blob/sdk2013/tier1/commandbuffer.cpp#L568

### DIFF
--- a/core/GameHooks.cpp
+++ b/core/GameHooks.cpp
@@ -51,7 +51,7 @@ SH_DECL_HOOK1_void(IServerGameClients, SetCommandClient, SH_NOATTRIB, false, int
 
 GameHooks::GameHooks()
 	: client_cvar_query_mode_(ClientCvarQueryMode::Unavailable),
-	  last_command_client_(-1)
+	  last_command_client_(0)
 {
 }
 


asherkin:
> This may have uncovered a bug, that we never set CommandClient to 0 when commands are invoked by the server. 

Why isn’t this being fixed instead?

KyleSanderson:
> Why isn’t this being fixed instead?
Bit of a rabbits hole. Technically (well, we'd like it to) the engine should be calling this function before executing commands, but this is the VSP interface that was given years back... Fixing this back to SM1.0-1.7 functionality seemed like the more followed path, with the caveat noted above that there's still atleast one case where this is potentially incorrect.

Can we land the revert? Just trying to modernize my workloads at the moment, this being an unexpected behaviour change.

psychonic:
If I'm understanding this correctly (after only looking through it about half as deeply as I probably should), the conflict is that internally the game uses -1 for commands from server and SM uses 0. And the current issue being that SM is using -1 instead of 0 for (only) the case where SetCommandClient hasn't been called yet.

Assuming all of that is correct, I'm for taking the PR as-is, reverting that change, rather than blowing up a bunch of plugins not expecting -1 in more cases, or adjusting all parts of SM to translate it before it gets to them.

asherkin:
That isn't my understanding, the game passes engine client index (or `-1` for server) then we +1 on our end to get our client ID (entity index, 0 for server).

In this case the client ID on our end is not getting set due to the function apparently not being called, this was silently failing before and attributing the command to the server on early calls, **but would be passing through a stale client ID if occurring after a command from a client.** That is a serious bug that we should not let regress. I think we should also explicitly set the ID back to -1 after processing a command (if we're always expecting a 1:1 relationship between SetCommandClient calls and command dispatches) to try and catch any other missing calls later.

This needs debugging work to figure out why `SetCommandClient` isn't being called in all cases, I am strongly against reverting the original change as it appears to have exposed an issue that needs fixing -
 not caused the issue.

KyleSanderson:
> the conflict is that internally the game uses -1 for commands 
Kind of, the game stores the player slot internally, which is -1 for the server. The return from gameinterface is +1'd in the engine: https://github.com/alliedmodders/hl2sdk/blob/3957adff10fe20d38a62fa8c018340bf2618742b/game/server/gameinterface.cpp#L277 . In SM we do this on the storage, not the return which is why I'm assuming this even happened to begin with (storing -1 though the SM function would result in 0 being returned).

soft-docs: https://github.com/alliedmodders/hl2sdk/blob/3957adff10fe20d38a62fa8c018340bf2618742b/tier1/commandbuffer.cpp#L565

KyleSanderson:
@asherkin so, this is still broken in trunk (my consumers are still on a pre-SM version of this change). I guess it's been broken for eons now, do we want to land this one? paper over in SP?

Issue #749, opened 2018-01-16
Feature request: Traceray natives for triggers


jason-e:
There is currently no way to do point/ray/hull testing against trigger entities. Associated SourceMod traceray natives make use of `IEngineTrace->TraceRay()` and `IEngineTrace->GetPointContents()` which, at least according to the implementation found [here](https://github.com/VSES/SourceEngine2007/blob/master/se2007/engine/enginetrace.cpp), explicitly ignore trigger entities through the use of the partition mask `PARTITION_ENGINE_SOLID_EDICTS` which includes ["every edict_t that isn't SOLID_TRIGGER or SOLID_NOT (and static props)"](https://github.com/ValveSoftware/source-sdk-2013/blob/master/mp/src/public/ispatialpartition.h#L30).

The enginetrace function `EnumerateEntities()` (with `triggers == true`) could be used to at least find triggers along a given ray/rayhull. True "collision" testing would also require the use of enginetrace's `ClipRayToEntity()`. Whether these two functions should be exposed directly or used in some higher-level function(s) is not important as long as point/ray/hull testing is made possible for triggers.

It is technically already possible to "find" triggers by hooking trigger entity touch events and moving some entity through a given area until they are fired, however that is not a great solution as it is imprecise and cannot be done instantaneously.

PR #754, opened 2018-01-21
Add new Traceray natives

SlidyBat:
Related to #749 
Adds TR_EnumerateEntities and TR_ClipRayToEntity natives, which will be very useful for having more control over tracerays. More specifically, this allows plugin developers to create rays that can hit trigger entities.

Here's an example of how using these natives would look like in a plugin:
```cpp
#include <sourcemod>
#include <sdktools>

public void OnPluginStart()
{
	RegConsoleCmd( "sm_trigtest", Command_TriggerTest );
}

public Action Command_TriggerTest( int client, int args )
{
	float startpos[3];
	GetClientEyePosition( client, startpos );
	float angles[3];
	GetClientEyeAngles( client, angles );
	
	// native void TR_EnumerateEntities(const float pos[3], const float vec[3], bool triggers, RayType rtype, TraceEntityEnumerator enumerator, any data=0);
	TR_EnumerateEntities( startpos, angles, true, RayType_Infinite, HitTrigger );
}

// return true to continue enumerating, or false to stop
public bool HitTrigger( int entity )
{
	char classname[64];
	GetEntityClassname( entity, classname, sizeof( classname ) );	
	if( StrContains( classname, "trigger" ) > -1 )
	{
		// native void TR_ClipRayToEntity(int flags, int entity, Handle tr=INVALID_HANDLE);
		TR_ClipRayToEntity( MASK_ALL, entity );
		
		float pos[3];
		TR_GetEndPosition( pos );
		
		PrintToChatAll( "Hit a trigger! Collided at (%.2f, %.2f, %.2f)", pos[0], pos[1], pos[2] );
		return false;
	}
	
	return true;
}
```

Also I think the diffs looks weird because VS told me that line endings were inconsistent and it "fixed" them.
EDIT: Re-did it without the line ending bs to fix the cancer diff.


Headline:
Can you go back and undo the white space changes so the diffs only contain what you’ve added? You’ve really nuked line endings here

asherkin:
Is `TR_ClipRayToEntity` ever useful outside of the `TR_EnumerateEntities` callback?

Does something like the following make more sense to be consistent with the other TR natives:
```
public Action Command_TriggerTest( int client, int args )
{
	float startpos[3];
	GetClientEyePosition( client, startpos );
	float angles[3];
	GetClientEyeAngles( client, angles );
	
	// native void TR_EnumerateEntities(const float pos[3], const float vec[3], bool triggers, RayType rtype, TraceEntityEnumerator enumerator, any data=0);
	TR_EnumerateEntities( startpos, angles, true, RayType_Infinite, HitTrigger );

	// TR_EnumerateEntities internally calls TR_ClipRayToEntity before returning when enumerator returns false.
	//TR_ClipRayToEntity( MASK_ALL, entity );
		
	float pos[3];
	TR_GetEndPosition( pos );
		
	PrintToChatAll( "Hit a trigger! Collided at (%.2f, %.2f, %.2f)", pos[0], pos[1], pos[2] );
}

// return true to continue enumerating, or false to stop
public bool HitTrigger( int entity )
{
	char classname[64];
	GetEntityClassname( entity, classname, sizeof( classname ) );	
	return StrContains( classname, "trigger" ) == -1;
}
```

This is making the assumption that after you've clipped the ray you'll always be returning false, is this assumption correct?

When is the `mask` param to `TR_ClipRayToEntity` needed? (other than `MASK_ALL`)

If `TR_ClipRayToEntity` is needed, I think it should probably be used with `TR_GetEntityIndex`, rather than inside the enumerator.

Only other thing I would like to see is replacing the boolean `triggers` param with a flag param - we're trying to move away from non-obvious boolean params as they make code harder to understand quickly.

jason-e:
Whether `ClipRayToEntity` should be called automatically or available directly is mostly just a matter of optimization, I think. Two scenarios come to mind:

- If you want to use `EnumerateEntities` with some break condition in the enumerator for efficiency but don't care about the actual point of collision, calling `ClipRayToEntity` automatically is a waste. Though maybe that "waste" is too little to care about, I have no idea.
- If you have a known entity in mind and just want to clip a ray to it, without direct access to `ClipRayToEntity` you would have to call `EnumerateEntities` with a special filter, which seems inconvenient - though I guess it's no less convenient than `TR_TraceRayFilter`.

> When is the mask param to TR_ClipRayToEntity needed? (other than MASK_ALL)

I'm not sure, but it's probably there for a reason, right? I guess if you start with a known entity but don't know what its contents type might be. Or maybe: can entities be composed of multiple collidables with differing contents types? That would absolutely make the mask parameter useful. [This 2007 implementation of IEngineTrace](https://github.com/VSES/SourceEngine2007/blob/master/se2007/engine/enginetrace.cpp) passes that mask value through to a lot of stuff, so it probably matters for something.

I feel like it makes the most sense to expose these functions as-is rather than make our own, potentially more restrictive variants. They are probably that way for a reason, even if we don't know exactly why right now. The other TR_ natives seem to leave their corresponding `enginetrace` functions unrestricted.

>Only other thing I would like to see is replacing the boolean triggers param with a flag param

I guess it makes code more readable, but makes it less obvious to write since you have to find the flag names, and there will only ever be two values needed. Not a big deal either way; better to be consistant if this is something SM already does in other places.

SlidyBat:
> This is making the assumption that after you've clipped the ray you'll always be returning false, is this assumption correct?

Clipping the ray doesn't always mean that you want to stop enumerating. The condition for stopping the enumerating could be dependent on the results of ClipRayToEntity (or something else entirely). 
e.g. You only want to stop enumerating if the position meets some requirements. I think it would be best if its left up to the people using these functions instead of taking away that option.

> When is the mask param to TR_ClipRayToEntity needed? (other than MASK_ALL)

This depends on the use case. Using different masks on entities could give different results depending on where it hits them. For example, say I wanted to clip the ray to a players bounding box rather than it's hitbox, I could use MASK_PLAYERSOLID for that. I tested this out myself using [this code](https://pastebin.com/Er3B9p0A). Here's an image showcasing it as well (red hits hitbox, green hits bbox): https://i.imgur.com/HjNZqEy.jpg


> Only other thing I would like to see is replacing the boolean triggers param with a flag param

Do you mean making an enum specifically for this? I only made it a boolean since [the function it calls](https://github.com/VSES/SourceEngine2007/blob/master/se2007/engine/enginetrace.cpp#L98) also uses a bool so I thought that would be the most straightforward way to do it.

Also I wasn't sure about my ClipRayToEntity, since it creates a new trace result out of the ray. When you want to use the global trace result you pass in INVALID_HANDLE, or pass in a handle and it gets filled with a new trace result. Should I change this behaviour to work like the other TR functions with a regular/Ex version?

SlidyBat:
I've pushed another commit that seperates ClipRayToEntity into 2 groups of natives:
- TR_ClipRayToEntity(Ex) - This now takes a start/end/raytype like the other Trace natives and initializes a new ray before calling ClipRayToEntity.
- TR_ClipCurrRayToEntity(Ex) - This works like the implementation before this commit where it just performs ClipRayToEntity on the current global ray.

This way the ClipRayToEntity function isn't tied down to EnumerateEntities and can also be used by itself.
I also split those to natives into regular/extended versions that return new trace results to make them more like the other natives.

Any thoughts?

jason-e:
That implementation looks solid in my opinion.

Just to clarify why SildyBat went with two versions of `ClipRayToEntity` (not including their Ex counterparts): 

At a minimum, it makes sense to have a version of `ClipRayToEntity` that mirrors the original (it accepts an arbitrary ray definition); but there is also potential for a need to call `ClipRayToEntity` as part of the `EnumerateEntities` enumerator - especially since the two tend to go together. Since we don't have access to the ray/trace instances in SourcePawn, having `TR_ClipCurrRayToEntity` makes that much easier. Without it we would need to pass something like a datapack to the enumerator and repeatedly initialize a new copy of the same ray.

Pelipoika:
Can we have this please?

Benoist3012:
Yeah please.

PR #756, opened 2018-01-25
Add more TF2 specific DMG_ definitions

Pelipoika:


plugins/include/sdkhooks.inc
```diff
@@ -74,6 +74,24 @@
 #if !defined DMG_CRIT
```
asherkin: You don’t want these inside this `#if` - you might want an `#if` for each one though (@psychonic?)

Pelipoika: The DMG_ defines above are defined the same way

psychonic: I agree with @asherkin 

The ones above were done that way because there was a popular include at the time that had all of them in it copied from the SDK like ours were.

For the TF2 ones, it's very possible to run into plugins where users have manually defined one or more of these on their own due to us never having them. It doesn't hurt to guard them.

Pelipoika: Done

plugins/include/sdkhooks.inc
```diff
@@ -76,6 +76,36 @@
 	#define DMG_CRIT DMG_ACID
 #endif
 
+#if !defined DMG_RADIUS_MAX
+	//No damage falloff
```
Headline: This is a nit, but by doing this we eat up lots of whitespace, perhaps we can move the comments inline?
```cpp
#if !defined DMG_CRIT
	#define DMG_CRIT DMG_ACID /* TF2 crits and minicrits */
#endif
```
Perhaps even take away the whitespace between each `#endif` and `#if`, but I'm not sure if smdevs woud like them slammed together that closely.
@asherkin ?

asherkin: That's actually a good point - they need to be doccomments for the parser to pick them up correctly, and the sanest way to do that is inline using `/**< comment here */` (the `DMG_CRIT` one should change as well).

I don't really care about the line count / whitespace.

PR #757, opened 2018-01-26
Fix ``KvSavePosition`` definition

fakuivan:
The current ``KvSavePosition`` definition does not reflect the behaviour seen in [core/smn_keyvalues.cpp#L973-L981](https://github.com/alliedmodders/sourcemod/blob/404e96ad456531e9029f3b02b243a5284efd9691/core/smn_keyvalues.cpp#L973-L981). This can cause problems if some procedure on sp expects ``KvSavePosition`` to always add one node to the stack, no matter if there is no higher node.


asherkin:
Does it make sense to just have it always push rather than requiring plugins to check?

fakuivan:
I'd rather just keep the logic intact just in case someone found out about the issue and addressed it on their code. 
The only case I could think of would be if a pluign assumed that if this returned false, the stack should be on the root node.

peace-maker (approved):
I'm also for fixing the documentation instead of changing the native's behavior after all the time it existed.

PR #763, opened 2018-02-02
Prevent use of primitive float operation functions

Headline:
Apparently people use the float natives for actual float arithmetic; they're not _really_ intended to be used directly. I've added a note to deter people from using these instead of their respective operators.

### DIFF
--- a/plugins/basevotes.sp
+++ b/plugins/basevotes.sp
@@ -394,7 +394,7 @@ void VoteMenuClose()
 
 float GetVotePercent(int votes, int totalVotes)
 {
-	return FloatDiv(float(votes),float(totalVotes));
+	return float(votes) / float(totalVotes);
 }
 
 bool TestVoteDelay(int client)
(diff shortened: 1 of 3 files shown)


asherkin:
I'd prefer "This native is internal implementation." rather than "This native only exists to be overloaded."

Does adding `#deprecated` work and not carry through to the operators?

Headline:
@asherkin I haven’t tested whether #depreicated will carry through, but depreicated implies removal so I’m not sure if that’s what we’d want to suggest.

asherkin:
We want people to stop using these natives. They are not used internally in modern plugins, the JIT emits float opcodes directly, they're definitely on the removal path if we break binary compat for some reason.

Headline:
The deprecation warning doesn't carry through to the overloads, so `#pragma deprecated` works as we'd want here. I also threw in your edit, thanks!

🚢

PR #764, opened 2018-02-04
Use msvc 2015 lib with 2017 (#764)

### LINKED ISSUE #761: Add MSVC 1912 support
Visual Studio 2017 version 15.5 added more support for more C++ 17 features, this caused MSVC's version to be bumped up to version 1912 (from previous version 1911).

I forgot PR #678 existed which is why it wasn't notified about the required change to support this new version.

Headline:
All we needed to do was bump the version, no changes were required.

This fixes #761

### DIFF
--- a/core/AMBuilder
+++ b/core/AMBuilder
@@ -86,7 +86,7 @@ for sdk_name in SM.sdks:
           vs_year = '2013'
         elif msvc_ver == 1900:
           vs_year = '2015'
-        elif msvc_ver in (1910, 1912):
+        elif msvc_ver >= 1910:
           vs_year = '2017'
         else:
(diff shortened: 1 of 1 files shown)


psychonic:
That's going to break it for 1911. We should probably just change it to >= 1910 for now.

Headline:
oh I skimmed over it too fast, I thought it was `elif msvc_ver in range(1910, 1912):`

Would you prefer this, or just go `>= 1910`

psychonic:
I'm not even positive that we can't just use the the 2015 (1900) lib with 2017. The internal VS version numbers are so close to each other due to the minimal ABI differences. Microsoft has been intentionally trying to make it less of an issue with recent versions.

PR #766, opened 2018-02-05
Add GameData methodmap

peace-maker:
Wrap gamedata natives in a pretty methodmap and update stock plugins/includes to use it.

This fixes bad documentation on `GameConfGetAddress` and not closing the gameconfig file if handle creation fails as well.

### DIFF
--- a/core/logic/smn_gameconfigs.cpp
+++ b/core/logic/smn_gameconfigs.cpp
@@ -67,7 +67,10 @@ static cell_t smn_LoadGameConfigFile(IPluginContext *pCtx, const cell_t *params)
 		return pCtx->ThrowNativeError("Unable to open %s: %s", filename, error);
 	}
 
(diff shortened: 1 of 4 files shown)


asherkin:
I'm not a fan of the name `GameConf`, but don't really have a better suggestion, opening up to the floor? @alliedmodders/sourcemod 

Implementation looks sounds, nice catch on the fixes.

peace-maker:
Could call it `GameData` since that's what people use to refer to it.

Fyren:
I think with the existing natives named the way they are, I'd prefer `GameConf`.

Headline:
Personally, I think that naming it `GameData` is a better choice even though it differs from the original native's naming scheme. It looks cleaner, and this is an evolving language anyway. 😛 

I just wanted to give my two cents, if that's welcome.

psychonic:
I have no strong feelings one way or the other on the naming. Everything else looks okay to me.

PR #767, opened 2018-02-13
Add new regex natives.

Drifter321:
Not sure how long this has been broken but currently MatchRegex doesn't actually return the number of matches. It instead returns 1 + the number of capture group matches. This simply fixes it to work how the docs currently say it should work. However, it is possible to add capture group support at some point but would require some effort along with new natives.

Some test outputs.

```
Matches 5
Match 0 is t
Match 1 is t
Match 2 is t
Match 3 is t
Match 4 is t

Matches 1
Match 0 is _

Matches 3
Match 0 is .
Match 1 is _
Match 2 is +

Matches 1
Match 0 is test
```

Test plugin

```
#pragma newdecls required
 
#include <sourcemod>
#include <regex>
 
public void OnPluginStart()
{
	Regex regex = CompileRegex("t", PCRE_UTF8);
	int filter = MatchRegex(regex, "test.string_doesnt+match");
	
	PrintToServer("Matches %i", filter)
	
	for(int i = 0; i < filter; i++)
	{
		char reply[100];
		
		GetRegexSubString(regex, i, reply, sizeof(reply));
		
		PrintToServer("Match %i is %s", i, reply);
	}
	
	Regex regex2 = CompileRegex("_", PCRE_UTF8);
	filter = MatchRegex(regex2, "test.string_doesnt+match");
	
	PrintToServer("\nMatches %i", filter)
	
	for(int i = 0; i < filter; i++)
	{
		char reply[100];
		
		GetRegexSubString(regex2, i, reply, sizeof(reply));
		
		PrintToServer("Match %i is %s", i, reply);
	}
	
	Regex regex3 = CompileRegex("[\\\\\\W|^_]", PCRE_UTF8);
	filter = MatchRegex(regex3, "test.string_doesnt+match");
	
	PrintToServer("\nMatches %i", filter)
	
	for(int i = 0; i < filter; i++)
	{
		char reply[100];
		
		GetRegexSubString(regex3, i, reply, sizeof(reply));
		
		PrintToServer("Match %i is %s", i, reply);
	}
	
	Regex regex4 = CompileRegex("tes(t)", PCRE_UTF8);
	filter = MatchRegex(regex4, "test.string_doesnt+match");
	
	PrintToServer("\nMatches %i", filter)
	
	for(int i = 0; i < filter; i++)
	{
		char reply[100];
		
		GetRegexSubString(regex4, i, reply, sizeof(reply));
		
		PrintToServer("Match %i is %s", i, reply);
	}
}
```

I'm not exactly sure if this is the best approach. But I am open to all feedback.

### DIFF
--- a/extensions/regex/CRegEx.cpp
+++ b/extensions/regex/CRegEx.cpp
@@ -42,7 +42,7 @@ RegEx::RegEx()
 	re = NULL;
 	mFree = true;
 	subject = NULL;
-	mSubStrings = 0;
+	mMatchCount = 0;
 }
 
 void RegEx::Clear ()
@@ -56,7 +56,7 @@ void RegEx::Clear ()
 	if (subject)
 		delete [] subject;
 	subject = NULL;
-	mSubStrings = 0;
+	mMatchCount = 0;
 }
 
 RegEx::~RegEx()
@@ -92,7 +92,7 @@ int RegEx::Compile(const char *pattern, int iFlags)
 	return 1;
 }
 
-int RegEx::Match(const char *str)
+int RegEx::Match(const char *str, unsigned int offset)
 {
 	int rc = 0;
 
@@ -105,37 +105,11 @@ int RegEx::Match(const char *str)
 	subject = new char[strlen(str)+1];
 	strcpy(subject, str);
 
-	unsigned int offset = 0;
 	unsigned int len = strlen(subject);
-	unsigned int matches = 0;
 
-	while (offset < len && (rc = pcre_exec(re, 0, subject, len, offset, 0, ovector, sizeof(ovector))) >= 0)
-	{
-		/* This also works for capture groups now, but assume rc is always 1 if we have any match.
-		Reference : https://stackoverflow.com/questions/1421785/how-can-i-use-pcre-to-get-all-match-groups
-		Example:
-		rc = 2 for
-		tes(t)
-		with string test
+	rc = pcre_exec(re, NULL, subject, len, offset, 0, mMatches[0].mVector, sizeof(mMatches[0].mVector));
 
-		0 = test
-		1 = t
-
-		for (int i = 0; i < rc; ++i)
-		{
-			printf("%2d: %.*s\n", i, ovector[2 * i + 1] - ovector[2 * i], str + ovector[2 * i]);
-		}
-		*/
-
-		char *substr_a = subject + ovector[2 * 0];
-		int substr_l = ovector[2 * 0 + 1] - ovector[2 * 0];
-
-		mMatches[matches] = ke::AString(substr_a, substr_l);
-		offset = ovector[1];
-		matches++;
-	}
-
-	if (rc < 0 && matches == 0) // rc < 0 when there is no more matches, so if we have matches dont error (Maybe check if rc == PCRE_ERROR_NOMATCH and if we have matches?)
+	if (rc < 0)
 	{
 		if (rc == PCRE_ERROR_NOMATCH)
 		{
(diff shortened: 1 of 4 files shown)


KyleSanderson:
This is tough to change because a for loop would have to be fixed at the plugin level. Fundamentally, even at a base level you're absolutely correct this is wrong. But people using < length would have to be fixed to <=. I'm sure there are other cases, but this to me is technical debt that can't be changed at this point.

asherkin:
My understanding is still that this is currently working exactly as expected and documented, the documentation just confuses people more familiar with modern regex implementations than Perl.

Global matching behaviour should be behind a new native, maybe `MatchAll`?

Drifter321:
After additional discussion it appears it was working correctly and the documentation was leading to lots of confusion and various features lacking.

MatchRegex has been reverted to how it originally was, but will now also accept an additional param for offset (defaulting to 0). GetRegexSubString now also takes an additional param to identify the match (defaulting to 0) to get substrings for.

MatchAll has been added. MatchCount, MatchOffset (Offset in the test plugin since i decided to rename it after), CaptureCount have also been added to provide info when using MatchAll, or when trying to manually loop using MatchRegex.

Test Plugin
https://paste.ee/p/CdFjB

And output
https://paste.ee/p/r0i44

I tried my best to clarify the documentation, but its probably still not great...

extensions/regex/CRegEx.cpp
```diff
@@ -106,7 +105,9 @@ int RegEx::Match(const char *str)
 	subject = new char[strlen(str)+1];
 	strcpy(subject, str);
 
-	rc = pcre_exec(re, NULL, subject, (int)strlen(subject), 0, 0, ovector, 30);
+	unsigned int len = strlen(subject);
+
+	rc = pcre_exec(re, NULL, subject, len, offset, 0, mMatches[0].mVector, sizeof(mMatches[0].mVector));
```
asherkin: This `sizeof` is incorrect, it needs to be the element count, not the byte size.

The PRCE documentation points this out about 100 times 😛

extensions/regex/CRegEx.cpp
```diff
@@ -119,10 +120,54 @@ int RegEx::Match(const char *str)
...
+
+	//save str
+	subject = new char[strlen(str) + 1];
+	strcpy(subject, str);
+
+	unsigned int offset = 0;
+	unsigned int len = strlen(subject);
+	unsigned int matches = 0;
+
+	while (offset < len && (rc = pcre_exec(re, 0, subject, len, offset, 0, mMatches[matches].mVector, sizeof(mMatches[matches].mVector))) >= 0)
```
asherkin: Ditto.

extensions/regex/CRegEx.h
```diff
@@ -28,10 +28,17 @@
...
  */
+#include <am-string.h>
 
 #ifndef _INCLUDE_CREGEX_H
 #define _INCLUDE_CREGEX_H
 
+struct RegexMatch
+{
+	int mSubStringCount;
+	int mVector[30];
```
asherkin: Let's double this, 10 captures seems limiting, we have the RAM.

plugins/include/regex.inc
```diff
@@ -149,40 +184,31 @@ native Regex CompileRegex(const char[] pattern, int flags = 0, char[] error="",
...
- * @param flags			General flags for the regular expression.
- * @param error			Error message, if applicable.
- * @param maxLen		Maximum length of the error buffer.
- * @return				Number of substrings found or -1 on failure.
+ * @note				str_id = 0 is the full captured string, anything else is the capture group index.
+ *						if MatchRegex is used match can only be 0
  */
+native bool GetRegexSubString(Handle regex, int str_id, char[] buffer, int maxlen, int match = 0);
+  
 stock int SimpleRegexMatch(const char[] str, const char[] pattern, int flags = 0, char[] error="", int maxLen = 0)
```
asherkin: It looks like the documentation got deleted for `SimpleRegexMatch`?

Drifter321: Oops, good catch must have deleted it when i deleted the new natives from non-methodmap.

PR #775, opened 2018-02-27
Fix CompileRegex not actually setting a valid error code.

Drifter321:
CompileRegex is currently setting the error code to the offset of where the error occurred rather than the actual error.

To fix this pcre_compile was replaced with pcre_compile2 and a map used to convert the pcre error to a simpler (posix) error code but maintaining pcre's error string and adding the error codes to the enum. Also changed the internal member names since it was rather confusing. In the future we can also add in the offset so we can pass the offset where the error occurred.

The reason for mapping the errors is that pcre's error codes dont seem to be documented at all(?) From pcre_internal.h the errors are named ERR0, ERR1 etc... and i couldn't find any more information. There is also over 80 of them (due to the amount of detail) so I think the posix error codes are a simpler approach.

Small test plugin and its output.
```
public void OnPluginStart()
{
	char test_string[] = "my9_test1+string2.wooooo5";
	
	RegexError errorcode;
	char error[128]

	Regex re = new Regex("([^_|\\\\][a-zA-z|+])([0-9]", _, error, sizeof(error), errorcode);
	
	PrintToServer("Error code is %i (%s)", view_as<int>(errorcode), error);
}

Error code is 11 (missing ))
```

### DIFF
--- a/extensions/regex/CRegEx.cpp
+++ b/extensions/regex/CRegEx.cpp
@@ -38,6 +38,7 @@
 RegEx::RegEx()
 {
 	mErrorOffset = 0;
+	mErrorCode = 0;
 	mError = NULL;
 	re = NULL;
 	mFree = true;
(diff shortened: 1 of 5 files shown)


404UserNotFound:
So this was actually a bug then? Glad to see it's gonna be fixed.

PR #777, opened 2018-03-03
collection of plugin cleanups

KyleSanderson:
Just doing drive-bys on first-party plugins.

### DIFF
--- a/plugins/basebans.sp
+++ b/plugins/basebans.sp
@@ -299,6 +299,13 @@ public Action Command_AddBan(int client, int args)
 		return Plugin_Handled;
 	}
 
+	AdminId tid = FindAdminByIdentity("steam", authid);
+	if (client && !CanAdminTarget(GetUserAdmin(client), tid))
+	{
+		ReplyToCommand(client, "[SM] %t", "No Access");
+		return Plugin_Handled;
+	}
+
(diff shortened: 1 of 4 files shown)


plugins/basebans.sp
```diff
@@ -86,7 +86,7 @@ public void OnPluginStart()
 	}
 }
 
-public void OnMapStart()
+public void OnConfigsExecuted()
```
asherkin: `LoadBanReasons` doesn't rely on anything from a config file, `OnMapStart` is a better choice.

plugins/rockthevote.sp
```diff
@@ -99,47 +92,40 @@ public void OnMapStart()
...
 }
 
 public void OnClientDisconnect(int client)
 {
-	if(IsFakeClient(client))
-		return;
+	g_Voted[client] = false;
 	
-	if(g_Voted[client])
+	if (g_Voted[client])
```
asherkin: This will never be hit.

KyleSanderson: Great catch, thanks

plugins/basebans.sp
KyleSanderson: g_BanReasonsPath is a config file :-)

Issue #771, opened 2018-02-22
Support for Zombie Panic! Source 3.0 and above


JoelTroch:
This issue is a follow-up to the discussion on IRC about SourceMod support for [Zombie Panic! Source](http://store.steampowered.com/app/17500/Zombie_Panic_Source/) (ZPS) 3.0 and above.

**Note worthy informations:**
- We now use the "**Source SDK 2013 Multiplayer**" version of the Source engine (previously it was "Source 2007").
- The next ZPS version ("3.0.7-Hotfix2") will contain Linux binaries with this [patch](https://forums.alliedmods.net/showpost.php?p=2131144&postcount=2) applied.
- Main AppID is `17500`, dedicated server AppID is `17505`.

JoelTroch:
We just released our "3.0.7-Hotfix2" patch that contains the Linux binaries with the SDK 2013 patch applied.

If you guys need help, just drop the word ^^

Snneaky:
When is it going to be finished? any news?

KyleSanderson:
Closing, fixed by PR #782.

PR #782, opened 2018-03-12
[ZPS] Update gamedata and engine migration

JoelTroch:
This pull request updates the gamedata files for Zombie Panic! Source, it also tells SourceMod that ZPS is now using the SDK 2013 engine rather than EP2's.

Related issue: #771


psychonic:
Thanks!

PR #784, opened 2018-03-19
Replace ke::SafeSprintf with ke::SafeStrcpy

Headline:
I've been going through looking for changes that I can sweep across the project to keep things consistent. This replaces every instance of `ke::SafeSprintf` which is only used for copying with `ke::SafeStrcpy`.


KyleSanderson:
As I'm sure you can tell safestrcpy came later. Extensions spin their own versions as well because we don't bring in stdlib, if you want to glance at those as well by all means. Otherwise this subtlety fixes some existing issues with double reformatting and similar; which is an improvement.

PR #786, opened 2018-03-19
Use MySQL 5.5 for MySQL extension.

psychonic:
@peace-maker Your wildest dream come true.

### DIFF
--- a/AMBuildScript
+++ b/AMBuildScript
@@ -183,7 +183,7 @@ class SMConfig(object):
         self.mysql_root['x86'] = builder.options.mysql_path
       else:
         for i in range(7):
-          self.mysql_root['x86'] = ResolveEnvPath('MYSQL5', 'mysql-5.' + str(i))
+          self.mysql_root['x86'] = ResolveEnvPath('MYSQL55', 'mysql-5.' + str(i))
           if self.mysql_root['x86']:
             break
       if not self.mysql_root['x86'] or not os.path.isdir(self.mysql_root['x86']):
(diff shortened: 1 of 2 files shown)


Bara:
With which version comes this?

asherkin:
Uh, none, it hasn’t been merged.

psychonic:
On our own buildslaves, we'll be using 5.5.54 on Windows, 5.5.58 on Linux, and 5.5.40 on Mac.

Intentionally not going higher, even in the 5.5 series, to not introduce extra dependencies.

Bara:
Okay good to know, but i mean with which sourcemod version (1.10 or for 1.9 too?). 😖

psychonic:
Undetermined as of yet.

Bara:
but isn't the version for linux 5.6? ( Line: https://github.com/alliedmodders/sourcemod/blob/33c74944ff112b92829cc9006e0231384e139b10/tools/checkout-deps.sh#L55 / https://github.com/alliedmodders/sourcemod/blob/33c74944ff112b92829cc9006e0231384e139b10/tools/checkout-deps.sh#L66 )

psychonic:
No. We don't use checkout-deps in the official builds. The MYSQL55 env var on the slave points to mysql-5.5.58-linux-glibc2.12-i686.

I didn't bother updating checkout-deps for it since it was already at a newer version than we build now (itself a newer one than we were building against before).

Doing self-compiles against 5.6 is fine. We're just keeping the lower one on official builds to not increase the GLIBC version requirement.

PR #788, opened 2018-03-21
There is no "menu" param & typos

Hexer10:


KyleSanderson:
Bonus points if you fix GetTitle as well right above it, as well as any others.

KyleSanderson:
Thanks bud

PR #791, opened 2018-03-28
Remove DB Configuration Locking

Headline:
As @KyleSanderson asked, I attempted to remove configuration locking in favor of a refcounted structure. I'm a multi-threading novice, and this should be heavily reviewed before any approval.

The idea behind this is that different threads can ask for a specific config, and that configuration file it returns will be grabbed from the latest parse. Because we now have two lists in `DatabaseConfBuilder`, any search will always happen on the fully populated list. The other one is just used while parsing happens, and it's ptr is copied over to the other member variable once parsing is done. A new list is created in it's place.


core/logic/Database.cpp
```diff
@@ -465,17 +323,16 @@ const DatabaseInfo *DBManager::FindDatabaseConf(const char *name)
...
+	ke::RefPtr<ConfDbInfoList> list(builder.GetConfigList());
+	for (size_t i = 0; i < list->length(); i++)
 	{
-		ConfDbInfo &conf = *(*iter);
-		if (conf.name == name)
+		ke::RefPtr<ConfDbInfo> current(list->at(i));
+		if (strcmp(current->name.chars(), name) == 0)
 		{
-			return &conf;
+			current->AddRef();
```
Headline: Without this line, we'll get heap corruption for when `info` in the snippet below goes out of scope. While this does solve this crash, it is fundamentally wrong. I'm not sure why this fixes the problem. Help.

https://github.com/Headline/sourcemod/blob/53dde59defa8b6029bc2f716952ee1a43841a2ef/core/logic/Database.cpp#L313-L322

core/logic/DatabaseConfBuilder.cpp
```diff
@@ -0,0 +1,206 @@
...
+ */
+#include "DatabaseConfBuilder.h"
+#include <bridge/include/ILogger.h>
+
+#define DBPARSE_LEVEL_NONE		0
+#define DBPARSE_LEVEL_MAIN		1
+#define DBPARSE_LEVEL_DATABASE	2
+
+DatabaseConfBuilder::DatabaseConfBuilder() :	m_CurInfoList(new ConfDbInfoList()), 
+												m_InfoList(new ConfDbInfoList())
```
KyleSanderson: what happened here?

core/logic/DatabaseConfBuilder.cpp
```diff
@@ -0,0 +1,206 @@
...
+	return m_DefDriver.chars();
+}
+
+void DatabaseConfBuilder::StartParse()
+{
+	SMCError err;
+	SMCStates states = {0, 0};
+	if ((err = textparsers->ParseFile_SMC(m_Filename.chars(), this, &states)) != SMCError_Okay)
+	{
+		logger->LogError("[SM] Detected parse error(s) in file \"%s\"", m_Filename);
```
KyleSanderson: .chars() since you converted to ke::AString

core/logic/Database.cpp
KyleSanderson: When you're invoking GetDatabaseConf you're not calling AddRef without this line, but you're calling Release by hand in the caller. Pair this with RAII or match the calls, otherwise you're going to get the assert this is probably generating for you.

core/logic/DatabaseConfBuilder.h
```diff
@@ -56,7 +56,31 @@ class ConfDbInfo : public ke::RefcountedThreadsafe<ConfDbInfo>
...
+	ConfDbInfo *GetDatabaseConf(const char *name) {
+		for (size_t i = 0; i < this->length(); i++)
+		{
+			ke::RefPtr<ConfDbInfo> current(this->at(i));
+			if (strcmp(current->name.chars(), name) == 0)
+			{
+				return current;
+			}
+		}
+		return NULL;
```
KyleSanderson: nullptr

core/logic/DatabaseConfBuilder.cpp
```diff
@@ -0,0 +1,207 @@
...
+		m_ParseState = DBPARSE_LEVEL_NONE;
+		return SMCResult_Halt;
+	}
+
+	return SMCResult_Continue;
+}
+
+void DatabaseConfBuilder::ReadSMC_ParseEnd(bool halted, bool failed)
+{
+	m_InfoList->Release();
```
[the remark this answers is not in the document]
Headline: How is this not thread-safe, exactly? We're ++refcount on it's creation and when we replace the reference a line down we'll no longer be able to --refcount, resulting in a never-destroyed obj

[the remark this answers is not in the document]
KyleSanderson: There needs to be a guarantee that m_InfoList is never invalid, once you release the final reference (this) obviously becomes invalid memory. You need to store infoList in the function, replace the ptr, and then perform the release. Alternatively, see if you can change this type to be a refptr.

core/logic/DatabaseConfBuilder.cpp
```diff
@@ -158,28 +158,15 @@ SMCResult DatabaseConfBuilder::ReadSMC_LeavingSection(const SMCStates *states)
...
-		cdb->realDriver = m_CurInfo->realDriver;
-		cdb->info.maxTimeout = m_CurInfo->info.maxTimeout;
-		cdb->info.port = m_CurInfo->info.port;
-
-		cdb->info.driver = cdb->driver.chars();
-		cdb->info.database = cdb->database.chars();
-		cdb->info.host = cdb->host.chars();
-		cdb->info.user = cdb->user.chars();
-		cdb->info.pass = cdb->pass.chars();
+	{
```
KyleSanderson: whitespace.

core/logic/DatabaseConfBuilder.cpp
```diff
@@ -0,0 +1,188 @@
...
+}
+
+void DatabaseConfBuilder::ReadSMC_ParseEnd(bool halted, bool failed)
+{
+	ConfDbInfoList *temp = m_InfoList;
+	m_InfoList = m_CurInfoList;
+	temp->Release();
+	
+	m_CurInfoList = nullptr;
+}
```
KyleSanderson: newline.

core/logic/DatabaseConfBuilder.cpp
```diff
@@ -0,0 +1,188 @@
...
+	return SMCResult_Continue;
+}
+
+void DatabaseConfBuilder::ReadSMC_ParseEnd(bool halted, bool failed)
+{
+	ConfDbInfoList *temp = m_InfoList;
+	m_InfoList = m_CurInfoList;
+	temp->Release();
+	
+	m_CurInfoList = nullptr;
```
KyleSanderson: I'd maybe rename these to m_parseInfoList or something and store them differently (in another private:), the cur stuff while I'm used to it now is really unclear to anyone doing a driveby.

core/logic/DatabaseConfBuilder.cpp
```diff
@@ -0,0 +1,188 @@
...
+	{
+		m_CurInfo->info.driver = m_CurInfo->driver.chars();
+		m_CurInfo->info.database = m_CurInfo->database.chars();
+		m_CurInfo->info.host = m_CurInfo->host.chars();
+		m_CurInfo->info.user = m_CurInfo->user.chars();
+		m_CurInfo->info.pass = m_CurInfo->pass.chars();
+		
+		/* Save it.. */
+		m_CurInfoList->append(m_CurInfo);
+
```
KyleSanderson: reset or free m_CurInfo here.

core/logic/DatabaseConfBuilder.cpp
```diff
@@ -0,0 +1,188 @@
...
+}
+
+DatabaseConfBuilder::~DatabaseConfBuilder()
+{
+	if (m_CurInfoList)
+		m_CurInfoList->Release();
+	m_InfoList->Release();
+}
+
+ConfDbInfoList *DatabaseConfBuilder::GetConfigList()
```
KyleSanderson: This needs to be refptr'd in the InfoList, bringing it in from the builder still violates the refptr's guarantee. One authoritative data source.

core/logic/DatabaseConfBuilder.cpp
Headline: why?

core/logic/DatabaseConfBuilder.cpp
Headline: I'm not entirely sure I understand, are you saying we should be doing something along the lines of 

```cpp
RefPtr<ConfDbInfoList> list(m_InfoList);
return list;
```
?

Headline: Github shows that this review comment is outdated, but no changes involving this have actually been made. Just a heads up

Headline:
@dvander here's your reminder :p

core/logic/DatabaseConfBuilder.h
```diff
@@ -0,0 +1,111 @@
...
+	void ReadSMC_ParseEnd(bool halted, bool failed);
+
+private:
+	unsigned int m_ParseLevel;
+	unsigned int m_ParseState;
+	ConfDbInfo *m_ParseCurrent;
+	ConfDbInfoList *m_ParseList;
+private:
+	ke::AString m_Filename;
+	ConfDbInfoList *m_InfoList;
```
dvander: This (and m_ParseList) should be in a RefPtr.

core/logic/DatabaseConfBuilder.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+#define DBPARSE_LEVEL_NONE		0
+#define DBPARSE_LEVEL_MAIN		1
+#define DBPARSE_LEVEL_DATABASE	2
+
+DatabaseConfBuilder::DatabaseConfBuilder()
+	: m_ParseList(nullptr),
+	  m_InfoList(new ConfDbInfoList())
+{
+	m_InfoList->AddRef();
```
dvander: Drop this manual AddRef + Release

dvander:
ACK, this should be much safer now w/ the thread lookup removed.

core/logic/DatabaseConfBuilder.h
```diff
@@ -0,0 +1,110 @@
...
+	ke::AString database;
+	IDBDriver *realDriver;
+	DatabaseInfo info;
+};
+
+class ConfDbInfoList : public ke::Vector<ConfDbInfo *>
+{
+	friend class DatabaseConfBuilder; // lets DatabaseConfBuilder use SetDefaultDriver
+public:
+	ke::AString& GetDefaultDriver() {
```
KyleSanderson: I almost wonder if we even need this exposed publicly still. Code that doesn't handle this will fallback to the default driver anyways; this seems like a well intentioned piece that we just never really used. Errors if a specific databases entry is outstanding just don't really make sense to me.

Headline: Are you suggesting to keep the default driver in `DBManager` where it was before? I like it's existence here that way a `ConfDbInfoList` contains every piece of information acquired from a single parse

Headline:
Some simple SQL tests produce expected results. Just another pair of eyes and we should be good to go.

I'd request a review, but tagging is the best I can do. @asherkin can you give this a look

core/logic/Database.cpp
```diff
@@ -343,17 +198,17 @@ void DBManager::RemoveDriver(IDBDriver *pDriver)
...
 
-	/* Make sure NOTHING references this! */
-	List<ConfDbInfo *>::iterator iter;
-	for (iter=m_confs.begin(); iter!=m_confs.end(); iter++)
+	ke::RefPtr<ConfDbInfoList> list(builder.GetConfigList());
+	for (size_t i = 0; i < list->length(); i++)
 	{
-		ConfDbInfo &db = *(*iter);
-		if (db.realDriver == pDriver)
+		ConfDbInfo *current = list->at(i);
```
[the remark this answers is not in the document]
Headline: **note for people driving by**: This review comment is outdated since we threw out the refptr stuff

core/logic/DatabaseConfBuilder.h
Headline: note for people driving by: This review comment is also outdated

core/logic/DatabaseConfBuilder.h
```diff
@@ -0,0 +1,122 @@
...
+	ke::AString driver;
+	ke::AString host;
+	ke::AString user;
+	ke::AString pass;
+	ke::AString database;
+	IDBDriver *realDriver;
+	DatabaseInfo info;
+};
+
+class ConfDbInfoList : public ke::Vector<ConfDbInfo *>
```
asherkin: Ok, scratch that last comment - I see that threaded access was removed - but it does look like we leak every `ConfDbInfo` created?

PR #793, opened 2018-03-30
Distribute common number phrases

stickz:
Translation URL: http://translator.mitchdempsey.com/sourcemod_plugins/247

### DIFF
--- a/plugins/include/string.inc
+++ b/plugins/include/string.inc
@@ -546,3 +546,32 @@ stock int ImplodeStrings(const char[][] strings, int numStrings, const char[] jo
 	}
 	return total;
 }
+
+/**
+ * Converts a small integer number between 1 and 12 to a word.
+ * Able to be used when translating across multiple languages.
+ *
+ * @param int		A number between 1 and 12.
+ * @return		The corresponding word belonging to the number.
+ */
+
+stock char NumberToWord(int num)
+{
+	char englishNumber[16];
+	switch (num)
+	{
+		case 1: Format(englishNumber, sizeof(englishNumber), "one");
+		case 2: Format(englishNumber, sizeof(englishNumber), "two");
+		case 3: Format(englishNumber, sizeof(englishNumber), "three");
+		case 4: Format(englishNumber, sizeof(englishNumber), "four");
+		case 5: Format(englishNumber, sizeof(englishNumber), "five");
+		case 6: Format(englishNumber, sizeof(englishNumber), "six");
+		case 7: Format(englishNumber, sizeof(englishNumber), "seven");
+		case 8: Format(englishNumber, sizeof(englishNumber), "eight");
+		case 9: Format(englishNumber, sizeof(englishNumber), "nine");
+		case 10: Format(englishNumber, sizeof(englishNumber), "ten");
+		case 11: Format(englishNumber, sizeof(englishNumber), "eleven");
+		case 12: Format(englishNumber, sizeof(englishNumber), "twelve");
+	}	
+	return englishNumber;
+}
--- a/plugins/include/string.inc
+++ b/plugins/include/string.inc
@@ -549,7 +549,9 @@ stock int ImplodeStrings(const char[][] strings, int numStrings, const char[] jo
 
 /**
  * Converts a small integer number between 1 and 12 to a word.
- * Able to be used when translating across multiple languages.
+ *
+ * Able to be used when translating if numbers.phrases is loaded.
+ * Example: PrintToChat(client, "Language Number: %t", NumberToWord(5));
  *
  * @param int		A number between 1 and 12.
  * @return		The corresponding word belonging to the number.
--- a/translations/numbers.phrases.txt
+++ b/translations/numbers.phrases.txt
@@ -1,0 +1,51 @@
+"Phrases"
+{
+	"one"
+	{
+		"en"		"one"
+	}
+	"two"
+	{
+		"en"		"two"
+	}
+	"three"
+	{
+		"en"		"three"
+	}
+	"four"
+	{
+		"en"		"four"
+	}
+	"five"
+	{
+		"en"		"five"
+	}
+	"six"
+	{
+		"en"		"six"
+	}
+	"seven"
+	{
+		"en"		"seven"
+	}
+	"eight"
+	{
+		"en"		"eight"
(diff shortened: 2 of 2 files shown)


TiBarification:
what is purpose for this?

stickz:
To immediately be able to convert an integer to a number word and have translations available.

Headline:
I’m not sure I see the purpose, and there’s also no logic in sm that would even utilize this.

stickz:
There is logic now. You can use ether %s for the English literal or %t for the translation on a PrintToChat().

plugins/include/string.inc
```diff
@@ -546,3 +546,32 @@ stock int ImplodeStrings(const char[][] strings, int numStrings, const char[] jo
...
+		case 5: Format(englishNumber, sizeof(englishNumber), "five");
+		case 6: Format(englishNumber, sizeof(englishNumber), "six");
+		case 7: Format(englishNumber, sizeof(englishNumber), "seven");
+		case 8: Format(englishNumber, sizeof(englishNumber), "eight");
+		case 9: Format(englishNumber, sizeof(englishNumber), "nine");
+		case 10: Format(englishNumber, sizeof(englishNumber), "ten");
+		case 11: Format(englishNumber, sizeof(englishNumber), "eleven");
+		case 12: Format(englishNumber, sizeof(englishNumber), "twelve");
+	}	
+	return englishNumber;
```
Headline: You can't return `char[]`, you'll have to write to a buffer that's passed in for this to even work as intended.

plugins/include/string.inc
```diff
@@ -546,3 +546,32 @@ stock int ImplodeStrings(const char[][] strings, int numStrings, const char[] jo
...
+ * @param int		A number between 1 and 12.
+ * @return		The corresponding word belonging to the number.
+ */
+
+stock char NumberToWord(int num)
+{
+	char englishNumber[16];
+	switch (num)
+	{
+		case 1: Format(englishNumber, sizeof(englishNumber), "one");
```
Headline: For this case and every other case below, you're not using the translation file you've just created.

plugins/include/string.inc
```diff
@@ -546,3 +546,32 @@ stock int ImplodeStrings(const char[][] strings, int numStrings, const char[] jo
 	}
 	return total;
 }
+
+/**
+ * Converts a small integer number between 1 and 12 to a word.
+ * Able to be used when translating across multiple languages.
```
Headline: "Able to be used when translating across multiple languages"

this can be accomplished if you use the translation file you've created

plugins/include/string.inc
stickz: Yes actually can on sourcemod. It was used a while back for strings.

This works: 'PrintToChatAll("Hello my number is %s" , NumberToWord(5));'

Headline: I'm aware in that niche case you can return `char[]`, but it limits it's application greatly.

plugins/include/string.inc
stickz: This works: PrintToChat(client, "Hello, number 5 in my language is: %t", NumberToWord(5));

plugins/include/string.inc
stickz: If you need a buffer, create one using Format. It's not worth creating two extra steps to translate.

asherkin:
I don’t really want to take this into SM’s current translation system, translating numbers into various locales is extremely complex.

asherkin:
Can you show how you’re using this in your plugins?

stickz:
I use it to cut back on translations and support more languages immediately.

Passing to other translations. (new)  [Plugin File](https://github.com/stickz/Redstone/blob/master/addons/sourcemod/scripting/nd_suicide_controls.sp#L112) , [Translation File](https://github.com/stickz/Redstone/blob/master/updater/nd_tools/translations/nd_tools.phrases.txt#L13-L20)

Passing with buffer. (outdated) [Plugin File](https://github.com/stickz/Redstone/blob/e1564975e1ab3cd739643b42f09c8f9337d4ca46/addons/sourcemod/scripting/nd_rockthevote.sp#L306), [Translation File](https://github.com/stickz/Redstone/blob/master/updater/nd_rockthevote/translations/nd_rockthevote.phrases.txt#L38-L47)

 Passing with buffer. (outdated) [Plugin File](https://github.com/stickz/Redstone/blob/d97f66e7192adabf870ebe2dc2c00bbf5e69735c/addons/sourcemod/scripting/nd_surrender.sp#L263), [Translation File] (https://github.com/stickz/Redstone/blob/master/updater/nd_surrender/translations/nd_surrender.phrases.txt#L43-L51)

404UserNotFound:
Why is this just limited to 12? That seems not very useful unless one was to manually expand it....unless I'm missing something here.

stickz:
It's not designed to go up all the way up to 32 or 64. That would create too many phrases to translate. And result in a ginormous switch statement. This is quick and dirty if you only need small numbers.  Once the percent of probable usage starts getting in the single digits, it's time to stop and implement a better solution. 10 wasn't enough and 12 only created two more phrases and made a big difference.

dvander:
This doesn’t really let you translate well... in fact the design discourages utility files like this. “four beers” and “four desks” translate different in Japanese (4-pon vs 4-tsu, see counter words). Or similarly in Vietnamese where bốn can change to tứ or một to nhất if the context is ordered.

stickz:
How many people will switch 12 different phrases for each print-out, to support numbers properly? I tried 2 then the numbers changed, when I updated my server configuration.

I see this as a starting point. If someone wants to create a full broken translation feature for sourcemod with artificial intelligence, that won't be me. Canada has one for parliament. It took three years to make.

dvander:
Why bother including twelve hardcoded numbers then, what kind of useful phrases do you have in mind that stop and twelve and don’t require context?

stickz:
I can realized this function can be done in three lines if the format of phrases are changed, by calling the phrase number + X. Now it's a question of how much people are willing to translate. 

Having an inaccurate context and avoiding unreasonably large numbers, is the only limiting factor now.  Many print-outs use small number quantities and multiple voting plugins have percent of players.

stickz:
Red is just Red, when the color is in a menu.

PR #794, opened 2018-03-30
Distribute common color phrases

stickz:
Did I miss anything common? Things like cyan or magenta which are textures need a separate file.

This is as simple as loading color.phrases and `PrintToChat(client, "My favorite color is %t.", "blue");`

### DIFF
--- a/translations/color.phrases.txt
+++ b/translations/color.phrases.txt
@@ -1,0 +1,55 @@
+"Phrases"
+{
+	"red"
+	{
+		"en"		"red"
+	}
+	"green"
+	{
+		"en"		"green"
+	}
+	"blue"
+	{
+		"en"		"blue"
+	}
+	"yellow"
+	{
+		"en"		"yellow"	
+	}
+	"orange"
+	{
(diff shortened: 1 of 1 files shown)


SlidyBat:
Why not have translations for every word in the dictionary?
I feel this isn't really within the scope of sourcemod, and I don't see this being used often to warrant it  becoming a part of sourcemod. If you feel this is needed for plugins you make then you should ship the translation files with your plugins.

stickz:
@SlidyBat This can be used in hundreds of sourcemod plugins. Anything from changing chat colors, to displaying trails, paintballs, changing model colors etc. I can't say the same for other dictionary words.

stickz:
If people like this concept, I can enumerate all the colors and match them with a RGB color code. Then when plugin developers are creating a menu, they can just flip through arrays and add them all at once. It will be pure laziness if plugins do not support this new method for translating and displaying colors.

404UserNotFound:
Why not release this as a custom include/translation file package in the Snippets/Tutorials section on the forums?

PR #803, opened 2018-04-12
Fix missing include.

dvander:
Mac build complains about abort(), for me.


peace-maker:
That's in a69d18ae1c394c42bda6db3d6dad7fba077d204f now.

PR #807, opened 2018-04-29
Modified rtv message format to check if vote count = 1. (Votes -> Vote)

JoinedSenses:


JoinedSenses:
Very minor and probably unnecessary change.

asherkin:
This isn't really worth a translation break to special-case - the way to go is for the translation system to handle pluralisation internally.

PR #808, opened 2018-05-01
Update Function Offsets For PVKII

Spirrwell:
-Updates offsets
-Removes "Reload" as no corresponding function actually exists

Extra Notes:

- There does not seem to be any "Reload" function, and the offsets in this gamedata update seem to have gone up by 2, and the corresponding (Windows) offset 273+2 is CBasePlayer::RemovePlayerItem() which is probably not "Reload"

- GroundEntChanged points to a network state change. Windows and Linux were previously pointing to two different versions of this function, one that takes a void* and one that does not. They are now both the same HOWEVER I am not sure if this is desired behavior

I have attached a copy of the offset dump which should correspond to Linux offsets. (The dump was done with the Mac server.dylib binary)

[server_dump_linux.txt](https://github.com/alliedmodders/sourcemod/files/1962813/server_dump_linux.txt)

### DIFF
--- a/gamedata/sdkhooks.games/game.pvkii.txt
+++ b/gamedata/sdkhooks.games/game.pvkii.txt
@@ -32,8 +32,8 @@
 			"GroundEntChanged"
 			{
 				"windows"	"180"
-				"linux"		"181"
-				"mac"		"181"
+				"linux"		"182"
+				"mac"		"182"
 			}
 			"OnTakeDamage"
 			{
@@ -58,6 +58,12 @@
 				"windows"	"339"
 				"linux"		"340"
 				"mac"		"340"
+			}
+			"Reload"
+			{
+				"windows"	"275"
+				"linux"		"276"
+				"mac"		"276"
 			}
 			"SetTransmit"
 			{
--- a/gamedata/sdkhooks.games/game.pvkii.txt
+++ b/gamedata/sdkhooks.games/game.pvkii.txt
@@ -7,147 +7,141 @@
 		{
 			"Blocked"
 			{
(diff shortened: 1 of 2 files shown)


peace-maker:
`Reload` is a function of `CBaseWeapon` not `CBasePlayer`.

psychonic:
> GroundEntChanged points to a network state change. Windows and Linux were previously pointing to two different versions of this function, one that takes a void* and one that does not. They are now both the same HOWEVER I am not sure if this is desired behavior

For overloaded functions, the order is reversed on Windows. That script mentioned in the link you gave doesn't account for that. I believe they were both pointing to the same function before.

Spirrwell:
Ah okay. It doesn't seem like IDAPython can be run in the freeware version of IDA so I can't use that python script to dump the vtable in the format seen here https://wiki.alliedmods.net/CPVK2Player_Offset_List_(Pirates,_Vikings,_and_Knights_II)

However, I'm quite sure all offsets have just gone up two due to the additions of CBaseEntity::IsDoor and CBaseEntity::IsChest, so I could fix those offsets under that assumption.

I would still like to know about that "Reload" offset though. It's an offset of CBaseCombatWeapon not CBasePlayer, so how does SourceMod know what vtable "Reload" belongs to? Or was it an error being in the game_pvkii.txt file?

psychonic:
> how does SourceMod know what vtable "Reload" belongs to?

It's hardcoded per call/hook.

Spirrwell:
Oh, alright then. Should be all good now then.

psychonic:
Thanks!

PR #813, opened 2018-05-08
Send ShowActivity messages to TV clients

jason-e:
`ShowActivity` and related functions are used to inform all clients about the usage of admin commands (such as bans or mutes). These messages are not sent to any fake clients, however they _should_ be sent to SourceTV/GOTV fake clients because messages they receive are actually visible to broadcast viewers and in demos recorded with `tv_record`. 

Player chat messages as well as SM functions like `PrintToChat`/`PrintToChatAll` broadcast to TV clients, but `ShowActivity` does not.

This fix could also be implemented by sending these messages to all fake clients, like `PrintToChat` does.

### DIFF
--- a/core/logic/smn_players.cpp
+++ b/core/logic/smn_players.cpp
@@ -1123,7 +1123,6 @@ static cell_t _ShowActivity(IPluginContext *pContext,
 	{
 		IGamePlayer *pPlayer = playerhelpers->GetGamePlayer(i);
 		if (!pPlayer->IsInGame()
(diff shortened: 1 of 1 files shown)


asherkin:
> This fix could also be implemented by sending these messages to all fake clients, like PrintToChat does.

I think I'd prefer that.

jason-e:
Done.

KyleSanderson:
Sorry, I think what Asher meant was IsReplay or IsSTV, not the removal of all fake clients.

PR #814, opened 2018-05-09
Add Profiler methodmap

Headline:
Think this is one of the last Handle types without a method map, decided to piece together a small patch to change that.

### DIFF
--- a/plugins/include/profiler.inc
+++ b/plugins/include/profiler.inc
@@ -41,31 +41,25 @@
 
 methodmap Profiler < Handle
 {
-	/**
-	 * Creates a new profile object.  The Handle must be freed
-	 * using delete or CloseHandle().
-	 */
+	// Creates a new profile object.  The Handle must be freed
+	// using delete or CloseHandle().
+	//
+	// @return		A new Profiler Handle.
 	public native Profiler();
 	
-	/**
-	 * Starts profiling.
-	 */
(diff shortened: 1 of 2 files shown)


plugins/include/profiler.inc
```diff
@@ -38,30 +38,62 @@
...
-native Handle CreateProfiler();
+native Profiler CreateProfiler();
 
 /**
  * Starts profiling.
  *
  * @param prof		Profiling object.
  * @error			Invalid Handle.
  */
-native void StartProfiling(Handle prof);
```
peace-maker: Thinking about changes like that - we don't have to change the tag here for old natives. This change will just make old code throw tag mismatch warnings if the profiler is just stored in a Handle variable.

peace-maker: Just the other way around of expecting a `Handle` param but passing a `methodmap X < Handle`.

```sourcepawn
methodmap CoolNewThing < Handle {
};

native void DoStuff(CoolNewThing t);
native void DoMore(Handle t);

public void OnPluginStart()
{
	Handle thing;
	DoStuff(thing); // warning 213: tag mismatch
	
	CoolNewThing newThing;
	DoMore(newThing); // ok
}
```

Headline:
My bad, thought those wouldn't throw mismatches. Hopefully all is well now.

PR #818, opened 2018-05-23
Deprecate GetMaxClients

RumbleFrog:
🎉

### DIFF
--- a/plugins/include/clients.inc
+++ b/plugins/include/clients.inc
@@ -226,7 +226,7 @@ forward void OnClientPostAdminFilter(int client);
 forward void OnClientPostAdminCheck(int client);
 
 /**
- * This function will be deprecated in a future release.  Use the MaxClients variable instead.
+ * This function is deprecated. Use the MaxClients variable instead.
  *
  * Returns the maximum number of clients allowed on the server.  This may 
  * return 0 if called before OnMapStart(), and thus should not be called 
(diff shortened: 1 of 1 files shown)


plugins/include/clients.inc
```diff
@@ -226,7 +226,7 @@ forward void OnClientPostAdminFilter(int client);
 forward void OnClientPostAdminCheck(int client);
 
 /**
- * This function will be deprecated in a future release.  Use the MaxClients variable instead.
+ * This function is deprecated. Use the MaxClients variable instead.
  *
  * Returns the maximum number of clients allowed on the server.  This may
```
Headline: You can remove this whole paragraph explaining it's use, along with the @return. It's only comment should be about it's deprecation.
```
 /**
   * This function is deprecated. Use the MaxClients variable instead.
   */
```

Headline: Actually, it looks like we don't do this anywhere else. It's fine as-is.

404UserNotFound:
Yay RumbleFrog

RumbleFrog:
Done 🎉

plugins/include/clients.inc
Headline: *Note*: Apparently we don't do this anywhere else, so feel free to ignore this comment. Thanks!

404UserNotFound: I've seen a few places where deprecation comment blocks exist. Let me take a quick gander and pull some examples up.

asherkin:
Thanks!

PR #824, opened 2018-06-05
Fix MySQL extension build using vs2015

peace-maker:
MySQL defines its own timespec_t. Official mysql builds are built using VS2013, so you still need to compile the client library yourself to be able to link it, but this change makes the dbi extension compatible with future versions already.


psychonic:
Thanks!

PR #825, opened 2018-06-05
MySQL: Support multiple result sets in prepared queries

### LINKED ISSUE #823: MySQL support for multiple result set from Stored Procedures
# Help us help you

# Environment
  * Operating System version: Ubuntu server 16.04 (But I'm pretty sure it happens on every os)
  * Game/AppID (with version if applicable): Counter-Strike: Global offensive (But I'm pretty sure it happens on every game that SourceMod Supports)
  * Current SourceMod version: 1.8.0
  * Current SourceMod snapshot: 6041
  * Current Metamod: Source snapshot: 1.10.7

# Description
Like I say in the title, I'm needing the support of get results from MySQL DB Stored Procedures.
I asked first on https://forums.alliedmods.net/showthread.php?p=2595241#post2595241 and asherkin told me the problem is in the driver.

> The actual driver supports a single result set per query and CALL always returns multiple so does not work.

In the post, I say I would like to help. But I'm not too familiar with C ++ syntax but I have experience with Java.
If there is something I can do please tell me.

Stored Procedure: 
```
`sp_load_characters`(steam_id INT)
BEGIN
	SELECT c.id, c.name, c.level, c.reset 
    FROM Players p RIGHT JOIN Characters c ON (p.id = c.idPlayer) 
    WHERE p.steamid = steam_id;
END
```

Code: 

```
 int steamid = GetSteamAccountID(client);
char error[255];
	Database db = SQL_DefConnect(error, sizeof(error));
	DBStatement hUserStmt = null;
	if (db == null){
		PrintToServer("Could not connect: %s", error);
	}
	
	
	/* Check if we haven't already created the statement */
	if (hUserStmt == null){
		char error2[255];
		hUserStmt = SQL_PrepareQuery(db, "CALL sp_load_characters(?)", error2, sizeof(error2));
		if (hUserStmt == null){
			PrintToServer(error2);
			
			delete hUserStmt;
			delete db;
			
			return Plugin_Handled;
		}
	}
	SQL_BindParamInt(hUserStmt, 0, steamid, false);
	if (!SQL_Execute(hUserStmt)){
		char errorr[255];
		SQL_GetError(hUserStmt, errorr, sizeof(errorr));
		PrintToServer(errorr);
		delete hUserStmt;
		delete db;
		
		return Plugin_Handled;
	}
	
	
	
	int rows = SQL_GetRowCount(query); 
	PrintToServer("%d", rows); //returns 0
	
	delete hUserStmt;
	delete db;
	return Plugin_Handled;
```

# Logs
No logs, just no resultset.

peace-maker:
Prepared statements can return multiple result sets in MySQL 5.5. That can happen when calling stored procedures using `CALL x();`.

This change removes the previous caching of result bindings, since the number of fields in a result can differ from result set to result set. This could potentially have a negative impact on performance of prepared statements always only returning one result set, since the result binding buffers are recreated everytime the statement is executed instead of once. That difference should be negligible though.

Fixes #823.

Usage is just like with normal queries:
```sourcepawn
// From https://dev.mysql.com/doc/refman/5.5/en/c-api-prepared-call-statements.html
public void OnPluginStart()
{
	char error[255];
	Database db = SQL_DefConnect(error, sizeof(error));
	if (db == null){
		PrintToServer("Could not connect: %s", error);
		return;
	}

	/*
	CREATE PROCEDURE p1(
	  IN p_in INT,
	  OUT p_out INT,
	  INOUT p_inout INT)
	BEGIN
	  SELECT p_in, p_out, p_inout;
	  SET p_in = 100, p_out = 200, p_inout = 300;
	  SELECT p_in, p_out, p_inout;
	END
	*/

	DBStatement hStmt = SQL_PrepareQuery(db, "CALL p1(?, ?, ?)", error, sizeof(error));
	if (!hStmt){
		PrintToServer(error);
		delete db;
		return;
	}
	
	/* assign values to parameters and execute statement */
	hStmt.BindInt(0, 10); /* p_in */
	hStmt.BindInt(1, 20); /* p_out */
	hStmt.BindInt(2, 30); /* p_inout */

	if (!SQL_Execute(hStmt)){
		SQL_GetError(hStmt, error, sizeof(error));
		PrintToServer(error);
		delete hStmt;
		delete db;
		return;
	}
	
	do 
	{
		int rows = SQL_GetRowCount(hStmt); 
		PrintToServer("Num rows: %d", rows);

		int num_fields = SQL_GetFieldCount(hStmt);
		while (SQL_FetchRow(hStmt))
		{
			for (int i = 0; i < num_fields; i++)
			{
				if (SQL_IsFieldNull(hStmt, i))
					PrintToServer("val[%d] = NULL", i);
				else
					PrintToServer("val[%d] = %d", i, SQL_FetchInt(hStmt, i));
			}
		}
	}
	while (SQL_FetchMoreResults(hStmt));

	delete hStmt;
	delete db;
}
```

Output:
```
Num rows: 1
val[0] = 10
val[1] = NULL
val[2] = 30
Num rows: 1
val[0] = 100
val[1] = 200
val[2] = 300
Num rows: 1
val[0] = 200
val[1] = 300
```

### DIFF
--- a/extensions/mysql/mysql/MyBoundResults.cpp
+++ b/extensions/mysql/mysql/MyBoundResults.cpp
@@ -69,18 +69,13 @@ enum_field_types GetTheirType(DBType type)
 	return MYSQL_TYPE_STRING;
 }
 
-MyBoundResults::MyBoundResults(MYSQL_STMT *stmt, MYSQL_RES *res)
-: m_stmt(stmt), m_pRes(res), m_Initialized(false), m_RowCount(0), m_CurRow(0)
+MyBoundResults::MyBoundResults(MYSQL_STMT *stmt, MYSQL_RES *res, unsigned int num_fields)
+: m_stmt(stmt), m_pRes(res), m_ColCount(num_fields), m_Initialized(false), m_RowCount(0), m_CurRow(0)
 {
 	/**
-	 * Important things to note here: 
-	 * 1) We're guaranteed at least one field.
-	 * 2) The field information should never change, and thus we
-	 *    never rebuild it.  If someone ALTERs the table during
-	 *    a prepared query's lifetime, it's their own death.
+	 * Important thing to note here: 
+	 * We're guaranteed at least one field.
 	 */
-
-	m_ColCount = (unsigned int)mysql_num_fields(m_pRes);
 
 	/* Allocate buffers */
 	m_bind = (MYSQL_BIND *)malloc(sizeof(MYSQL_BIND) * m_ColCount);
--- a/extensions/mysql/mysql/MyBoundResults.h
+++ b/extensions/mysql/mysql/MyBoundResults.h
@@ -55,7 +55,7 @@ class MyBoundResults :
 {
 	friend class MyStatement;
 public:
-	MyBoundResults(MYSQL_STMT *stmt, MYSQL_RES *res);
+	MyBoundResults(MYSQL_STMT *stmt, MYSQL_RES *res, unsigned int num_fields);
 	~MyBoundResults();
 public: //IResultSet
 	unsigned int GetRowCount();
--- a/extensions/mysql/mysql/MyStatement.cpp
+++ b/extensions/mysql/mysql/MyStatement.cpp
@@ -33,7 +33,7 @@
 #include "MyBoundResults.h"
 
 MyStatement::MyStatement(MyDatabase *db, MYSQL_STMT *stmt)
-: m_mysql(db->m_mysql), m_pParent(db), m_stmt(stmt), m_rs(NULL), m_Results(false)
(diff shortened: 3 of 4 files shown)


Codes97:
@asherkin When will we be able to use this? (since which version of sourcemod)

asherkin:
@Codes97 1.10 build 6287

Codes97:
@asherkin Sorry I still bothering, but which stable build would it be? And how's the process? how many dev builds do you do to make a stable one?It depends on changes? testing?

Bara:
1.10 isn't the current dev-Version ( that's [1.9](http://www.sourcemod.net/downloads.php?branch=dev) ). So it could take some months. If you want that build you can [download it here](http://www.sourcemod.net/downloads.php?branch=master) but only available for windows.

Codes97:
Thanks Bara, but I need it for linux, no problem, I'll wait

PR #835, opened 2018-06-27
Update sm_map where args == 0 to display menu

JoinedSenses:
If sm_map args == 0, display menu instead of returning "sm_map <map>" message to client.

### DIFF
--- a/plugins/basecommands/map.sp
+++ b/plugins/basecommands/map.sp
@@ -85,7 +85,14 @@ public Action Command_Map(int client, int args)
 {
 	if (args < 1)
 	{
-		g_MapList.Display(client, MENU_TIME_FOREVER);
+		if (client != 0)
+		{
+			g_MapList.Display(client, MENU_TIME_FOREVER);
+		}
+		else
+		{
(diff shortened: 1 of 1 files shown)


Headline:
If we were to go with this idea, should we also consider adding similar behaviors to commands like `sm_ban`, `sm_kick`, etc?

plugins/basecommands/map.sp
```diff
@@ -85,7 +85,7 @@ public Action Command_Map(int client, int args)
 {
 	if (args < 1)
 	{
-		ReplyToCommand(client, "[SM] Usage: sm_map <map>");
+		g_MapList.Display(client, MENU_TIME_FOREVER);
```
Headline: this will throw invalid client index 0 if console tries to sm_map

JoinedSenses:
I've modified the behavior again to consider client index 0. Personally, I think it would be useful to implement a menu to appear when args == 0 for kick/ban, however the message shown for how to format the ban arguments is useful to know (sm_ban playerName timeInMinutes "Reason"). Another idea is to display both the message and the menu if args == 0 for sm_ban

PR #836, opened 2018-06-27
Update sm_ban where args == 0 to display menu

JoinedSenses:
A modification to sm_ban to display the ban menu to a client when args == 0.


KyleSanderson:
Use one PR.

PR #838, opened 2018-06-27
Update to sm_ban, sm_kick, & sm_map where args == 0 to display menu

JoinedSenses:


plugins/basebans/ban.sp
```diff
@@ -267,6 +267,10 @@ public Action Command_Ban(int client, int args)
 {
 	if (args < 2)
 	{
+		if (client != 0 && args == 0)
+		{
+			DisplayBanTargetMenu(client);
+		}
 		ReplyToCommand(client, "[SM] Usage: sm_ban <#userid|name> <minutes|0> [reason]");
```
Headline: Is it intended to error w/ usage when an argument-less `sm_ban` would now be considered valid? 

ditto for the other commands, as well

JoinedSenses: I left in the replytocommand as a reminder of the usage just in case it's needed and also as a fallback where client ==0.

Headline: I don't think I'm alone in thinking that we probably shouldn't print usage errors if it's gonna be valid. Other than that this lgtm

KyleSanderson:
At the very least if we're even going to consider merging something similar to this (it should be a convar) the cmdsrc needs to be evaluated.

Benoist3012:
And what if I'm a newbie to SM and wanna know the params of sm_ban how do I check that since you will literally slam a menu I don't want in my face. I suggest making plural version of these commands (sm_bans, sm_maps, sm_kicks ect...) doing what you suggested.

JoinedSenses:
Haha "slam a menu"? That's a bit dramatic. The method I used to add the menus will still post the usage parameters to chat/console, and will also display the menu. I'm not sure what cmdsrc is, and by cvar, do you mean adding an option to enable/disable the menus from displaying when args == 0?

plugins/basebans/ban.sp
JoinedSenses: If sm_map and sm_kick were to at least be consistent with sm_nominate, then I suppose the usage parameter return could be removed.

JoinedSenses:
Also, see above conversation with Headline regarding consistency with sm_nominate

JoinedSenses:
On top of everything else, I have now also renamed the translation file for basecommands to be consistent with every other translation file (plugin.basecommands -> basecommands.phrases)

asherkin:
I'd be moderately OK (but still not a fan) of doing this for chat commands, but console usage of these should definitely not be opening a menu.

Is there a driving force behind this? PRs with no description are rather annoying (at the very least, every feature addition needs a use case.)

JoinedSenses:
This change makes it so /kick, /ban, and /map open the already created methods for displaying their menus. 

The reason for the add is to take advantage of menus that already exist and to make the commands easier to use.

The client == 0 check prevents them from opening if it was ran via rcon, sm_rcon, or server command. Client auth is also checked because its a registered admin command.

Usage params will also still display for each method.

plugins/basebans/ban.sp
Headline: I think that's best. It'd be annoying if users typed `!ban Headline 0 "lemon"` and got hit with usage errors every time.

JoinedSenses:
The latest change requested from Headline will now only display usage params if client == 0 && args == 0, or if args is greater than 0 and less than 2 for sm_ban

JoinedSenses:
With these changes, the menu does only show on 0 args. No idea what you mean by the reply source? What sources should be checked for ans how? It currently runs a check against client == 0, which would occur on rcon, sm_rcon, and servercommand.

JoinedSenses:
My previous comment was regarding usage params, not the menu.

JoinedSenses:
New commit, which now checks both reply source and client. Tested - useage params always display when issued via console and menu would not display.

Menu will only display if there are 0 arguments, client != 0, and the command was sent in chat.

asherkin:
Thanks! Nice work.

PR #847, opened 2018-07-07
Changed "entiy" to "entity" in IsEntNetworkable

Psykotikism:


plugins/include/entity.inc
```diff
@@ -125,7 +125,7 @@ native bool IsValidEdict(int edict);
  * @param edict			Index of the entity.
```
50Wliu: The param name should also be changed.

psychonic:
Thanks!

PR #848, opened 2018-07-09
Match old DataPack behavior when overwriting data

Headline:
```cpp
DataPack dp = new DataPack();
dp.WriteCell(3);
dp.Reset();
dp.WriteCell(4);
delete dp;
```
Instead of the data being overwritten, the current behavior inserts 4 before 3. This is a problem because before #688 DataPacks would overwrite. This adds parameters to every native, default false, which determines whether the data should be inserted or overwritten that way old plugins wont break, and if someone really wants to insert they may.

### DIFF
--- a/core/logic/CDataPack.cpp
+++ b/core/logic/CDataPack.cpp
@@ -230,6 +230,15 @@ bool CDataPack::RemoveItem(size_t pos)
 	{
 		return false;
 	}
+	if (pos == 0)
+	{
+		if (position == elements.length())
+		{
+			return false;
+		}
+		elements.remove(position--);
+		return true;
+	}
 
 	elements.remove(pos-1);
 	--position;
--- a/core/logic/CDataPack.cpp
+++ b/core/logic/CDataPack.cpp
@@ -235,7 +235,7 @@ bool CDataPack::RemoveItem(size_t pos)
 	{
 		pos = position;
 	}
-	else if (pos >= elements.length())
+	else if (pos > elements.length())
 	{
 		return false;
 	}
--- a/core/logic/CDataPack.cpp
+++ b/core/logic/CDataPack.cpp
@@ -224,7 +224,15 @@ void *CDataPack::ReadMemory(size_t *size) const
 	return ptr;
 }
 
-void CDataPack::RemoveNextItem()
+void CDataPack::RemoveItem(size_t pos)
 {
-	elements.remove(position--);
+	if (pos < 0)
+	{
+		elements.remove(position--);
+		return;
+	}
+	
+	assert(pos > 0 && pos <= elements.length());
+	elements.remove(pos-1);
+	--position;
 }
--- a/core/logic/CDataPack.cpp
+++ b/core/logic/CDataPack.cpp
@@ -226,21 +226,21 @@ void *CDataPack::ReadMemory(size_t *size) const
 
 bool CDataPack::RemoveItem(size_t pos)
 {
-	if (pos > elements.length())
+	if (!elements.length())
 	{
 		return false;
 	}
-	if (pos == 0)
+
+	if (pos == static_cast<size_t>(-1))
 	{
-		if (position == elements.length())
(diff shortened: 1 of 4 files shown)


asherkin:
Does this affect 1.9 and 1.10?

Headline:
only 1.10 has the behavior change

asherkin:
I'm not a fan of `RemoveNextItem`, I think the add/overwrite behaviour should live on `CDataPack` directly, thoughts?

Also random ping of @Drifter321 as I see #688 completely nuked the public interface, which CSS:DM use(s|d).

Drifter321:
I totally forgot CSS:DM used that interface... welp.

Headline:
@asherkin if we add overwrite parameters to CDataPack write methods then we'd have to change IDataPack and it's a bit more invasive. I made a trial run of that change [here](https://github.com/Headline/sourcemod/pull/15/files)

KyleSanderson:
Syncing up from IRC during early minutes of the PR. What I think we should do is add RemoveItem and give it an element param that defaults to -1 which deletes the current element in the pack. Modifying the cpp iface is brutal for the aforementioned reasons.

Headline:
I've implemented KyleS's suggestion which hopefully makes RemoveItem's addition to CDataPack feel more appropriate.

core/logic/CDataPack.cpp
```diff
@@ -223,3 +223,16 @@ void *CDataPack::ReadMemory(size_t *size) const
 
 	return ptr;
 }
+
+void CDataPack::RemoveItem(size_t pos)
+{
+	if (pos < 0)
```
asherkin: `size_t` is unsigned, this shouldn't build (but does appear to - I'm guessing we're ignoring the warning).

core/logic/CDataPack.cpp
```diff
@@ -223,3 +223,16 @@ void *CDataPack::ReadMemory(size_t *size) const
...
+
+void CDataPack::RemoveItem(size_t pos)
+{
+	if (pos < 0)
+	{
+		elements.remove(position--);
+		return;
+	}
+	
+	assert(pos > 0 && pos <= elements.length());
```
asherkin: Should probably have this return false rather than asserting, it is public API after all.

core/logic/smn_datapacks.cpp
```diff
@@ -99,6 +99,12 @@ static cell_t smn_WritePackCell(IPluginContext *pContext, const cell_t *params)
 		return pContext->ThrowNativeError("Invalid data pack handle %x (error %d).", hndl, herr);
 	}
 
+	bool overwrite = (params[0] >= 3) ? params[3] : true;
+	if (overwrite && pDataPack->GetPosition() != pDataPack->GetCapacity())
```
asherkin: And then you get to drop the sanity check here and in all the other funcs.

plugins/include/datapack.inc
```diff
@@ -49,23 +49,27 @@ methodmap DataPack < Handle
 
 	// Packs a normal cell into a data pack.
 	//
-	// @param cell		Cell to add.
-	public native void WriteCell(any cell);
+	// @param cell			Cell to add.
+	// @param overwrite		Whether existing data should be overwritten.
+	public native void WriteCell(any cell, bool overwrite = true);
```
asherkin: As KyleS mentioned on IRC, I think default-false params are a bit easier for people to wrap their head around, so lets go with `insert` here and flip the behaviour?

Headline:
@asherkin didn't follow the review exactly to spec, but instead of the implied behavior we'll just leave RemoveItem to deal with PackPos *only*, and the logic for whether an item is removed will exist in the native definition if that's better.

KyleSanderson:
Please test my modifications when you have a moment to ensure the functionality is the same. We should be okay to land this afterwards.

KyleSanderson:
Thanks bud.

PR #850, opened 2018-07-16
Fix memory leak regression from #848.

KyleSanderson:
Agh. We focused a little too heavily on the regression from https://github.com/alliedmodders/sourcemod/pull/688 and missed a memory leak from not properly freeing dynamic objects when they're overwritten in https://github.com/alliedmodders/sourcemod/pull/848. All good, shower thoughts drove this out.

### DIFF
--- a/core/logic/CDataPack.cpp
+++ b/core/logic/CDataPack.cpp
@@ -225,7 +225,7 @@ bool CDataPack::RemoveItem(size_t pos)
 		return false;
 	}
 
-	switch (elements[index].type)
+	switch (elements[pos].type)
 	{
 		case CDataPackType::Raw:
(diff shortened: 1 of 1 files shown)


core/logic/CDataPack.cpp
```diff
@@ -65,25 +65,9 @@ CDataPack::Free(IDataPack *pack)
...
-
-			case CDataPackType::String:
-			{
-				delete elements[index].pData.sval;
-				elements.remove(index--);
-				break;
-			}
-		}
-	}
+	} while (this->RemoveItem());
```
Headline: nit: `while (this->RemoveItem()) {};`
<3

PR #852, opened 2018-07-18
Add Clone() implementation for StringMap

### LINKED ISSUE #796: Feature Request: CloneTrie() and CloneStack()
We have [CloneArray()](https://sm.alliedmods.net/new-api/adt_array/CloneArray), but we don't have `CloneTrie()` and `CloneStack()`. Why?

Headline:
We implemented a clone for `ArrayList`, but `ArrayStack` and `StringMap` were left without one. This pull request adds deep-copy cloning to said structures. Construct generally looks like
```cpp
ArrayStack mystack = // something, lets assume it also has data
ArrayStack copy = mystack.Clone();

delete mystack;
delete copy;
```

Fixes #796

### DIFF
--- a/core/logic/smn_adt_trie.cpp
+++ b/core/logic/smn_adt_trie.cpp
@@ -658,7 +658,7 @@ static cell_t CloneTrie(IPluginContext *pContext, const cell_t *params)
 
 	for (StringHashMap<Entry>::iterator it = pOldTrie->map.iter(); !it.empty(); it.next())
 	{
-		const char *key = it->key.chars();
+		const char *key = it->key.c_str();
 
 		StringHashMap<Entry>::Insert insert = pNewTrie->map.findForAdd(key);
 		if (pNewTrie->map.add(insert, key))
@@ -670,7 +670,7 @@ static cell_t CloneTrie(IPluginContext *pContext, const cell_t *params)
 			}
 			else if (result->value.isString())
 			{
-				insert->value.setString(result->value.chars());
+				insert->value.setString(result->value.c_str());
 			}
 			else if (result->value.isArray())
 			{
--- a/core/logic/smn_adt_trie.cpp
+++ b/core/logic/smn_adt_trie.cpp
@@ -636,6 +636,57 @@ static cell_t GetTrieSnapshotKey(IPluginContext *pContext, const cell_t *params)
 	return written;
 }
 
+static cell_t CloneTrie(IPluginContext *pContext, const cell_t *params)
+{
+	HandleError err;
+	HandleSecurity sec = HandleSecurity(pContext->GetIdentity(), g_pCoreIdent);
+
+	CellTrie *pOldTrie;
+	if ((err = handlesys->ReadHandle(params[1], htCellTrie, &sec, (void **)&pOldTrie))
+		!= HandleError_None)
+	{
+		return pContext->ThrowNativeError("Invalid Handle %x (error %d)", params[1], err);
+	}
+
+	CellTrie *pNewTrie = new CellTrie;
+	Handle_t hndl = handlesys->CreateHandle(htCellTrie, pNewTrie, pContext->GetIdentity(), g_pCoreIdent, NULL);
+	if (!hndl)
+	{
+		delete pNewTrie;
+		return hndl;
+	}
+
+	for (StringHashMap<Entry>::iterator it = pOldTrie->map.iter(); !it.empty(); it.next())
+	{
+		const char *key = it->key.chars();
+
+		StringHashMap<Entry>::Insert insert = pNewTrie->map.findForAdd(key);
+		if (pNewTrie->map.add(insert, key))
+		{
+			StringHashMap<Entry>::Result result = pOldTrie->map.find(key);
+			if (result->value.isCell())
+			{
+				insert->value.setCell(result->value.cell());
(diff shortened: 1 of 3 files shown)


asherkin:
> This pull request adds deep-copy cloning to said structures.

The implementation is a shallow copy (which I think is all we can sanely do and matches ArrayList), was something else intended?

plugins/include/adt_stack.inc
```diff
@@ -54,6 +54,14 @@ methodmap ArrayStack < Handle
 	//                     new Array[X][32]
 	public native ArrayStack(int blocksize=1);
 
+	// Clones an array stack, returning a new handle with the same size and data.
+	// This should NOT be confused with CloneHandle. This is a completely new
+	// handle with the same data but no relation to the original. It should
```
KyleSanderson: be closed*?

core/logic/smn_adt_trie.cpp
```diff
@@ -635,6 +635,51 @@ static cell_t GetTrieSnapshotKey(IPluginContext *pContext, const cell_t *params)
...
+				insert->value.setString(result->value.chars());
+			}
+			else if (result->value.isArray())
+			{
+				insert->value.setArray(result->value.array(), result->value.arrayLength());
+			}
+		}
+	}
+
+	Handle_t hndl = handlesys->CreateHandle(htCellTrie, pNewTrie, pContext->GetIdentity(), g_pCoreIdent, NULL);
```
KyleSanderson: Create the handle before you do all that work and toss it away.

core/logic/smn_adt_trie.cpp
```diff
@@ -635,6 +635,51 @@ static cell_t GetTrieSnapshotKey(IPluginContext *pContext, const cell_t *params)
...
+	{
+		return pContext->ThrowNativeError("Invalid Handle %x (error %d)", params[1], err);
+	}
+
+	CellTrie *pNewTrie = new CellTrie;
+	for (auto it = pOldTrie->map.iter(); !it.empty(); it.next())
+	{
+		const char *key = it->key.chars();
+
+		auto result = pOldTrie->map.find(key);
```
KyleSanderson: Result isn't used if the trie can't add the key, move this into the branch.

I also really dislike auto because it subtly hides bugs.

core/logic/smn_adt_trie.cpp
```diff
@@ -635,6 +635,51 @@ static cell_t GetTrieSnapshotKey(IPluginContext *pContext, const cell_t *params)
...
+				insert->value.setCell(result->value.cell());
+			}
+			else if (result->value.isString())
+			{
+				insert->value.setString(result->value.chars());
+			}
+			else if (result->value.isArray())
+			{
+				insert->value.setArray(result->value.array(), result->value.arrayLength());
+			}
```
KyleSanderson: Add an else and throw a native error saying the type is unhandled and the user should file so we can catch bugs in the future

plugins/include/adt_trie.inc
```diff
@@ -52,6 +52,14 @@ methodmap StringMap < Handle
 	// The StringMap must be freed via delete or CloseHandle().
 	public native StringMap();
 
+	// Clones a string map, returning a new handle with the same size and data.
+	// This should NOT be confused with CloneHandle. This is a completely new
+	// handle with the same data but no relation to the original. It should
```
KyleSanderson: same be closed* quirk

public/sm_trie_tpl.h
```diff
@@ -48,7 +48,7 @@ enum NodeType
  * @brief DEPRECATED. This class scales extremely poorly; insertion scales
  * quadratic (O(n^2)) with respect to the number of elements in the table.
  * Only use this class if you have less than 200 elements or so. Otherwise,
- * use StringHashMap in sm_hashtable.h which scales linearly and has comparable
+ * use StringHashMap in sm_stringhashmap.h which scales linearly and has comparable
```
KyleSanderson: nice 🐈ch.

Headline:
@KyleSanderson I’m unsure if ArrayStack’s clone is used (or if people use ArrayStack at all), but I think being able to clone a StringMap is a really good addition into core. I can’t double check, but I don’t think it’s possible to distinguish a key’s mapped type so this couldn’t even be done native in sp, so a clone couldn’t be accomplished prior to this pr.

KyleSanderson:
Stack is used by many; cloning a stack I'm not sure on as it likely predates most of us. The trick is to jump on BZ (or now here) and looking at why something was added. Nice-to-haves are legit, just wanted some lineage.

All good.

Headline:
Ah, my bad. All should be well now.

core/logic/smn_adt_trie.cpp
```diff
@@ -635,6 +635,57 @@ static cell_t GetTrieSnapshotKey(IPluginContext *pContext, const cell_t *params)
...
+			else
+			{
+				delete pNewTrie;
+				return pContext->ThrowNativeError("Unhandled data type encountered, file a bug");
```
KyleSanderson: This needs to be way more constructive. Mention this PR directly.

peace-maker:
@Headline now that #1304 was merged, this could be reduced to only adding a `StringMap.Clone()`?

PR #854, opened 2018-07-22
Add CreateDirectory return value documentation

hydrogen-mvm:
For reference the implementation is here: https://github.com/alliedmodders/sourcemod/blob/master/core/logic/smn_filesystem.cpp#L725

Let me know if I should change the comment or something else. Thanks.


Headline:
Nice catch, thanks!

PR #855, opened 2018-07-22
Fix English Style

hydrogen-mvm:
Fixed some grammar and word choice issues. I have to write a lot of prose like this for my MvM contests and campaigns, so I gave this a good proofread, since I noticed some issues when reading it before making my first pull request.

That said, please do look over it and see if I missed any issues or (worse) introduced a mistake myself. Let me know if any changes need to be made or reverted.

Thanks.


Fyren:
Thanks for proofreading.

PR #856, opened 2018-07-22
Allow client 0 in FormatUserLogText

ddhoward:
Modifying FormatUserLogText to allow client 0 to be passed.

### DIFF
--- a/plugins/include/helpers.inc
+++ b/plugins/include/helpers.inc
@@ -45,31 +45,7 @@
  */
 stock void FormatUserLogText(int client, char[] buffer, int maxlength)
 {
-	char auth[32];
-	char name[MAX_NAME_LENGTH];
-	int userid;
-	
-	if (client)
-	{
-		userid = GetClientUserId(client);
-		if (!GetClientAuthId(client, AuthId_Engine, auth, sizeof(auth)))
-		{
-			strcopy(auth, sizeof(auth), "UNKNOWN");
-		}
-		if (!GetClientName(client, name, sizeof(name)))
-		{
-			strcopy(name, sizeof(name), "UNKNOWN");
-		}
-	}
-	else
-	{
-		strcopy(auth, sizeof(auth), "Console");
-		strcopy(name, sizeof(name), "Console");
-	}
-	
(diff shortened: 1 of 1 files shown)


plugins/include/helpers.inc
```diff
@@ -47,20 +47,29 @@ stock void FormatUserLogText(int client, char[] buffer, int maxlength)
...
+		if (!GetClientName(client, name, sizeof(name)))
+		{
+			strcopy(name, sizeof(name), "UNKNOWN");
+		}
 	}
-	if (!GetClientName(client, name, sizeof(name)))
+	else
 	{
-		strcopy(name, sizeof(name), "UNKNOWN");
+		strcopy(auth, sizeof(auth), "Console");
```
KyleSanderson: We probably want CONSOLE here? I'm not sure, do most games have it as a name and not an entire capitalized string now?

ddhoward: It's `Console` in TF2, as well as in the %L and %N explanations on the wiki:
https://wiki.alliedmods.net/Format_Class_Functions_(SourceMod_Scripting)#Format_Specifiers

I am not sure about other games.

ddhoward: Honestly, maybe the whole function should just be marked as deprecated and the body replaced with a `FormatEx(buffer, maxlength, "%L", 0);`

KyleSanderson:
Well you know what, we should just be using %N here regardless instead of filling a magical buffer.

ddhoward:
For some reason I was under the impression that %N would have the same behavior as `GetClientName()` in regards to client 0. In reality, %N results in `Console` whereas GetClientName() gives the current value of the `hostname` cvar. So you're totally right.

Regardless, is there any reason why the whole thing can't just be replaced with:
```stock void FormatUserLogText(int client, char[] buffer, int maxlength)
{
	FormatEx(buffer, maxlength, "%L", client);
}```
This should guarantee consistency.

KyleSanderson:
This may be \"%L\" but yeah this is an ancient func in comparison to what actually landed.

asherkin:
We should mark it as deprecated alongside this change I think.

PR #857, opened 2018-07-24
Allow any returns in native callbacks

Headline:
Native callbacks currently require plugin authors to retag their returns to an `int`, but by adding `any` as another acceptable prototype we allow them to return without needing to retag.

```cpp
public any Native_Something(Handle plugin, int numParams)
{
    float val =5.0;
    return val; // no tag mismatch
}
```

Thanks asherkin for bringing this change to light

### DIFF
--- a/plugins/adminmenu.sp
+++ b/plugins/adminmenu.sp
@@ -162,9 +162,9 @@ public void DefaultCategoryHandler(TopMenu topmenu,
 	}
 }
 
-public int __GetAdminTopMenu(Handle plugin, int numParams)
+public any __GetAdminTopMenu(Handle plugin, int numParams)
 {
-	return view_as<int>(hAdminMenu);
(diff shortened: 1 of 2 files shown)


KyleSanderson:
I guess we still need int in there for old code? is this correct?

Headline:
The typeset has the older int prototype, this shouldn’t break anything older. Many of our built-in plugins still use the `int` counterpart and they compile

KyleSanderson:
Why not just change it to any then? Does this give some micro-optimization?

Headline:
@KyleSanderson if we changed `int` to `any` and kept the typedef, then every callback would have to be rewritten. While int, float, char can coherse to `any`, they're still different types and the compiler errors out w/ prototype mismatch. We need that typeset(funcenum) for this change to be possible.

PR #860, opened 2018-07-26
Flip sm_debug_connect to 1

KyleSanderson:
Asher pointed out almost immediately after this landed that the logging param should be 1 by default. This fixes that, however many years later now.


asherkin:
Who reviewed this?

KyleSanderson:
Flipping the convar? Myself. Adding this functionality? It's a sec bug on bz.

PR #862, opened 2018-07-31
Fix DataPack Overwrite Regression

Headline:
Some tests passed with the implementation prior to this commit, but those were edge cases. Good older behavior is now fully restored

### DIFF
--- a/core/logic/CDataPack.cpp
+++ b/core/logic/CDataPack.cpp
@@ -241,6 +241,5 @@ bool CDataPack::RemoveItem(size_t pos)
 	}
 
 	elements.remove(pos);
-	--position;
 	return true;
 }
--- a/core/logic/CDataPack.cpp
+++ b/core/logic/CDataPack.cpp
@@ -219,10 +219,14 @@ bool CDataPack::RemoveItem(size_t pos)
 	{
 		pos = position;
 	}
-
 	if (pos >= elements.length())
 	{
 		return false;
+	}
+
+	if (pos < position) // we're deleting under us, step back
+	{
+		position--;
 	}
 
 	switch (elements[pos].type)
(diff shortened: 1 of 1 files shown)


core/logic/CDataPack.cpp
```diff
@@ -219,12 +219,16 @@ bool CDataPack::RemoveItem(size_t pos)
 	{
 		pos = position;
 	}
-
 	if (pos >= elements.length())
 	{
 		return false;
 	}
 
+	if (pos < position) // we're deleting under us, step back
```
asherkin: Should this not be `<=`? My brain has trouble following the logic here so I can't be sure.

Headline: if `pos = position`, we're still deleting in front of us. (`position` is the index to the *next* read/write index in the array.

core/logic/CDataPack.cpp
```diff
@@ -219,12 +219,16 @@ bool CDataPack::RemoveItem(size_t pos)
...
 	}
-
 	if (pos >= elements.length())
 	{
 		return false;
 	}
 
+	if (pos < position) // we're deleting under us, step back
+	{
+		position--;
```
KyleSanderson: pre to post?

core/logic/CDataPack.cpp
KyleSanderson: why not just set this to set position to pos?

Headline: if we’re deleting far behind us, we shouldn’t  change our pack position. Only scoot us down to account for it.

if cell func string was written, and we deleted cell, our position shouldn’t reset to be 0 and we should still be after the func write

core/logic/CDataPack.cpp
Headline: no comprendo

PR #685, opened 2017-09-17
Add LogStackTrace Native (#684)

Headline:
In order for this to be merged, [this pr](https://github.com/alliedmodders/sourcepawn/pull/140) for sourcepawn must be merged. 

Essentially this adds pull request adds LogStackTrace, which will serve as a debugging function that outputs a stack trace where the native is called, but does not halt execution. For more info: see #684

```cpp
#include <sourcemod>

public void OnPluginStart()
{
	SomeFunction();
}

void SomeFunction()
{
	SomeFunction2();
}

void SomeFunction2()
{
	LogStackTrace("LogStackTrace%i %.2f", 4, 10.0)
	SetFailState("SetFailState %i %.2f", 4, 10.0)
}
```
will output in errors

```
L 09/17/2017 - 15:07:55: [SM] Stack trace requested: LogStackTrace 4 10.00
L 09/17/2017 - 15:07:55: [SM] Call stack trace:
L 09/17/2017 - 15:07:55: [SM]   [0] LogStackTrace
L 09/17/2017 - 15:07:55: [SM]   [1] Line 15, C:\Users\Micha\Desktop\sourcemod\sourcemod\build\package\addons\sourcemod\Test.sp::SomeFunction2
L 09/17/2017 - 15:07:55: [SM]   [2] Line 10, C:\Users\Micha\Desktop\sourcemod\sourcemod\build\package\addons\sourcemod\Test.sp::SomeFunction
L 09/17/2017 - 15:07:55: [SM]   [3] Line 5, C:\Users\Micha\Desktop\sourcemod\sourcemod\build\package\addons\sourcemod\Test.sp::OnPluginStart
L 09/17/2017 - 15:07:55: [SM] Exception reported: SetFailState 4 10.00
L 09/17/2017 - 15:07:55: [SM] Blaming: test.smx
L 09/17/2017 - 15:07:55: [SM] Call stack trace:
L 09/17/2017 - 15:07:55: [SM]   [0] SetFailState
L 09/17/2017 - 15:07:55: [SM]   [1] Line 16, C:\Users\Micha\Desktop\sourcemod\sourcemod\build\package\addons\sourcemod\Test.sp::SomeFunction2
L 09/17/2017 - 15:07:55: [SM]   [2] Line 10, C:\Users\Micha\Desktop\sourcemod\sourcemod\build\package\addons\sourcemod\Test.sp::SomeFunction
[...]

PR #863, opened 2018-07-31
Fix LogStackTrace Anomaly

Headline:
These changes should've been included in #685, but it appears my git client didn't stage them for commit. Not sure specifically what happened but this has already been r'd, re-ran tests and all still checkout okay. We should also cherry-pick this to 1.9.

### DIFF
--- a/plugins/include/sourcemod.inc
+++ b/plugins/include/sourcemod.inc
@@ -313,6 +313,16 @@ native void SetFailState(const char[] string, any ...);
 native void ThrowError(const char[] fmt, any ...);
 
 /**
+ * Logs a stack trace from the current function call. Code 
+ * execution continues after the call
+ *
+ * @param fmt		String format.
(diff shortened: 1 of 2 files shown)


plugins/include/sourcemod.inc
```diff
@@ -312,6 +312,16 @@ native void SetFailState(const char[] string, any ...);
  */
 native void ThrowError(const char[] fmt, any ...);
 
+/**
+ * Logs a stack trace from the current function call. Code 
+ * execution continues after the call
+ *
+ * @param fmt		String format.
+ * @param ...		Format arguments.
```
asherkin: It'd be nice to note what these args are actually for (I'm guessing just contextual information for the log?)

Headline:
oh; I thought this landed before 1.10, I was wrong. No cherry-pick needed 🥂

PR #864, opened 2018-07-31
Burninate IDataPack

asherkin:
This doesn't break any extensions NOT using IDataPack, and we do not know of any that are.

* The extension storage utility of this interface has been broken for the last 9 months, with ISourceMod::CreateDataPack being disabled.
* The plugin interop utility of this interface (its stated purpose) has been broken for the last 11+ years, with ISourceMod::GetDataPackHandleType being disabled.

I imagine it only survived the first cleanup 11 years ago because CSS:DM was using it internally, which it has now been migrated away from.

Compiled all the included extensions without changes (API compat), and loaded extensions built pre-change without issue (ABI compat).


peace-maker:
You could copy the documentation from IDataPack.h to the CDataPack header.

bridge/include/BridgeAPI.h
```diff
@@ -35,7 +35,7 @@ namespace SourceMod {
 
 // Add 1 to the RHS of this expression to bump the intercom file
 // This is to prevent mismatching core/logic binaries
-static const uint32_t SM_LOGIC_MAGIC = 0x0F47C0DE - 56;
+static const uint32_t SM_LOGIC_MAGIC = 0x0F47C0DE - 57;
```
Headline: If we fwd-decl'd `IDataPack `in the SourceMod namespace and didn't change the iface declarations of `CreateDataPack `and `FreeDataPack` in sourcemod.h, could we prevent bumping this?

asherkin: The bump here is because I've removed the functions from the bridge (which were only being used by ISourceMod), we could just leave them in the bridge (even just as null ptrs) to avoid bumping this, but logic bridge iface bumps are not a problem, it is just to prevent against accidents.

Headline: ah; makes sense. 👍

dvander:
:+1:

PR #866, opened 2018-08-04
Individualize NameHashSet Hashing & Revisit #709 (#740) (1.9-dev)

Headline:
*this is a merge into 1.9-dev, not master*

This is a clone of #740, but without the amtl ke::AString lowercase which was implemented in a new version of amtl that 1.9-dev isn't pinned to. Updating this pin and moving fixes is beyond what should go in 1.9, and this fixes a annoying and user-impactful bug with reload/unloading plugins on windows.

Currently in 1.9, once a plugin is loaded into the pluginsys, they must be used with lowercase characters *only*, since pr #709 ignorantly modified their names. 

```
// test.smx exists in /plugins/
sm plugins load TEST.smx // successful
sm plugins unload TEST.smx // TEST.smx not found, it's actually test.smx
```

This pr fixes that error by converting *all* lookups, not just loads.

Edit: plugin loading tests workout great 👍


Dispenz0r:
Hopefully expecting these changes to hit the dev build soon, causing a lot of troubs for my servers.

Headline:
@Dispenz0r the temporary fix is to just use lowercase always, regardless of what the plugin's actual casing is. This only impacts windows & osx builds

PR #867, opened 2018-08-06
Mark new methodmap natives as optional

peace-maker:
The `TopMenu` methodmap natives weren't marked as optional if the extension was optional like the other normal natives.

The `Regex` natives weren't marked as optional at all before if the regex extension was included optionally.

This makes the error message cleaner in case topmenus aren't loaded and won't prevent plugins from loading if they're supposed to have an optional dependency.
```
adminmenu.smx (Admin Menu): Required extension "TopMenus" file("topmenus.ext") not running
basebans.smx (Basic Ban Commands): Native "TopMenu.Display" was not found
```

### DIFF
--- a/plugins/include/regex.inc
+++ b/plugins/include/regex.inc
@@ -273,3 +273,19 @@ public Extension __ext_regex =
 	required = 0,
 #endif
 };
+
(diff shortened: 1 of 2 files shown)


asherkin:
Thanks!

PR #868, opened 2018-08-16
Fix SetClanTag CS:GO Windows Signature

Headline:
👍


Benoist3012:
Dumb question why not wildcard this byte instead, if it changed, it may change again again (x2) later? Or does that make the signature no longer unique?

asherkin:
> Edit: @Benoist3012 It'd make the signature not unique, so a wildcard wouldn't work.

That probably means the signature should be a bit longer anyway, something to think about next time it needs updating anyway.

PR #869, opened 2018-08-16
Add new weapons to CSWeaponID enum.

Drifter321:
It is August once again and its time for valve to add new weapons and change how we handle weapon id's once again. Luckily the previous changes seem to have helped. However, they have added a new weapon (weapon_mp5sd) in the middle of the item definitions instead of at the end (itemdef 23). What I did to fix this was change the new id to point to the CS:S equivalent for mp5 (id 18). Luckily, that wasnt used in CS:GO either so we can get away with this. I'm not sure if there is a better approach for this or not. 

One thing i noticed however, is that SMCSWeapon_G3SG1 (id 23, itemdef 11) is already in the game so this can get confusing if passing the wrong id/itemdef around to natives (pass the itemdef to one expecting an id and you get the g3sg1). 

My other concern is if they add more weapons in the middle if we will clash at somepoint (i dont think this is possible but throwing it out there) The only case i can see this happening is if they actually add the mp5navy we will then not have an id for it since its used by this one. Another is they can add one in the middle for a weapon we never had (let say they added an rpg at 22) we wouldnt be able to map it to an id unless we created our own id.

Unfortunately, this change can also break plugins (Weapon restrict for example... again :rage:) I used a map of weaponid to classname the classname ofc, doesnt match anymore since they use different classnames (weapon_mp5navy, weapon_mp5sd).

Open to feedback, suggestions, or whatever.


KyleSanderson:
Why not make these const strings and do a trie lookup? that way we can avoid this weird index issue?

Drifter321:
Not sure I understand, change what to const strings? The id issue will still persist since I cant change how that sutff works (without breaking compat).

KyleSanderson:
New API, consts would bolt onto strings. Alternatively, doesn't TF have a stable mechanism for something similar using pubvars?

PR #870, opened 2018-08-18
Remove `MAPLIST_FLAG_NO_DEFAULT` misuses

asherkin:
These two base plugins are using the `MAPLIST_FLAG_NO_DEFAULT` when they shouldn't, which causes much user (and SM developer) confusion as it is completely counter to the documentation.

Raised by @DoctorOcsid on Discord.

### DIFF
--- a/plugins/basecommands/map.sp
+++ b/plugins/basecommands/map.sp
@@ -133,7 +133,7 @@ int LoadMapList(Menu menu)
 	if ((map_array = ReadMapList(g_map_array,
 			g_map_serial,
 			"sm_map menu",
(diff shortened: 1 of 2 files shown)


asherkin:
These 2 currently have a separate file specified: https://github.com/alliedmodders/sourcemod/blob/17700e45b0f8984d06e3d1773fb90f3d5fca7396/configs/maplists.cfg#L30-L40

The problem is that upon removal of that configuration (or if that file can't be loaded), they fall back to loading all the maps rather than the documented behaviour of try "default" then "mapcyclefile".

PR #873, opened 2018-08-21
Restore SQL_CheckConfig Behavior

### LINKED ISSUE #865: SQL_CheckConfig always returns true (dev)
```
    SourceMod Version: 1.10.0.6322
    SourcePawn Engine: 1.10.0.6322, jit-x86 (build 1.10.0.6322)
    SourcePawn API: v1 = 4, v2 = 12
    Compiled on: Aug  2 2018 14:34:15
    Built from: https://github.com/alliedmodders/sourcemod/commit/3511d2a4
    Build ID: 6322:3511d2a4
    http://www.sourcemod.net/
```

```
#include <sourcemod>
#include <dbi>

public void OnPluginStart()
{
    SetFailState( "Result: %i", SQL_CheckConfig( "idontexist" ) ); // Result: 1
}
```

Headline:
~~This is a minimal change, more can be done.~~

Basically, `SQL_CheckConfig` used to fallback to "default" configuration and return true. This is different from old behavior and would break older plugins. Contrary to what we thought, Database connections doesn't fall-back to default configuration.

~~Because of this, I'm not opposed to removing the default parameter in SQL_CheckConfig entirely and just make it always pass true into the FindDatabaseConf method (which now disables default fallback), but I decided to throw this up here and get the conversation started~~

Disregarding the strategy above, we'll just do a list lookup directly and discard any fall through, as connect currently doesn't fall through either.

Fixes #865

### DIFF
--- a/core/logic/smn_database.cpp
+++ b/core/logic/smn_database.cpp
@@ -1425,7 +1425,7 @@ static cell_t SQL_CheckConfig(IPluginContext *pContext, const cell_t *params)
 	char *name;
 	pContext->LocalToString(params[1], &name);
 
-	return (g_DBMan.FindDatabaseConf(name) != NULL) ? 1 : 0;
+	return (g_DBMan.GetDatabaseConf(name) != nullptr) ? 1 : 0;
 }
 
 static cell_t SQL_ConnectCustom(IPluginContext *pContext, const cell_t *params)
--- a/core/logic/smn_database.cpp
+++ b/core/logic/smn_database.cpp
@@ -315,7 +315,7 @@ public:
 		strncopy(dbname, _dbname, sizeof(dbname));
 		me = scripts->FindPluginByContext(m_pFunction->GetParentContext()->GetContext());
 		
-		m_pInfo = g_DBMan.GetDatabaseConf(dbname);
+		m_pInfo = g_DBMan.FindDatabaseConf(dbname);
 		if (!m_pInfo)
 		{
 			g_pSM->Format(error, sizeof(error), "Could not find database config \"%s\"", dbname);
@@ -333,7 +333,7 @@ public:
 	{
 		if (m_pInfo)
 		{
-			m_pDatabase = m_pDriver->Connect(&m_pInfo->info, false, error, sizeof(error));
+			m_pDatabase = m_pDriver->Connect(m_pInfo, false, error, sizeof(error));
 		}
 	}
 	void CancelThinkPart()
@@ -384,7 +384,7 @@ public:
 		delete this;
 	}
 private:
-	ke::RefPtr<ConfDbInfo> m_pInfo;
+	const DatabaseInfo *m_pInfo;
 	IPlugin *me;
 	IPluginFunction *m_pFunction;
 	IDBDriver *m_pDriver;
--- a/core/logic/smn_database.cpp
+++ b/core/logic/smn_database.cpp
@@ -315,7 +315,7 @@ public:
 		strncopy(dbname, _dbname, sizeof(dbname));
 		me = scripts->FindPluginByContext(m_pFunction->GetParentContext()->GetContext());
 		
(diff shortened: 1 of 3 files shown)


core/logic/smn_database.cpp
```diff
@@ -1425,7 +1425,7 @@ static cell_t SQL_CheckConfig(IPluginContext *pContext, const cell_t *params)
 	char *name;
 	pContext->LocalToString(params[1], &name);
 
-	return (g_DBMan.FindDatabaseConf(name) != NULL) ? 1 : 0;
+	return (g_DBMan.FindDatabaseConf(name, (params[0] >= 2)?!params[2]:true) != nullptr) ? 1 : 0;
```
KyleSanderson: What's going on here?

Headline: I could make it more readable, but if the param is even there, we pass it in, but `not`ing it since we're converting from `default_fallback` (true to enable default) to `nodefault` (true to disable default), otherwise just pass true to match old behavior

public/IDBDriver.h
```diff
@@ -819,9 +819,11 @@ namespace SourceMod
 		 * and all pointers contained therein should be considered volatile.
 		 *
 		 * @param name			Named database info.
+		 * @param nodefault		Whether the search failures should fallback 
+		 *						to default configuration, if it exists.
 		 * @return				DatabaseInfo pointer.
 		 */
-		virtual const DatabaseInfo *FindDatabaseConf(const char *name) =0;
+		virtual const DatabaseInfo *FindDatabaseConf(const char *name, bool nodefault = false) =0;
```
peace-maker: That change isn't backwards compatible. Extensions compiled against an older version of the interface will fail to call this function while they'd load fine.

You'd have to override `IsVersionCompatible` to stop old extensions from loading on newer sm versions (which is bad), add a new function to the end of the interface or just don't change the interface at all with the old behavior preserved for consumers.

CrazyHackGUT:
Now this breaks default behavior. You should check default section too.

peace-maker:
I think you got mixed up with what the problem was.

KyleSanderson:
@CrazyHackGUT the real issue is Connect doesn't use the same function as CheckConfig and this is the real bug as P-M has pointed out. It should be a one-line change.

asherkin:
Given that the original behaviour was to not fall back automatically, and changing that behaviour didn't work correctly (affecting the rarely used CheckConfig, but not the oft-used Connect,) I think the only change made here should be to rip out all fallback-to-default behaviour entirely.

It has become clear that is not the expected behaviour and very, very few plugins outside of our base set implement it, almost nobody sets up the default configuration and it is "valid" but non-working out of the box.

I think it is going to be difficult to achieve consensus on changes here, which means we should always lean towards restoring long-standing behaviour imo.

Headline:
@asherkin I do still enjoy the idea of allowing Connect to fall into `default`, it can be very handy for installing new plugins and not bothering to create your own db entry (assuming said plugin doesn't check for SQL_CheckConfig). 

This one is kind of up in the air for me, though. I don't particularly stand firmly in any direction. Most plugins that really care do the check anyway, so I don't see the fallback causing any real damage.

For lineage, the reason the fallback was created in the first place was to clean up logic elsewhere in the sm codebase that did the default check by-hand and wasn't really mean't to be seen in plugins. Now that we've gotten a taste of it, however, I don't think I mind.

asherkin:
Me too, but I think we need to attack it properly after the bug fixes.

Probably alongside removing but documenting "default" in the database configuration we ship, which'll be nice for new installs explicitly telling them they can only setup one config, and be unlikely to break existing installs since they never update the database config anyway.

Plus, I want to have another shot at the CheckConfig-doesn't-match-Connect argument if we're doing it, I feel pretty strongly that they should be aligned by default or plugins can accidentally opt-out of fallback, hurting users - and moderately-strongly that they should always be aligned to preserve end-user interests first and foremost.

PR #877, opened 2018-08-26
Added plugin version print to stack trace

TiBarification:
This change refers and closes #876

Issue #876, opened 2018-08-26
Feature request: plugin version


TiBarification:
Add please plugin version print on plugin stack trace. It can help developers to support their plugins and quickly found corrupted section in plugin by knewing a version.

Example:
```
L 08/26/2018 - 14:43:21: [SM] Exception reported: Client index 1 is a bot
L 08/26/2018 - 14:43:21: [SM] Blaming: testplugin.smx (Version 1.1.1)
L 08/26/2018 - 14:43:21: [SM] Call stack trace:
L 08/26/2018 - 14:43:21: [SM]   [0] ThrowNativeError
....
```

asherkin:
I'm not a fan of doing this, especially with the changes proposed in #877.

While in some cases the version number here might actually be useful, there are a number of cases that it can not cover (code changes without matching version bumps, natives provided by plugins where the source of the issue might be in the calling plugin) - and I'm not a fan of adding yet more information to error reports that doesn't serve a very clear purpose, users cause enough issues cutting off the "Exception reported:" line with the actual error message half the time.

I can't find it at the moment, but there is an open issue somewhere that talks about adding a new command to generate a package of debug info to use when reporting problems - I'd prefer to go down that route.

PR #878, opened 2018-08-26
Added plugin version print to stack trace

TiBarification:
In case of what happen in #877, i created this branch, it depends on master.

This changes related to #876 and will add version next to plugin name.

### DIFF
--- a/core/logic/DebugReporter.cpp
+++ b/core/logic/DebugReporter.cpp
@@ -34,6 +34,7 @@
 #include <stdarg.h>
 #include "DebugReporter.h"
 #include "Logger.h"
+#include "sm_stringutil.h"
 
 DebugReport g_DbgReporter;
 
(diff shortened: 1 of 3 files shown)


core/logic/DebugReporter.cpp
```diff
@@ -34,7 +34,7 @@
 #include <stdarg.h>
 #include "DebugReporter.h"
 #include "Logger.h"
-#include "sm_stringutil.h"
+#include "../sm_stringutil.h"
```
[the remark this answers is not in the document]
Headline: `#include <sm_stringutil.h>` and adding core to the logic/AMBuilder cxx includes will do work fine, too.

@asherkin anything wrong with that?

asherkin:
`logic` should definitely not have `core` in its include path, that could very easily cause lots of trouble.

core/logic/DebugReporter.cpp
```diff
@@ -103,14 +111,19 @@ void DebugReport::GenerateCodeError(IPluginContext *pContext, uint32_t code_addr
 	ke::SafeVsprintf(buffer, sizeof(buffer), message, ap);
 	va_end(ap);
 
-	const char *plname = pluginsys->FindPluginByContext(pContext->GetContext())->GetFilename();
+	IPlugin* plugin = pluginsys->FindPluginByContext(pContext->GetContext());
+	const char *plname = plugin->GetFilename();
+	const char *plversion = plugin->GetPublicInfo()->version;
 	const char *error = g_pSourcePawn2->GetErrorString(err);
-
+	if (!IS_STR_FILLED(plversion))
```
Headline: We're flailing around a lot to get this macro in. I don't think we want to add this into public/, so checking `plversion[0] == '\0'` by hand here is much preferred at this point.

asherkin:
The discussion (or lack thereof) in #876 needs to be addressed.

core/logic/DebugReporter.cpp
```diff
@@ -68,14 +69,21 @@ void DebugReport::GenerateErrorVA(IPluginContext *ctx, cell_t func_idx, int err,
...
 	ke::SafeVsprintf(buffer, sizeof(buffer), message, ap);
 
-	const char *plname = pluginsys->FindPluginByContext(ctx->GetContext())->GetFilename();
+	IPlugin* plugin = pluginsys->FindPluginByContext(ctx->GetContext());
+	const char *plname = plugin->GetFilename();
+	const char *plversion = plugin->GetPublicInfo()->version;
+	
+	if (!IS_STR_FILLED(plversion))
+	{
+		plversion = "Not specified";
```
stickz: Should the version section be removed from error log, if there's none specified?

PR #885, opened 2018-09-23
More TR natives

SlidyBat:
Adds some more TR natives that expose more properties of the trace results. Here's a short summary of the new natives:
```
TR_GetStartPosition - Self explanatory, added this in just for completion since GetEndPosition exists.
TR_GetFractionLeftSolid - Similar to GetFraction, but returns the fraction of when the trace left a solid (assuming it started in a solid)
TR_GetDispFlags - Returns the displacement flags of a surface. See DISPSURF_FLAG_* for possible values.
TR_GetSurfaceName - Returns the name of the surface that was hit.
TR_GetSurfaceProps - Returns the surface properties index of the surface that was hit. Not sure if this is very useful alone since it might require other non-TR related functions to be exposed to get any real info from the index.
TR_GetSurfaceFlags - Returns the surface flags of the surface that was hit. See SURF_* for possible values.
TR_GetPhysicsBone - Returns the physics bone index, if one was hit.
TR_AllSolid - Returns whether or not the entire trace was in a solid area.
TR_StartSolid - Returns whether or not the start of the trace was in a solid area.
```

In addition to the new natives, I also changed the existing `TR_EnumerateEntities` native to accept partition flags instead of the `triggers` bool value. This means that the traces made with `TR_EnumerateEntities` are no longer restricted to just "solids or triggers". Instead you can specify any combination of partition flags to use.

The biggest bonus of doing this is that you can now hit non-solid entities with traces. To showcase this (as well as a few of the new natives), here's the output of a little test I made that traces against a non-solid brush, something that was impossible to do previously.
```
Entity: 77 (func_brush)
    Start Pos: 843.38 -185.22 320.03
    Did Hit: true
    All Solid: false
    Start Solid: false
    Fraction Left Solid: 0.000000
    Displacement Flags: 00000000
    Surface Name: DEV/DEV_MEASUREGENERIC01B
    Surface Props: 41
    Surface Flags: 0000000000000000
    Physics Bone: 0
```

For reference, or if anyone else wants to play around with this, here's the plugin I used to do this: https://pastebin.com/xYXvrfa4

Another thing to note is that the partition flag values used in the inc file are a bit different to the original SDK flags. This was done to maintain backwards compatibility so that old plugins that use `true`/`false` still keep their expected behaviour.

### DIFF
--- a/extensions/sdktools/trnatives.cpp
+++ b/extensions/sdktools/trnatives.cpp
@@ -884,7 +884,7 @@ static cell_t smn_TRGetSurfaceName(IPluginContext *pContext, const cell_t *param
 		return pContext->ThrowNativeError("Invalid Handle %x (error %d)", params[1], err);
 	}
 
-	pContext->StringToLocal(params[2], params[3], tr->surface.name);
+	pContext->StringToLocal(params[1], params[2], tr->surface.name);
 
 	return 1;
 }
--- a/extensions/sdktools/trnatives.cpp
+++ b/extensions/sdktools/trnatives.cpp
@@ -117,16 +117,13 @@ enum
 // For backwards compatibility, old EnumerateEntities functions accepted bool instead of flags
 int TranslatePartitionFlags(int input)
 {
-	if (input == 0)
+	switch (input)
 	{
+	case 0:
 		return PARTITION_ENGINE_SOLID_EDICTS;
-	}
-	else if (input == 1)
-	{
+	case 1:
 		return PARTITION_ENGINE_TRIGGER_EDICTS;
-	}
-	else
-	{
+	default:
 		return input >> 1;
 	}
 }
--- a/extensions/sdktools/trnatives.cpp
+++ b/extensions/sdktools/trnatives.cpp
@@ -877,7 +877,7 @@ static cell_t smn_TRGetSurfaceName(IPluginContext *pContext, const cell_t *param
 	HandleError err;
 	HandleSecurity sec(pContext->GetIdentity(), myself->GetIdentity());
 
-	if (params[1] == BAD_HANDLE)
+	if (params[3] == BAD_HANDLE)
 	{
 		tr = &g_Trace;
 	} else if ((err = handlesys->ReadHandle(params[1], g_TraceHandle, &sec, (void **)&tr)) != HandleError_None) {
--- a/extensions/sdktools/trnatives.cpp
+++ b/extensions/sdktools/trnatives.cpp
@@ -874,14 +874,14 @@ static cell_t smn_TRGetSurfaceName(IPluginContext *pContext, const cell_t *param
 	HandleError err;
 	HandleSecurity sec(pContext->GetIdentity(), myself->GetIdentity());
 
-	if (params[3] == BAD_HANDLE)
+	if (params[1] == BAD_HANDLE)
 	{
 		tr = &g_Trace;
 	} else if ((err = handlesys->ReadHandle(params[1], g_TraceHandle, &sec, (void **)&tr)) != HandleError_None) {
-		return pContext->ThrowNativeError("Invalid Handle %x (error %d)", params[3], err);
+		return pContext->ThrowNativeError("Invalid Handle %x (error %d)", params[1], err);
 	}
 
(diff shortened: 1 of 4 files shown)


extensions/sdktools/trnatives.cpp
```diff
@@ -114,6 +114,23 @@ enum
 	RayType_Infinite
 };
 
+// For backwards compatibility, old EnumerateEntities functions accepted bool instead of flags
+int TranslatePartitionFlags(int input)
+{
+	if (input == 0)
```
Headline: nit: maybe a small switch here instead of `else if`s?

plugins/include/sdktools_trace.inc
```diff
@@ -480,6 +551,71 @@ native void TR_GetEndPosition(float pos[3], Handle hndl=INVALID_HANDLE);
...
+/**
+ * Returns the name of the surface that was hit.
+ *
+ * @param buffer			Buffer to store surface name in
+ * @param maxlen			Maximum length of output buffer
+ * @param hndl			A trace Handle, or INVALID_HANDLE to use a global trace result.
+ * @noreturn
+ * @error				Invalid Handle.
+ */
+native void TR_GetSurfaceName(char[] buffer, int maxlen, Handle hndl=INVALID_HANDLE);
```
KyleSanderson: The handle or context value must be first to make this into an Object if desired in the future.

asherkin: Looks like `TR_GetEndPosition` is the only one of the original set that catches us out here.

SlidyBat: Yeah I specifically moved the handle to last param in a later commit to keep it consistent with `TR_GetEndPosition`. Should I also change `TR_GetStartPosition` or keep that as is?

extensions/sdktools/trnatives.cpp
```diff
@@ -776,6 +806,28 @@ static cell_t smn_TRGetPlaneNormal(IPluginContext *pContext, const cell_t *param
 	return 1;
 }
 
+static cell_t smn_TRGetStartPosition(IPluginContext *pContext, const cell_t *params)
+{
+	sm_trace_t *tr;
+	HandleError err;
+	HandleSecurity sec(pContext->GetIdentity(), myself->GetIdentity());
+
+	if (params[2] == BAD_HANDLE)
```
KyleSanderson: Still broken...

plugins/include/sdktools_trace.inc
```diff
@@ -103,7 +103,61 @@
...
+/**
+ * @section Surface flags.
+ */
+ 
+#define	SURF_LIGHT		0x0001		// value will hold the light strength
+#define	SURF_SKY2D		0x0002		// don't draw, indicates we should skylight + draw 2d sky but not draw the 3D skybox
+#define	SURF_SKY		0x0004		// don't draw, but add to skybox
+#define	SURF_WARP		0x0008		// turbulent water warp
+#define	SURF_TRANS		0x0010
+#define SURF_NOPORTAL	0x0020	// the surface can not have a portal placed on it
```
KyleSanderson: spacing.

plugins/include/sdktools_trace.inc
```diff
@@ -103,7 +103,61 @@
...
+#define	SURF_TRIGGER	0x0040	// FIXME: This is an xbox hack to work around elimination of trigger surfaces, which breaks occluders
+#define	SURF_NODRAW		0x0080	// don't bother referencing the texture
+
+#define	SURF_HINT		0x0100	// make a primary bsp splitter
+
+#define	SURF_SKIP		0x0200	// completely ignore, allowing non-closed brushes
+#define SURF_NOLIGHT	0x0400	// Don't calculate light
```
KyleSanderson: ditto spacing issue for the entire file.

extensions/sdktools/trnatives.cpp
```diff
@@ -114,6 +114,20 @@ enum
 	RayType_Infinite
 };
 
+// For backwards compatibility, old EnumerateEntities functions accepted bool instead of flags
+int TranslatePartitionFlags(int input)
```
KyleSanderson: `static int` if it's just for this file.

KyleSanderson:
Thank you

PR #894, opened 2018-10-04
Fix regression in vstk size

Headline:
This is the beginning of fixing what's left to be done for the 10/3/18 CS:GO update that broke us. 

Essentially, this fixes an obvious OOB error I made, however does not completely fix issues that sourcemod is currently having.

@Drifter321 if you want to give this a shot, here's where I'm at. I'm quite stuck at the moment, though.
```sourcepawn
#include <cstrike>

public void OnPluginStart()
{
	CS_TerminateRound(0.0, CSRoundEnd_CTWin, true);
}

public Action CS_OnTerminateRound(float& delay, CSRoundEndReason& reason)
{
	PrintToServer("%.2f | REASON: %d", delay, reason);
}
```

works currently OK, however default plugins seem to fail oddly, and I'm not sure of the cause although I wouldn't be surprised if it was just one of my mistakes. 

default plugins fail with this
```
Critical error detected c0000374
(39f8.444c): Break instruction exception - code 80000003 (first chance)
eax=00000000 ebx=77d558c0 ecx=c0000374 edx=00b7c8f1 esi=00000002 edi=00f90000
eip=77d1882c esp=00b7ca60 ebp=00b7caf4 iopl=0         nv up ei pl zr na pe nc
cs=0023  ss=002b  ds=002b  es=002b  fs=0053  gs=002b             efl=00000246
ntdll!RtlReportCriticalFailure+0x4b:
77d1882c cc              int     3
```
and here's a disgusting trace. https://pastebin.com/JBykeyS5

EDIT: Feel free to takeover this pr with those fixes

### DIFF
--- a/extensions/cstrike/natives.cpp
+++ b/extensions/cstrike/natives.cpp
@@ -321,10 +321,10 @@ static cell_t CS_TerminateRound(IPluginContext *pContext, const cell_t *params)
 	{
 		REGISTER_NATIVE_ADDR("TerminateRound",
 			PassInfo pass[2]; \
-			pass[0].flags = PASSFLAG_BYVAL; \  // delay
+			pass[0].flags = PASSFLAG_BYVAL; \
 			pass[0].type = PassType_Basic; \
 			pass[0].size = sizeof(float); \
-			pass[1].flags = PASSFLAG_BYVAL; \ // reason
+			pass[1].flags = PASSFLAG_BYVAL; \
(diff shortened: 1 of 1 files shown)


SlidyBat:
I tested these changes myself on a local dedicated server on Windows and I haven't experienced any crashes. Seems to work perfectly. It'd be good if someone else could test this to confirm whether the crashes are a mistake on Headline's end.

PR #895, opened 2018-10-04
Fix build error for the cstrike extention

neko-pm:
Fix for build error C2017: illegal escape sequence, as seen here https://builds.alliedmods.net/sm/builders/windows-1.10/builds/98/steps/build/logs/stdio


SlidyBat:
This is already fixed by #894

neko-pm:
Oh, my bad. Sorry for not spotting that.

PR #903, opened 2018-10-09
Improve CreateNative failure message

asherkin:
This confuses everyone.


KyleSanderson:
Docs are definitely an improvement. Do we care that we call these Fake Natives internally; but Dynamic Natives to pawn? Anyone digging through code will know them by the former name.

asherkin:
I belive the intention is that they become just "natives" in the future, and the term "fake" has confused people in the past, I figured "dynamic" was clear but happy to change/remove. I think dropping the qualifier would make the most sense if anything.

KyleSanderson:
Yeah; if Fake/Dynamic are ephemeral ("") qualifiers let's drop them entirely to userspace.

PR #904, opened 2018-10-09
Update CS:GO gamedata.

Drifter321:


Benoist3012:
Before merging I think we should consider what has been said on #868 , make the signature longer and wildcard that byte. That's second time it broke now and on the same byte.

Headline:
@Benoist3012 unfortunately that can't be done due to the nearly identical function below, any wild carding will match both

Drifter321:
The byte that changed is the offset that is used for getclantag native. This insures nothing weird changed in the function. Also, as mentioned on windows this is inline strcpy so it can match others.

Benoist3012:
So I get that making it longer is impossible? And I guess trying to replicate the function is a dead end.

PR #906, opened 2018-10-10
Handle NULLs in SDKCall string return (Fixes #874)

### LINKED ISSUE #874: SDKCall crashes on null char pointers
# Environment
  * Operating System version: Linux
  * Game/AppID (with version if applicable): Team Fortress 2
  * SourceMod version (tested): 1.9.0.6238

# Description

SDKCalls that expect a char pointer but receive a null pointer instead seems to result in a segfault.  This was observed when calling TF2's `GetTFConditionName()` function on non-existent conditions.  The function itself has bounds checks and returns null on invalid conditions.

# Problematic Code

[Available as a gist here.](https://gist.github.com/nosoop/781c73a9e6583f05ce2bd3a465ad017b)  Prints out addresses by default (breaking out of the loop after cond 127 as expected); uncomment and comment accordingly to get the string returned instead (crashing instead of exiting the loop).

asherkin:
`SDKCall` has existing semantics that a `-1` retval indicates null, so use that and write an empty string to the buffer.

Consumers can tell the difference between `""` and `NULL` by checking if the return value is `0` or `-1`.

### DIFF
--- a/extensions/sdktools/vcaller.cpp
+++ b/extensions/sdktools/vcaller.cpp
@@ -449,6 +449,11 @@ static cell_t SDKCall(IPluginContext *pContext, const cell_t *params)
 			cell_t *addr;
 			size_t written;
 			pContext->LocalToPhysAddr(params[retparam+1], &addr);
+			if (!(*(char **)vc->retbuf))
+			{
+				pContext->StringToLocalUTF8(params[retparam], *addr, "", &written);
+				return -1;
+			}
 			pContext->StringToLocalUTF8(params[retparam], *addr, *(char **)vc->retbuf, &written);
 			return (cell_t)written;
 		} else if (vc->retinfo->vtype == Valve_Vector


extensions/sdktools/vcaller.cpp
```diff
@@ -449,6 +449,11 @@ static cell_t SDKCall(IPluginContext *pContext, const cell_t *params)
 			cell_t *addr;
 			size_t written;
 			pContext->LocalToPhysAddr(params[retparam+1], &addr);
+			if (!(*(char **)vc->retbuf))
```
Headline: `static_cast<>` ?

asherkin: I'm quite in-favour of consistently using C-style casts rather than mixing them, especially in a behemoth like this function.

Headline: i don’t think consistency is always a valid reason to avoid an approach that may be better, but we can stick to c-style casts as long as it’s already plagued in there everywhere.

PR #907, opened 2018-10-10
Use natural sorting for map lists (Fixes #892)

### LINKED ISSUE #892: Wrong Sorting Style For Choose Map sm_admin
in sourcemod/plugins/basecommands/map.sp could we have it sort properly?

current sort order is ABC
```
c10m1
c10m2
c10m3
c10m4
c10m5
c11m1
```

I need it to be 123
```
c1m1
c1m2
c1m3
c1m4
c2m1
c2m2
```

asherkin:
This produces consistently better results, especially for games such as L4D(2).

### DIFF
--- a/core/logic/smn_maplists.cpp
+++ b/core/logic/smn_maplists.cpp
@@ -321,9 +321,109 @@ public:
 		delete m_pCurMapList;
 		m_pCurMapList = NULL;
 	}
+	static bool alphanum_isdigit(const char c)
+	{
+		return c >= '0' && c <= '9';
+	}
+	static int alphanum_impl(const char *l, const char *r)
+	{
+		/**
+		 * http://www.davekoelle.com/files/alphanum.hpp
+		 *
+		 * compare l and r with strcmp() semantics, but using
+		 * the "Alphanum Algorithm". This function is designed to read
+		 * through the l and r strings only one time, for
+		 * maximum performance. It does not allocate memory for
+		 * substrings.
+		 *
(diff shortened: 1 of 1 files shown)


core/logic/smn_maplists.cpp
```diff
@@ -321,9 +321,103 @@ class MapLists : public SMGlobalClass, public ITextListener_SMC
...
+					{
+						mode = NUMBER;
+						break;
+					}
+					// if only the left character is a digit, we have a result
+					if (l_digit) return -1;
+					// if only the right character is a digit, we have a result
+					if (r_digit) return +1;
+					// compute the difference of both characters
+					const int diff=l_char - r_char;
```
Drifter321: Nit : add spaces around "=" to match everything else.

core/logic/smn_maplists.cpp
```diff
@@ -321,9 +321,103 @@ class MapLists : public SMGlobalClass, public ITextListener_SMC
...
+				}
+			}
+			else // mode == NUMBER
+			{
+				// get the left number
+				unsigned long l_int = 0;
+				while (*l && alphanum_isdigit(*l))
+				{
+					// TODO: this can overflow
+					l_int=l_int * 10 + *l - '0';
```
Drifter321: ditto

asherkin:
> Only question I have is should we be concerned with the possibility of overflows?

I think we can probably survive slightly incorrect map sorting past 4,294,967,295 alphabetically-identical maps.

PR #908, opened 2018-10-10
Add a note about FormatTime platform dependence

asherkin:
This has been another constant source of confusion since the C99 ones were added to the strftime documentation we link.

It'd be nice to have a consistent implementation inside SM in the future.

### DIFF
--- a/plugins/include/sourcemod.inc
+++ b/plugins/include/sourcemod.inc
@@ -335,6 +335,10 @@ native int GetTime(int bigStamp[2]={0,0});
  *
  * See this URL for valid parameters: 
  * http://cplusplus.com/reference/clibrary/ctime/strftime.html
+ * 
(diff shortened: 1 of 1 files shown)


plugins/include/sourcemod.inc
```diff
@@ -335,6 +335,10 @@ native int GetTime(int bigStamp[2]={0,0});
  *
  * See this URL for valid parameters: 
  * http://cplusplus.com/reference/clibrary/ctime/strftime.html
+ * 
+ * Note that available parameters depends on support from your operating system.
+ * In particular, ones highlighted in yellow on that page are not currently
+ * available on Windows and should be avoided for portable plugins.
```
asherkin: I thought that "operating system" was probably more clear to users than "C standard library".

Issue #912, opened 2018-10-27
SDKHooks:  OnLevelInit stops being called when the last registered plugin is reloaded


nosoop:
# Environment
  * Operating System version: Linux
  * Game/AppID (with version if applicable): Team Fortress 2
  * Tested against SourceMod stable 1.9.0.6259, development 1.10.0.6353
  * Tested against MetaMod:Source snapshot 1.11.0-dev+1116

# Description
When the last plugin using `OnLevelInit` is unloaded (or reloaded), the forward never fires again.

# Problematic Code (or Steps to Reproduce)
Below is `levelinit_check.sp`; all it does is report that `OnLevelInit` is called and logs it to a text file.
```
#pragma semicolon 1
#include <sourcemod>

#include <sdkhooks>

#pragma newdecls required
#pragma dynamic 1048576

public Action OnLevelInit(const char[] mapName, char mapEntities[2097152]) {
	LogToFile("level_inits.txt", "-------- Level init on %s --------", mapName);
}
```

The test case assumes that no other plugin using `OnLevelInit` is loaded in.

If the plugin is installed before the server starts up:

1.  Message is correctly displayed on level init.
2.  Perform mapchange; message is correctly displayed on level init.
3.  Reload plugin via `sm plugins reload levelinit_check`.
4.  Perform mapchange, message is (incorrectly) not displayed.

Without the plugin present:

1.  Message is correctly not displayed on the first map.
2.  Install plugin and load via `sm plugins load levelinit_check`.
3.  Perform mapchange; message is correctly displayed.
4.  Reload plugin via `sm plugins reload levelinit_check`.
5.  Perform mapchange, message is (incorrectly) not displayed.

asherkin:
Looks like the only possible cause of this I can see is if `OnPluginLoaded` is being called before forwards are re-bound.

Does it work if you unload and then load the plugin rather than using reload?

nosoop:
The plugin does not report any output if the plugin is unloaded and reloaded before performing a map change (in the case that the message was previously displayed on level init).

PR #916, opened 2018-10-28
SDKHooks: Reset global hookid when unhooking in SH.

KyleSanderson:
This is an ancient bug in SDKHooks (I actually mailed PM about it before I knew better) where we don't reset the global hookids when we unhook functions. This results in double-frees and oddness when the last users of these forwards unload; and the same or new user comes into the fold using the forward again.

This fixes https://github.com/alliedmodders/sourcemod/issues/912
This is also untested.

### DIFF
--- a/extensions/sdkhooks/extension.cpp
+++ b/extensions/sdkhooks/extension.cpp
@@ -328,6 +328,7 @@ void SDKHooks::SDK_OnAllLoaded()
 	if (hook != 0) \
 	{ \
 		SH_REMOVE_HOOK_ID(hook); \
+		hook = 0; \
 	}
 
 void SDKHooks::SDK_OnUnload()


asherkin:
Other than possibly being a bit surprising, is there any reason for `SH_REMOVE_HOOK_ID` not to do this itself?

KyleSanderson:
> Other than possibly being a bit surprising, is there any reason for `SH_REMOVE_HOOK_ID` not to do this itself?

https://github.com/alliedmodders/sourcemod/blob/9abac06e85740fa7c8a74907168abd13648d9848/public/vtable_hook_helper.h#L66

While internally I think we'd be okay this could subtly break plugins if they're storing hookids elsewhere. This would only be an issue on recompilation; but I think it's too late to change this behaviour now (and I like it better as-is, anyways).

Issue #875, opened 2018-08-23
Feature Request: FormatTime() and time correcting


CrazyHackGUT:
Add a optional boolean parameter for correcting formatted time with default value `true` (for compatibility).
Some web-applications require time in `%Y-%m-%dT%H:%M:%SZ` format (result example: `2001-08-23T14:55:02Z`).
SourceMod edits time before formatting with considering current Timezone and console variable `sm_time_adjustment`. This is good, but sometimes it is not necessary.

PR #917, opened 2018-10-29
FormatTime() and time correcting #875

komashchenko:
Implementing time formatting without taking into account time zone changes
compatibility old plugins not be broken

### DIFF
--- a/plugins/include/sourcemod.inc
+++ b/plugins/include/sourcemod.inc
@@ -375,9 +375,10 @@ native int GetTime(int bigStamp[2]={0,0});
  * @param maxlength		Maximum length of output string buffer.
  * @param format		Formatting rules (passing NULL_STRING will use the rules defined in sm_datetime_format).
  * @param stamp			Optional time stamp.
+ * @param localtime		Requires use OS timezone settings.
  * @error			Buffer too small or invalid time format.
  */
-native void FormatTime(char[] buffer, int maxlength, const char[] format, int stamp=-1);
+native void FormatTime(char[] buffer, int maxlength, const char[] format, int stamp=-1, bool localtime = true);
 
 /**
  * Loads a game config file.
--- a/core/logic/smn_core.cpp
+++ b/core/logic/smn_core.cpp
@@ -202,7 +202,8 @@ static cell_t FormatTime(IPluginContext *pContext, const cell_t *params)
 #endif
 
 	time_t t = (params[4] == -1) ? g_pSM->GetAdjustedTime() : (time_t)params[4];
-	size_t written = strftime(buffer, params[2], format, localtime(&t));
+	struct tm *tm_t = (params[0] > 4 && params[5] == 0) ? gmtime(&t) : localtime(&t);
+	size_t written = strftime(buffer, params[2], format, tm_t);
 
 #if defined SUBPLATFORM_SECURECRT
 	_set_invalid_parameter_handler(handler);


plugins/include/sourcemod.inc
```diff
@@ -375,9 +375,10 @@ native int GetTime(int bigStamp[2]={0,0});
  * @param maxlength		Maximum length of output string buffer.
  * @param format		Formatting rules (passing NULL_STRING will use the rules defined in sm_datetime_format).
  * @param stamp			Optional time stamp.
+ * @param localtime		Requires use OS timezone settings.
```
SM91337:
```suggestion
 * @param localtime		Use OS timezone.
```

SM91337:
> Implementing time formatting without taking into account time zone changes
> compatibility old plugins not be broken

```suggestion
> Implementing ability to use local time formatting without breaking compatibility with existing plugins.
```

peace-maker:
What's your use case?

komashchenko:
> What's your use case?

Necessary to record time UTC in database.

Benoist3012:
I would have a use for this one too

plugins/include/sourcemod.inc
psychonic: I agree with this. The wording in the comment needs to be improved.

Other than that, looks good to me.

asherkin:
It’s so people who rent severs from a GSP can set the timezone correctly.

DarkDeviL:
> 
> 
> > What's your use case?
> 
> Necessary to record time UTC in database.

MySQL/MariaDB UTC_TIMESTAMP(), maybe ?

> 
> 
> It’s so people who rent severs from a GSP can set the timezone correctly.

If things should change, it would make more sense to make an entry in addons/sourcemod/configs/core.cfg for the server's timezone, similar to the Linux system's /etc/timezone, e.g. "America/New_York" or "Europe/London".

A similar per-player setting could then be introduced, so it can also be adjusted for each individual player.

The latest "localtime" bool variable of FormatTime could then be the client id, of which it should be changed according to, e.g. "0" (which would be default) would imply the server, and be relative to the setting from core.cfg, where anything else would point towards a specific client id.

Or, maybe for simplicity, the latest "localtime" variable could simply just be a text string, e.g. "America/New_York" or "Europe/London".

KyleSanderson:
@asherkin before I ping komashchenko on this one - how do you feel about it?

I can see it being useful, but additionally with FormatTime I always included the TZ as end-consumers already had a hint about where things were hosted (per latency to the box). You're right for GSPs this scenario is useful. 

Anyway, I'm conflicted because it's legitimately changing time underneath people. In a totally amateur plugin I wrote in 2009 I have `FormatTime(Clock, sizeof(Clock), "%I:%M:%S%p %Z", TimeCache);` which to me shows this is beyond user error.

komashchenko:
@KyleSanderson At the time of creation, I wanted to use it with [TZD](https://github.com/komashchenko/TimeZoneDatabase), since I was getting the player's local time, then when formatting via FormatTime it turned out to be wrong, so I just added myself the [TZD_FormatTime ](https://github.com/komashchenko/TimeZoneDatabase/blob/master/TimeZoneDatabase.inc#L58) function

awillinger:
To reactivate this old PR: I would also have a use case for this: Using SourceTV Manager, I've written a plugin which automatically starts/stops recording of server-side demos and uploads them to a FTP server so they can be downloaded. 

To avoid people having to know which time zone the server is in, I would like to store the recording start time as a UTC string in the filename, which currently is not possible, as FormatTime uses the server's timezone, which can either be EST or CE(S)T in my case. Moreover, I've also written a script which automatically links said demos to bans in Sourcebans for evidence purposes and which currently requires extra (and imo unnecessary) code for timezone adjustment per server.

Not sure what else is preventing this PR from being merged, doesn't seem like it breaks bcompat.

PR #922, opened 2018-11-06
Switch CS:GO Clantag set/get to use netprops + offset over sig + offset.

Drifter321:
So since this stupid sig keeps breaking... Here is a change to switch to grabbing a netprop address and using an offset to get the address of the clantag. Completely untested, will test if no one does prior later this week.

### DIFF
--- a/extensions/cstrike/natives.cpp
+++ b/extensions/cstrike/natives.cpp
@@ -772,11 +772,11 @@ static inline cell_t GetPlayerStringVar(IPluginContext *pContext, const cell_t *
 		return pContext->ThrowNativeError("Client index %d is not valid", params[1]);
 	}
 
-	char **pVar = GetPlayerVarAddressOrError<char *>(varName, pContext, pPlayer);
+	char *pVar = GetPlayerVarAddressOrError<char>(varName, pContext, pPlayer);
 	if (pVar)
 	{
 		size_t len;
-		pContext->StringToLocalUTF8(params[2], params[3], *pVar, &len);
+		pContext->StringToLocalUTF8(params[2], params[3], pVar, &len);
 		return len;
 	}
 
@@ -800,13 +800,13 @@ static inline cell_t SetPlayerStringVar(IPluginContext *pContext, const cell_t *
 		return pContext->ThrowNativeError("Failed to locate %s offset in gamedata", szSizeName);
 	}
 
-	char **pVar = GetPlayerVarAddressOrError<char *>(varName, pContext, pPlayer);
+	char *pVar = GetPlayerVarAddressOrError<char>(varName, pContext, pPlayer);
 
 	if (pVar)
 	{
 		char *newValue;
 		pContext->LocalToString(params[2], &newValue);
-		Q_strncpy(*pVar, newValue, maxlen);
+		Q_strncpy(pVar, newValue, maxlen);
 	}
 
 	return 1;
--- a/extensions/cstrike/natives.cpp
+++ b/extensions/cstrike/natives.cpp
@@ -521,88 +521,6 @@ static cell_t CS_GetWeaponPrice(IPluginContext *pContext, const cell_t *params)
 }
 #endif
 
-static cell_t CS_GetClientClanTag(IPluginContext *pContext, const cell_t *params)
-{
-#if SOURCE_ENGINE == SE_CSGO
-	return GetPlayerStringVar(pContext, params, "ClanTag");
-#else
-	static void *addr;
-	if (!addr)
-	{
-		if (!g_pGameConf->GetMemSig("SetClanTag", &addr) || !addr)
-		{
-			return pContext->ThrowNativeError("Failed to locate function");
-		}
-	}
-
-	CBaseEntity *pEntity;
-	if (!(pEntity = GetCBaseEntity(params[1], true)))
-	{
-		return pContext->ThrowNativeError("Client index %d is not valid", params[1]);
-	}
-
-	static int tagOffsetOffset = -1;
-	static int tagOffset;
-
-	if (tagOffsetOffset == -1)
-	{
-		if (!g_pGameConf->GetOffset("ClanTagOffset", &tagOffsetOffset))
-		{
-			tagOffsetOffset = -1;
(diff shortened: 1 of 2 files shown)


e54385991:
L 11/07/2018 - 08:01:54: [SM] Exception reported: Failed to locate ClanTag offset in gamedata

e54385991:
https://github.com/alliedmodders/sourcemod/pull/922/commits/517f7d32cb53f28180a89b528e28e81481c35f57
Crashes when server starting

Headline:
Call stack


windbg


thanks to sneak because he's a cutie

extensions/cstrike/natives.cpp
```diff
@@ -800,13 +800,13 @@ static inline cell_t SetPlayerStringVar(IPluginContext *pContext, const cell_t *
...
 
-	char **pVar = GetPlayerVarAddressOrError<char *>(varName, pContext, pPlayer);
+	char *pVar = GetPlayerVarAddressOrError<char>(varName, pContext, pPlayer);
 
 	if (pVar)
 	{
 		char *newValue;
 		pContext->LocalToString(params[2], &newValue);
-		Q_strncpy(*pVar, newValue, maxlen);
+		Q_strncpy(pVar, newValue, maxlen);
```
Headline: `ke::SafeStrcpy` ?

extensions/cstrike/natives.cpp
```diff
@@ -838,6 +764,136 @@ static inline cell_t SetPlayerVar(IPluginContext *pContext, const cell_t *params
...
+	return len;
+#endif
+}
+
+static cell_t CS_SetClientClanTag(IPluginContext *pContext, const cell_t *params)
+{
+#if SOURCE_ENGINE == SE_CSGO
+	return SetPlayerStringVar(pContext, params, "ClanTag");
+#else
+	static ICallWrapper *pWrapper = NULL;
```
Headline: NULL -> nullptr

extensions/cstrike/natives.cpp
```diff
@@ -838,6 +764,136 @@ static inline cell_t SetPlayerVar(IPluginContext *pContext, const cell_t *params
...
+
+	if (!pWrapper)
+	{
+		REGISTER_NATIVE_ADDR("SetClanTag",
+			PassInfo pass[1]; \
+			pass[0].flags = PASSFLAG_BYVAL; \
+			pass[0].type = PassType_Basic; \
+			pass[0].size = sizeof(char *); \
+			pWrapper = g_pBinTools->CreateCall(addr, CallConv_ThisCall, NULL, pass, 1))
```
Headline: ditto

extensions/cstrike/natives.cpp
```diff
@@ -838,6 +764,136 @@ static inline cell_t SetPlayerVar(IPluginContext *pContext, const cell_t *params
...
+	pContext->LocalToString(params[2], &szNewTag);
+
+	unsigned char vstk[sizeof(CBaseEntity *) + sizeof(char *)];
+	unsigned char *vptr = vstk;
+
+	*(CBaseEntity **)vptr = pEntity;
+	vptr += sizeof(CBaseEntity *);
+	*(char **)vptr = szNewTag;
+
+	pWrapper->Execute(vstk, NULL);
```
Headline: ditto

sneak-it:
No more crashes with latest commit (windows)! :)

Drifter321:
As mentioned on IRC, I'm going to ignore the nits, the main goal of this pr is to actually fix and is currently inline with the remainder of the extension. A separate PR should be done fixing all of them at once after this gets merged.

Drifter321:
@asherkin I have no preference. I have the sig already fixed locally if we want to keep 1.9 the same.

extensions/cstrike/natives.cpp
```diff
@@ -838,6 +764,136 @@ static inline cell_t SetPlayerVar(IPluginContext *pContext, const cell_t *params
...
+static inline cell_t GetPlayerStringVar(IPluginContext *pContext, const cell_t *params, const char *varName)
+{
+	CBaseEntity *pPlayer = GetCBaseEntity(params[1], true);
+	if (!pPlayer)
+	{
+		return pContext->ThrowNativeError("Client index %d is not valid", params[1]);
+	}
+
+	char *pVar = GetPlayerVarAddressOrError<char>(varName, pContext, pPlayer);
+	if (pVar)
```
KyleSanderson: style would be !pVar here but this is okay too.

extensions/cstrike/natives.cpp
```diff
@@ -838,6 +764,136 @@ static inline cell_t SetPlayerVar(IPluginContext *pContext, const cell_t *params
...
+	if (!pPlayer)
+	{
+		return pContext->ThrowNativeError("Client index %d is not valid", params[1]);
+	}
+
+	char szSizeName[128];
+	g_pSM->Format(szSizeName, sizeof(szSizeName), "%sSize", varName);
+
+	int maxlen = 0;
+	if(!g_pGameConf->GetOffset(szSizeName, &maxlen))
```
KyleSanderson: Crunch.

Headline:
@asherkin Is it possible that we can take these changes into both, but also update that gamedata signature so people don’t have to update entire sm installations for this signature change in particular?

KyleSanderson:
I'm not too worked up about any of this (I don't like the style of the existing code now that was shuffled; but that's okay).

My preference would be for 1.9 to preserve the gamedata method and 1.10 we move onwards and forwards with this. If it breaks irrevocably we'll backport or pin 1.10.

Drifter321:
Updated the gamedata in 1.9 this time it will hopefully not break as easily. Instead of matching the offset of the member to make it unique, it now matches the max length of the member. So this will now be 1.10/central only.

sneak-it:
Could 1.10 also get the updated gamedata committed and/or pushed through the updater since this PR hasn't been merged yet?

Drifter321:
> Could 1.10 also get the updated gamedata committed and/or pushed through the updater since this PR hasn't been merged yet?

I rather just merge this in. I think we are good to do so, @psychonic any objections?

PR #924, opened 2018-11-14
Added DMG_LASTGENERICFLAG and DMG_HEADSHOT defines

dragokas:
According to https://developer.valvesoftware.com/wiki/Damage_types

### DIFF
--- a/plugins/include/sdkhooks.inc
+++ b/plugins/include/sdkhooks.inc
@@ -69,6 +69,8 @@
 #define DMG_BLAST_SURFACE					(1 << 27)	/**< A blast on the surface of water that cannot harm things underwater */
 #define DMG_DIRECT								(1 << 28)
 #define DMG_BUCKSHOT							(1 << 29)	/**< not quite a bullet. Little, rounder, different. */
+#define DMG_LASTGENERICFLAG							(1 << 30)
+#define DMG_HEADSHOT							(1 << 31)	/**< Damage from a headshot. */
 #endif
 
 #if !defined DMG_CRIT


Bara:
Do you've test that in CSGO?
https://github.com/alliedmodders/sourcemod/blob/master/plugins/include/cstrike.inc#L49

dragokas:
I can't test it in CSGO.
I tested it in Left4dead, where SDKHook_OnTakeDamage returned me DMG_LASTGENERICFLAG in some cases.
As about headshots, SDKHook in L4D returns -2147483646 mean (1 << 31) | DMG_BULLET
So, CSGO constants are not applicable for some another games.

Bara:
So in CSGO is DMG_LASTGENERICFLAG = Headshot

plugins/include/sdkhooks.inc
```diff
@@ -69,6 +69,8 @@
 #define DMG_BLAST_SURFACE					(1 << 27)	/**< A blast on the surface of water that cannot harm things underwater */
 #define DMG_DIRECT								(1 << 28)
 #define DMG_BUCKSHOT							(1 << 29)	/**< not quite a bullet. Little, rounder, different. */
+#define DMG_LASTGENERICFLAG							(1 << 30)
```
Headline: This is not necessary to be included, it only exists as an base point for other definitions near the msb.

plugins/include/sdkhooks.inc
```diff
@@ -69,6 +69,8 @@
...
+#define DMG_HEADSHOT							(1 << 31)	/**< Damage from a headshot. */
```
Headline: We should probably define what games do or don't support this

Headline:
closing this as it is stale, if you’d like to reopen this pull request please address required changes and reopen it once you’re ready

thank you!

PR #928, opened 2018-12-07
Add CS_SLOT_MELEE

Bara:
I would keep the CS_SLOT_KNIFE definition to don't break the exist plugins.


eyal282:
Zeus is not melee I Believe. Although it deals 0 damage at a fairly close distances

Bara:
But you've items like spanner, axe and hammer as melee weapons/items.

eyal282:
Fair enough.

PR #929, opened 2018-12-08
Add new weapon id's/itemdefs.

Drifter321:
This adds the new items/weapons to the weaponid enum. It should be noted that tablet uses the "utility" slot (I havent checked which one this actually is), the slot defines may need a new addition.


plugins/include/cstrike.inc
```diff
@@ -141,6 +141,17 @@ enum CSWeaponID
...
 	CSWeapon_REVOLVER = 64,
 	CSWeapon_TAGGRENADE = 68,
+	CSWeapon_FISTS = 69,
+	CSWeapon_BREACHCHARGE = 70,
+	CSWeapon_TABLET = 72,
+	CSWeapon_MELEE = 74,
+	CSWeapon_AXE = 75,
+	CSWeapon_HAMMER = 76,
+	CSWeapon_SPANNER = 78,
+	CSWeapon_KNiFE_GHOST = 80,
```
GoD-Tony: KNiFE -> KNIFE

PR #934, opened 2018-12-15
Inauguration of SourcePawn enum structs

dvander:
This brings in enum struct support.


dvander:
Note: This should not be merged yet. I want to make one additional syntax change that requires a & decorator on arguments, so pass-by-ref is not implicit.

dvander:
This is ready to go.

Headline:
We have lots of playerstate info held in the form of arrays for each property. I've added `PlayerState` and `PlayerInfo` enum structs that clean up our base plugins a little bit. Just waiting on another set of 👀 to make sure there was no mistakes

404UNFca:
This is the PR I've been waiting for.

PR #940, opened 2018-12-28
Patch-up InternalFilterCommandTarget during connection phase.

KyleSanderson:
As a christmas gift to a peer I've finally looked at this problem. During the connection phase immunity rules are ignored and as a result a "lower level admin" can kick a "higher level admin" or queue a bunch of timers on them while they're still downloading the level. By adding these additional checks to InternalFilterCommandTarget I think we can fix this everywhere while not introducing any additional issues.

This is untested against master as my production gear is running an older version of core.

### DIFF
--- a/core/PlayerManager.cpp
+++ b/core/PlayerManager.cpp
@@ -1499,10 +1499,23 @@ int PlayerManager::InternalFilterCommandTarget(CPlayer *pAdmin, CPlayer *pTarget
 
 	if (pAdmin != NULL)
 	{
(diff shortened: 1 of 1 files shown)


asherkin:
The existing behaviour here is intentional, with this change you can't kick or ban someone spoofing a privileged user's SteamID.

KyleSanderson:
sm_rcon is the sudo of sourcemod. No matter what the user sets as their AuthID the user can always be removed by the console.

KyleSanderson:
@asherkin do you still feel the same way about this?

PR #943, opened 2018-12-29
Add natives to work with 64 bit Protobuf values

komashchenko:


core/smn_protobuf.cpp
```diff
@@ -82,6 +82,32 @@ static cell_t smn_PbReadInt(IPluginContext *pCtx, const cell_t *params)
 	return ret;
 }
 
+static cell_t smn_PbReadInt64(IPluginContext *pCtx, const cell_t *params)
+{
+	GET_MSG_FROM_HANDLE_OR_ERR();
+	GET_FIELD_NAME_OR_ERR();
+	
+	int64 *ret;
+	pCtx->LocalToPhysAddr(params[3], reinterpret_cast<cell_t **>(&ret));
```
asherkin: This isn't strictly legal as `cell_t` and `int64` have differing alignment requirements (4 and 8 bytes respectively) and this could end up unaligned. (Which is just a perf penalty on x86, but can be fatal on other architectures.)

It would be better to use a temporary `int64` variable and combine manually.

This comment applies generally throughout the PR.

komashchenko:
@asherkin As I understand, do not need to add legacy API ?
https://github.com/alliedmodders/sourcemod/pull/946#discussion_r245096489

core/smn_protobuf.cpp
```diff
@@ -747,6 +827,7 @@ static cell_t smn_PbAddMessage(IPluginContext *pCtx, const cell_t *params)
 REGISTER_NATIVES(protobufnatives)
 {
 	{"PbReadInt",					smn_PbReadInt},
+	{"PbReadInt64",					smn_PbReadInt64},
```
peace-maker: You missed those from the non-methodmap api.

kireevvv:
Add this function is very necessary

core/smn_protobuf.cpp
```diff
@@ -553,6 +614,25 @@ static cell_t smn_PbAddInt(IPluginContext *pCtx, const cell_t *params)
...
 
+static cell_t smn_PbAddInt64(IPluginContext *pCtx, const cell_t *params)
+{
+	GET_MSG_FROM_HANDLE_OR_ERR();
+	GET_FIELD_NAME_OR_ERR();
+	
+	cell_t *value;
+	pCtx->LocalToPhysAddr(params[3], &value);
+	int64 temp;
+	((cell_t *)&temp)[0] = value[0];
```
KyleSanderson: Nit: static_cast please.

PR #944, opened 2018-12-30
Fix basecomm throwing errors on console native invocation...

KyleSanderson:
Basecomm natives error out when a friendly caller passes in the console index; this PR intends to resolve this undocumented issue.


KyleSanderson:
:-/...

PR #946, opened 2019-01-03
Added a method to get the value of whether an event's broadcasting di…

OkyHp:
…sabled or not

### DIFF
--- a/core/smn_events.cpp
+++ b/core/smn_events.cpp
@@ -438,6 +438,22 @@ static cell_t sm_SetEventBroadcast(IPluginContext *pContext, const cell_t *param
 	return 1;
 }
 
+static cell_t sm_GetEventBroadcast(IPluginContext *pContext, const cell_t *params)
+{
+	Handle_t hndl = static_cast<Handle_t>(params[1]);
+	HandleError err;
+	EventInfo *pInfo;
+	HandleSecurity sec(pContext->GetIdentity(), g_pCoreIdent);
+
+	if ((err=handlesys->ReadHandle(hndl, g_EventManager.GetHandleType(), &sec, (void **)&pInfo))
+		!= HandleError_None)
+	{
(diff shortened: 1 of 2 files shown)


asherkin:
This information is already included in the callback arguments - what are you doing that needs this info from the event handle?

OkyHp:
There is one problem: in another plugin, the notification for the event is disabled, but in another plugin, in the same event, it returns a different value when checked.
P.S. Sorry for my English.

core/smn_events.cpp
```diff
@@ -466,6 +483,7 @@ REGISTER_NATIVES(gameEventNatives)
 	{"Event.GetInt",		sm_GetEventInt},
 	{"Event.GetFloat",		sm_GetEventFloat},
 	{"Event.GetString",		sm_GetEventString},
+	{"Event.BroadcastDisabled.get", sm_GetEventBroadcast},
```
asherkin: Could you move this down under `Event.BroadcastDisabled.set` please. It's more readable to keep the properties together.

plugins/include/events.inc
```diff
@@ -338,3 +339,12 @@ native void GetEventName(Handle event, char[] name, int maxlength);
...
 native void SetEventBroadcast(Handle event, bool dontBroadcast);
+
+/**
+ * Returns whether an event's broadcasting disabled or not.
+ *
+ * @param event         Handle to an event from an event hook.
+ * @return				The boolean value of the whether an event's broadcasting disabled.
+ * @error               Invalid Handle.
+ */
+native bool GetEventBroadcast(Handle event);
```
asherkin: We're trying to avoid adding any more "legacy" API where possible, so could you remove this one and keep only the MethodMap property.

OkyHp: Fixed all comments

asherkin:
Thanks for the contribution!

PR #949, opened 2019-01-13
Prevent FrameIterator OOB Errors

Headline:
Before this patch, we'd check OOB before incrementing which makes no sense. We'll increment always this time, and change our OOB check to account for that.

Prior versions of the FrameIterator never worked, so we should probably cherry-pick this to 1.9


KyleSanderson:
This is a regression... You should not be able to slam Next until this overflows...

PR #953, opened 2019-01-21
Add CS_DMG_THROW to cstrike.inc

Bara:
I added this in cstrike.inc, because I don't know which other games has throwable weapons/items which make damage.

I already use this in one of [my plugins](https://github.com/Bara/LR-Melee/blob/master/scripting/lastrequest_melee.sp#L230) and it works pretty good.


asherkin:
This already exists in the `sdkhooks` include as `DMG_CLUB` (and the name makes sense, so I don't think we need an extra definition).

PR #961, opened 2019-02-07
Incorperate Licensing Into Project Tree

Headline:
We have not shipped the license file itself along with the root of the project tree, which is probably something that we should have already been doing. Along with that, I have added a section into the README.md file due to the guidance of the creators of GPL. 

To quote, `Just putting a copy of the GNU GPL in a file in your repository does not explicitly state that the code in the same repository may be used under the GNU GPL.`


Headline:
All it was is a file move, not sure if it’s just github being naughty

PR #974, opened 2019-03-26
SQL Admin Manager refinements and standardisation

b3none:
Changes:
- Use the `sizeof()` function to declare the size of the for loops as opposed to hard-coding the value because it feels cleaner to do it that way.
- Dynamically pad the variables storing multiple values.

### DIFF
--- a/plugins/sql-admin-manager.sp
+++ b/plugins/sql-admin-manager.sp
@@ -31,11 +31,10 @@
  * Version: $Id$
  */
 
-/* We like semicolons */
-#pragma semicolon 1
-
 #include <sourcemod>
 
+/* We like semicolons */
+#pragma semicolon 1
 #pragma newdecls required
 
 #define CURRENT_SCHEMA_VERSION		1409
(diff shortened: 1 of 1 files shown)


plugins/sql-admin-manager.sp
```diff
@@ -494,7 +493,7 @@ public Action Command_SetAdminGroups(int client, int args)
 			int gid = SQL_FetchInt(hFindQuery, 0);
 			
 			hAddQuery.BindInt(0, gid);
-			hAddQuery.BindInt(1, ++inherit_order);
+			hAddQuery.BindInt(1, inherit_order++);
```
Headline: This is a change of behavior, is this intended?

b3none: That's totally my bad. Will revert.

Headline:
Good catch, thanks for this.

I think we could go a step further and remove the hard coded 1st dim size.

`char[7][] ...` -> `char[][] ...`

plugins/sql-admin-manager.sp
```diff
@@ -31,17 +31,16 @@
...
 
+/* We like semicolons */
+#pragma semicolon 1
 #pragma newdecls required
 
 #define CURRENT_SCHEMA_VERSION		1409
 #define SCHEMA_UPGRADE_1			1409
 
-int current_version[4] = {1, 0, 0, CURRENT_SCHEMA_VERSION};
+int current_version[] = {1, 0, 0, CURRENT_SCHEMA_VERSION};
```
MitchDizzle: Weird design question but why is it '1.0.0.%VERSION%" why not just the version without the prefix of '1.0.0.'?

KyleSanderson:
Thank you.

PR #977, opened 2019-04-06
sm_cookies now display a number before the cookie (without arguments) similar to sm_help

eyal282:
[001] cookieName - Cookie's description
[002] anotherCookieName - Another Cookie's description.

This is extremely effective when the description reaches a new line.

On top of that, I also made sm_cookies <name> show the cookie's description rather than just the value in a manner similar to writing the name of a cvar.

THIS PULL REQUEST REQUIRES SOME DEGREE OF TESTING. I hope someone will have the spare time to test out everything possible and give some input on the change.


plugins/clientprefs.sp
```diff
@@ -124,13 +126,37 @@ public Action Command_Cookie(int client, int args)
...
+		description[0] = '\0';
+		
+		while (ReadCookieIterator(iter, 
+								CookieName, 
+								sizeof(CookieName),
+								access, 
+								description, 
+								sizeof(description)) != false) // We're allowed to re-use access since we're about to return anyways.
+		{
+			if(StrEqual(CookieName, name, true))
```
KyleSanderson: Inconsistent formatting on a control statement.

plugins/clientprefs.sp
```diff
@@ -124,13 +126,37 @@ public Action Command_Cookie(int client, int args)
...
+			{
+				TrimString(description);
+				if(description[0] != EOS)
```
KyleSanderson: Formatting.

plugins/clientprefs.sp
eyal282: I'm bad at Github's layout, what's the issue exactly? Bad names for the variables?

MitchDizzle: Is this because he didn't put a space between `if` and `(` ?

plugins/clientprefs.sp
KyleSanderson: Still crunched :(

plugins/clientprefs.sp
```diff
@@ -124,13 +126,37 @@ public Action Command_Cookie(int client, int args)
 	
 	if (args == 1)
 	{
+		Handle iter = GetCookieIterator();
+		
 		GetClientCookie(client, cookie, value, sizeof(value));
 		ReplyToCommand(client, "[SM] %t", "Cookie Value", name, value);
 		
+		char CookieName[30];
+		CookieName[0] = '\0';
```
Headline: I do not believe this is necessary.

Ditto on line 137

MitchDizzle: Well seeing how all the other comments are about the formatting of the current changes then this is absolutely 100% necessary.  This is same code that is farther up on the script, changing this would be changing the coding style, making it inconsistent. Starts to lean out of the scope of this Pull Request.

https://github.com/alliedmodders/sourcemod/blob/ebeb40f8a0b257081e1555239d9036113a1afb6a/plugins/clientprefs.sp#L69-L72

plugins/clientprefs.sp
```diff
@@ -84,7 +86,7 @@ public Action Command_Cookie(int client, int args)
 		{
 			if (access < CookieAccess_Private)
 			{
-				PrintToConsole(client, "%s - %s", name, description);
+				PrintToConsole(client, "[%03d] %s - %s", count++, name, description);
```
MitchDizzle: I would almost say if we're going to make it more like sm_help then this should be paginated instead of printing all cookies to the player as separate messages, to lessen the amount sent to the player at one time. This would be a bit more advanced so maybe a different Pull Request

plugins/clientprefs.sp
asherkin: It looks like removing these lines was just missed as part of #506 when converting them from `decl`.

Headline: Ah; we can correct these if they exist elsewhere in another pr

eyal282:
I think I solved all issues, sorry for my lack of scripting skills when it comes to making it look sane, I come from AmxModX and learning a to format in a generally accepted way can be sometimes difficult to me.

plugins/clientprefs.sp
eyal282: I'm am a bit confused about what you're asking me to do but I replaced all '\0' with EOS, as it's a better symbol with a meaning easier to remember for newer sourcemod coders trying to understand what it means ( End Of String )

eyal282:
I should mention I have not a slight clue about anything related to the "Some checks haven't completed yet" of the "continuous-integration" error

eyal282:
Anyways, are you okay with my changes or there is anything I forgot / new changes needed?

plugins/clientprefs.sp
MitchDizzle: It should be
```cpp
if (description[0] != EOS)
{
    ReplyToCommand(client, "- %s", description);
}
```

plugins/clientprefs.sp
MitchDizzle: This is still not resolved, inconsistent formatting on a control statement just means your if statement needs to look like this:
```cpp
if (StrEqual(CookieName, name, true))
{
    //etc.
}
```

eyal282:
Changed the while's formatting according to advice in sourcemod's discord channel.

plugins/clientprefs.sp
```diff
@@ -67,24 +67,26 @@ public Action Command_Cookie(int client, int args)
...
-		while (ReadCookieIterator(iter, 
-								name, 
-								sizeof(name),
-								access, 
-								description, 
-								sizeof(description)) != false)
+		int count = 1;
+		
+		while
+		(ReadCookieIterator(
```
MitchDizzle: Some how you keep making this look even worse...

plugins/clientprefs.sp
```diff
@@ -120,17 +122,40 @@ public Action Command_Cookie(int client, int args)
...
+		sizeof(CookieName),
+		access, 
+		description, 
+		sizeof(description)) != false) // We're allowed to re-use access since we're about to return anyways.
+		{
+			if (StrEqual(CookieName, name, true))
+			{
+				TrimString(description);
+				if (description[0] != EOS)
+					ReplyToCommand(client, "- %s", description);
```
MitchDizzle: Still not fixed. I've already commented on how to fix this.

plugins/clientprefs.sp
eyal282: Can you give your own idea on how to fix how it looks?

Headline: `while (...` not `while\n`

eyal282: Done.

plugins/clientprefs.sp
eyal282: I'm sorry, what did I do wrong here?

plugins/clientprefs.sp
eyal282: So what's the problem that remains here?

plugins/clientprefs.sp
eyal282: Now?

plugins/clientprefs.sp
eyal282: Now it's good?

PR #979, opened 2019-04-09
Block loading of incompatible IBinTools interface

peace-maker:
In #705 SourceMod received support for x64 binaries. The `IBinTools` interface was updated to call functions in 64bit binaries. The `PassInfo` struct's size was increased and the `Create[V]Call()` functions signatures changed, thus making the interface incompatible for consumers which were compiled against an earlier version.

`IBinTools::IsVersionCompatible` wasn't adjusted to that fact, so extensions compiled against pre SM 1.10 could request an `IBinTools` interface pointer, but crash when they try to use it.

This change makes requests to older interface versions invalid, thus letting `RequestInterface` return `NULL` for older extensions. It doesn't fix the backwards incompatibility, but at least makes the problem more blatant, so extensions can handle it themselves.

### DIFF
--- a/public/extensions/IBinTools.h
+++ b/public/extensions/IBinTools.h
@@ -37,6 +37,8 @@
 
 #define SMINTERFACE_BINTOOLS_NAME		"IBinTools"
 #define SMINTERFACE_BINTOOLS_VERSION	4
+// Backwards incompatible change for x64 support.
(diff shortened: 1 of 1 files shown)


KyleSanderson:
Ah shoot, thank you very much.

PR #983, opened 2019-04-23
Create menu if multiple nom matches found

JoinedSenses:
This change checks the nomination against the map arraylist. If the nomination matches multiple results, those results are then added to a menu to allow the client to select available maps. The maps added to the menu go against the same checks as the normal nomination menu and wont allow nomination of disabled maps. ~~There is also a defined MATCHED_INDEXES_MAX of 7, which means there will only be one page of matched maps.~~ Update: Added convar sm_nominate_maxfound.

The main reason for this change is to allow for more flexibility of map nominations. In its current state, the plugin only attempts to nominate the very first match.

Example image of /nominate jump_b
Nom Menu

Image demonstrating that disabled maps cannot be selected. 
Command used: /nominate jump_
Nom Menu2

I would also like feedback on the idea of registering command "sm_nom" with the same callback as "sm_nominate".

### DIFF
--- a/plugins/nominations.sp
+++ b/plugins/nominations.sp
@@ -34,8 +34,6 @@
 #include <sourcemod>
 #include <mapchooser>
 
-#define MATCHED_INDEXES_MAX 7
-
 #pragma semicolon 1
 #pragma newdecls required
 
--- a/plugins/nominations.sp
+++ b/plugins/nominations.sp
@@ -34,6 +34,8 @@
 #include <sourcemod>
 #include <mapchooser>
 
+#define MATCHED_INDEXES_MAX 7
+
 #pragma semicolon 1
 #pragma newdecls required
 
@@ -178,7 +180,7 @@ public void OnClientSayCommand_Post(int client, const char[] command, const char
 	{
 		ReplySource old = SetCmdReplySource(SM_REPLY_TO_CHAT);
 		
-		AttemptNominate(client);
+		OpenNominationMenu(client);
 		
 		SetCmdReplySource(old);
 	}
@@ -193,18 +195,87 @@ public Action Command_Nominate(int client, int args)
 	
 	if (args == 0)
 	{
-		AttemptNominate(client);
+		OpenNominationMenu(client);
 		return Plugin_Handled;
 	}
 	
 	char mapname[PLATFORM_MAX_PATH];
 	GetCmdArg(1, mapname, sizeof(mapname));
-	
-	if (FindMap(mapname, mapname, sizeof(mapname)) == FindMap_NotFound)
+
+	ArrayList results = new ArrayList();
+	int matches = FindMatchingMaps(g_MapList, results, mapname);
+
+	char mapResult[PLATFORM_MAX_PATH];
+
+	if (matches <= 0)
+	{
+		ReplyToCommand(client, "%t", "Map was not found", mapname);
+	}
+	else if (matches > 1)
+	{
+		// Display results to the client and end
+		Menu menu = new Menu(MenuHandler_FoundMaps, MENU_ACTIONS_DEFAULT|MenuAction_DrawItem|MenuAction_DisplayItem);
+		menu.SetTitle("Select map");
+		
+		for (int i = 0; i < results.Length; i++)
+		{
+			g_MapList.GetString(results.Get(i), mapResult, sizeof(mapResult));
+			menu.AddItem(mapResult, mapResult);
+		}
+
+		menu.Display(client, 30);
+		ReplyToCommand(client, "[SM] %t", "Multiple Matches Found");
+	}
+	// One result
+	else if (matches == 1)
+	{
+		// Get the result and nominate it
+		g_MapList.GetString(results.Get(0), mapResult, sizeof(mapResult));
+		AttemptNominate(client, mapResult, sizeof(mapResult));
+	}
+
+	delete results;
+
+	return Plugin_Handled;
+}
+
+int FindMatchingMaps(ArrayList mapList, ArrayList results, const char[] input){
+	int map_count = mapList.Length;
+
+	if (!map_count)
+	{
+		return -1;
+	}
+
+	int matches = 0;
+	char map[PLATFORM_MAX_PATH];
+
+	for (int i = 0; i < map_count; i++)
+	{
+		mapList.GetString(i, map, sizeof(map));
+		if (StrContains(map, input) != -1)
+		{
+			results.Push(i);
+			matches++;
+
+			if (matches >= MATCHED_INDEXES_MAX)
+			{
+				break;
+			}
+		}
+	}
+
+	return matches;
+}
+
+void AttemptNominate(int client, const char[] map, int size)
+{
+	char mapname[PLATFORM_MAX_PATH];
+	if (FindMap(map, mapname, size) == FindMap_NotFound)
 	{
 		// We couldn't resolve the map entry to a filename, so...
 		ReplyToCommand(client, "%t", "Map was not found", mapname);
-		return Plugin_Handled;		
+		return;		
 	}
 	
 	char displayName[PLATFORM_MAX_PATH];
@@ -214,7 +285,7 @@ public Action Command_Nominate(int client, int args)
 	if (!g_mapTrie.GetValue(mapname, status))
 	{
 		ReplyToCommand(client, "%t", "Map was not found", displayName);
-		return Plugin_Handled;		
+		return;		
 	}
 	
 	if ((status & MAPSTATUS_DISABLED) == MAPSTATUS_DISABLED)
@@ -234,7 +305,7 @@ public Action Command_Nominate(int client, int args)
 			ReplyToCommand(client, "[SM] %t", "Map Already Nominated");
 		}
 		
-		return Plugin_Handled;
+		return;
 	}
 	
 	NominateResult result = NominateMap(mapname, false, client);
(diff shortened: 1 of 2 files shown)


plugins/nominations.sp
```diff
@@ -34,6 +34,8 @@
 #include <sourcemod>
 #include <mapchooser>
 
+#define MATCHED_INDEXES_MAX 7
```
Headline: Why have a defined maximum? Could this be a ConVar that defaults to `0` which signifies infinite?

plugins/nominations.sp
```diff
@@ -193,18 +195,87 @@ public Action Command_Nominate(int client, int args)
...
+
+void AttemptNominate(int client, const char[] map, int size)
+{
+	char mapname[PLATFORM_MAX_PATH];
+	if (FindMap(map, mapname, size) == FindMap_NotFound)
 	{
 		// We couldn't resolve the map entry to a filename, so...
 		ReplyToCommand(client, "%t", "Map was not found", mapname);
-		return Plugin_Handled;		
+		return;
```
Headline: Good catch

plugins/nominations.sp
```diff
@@ -333,7 +403,7 @@ void BuildMapMenu()
 	delete excludeMaps;
 }
 
-public int Handler_MapSelectMenu(Menu menu, MenuAction action, int param1, int param2)
+int MenuHandler_MapSelect(Menu menu, MenuAction action, int param1, int param2)
```
Headline: This should be `public`

plugins/nominations.sp
```diff
@@ -386,9 +456,6 @@ public int Handler_MapSelectMenu(Menu menu, MenuAction action, int param1, int p
 			{
 				return ITEMDRAW_DISABLED;	
 			}
-			
-			return ITEMDRAW_DEFAULT;
```
Headline: Let's keep this; being explicit is nice

plugins/nominations.sp
```diff
@@ -426,10 +493,76 @@ public int Handler_MapSelectMenu(Menu menu, MenuAction action, int param1, int p
...
 			}
-			
-			return 0;
 		}
 	}
 	
 	return 0;
 }
+
+int MenuHandler_FoundMaps(Menu menu, MenuAction action, int param1, int param2) {
```
Headline: `public`

plugins/nominations.sp
```diff
@@ -426,10 +493,76 @@ public int Handler_MapSelectMenu(Menu menu, MenuAction action, int param1, int p
...
+	switch (action) {
+		case MenuAction_Select:
+		{
+			char mapname[96];
```
Headline: `PLATFORM_MAX_PATH`

plugins/nominations.sp
```diff
@@ -426,10 +493,76 @@ public int Handler_MapSelectMenu(Menu menu, MenuAction action, int param1, int p
...
+			if (!g_mapTrie.GetValue(map, status))
+			{
+				LogError("Menu selection of item not in trie. Major logic problem somewhere.");
+				return ITEMDRAW_DEFAULT;
+			}
+			
+			if ((status & MAPSTATUS_DISABLED) == MAPSTATUS_DISABLED)
+			{
+				return ITEMDRAW_DISABLED;	
+			}
```
Headline: I've seen this before; can we avoid code duplication?

plugins/nominations.sp
```diff
@@ -426,10 +493,76 @@ public int Handler_MapSelectMenu(Menu menu, MenuAction action, int param1, int p
...
+		}
+		case MenuAction_DisplayItem:
+		{
+			char mapname[128];
```
Headline: `PLATFORM_MAX_PATH`

plugins/nominations.sp
JoinedSenses: public isnt required on menu callbacks for whatever reason

plugins/nominations.sp
JoinedSenses: I will reduce this bit of code.

plugins/nominations.sp
404UNFca: Plus IIRC, the compiler assumes everything is `public` anyway.

Headline: That’s true, but for consistency let’s keep it

404UNFca: Yeahhhhh, you're right. Everything I've seen in the base SourceMod files always has `public`.

plugins/nominations.sp
JoinedSenses: This was actually part of the command callback. I made it into a void function.

plugins/nominations.sp
JoinedSenses: Addressed. Added ConVar "sm_nominate_maxfound".

plugins/nominations.sp
JoinedSenses: Addressed. Re-added public.

plugins/nominations.sp
JoinedSenses: Addressed. Readded.

plugins/nominations.sp
JoinedSenses: Addressed. Readded.

plugins/nominations.sp
JoinedSenses: Addressed. Array size updated.

plugins/nominations.sp
JoinedSenses: Addressed. The plugin has been reduced to a single callback, reducing code.

plugins/nominations.sp
JoinedSenses: Addressed. Array sized updated.

Headline:
Last thing that could probably use some discussion (aka don't implement it yet), we might want to add a check so player console & server console behavior is unaffected. I think this should only happen when a player executes !nominate in chat.

JoinedSenses:
I went ahead and went against your message and added an implementation to demonstrate what I had in mind for checking source.

If reply source is console, menus wont be generated and it will attempt to nominate as normal.

plugins/nominations.sp
```diff
@@ -190,21 +192,111 @@ public Action Command_Nominate(int client, int args)
...
+		
 		return Plugin_Handled;
 	}
 	
 	char mapname[PLATFORM_MAX_PATH];
 	GetCmdArg(1, mapname, sizeof(mapname));
-	
-	if (FindMap(mapname, mapname, sizeof(mapname)) == FindMap_NotFound)
+
+	ArrayList results = new ArrayList();
```
Headline: is this always empty?

plugins/nominations.sp
```diff
@@ -190,21 +192,111 @@ public Action Command_Nominate(int client, int args)
...
+		menu.SetTitle("Select map");
+		
+		for (int i = 0; i < results.Length; i++)
+		{
+			g_MapList.GetString(results.Get(i), mapResult, sizeof(mapResult));
+			menu.AddItem(mapResult, mapResult);
+		}
+
+		menu.Display(client, 30);
+		ReplyToCommand(client, "[SM] %t", "Multiple Matches Found");
```
Headline: I don't think it's necessary for a new translation to exist here. We're already showing them the menu.

plugins/nominations.sp
JoinedSenses: no

https://github.com/alliedmodders/sourcemod/blob/d7622489f2ac5d2ba2f4581d8ebf694270070014/plugins/nominations.sp#L279

plugins/nominations.sp
JoinedSenses: necessary? maybe not. there are a lot of translations that aren't necessary. i do think the acknowlegement is nice.

plugins/nominations.sp
JoinedSenses: Results is the arraylist of the found matching maps based on the /nominate args.

plugins/nominations.sp
Headline: The problem is that this will need to be translated to other languages, and it'd probably be better overall to just omit this line and not have to worry about that

JoinedSenses: Removed.

plugins/nominations.sp
```diff
@@ -190,21 +192,109 @@ public Action Command_Nominate(int client, int args)
...
+
+		menu.Display(client, 30);
+	}
+
+	delete results;
+
+	return Plugin_Handled;
+}
+
+int FindMatchingMaps(ArrayList mapList, ArrayList results, const char[] input)
```
KyleSanderson: results should be a reference.

plugins/nominations.sp
```diff
@@ -190,21 +192,109 @@ public Action Command_Nominate(int client, int args)
...
+		// Get the result and nominate it
+		g_MapList.GetString(results.Get(0), mapResult, sizeof(mapResult));
+		AttemptNominate(client, mapResult, sizeof(mapResult));
+	}
+	else if (matches > 1)
+	{
+		if (source == SM_REPLY_TO_CONSOLE)
+		{
+			// if source is console, attempt instead of displaying menu.
+			AttemptNominate(client, mapname, sizeof(mapname));
```
KyleSanderson: You're leaking results here causing an un-authenticated DoS...

plugins/nominations.sp
Headline: @JoinedSenses looks like this is the only thing holding the PR back

asherkin: I'm not sure I follow this requested change - why should `results` be passed by-ref, it's a Handle and the caller is responsible for freeing it?

Headline: probably because it’s an out param and it’s just a readability thing

KyleSanderson: This wasn't the problem, just part of the review of the DoS.

asherkin:
Thanks for this!

PR #985, opened 2019-05-01
Update CS:GO CSWeaponDropBB signatures

Technoblazed:
Fixes gamedata changes from the 4/30/19 CS:GO update. Found by rachnus in the sm discord.


TheExtr1m:
Hello. These signatures do not work, just updated on their SM server and does not work.

Fohdeesha:
Same issue as TheExtr1m, with these new signatures I'm still getting errors such as the below: 

`SM] Exception reported: Failed to lookup CSWeaponDropBB signature.`

Ayuto:
This PR actually hasn't changed the signatures at all (except removing the byte 56, but that is still a valid part of the signature. It's just not required to find a unique match). The first 6 bytes (2A) are just placeholders. SM's signature scanner ignores/skips them. I guess that has been done to still be able to find the signature even if the function has been hooked by a different server plugin.

It seems like the Windows signature is unchanged. The new Linux signature is:
``55 89 E5 56 53 83 EC 70 8B 5D 08 0F B6 75 10``

psychonic:
@Ayuto thanks. Fixing it up now.

PR #995, opened 2019-05-02
Update Reload offset for CS:GO

qubka:
[engine.csgo.txt](https://github.com/alliedmodders/sourcemod/files/3139014/engine.csgo.txt)


Headline:
This must be malformed

PR #1003, opened 2019-05-09
Add Sort and SortCustom to ArrayList methodmap

JoinedSenses:
Please correct me if this implementation is incorrect. Attempted to add Sort and SortCustom to ArrayList methodmap.


asherkin:
Thanks!

PR #1004, opened 2019-05-09
Created GlobalForward and Forward methodmaps

JoinedSenses:
This PR introduces types GlobalForward and Forward by creating methodmaps for the two. Pretty straight forward. As usual, please let me know if something looks off or could use improvement.


plugins/include/functions.inc
```diff
@@ -111,6 +111,93 @@ enum ExecType
...
+	 *
+	 * @note Cannot be used during an incomplete call.
+	 *
+	 * @param plugin            Handle of the plugin that contains the function.
+	 *                          Pass INVALID_HANDLE to specify the calling plugin.
+	 * @param func              Function to add to forward.
+	 * @return                  True on success, false otherwise.
+	 * @error                   Invalid or corrupt private forward handle, invalid or corrupt plugin handle, or invalid function.
+	 */
+	public native bool Add(Handle plugin, Function func);
```
KyleSanderson: Can you add plugin=INVALID_HANDLE to make this nice(r)?

JoinedSenses: I would have if the original natives were the same. If we were to modify if for Add, Remove, and RemovAll, then does that mean we should modify the original natives as well?

JoinedSenses:
Another thought I have regarding this is to add all `Call_` functions as static to the methodmap

plugins/include/functions.inc
```diff
@@ -111,6 +111,93 @@ enum ExecType
...
+	public native GlobalForward(const char[] name, ExecType type, ParamType ...);
+
+	/**
+	 * Starts a call to functions in a forward's call list.
+	 *
+	 * @note Cannot be used during an incomplete call.
+	 *
+	 * @error                   Invalid or corrupt forward handle or called before another call has completed.
+	 */
+	public native void StartCall();
```
asherkin: I'm really not a fan off this being on the methodmap, it feels awkward and detached from the rest of the function calling API. I think we can probably look at cleaning this up in the future by adding a varargs based wrapper on the methodmap, but not as part of these changes.

plugins/include/functions.inc
asherkin: I think these should be `AddFunction` / `RemoveFunction` / `RemoveAllFunctions`, without the suffix it isn't clear what is being added/removed (and the property is called `FunctionCount`).

asherkin: Adding a default value for a non-trailing param doesn't really make sense - I know SP has named params, but in that case `plugin` being `null` is super rarely used anyway it is probably good to be explicit, it's almost never the desired use.

plugins/include/functions.inc
```diff
@@ -552,4 +639,4 @@ typedef RequestFrameCallback = function void (any data);
  * @param Function	Function to call on the next frame.
  * @param data			Value to be passed on the invocation of the Function.
  */
-native void RequestFrame(RequestFrameCallback Function, any data=0);
+native void RequestFrame(RequestFrameCallback Function, any data=0);
```
asherkin: EOF newline has gone missing.

plugins/include/functions.inc
```diff
@@ -111,6 +111,93 @@ enum ExecType
...
+	 */
+	public native void StartCall();
+
+	// Returns the number of functions in a global or private forward's call list.
+	property int FunctionCount {
+		public native get();
+	}
+};
+
+methodmap Forward < GlobalForward {
```
asherkin: How do you feel about `PrivateForward`? That is what we call it in the doc comment and it feels more balanced against `GlobalForward` (which is generally the one people are using anyway). That said, calling them "private" and "global" forwards has never really made sense and this is a good opportunity to change the terminology, some nomenclature around "managed" might be clearer?

plugins/include/functions.inc
KyleSanderson: Yeah I get it, I'm one of the tiny consumers that heavily use private forwards and similar, the marshaller owns the fwds and clones them out where people add themselves to the fwd.

asherkin: That is an interesting API design choice, are the clones owned by the calling plugin then? What's to stop one treading on other's toes by removing all other functions from the forward?

KyleSanderson: Communism... :-)

plugins/include/functions.inc
JoinedSenses: Could you give an example of what a varargs based wrapper might look like, in both implementation and use? Mostly since I'm not sure I understand.

plugins/include/functions.inc
JoinedSenses: What is the EOL needed for here?

plugins/include/functions.inc
JoinedSenses: It makes sense to have `PrivateForward` instead of just `Forward`. If not "private" and "global", you're suggesting "managed" and "unmanaged"? Something about these terms don't sit right with me, perhaps because it's terminology i've already come to terms with. That and "private"/"global" are referenced in a ton of places already, so it might be weird going back on that.

plugins/include/functions.inc
JoinedSenses: This has been addressed in my latest commit: https://github.com/alliedmodders/sourcemod/pull/1004/commits/da4248c70560db2c58c6c14476303102ec8b07ae

plugins/include/functions.inc
JoinedSenses: https://github.com/alliedmodders/sourcemod/blob/master/plugins/include/events.inc#L342

plugins/include/functions.inc
Headline: Woah; what happened?

JoinedSenses: Ill fix tonight, uncaught change when reverting whitespace commit

JoinedSenses: This oversight has been addressed in the latest commit. https://github.com/alliedmodders/sourcemod/pull/1004/commits/26eb127a93e8bcc0c84303e6de415f2925257ecc

PR #1005, opened 2019-05-09
TFCond Descriptions

Batfoxkid:
Adds comments about the conditions in tf2.inc. This is mainly for the [API](https://sm.alliedmods.net/new-api/tf2/TFCond "TFCond · tf2 · SourceMod Scripting API Reference") to show description of the condition. Just a small thing.


KyleSanderson:
Where were these sourced from?

Batfoxkid:
> Where were these sourced from?

https://wiki.teamfortress.com/wiki/Cheats#addcond
Main place and my own experience using them.

asherkin:
I think I would prefer this without the numbers - if the numeric values are useful they should be specified as part of the enum or displayed by the doc tool. Thoughts?

Batfoxkid:
Probably yeah, I do like some way of showing numeric values so that way for people who don't want to use the full name and instead do view_as<Enum>(#)

Issue #1007, opened 2019-05-11
Wrong formating to "Vote Delay Minutes" when delay above 60s


PlayBoy31:
##

[delay formating.zip](https://github.com/alliedmodders/sourcemod/files/3169320/delay.formating.zip)

 Description
set sm_vote_delay above 60s, 900 for example

# Problematic Code (or Steps to Reproduce)
Reply to command after CheckVoteDelay() when delay > 60 on basevotes.sp or funvotes.sp are wrong.

`ReplyToCommand(client, "[SM] %t", "Vote Delay Minutes", delay % 60);`

must be replaced by:

ReplyToCommand(client, "[SM] %t", "Vote Delay Minutes", **(delay / 60)** % 60);

otherwise, seconds are only displayed when its above minute ...

Issue is present on **basevotes.sp** and on **funvotes.sp**

fixed files attached to this issue

[delay formating.zip](https://github.com/alliedmodders/sourcemod/files/3169321/delay.formating.zip)

KyleSanderson:
Can you issue a Pull Request instead of the zip? I'm on a handset and can't review...

PlayBoy31:
> 
> 
> Can you issue a Pull Request instead of the zip? I'm on a handset and can't review...

done

KyleSanderson:
thx let's move there

PR #1009, opened 2019-05-11
Fix wrong formating to "Vote Delay Minutes" when delay above 60s

PlayBoy31:
See https://github.com/alliedmodders/sourcemod/issues/1007

### DIFF
--- a/plugins/basevotes.sp
+++ b/plugins/basevotes.sp
@@ -405,7 +405,7 @@ bool TestVoteDelay(int client)
  	{
  		if (delay > 60)
  		{
- 			ReplyToCommand(client, "[SM] %t", "Vote Delay Minutes", (delay / 60) % 60);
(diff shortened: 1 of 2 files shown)


asherkin:
Unless I'm being silly, shouldn't the `%` just be `/` here?

KyleSanderson:
We still want it to wrap around 60 as that's our current unit of time. FormatTime is likely the most correct fix here.

PlayBoy31:
FormatTime is not appropriate and a time stamp must be used

Some solutions:
1) limit sm_vote_delay to 3600 seconds to cover the only corner case
2) or move function `int FormatTimeDuration(char[] buffer, int maxlen, int time)` from `nextmap.sp` to another place (maybe in `timers.inc` as stock) and use `FormatTimeDuration` to format the delay on basevotes.sp and funvotes.sp

PlayBoy31:
> 
> 
> Unless I'm being silly, shouldn't the `%` just be `/` here?

Ok corrected as your suggest

PR #1011, opened 2019-05-12
Update CS:GO "Reload" offset.

quasemago:


asherkin:
Do you know what was added/removed in the vtable? It is fairly suspicious for only one function to change (or were the others already updated and this one missed?)

psychonic:
Reload is the one function in the file that’s from CBaseCombatWeapon instead of CBasePlayer. It could have changed independently.

psychonic:
Thanks!

PR #1012, opened 2019-05-12
Added cookie methodmap (clientprefs)

JoinedSenses:
This PR adds a methodmap for clientprefs cookies without breaking changes, considering that the cookie Handle is not always the first param for clientprefs functions.

### DIFF
--- a/extensions/clientprefs/natives.cpp
+++ b/extensions/clientprefs/natives.cpp
@@ -471,25 +471,81 @@ cell_t GetClientCookieTime(IPluginContext *pContext, const cell_t *params)
 	return value;
 }
 
+static cell_t Cookie_Set(IPluginContext *pContext, const cell_t *params)
+{
+	// This version takes (handle, client, value). The original is (client, handle, value).
+	cell_t new_params[4] = {
+		3,
+		params[2],
+		params[1],
+		params[3]
+	}
+
+	return SetClientPrefCookie(pContext, new_params);
+}
+
+static cell_t Cookie_Get(IPluginContext *pContext, const cell_t *params)
+{
+	// This verson takes (handle, client, buffer, maxlen). The original is (client, handle, buffer, maxlen).
+	cell_t new_params[5] = {
+		4,
+		params[2],
+		params[1],
+		params[3],
+		params[4]
+	}
+
+	return GetClientPrefCookie(pContext, new_params);
+}
+
+static cell_t Cookie_SetByAuthId(IPluginContext *pContext, const cell_t *params)
+{
+	// Thisv ersion takes (handle, authid, value). The original is (authid, handle, value).
+	cell_t new_params[4] = {
+		3,
+		params[2],
+		params[1],
+		params[3]
+	}
+
+	return SetAuthIdCookie(pContext, new_params);
+}
+
+static cell_t Cookie_GetClientTime(IPluginContext *pContext, const cell_t *params)
+{
+
+	cell_t new_params[3] = {
+		2,
+		params[2],
+		params[1],
+	}
+
+	return GetClientCookieTime(pContext, new_params);
+}
+
 sp_nativeinfo_t g_ClientPrefNatives[] = 
 {
-	{"RegClientCookie",				RegClientPrefCookie},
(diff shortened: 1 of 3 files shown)


JoinedSenses:
If there is a better solution for handling the methods where the Handle is not the first param, please let me know.

psychonic:
> If there is a better solution for handling the methods where the Handle is not the first param, please let me know.

I think for other similar cases, we did the same shim but on the C++ side to give us more flexibility for future.

JoinedSenses:
Happen to have any examples? I spent an hour or two digging for similar cases, but couldnt find anything, although good chance I overlooked it.

By doing it on C++ side, how would it be handled? Creating a new native callback specifically for the methodmap?

psychonic:
Yeah, you'd create a native that just either calls the other native (similar to what you did), or one that calls the same internal function that the other native wraps (with the difference in parameters).

f020b5682ecc0de51a is an example

plugins/include/clientprefs.inc
```diff
@@ -82,18 +82,89 @@ enum CookieMenuAction
...
  *
  * Handles returned by RegClientCookie can be closed via CloseHandle() when
  * no longer needed.
  *
- * @param name			Name of the new preference cookie.
- * @param description	Optional description of the preference cookie.
- * @param access		What CookieAccess level to assign to this cookie.
- * @return				A handle to the newly created cookie. If the cookie already
+ * @param name          Name of the new preference cookie.
+ * @param description   Optional description of the preference cookie.
```
JoinedSenses: All of these formatting changes have been accepted. Just readding them here.

JoinedSenses:
Merge conflict existing due to moving typedef to top of inc, allowing it to be defined for the methodmap.

extensions/clientprefs/natives.cpp
```diff
@@ -471,20 +471,82 @@ cell_t GetClientCookieTime(IPluginContext *pContext, const cell_t *params)
 	return value;
 }
 
+static cell_t Cookie_Set(IPluginContext *pContext, const cell_t *params)
+{
+	// This version takes (handle, client, value). The original is (client, handle, value).
+	cell_t new_params[4] = {
```
Headline: can you mark this `const`?

JoinedSenses: Should this then be made const as well, the commit change I used for reference? https://github.com/alliedmodders/sourcemod/blob/master/core/smn_keyvalues.cpp#L1107

Headline: Yes please!

JoinedSenses: Addressed https://github.com/alliedmodders/sourcemod/pull/1012/commits/60e2a3032192b3a700b2f978d53c2439c9680a3d

JoinedSenses:
Tested with:

```
#include <sourcemod>
#include <clientprefs>

Cookie g_hTestCookie;

public void OnPluginStart() {
	RegAdminCmd("sm_testcookie", cmdTestCookie, ADMFLAG_ROOT);

	g_hTestCookie = new Cookie("TestCookie", "", CookieAccess_Private);
}

public Action cmdTestCookie(int client, int args) {
	if (!client) {
		return Plugin_Handled;
	}

	Cookie testcookie = Cookie.Find("TestCookie");
	PrintToChat(client, "%sFound", testcookie == null ? "Not " : "");
	delete testcookie;

	g_hTestCookie.Set(client, "TestValue");
	CreateTimer(5.0, timerCookie, GetClientUserId(client));
	return Plugin_Handled;
}

Action timerCookie(Handle timer, int userid) {
	int client = GetClientOfUserId(userid);
	if (!client) {
		return Plugin_Continue;
	}

	char buffer[10];
	g_hTestCookie.Get(client, buffer, sizeof(buffer));

	PrintToChat(client, "Result: %s", buffer);
	PrintToChat(client, "Time: %i", g_hTestCookie.GetClientTime(client));
	PrintToChat(client, "Level: %i", g_hTestCookie.AccessLevel);
	return Plugin_Continue;
}
```

Results:
```
Found
Result: TestValue
Time: 1559265624
Level: 2
```

PR #1017, opened 2019-05-15
Re-add missing GetEconItemView param (fixes #1001).

### LINKED ISSUE #1001: 1.10 - build 6405 CSGO CS_OnBuyCommand don't work
* OS:Windows Server 2016
  * GameAPPID:CS:GO/730
  * Current SourceMod version:1.10
  * Current SourceMod snapshot: 6405
  * Current Metamod: Source snapshot:1127

  - [√ ] I have updated SourceMod to the [latest snapshot](https://www.sourcemod.net/downloads.php?branch=dev) and it still happens.
  - [√ ] I have updated SourceMM to the [latest snapshot](https://sourcemm.net/downloads.php?branch=dev) and it still happens.

# Description

public Action CS_OnBuyCommand(int client, const char[] weapon)
{
	PrintToServer("Buy:%s",weapon);
       return Plugin_Handled;
}

No effect..

psychonic:


TheDS1337:
You may also wanna change these two to PassType_Float:
https://github.com/alliedmodders/sourcemod/blob/3e705ae9f77aa1e1910425d59a4ebfba447fd0d2/extensions/cstrike/natives.cpp#L308

and 

https://github.com/alliedmodders/sourcemod/blob/3e705ae9f77aa1e1910425d59a4ebfba447fd0d2/extensions/cstrike/natives.cpp#L330

psychonic:
@TheDS1337 good catch. Do you mind filing a separate issue (or PR?) since it's not directly related to this change?

I think that the difference there only matters for return type, which would explain why that never caused an issue, but it's still wrong and should get fixed.

TheDS1337:
> @TheDS1337 good catch. Do you mind filing a separate issue (or PR?) since it's not directly related to this change?
> 
> I think that the difference there only matters for return type, which would explain why that never caused an issue, but it's still wrong and should get fixed.

Sure, I will do it

PR #1020, opened 2019-05-23
XMM & GP registers support for BinTools on Windows

qubka:
Hello everyone, it just my small opinion about how BinTools can be improved for windows calls with some XMM register which passes float values. Functions like : (OnTerminateRound can be one of the examples of that kind func)

### DIFF
--- a/extensions/cstrike/natives.cpp
+++ b/extensions/cstrike/natives.cpp
@@ -305,7 +305,7 @@ static cell_t CS_TerminateRound(IPluginContext *pContext, const cell_t *params)
 		REGISTER_NATIVE_ADDR("TerminateRound",
 			PassInfo pass[2]; \
 			pass[0].flags = PASSFLAG_BYVAL; \
-			pass[0].type = PassType_Basic; \
+			pass[0].type = PassType_Float; \
 			pass[0].size = sizeof(float); \
 			pass[1].flags = PASSFLAG_BYVAL; \
 			pass[1].type = PassType_Basic; \
@@ -327,7 +327,7 @@ static cell_t CS_TerminateRound(IPluginContext *pContext, const cell_t *params)
 		REGISTER_NATIVE_ADDR("TerminateRound",
 			PassInfo pass[4]; \
 			pass[0].flags = PASSFLAG_BYVAL; \
-			pass[0].type = PassType_Basic; \
+			pass[0].type = PassType_Float; \
 			pass[0].size = sizeof(float); \
 			pass[1].flags = PASSFLAG_BYVAL; \
 			pass[1].type = PassType_Basic; \
--- a/extensions/sdktools/outputnatives.cpp
+++ b/extensions/sdktools/outputnatives.cpp
@@ -337,7 +337,7 @@ static cell_t FireEntityOutput(IPluginContext *pContext, const cell_t *params)
 		pass[6].type = PassType_Basic;
 		pass[6].size = sizeof(CBaseEntity *);
 		pass[7].flags = PASSFLAG_BYVAL;
-		pass[7].type = PassType_Basic;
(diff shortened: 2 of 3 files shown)


extensions/sdktools/vcaller.cpp
```diff
@@ -243,7 +243,12 @@ static cell_t PrepSDKCall_AddParameter(IPluginContext *pContext, const cell_t *p
 	DecodePassMethod(info->vtype, method, info->type, info->flags);
 	info->decflags = params[3] | VDECODE_FLAG_BYREF;
 	info->encflags = params[4];
-
+#ifdef PLATFORM_WINDOWS
+	info->reg = params[5];
```
KyleSanderson: You need to make sure this param is present, this will crash any callers not compiled against the new include.

qubka: Thanks for your reply, I fixed that mistake.

public/extensions/IBinTools.h
```diff
@@ -93,14 +152,19 @@ namespace SourceMod
...
-			: type(type), flags(flags), size(size), fields(fields), numFields(numFields)
+			: type(type), reg(RegisterType_None), flags(flags), size(size), fields(fields), numFields(numFields)
+		{ }
+		
+		PassInfo(PassType type, RegisterType reg, unsigned int flags, size_t size, ObjectField *fields, unsigned int numFields)
+			: type(type), reg(reg), flags(flags), size(size), fields(fields), numFields(numFields)
 		{ }
 		
 		PassType type;			/**< PassType value */
+		RegisterType reg;		/**< RegisterType value */
```
peace-maker: Remember to bump the interface version as well as the minimal interface version.
This is another breaking change like in #705 that was addressed in #979 recently.

extensions/bintools/jit_call.cpp
```diff
@@ -229,6 +229,22 @@ inline void Write_PushFloat(JitWriter *jit, const SourceHook::PassInfo *info, un
 		{
 		case 4:
 			{
+				#ifdef PLATFORM_WINDOWS
```
Headline: all of our preprocessor directives live with no indentation, this should be scooted back to the far left

extensions/bintools/jit_compile.h
```diff
@@ -71,6 +71,49 @@ inline jit_uint8_t _DecodeRegister3(jit_uint32_t val)
 	return 0xFF;
 }
 
+inline jit_uint8_t _DecodeRegister4(SourceHook::PassInfo::RegisterType reg)
+{
+	switch (reg)
+	{
+	case SourceHook::PassInfo::RegisterType_XMM0:
+		{
```
Headline: can we avoid branching here to make this more readible?

also, the owning `case` (and the others) should probably be tabbed over instead of being inline with the switch statement above

public/extensions/IBinTools.h
```diff
@@ -65,6 +65,65 @@ namespace SourceMod
...
+		RegisterType_XMM1,
+		RegisterType_XMM2,
+		RegisterType_XMM3,
+		RegisterType_XMM4,
+		RegisterType_XMM5,
+		RegisterType_XMM6,
+		RegisterType_XMM7,
+
+		// 8-bit general purpose registers
+	    RegisterType_AL,
```
Headline: nit: spacing

Headline:
There’s no way we could take this as-is. Like KyleS said, only make one PR per feature you add. 

Also, all new code must be AlliedModders copyright, always.

qubka:
> There’s no way we could take this as-is. Like KyleS said, only make one PR per feature you add.
> 
> Also, all new code must be AlliedModders copyright, always.

Sorry about that, I will make another PR with serverOS native

PR #1022, opened 2019-05-27
Fix Miscellaneous Regressions and UBs

Headline:
Ran the cppcheck static analyzer against /core/ to check if we were doing any ugliness that I didn't know about. Turns out there's a few situations where we were regressed.

First, and likely the most significant, was `CExtension::FindNextDependency`. This method must not have worked correctly for many many years. All operations were mistakenly done on a default-constructed iterator. 

Secondly, `realloc` can fail. By doing assignment directly on the result of `realloc` we miss the opportunity to check if the retval is null or not. If it *is* null, then our memory allocated previously is still valid and we've just successfully leaked it. For the case involving `InternalFormat`, an exception was added for realloc failures.

Finally, `1<<31` is undefined behavior as `1` is an int32. 

Also switched some casts to static_cast because this is c++, after all...

### DIFF
--- a/core/logic/stringutil.cpp
+++ b/core/logic/stringutil.cpp
@@ -381,7 +381,11 @@ public:
     {
         if (len > max_size)
         {
-            buffer = (char *)realloc(buffer, len);
+            auto *newbuffer = static_cast<char *>(realloc(buffer, len));
+            if (!newbuffer)
+                return nullptr;
+
+            buffer = newbuffer;
             max_size = len;
         }
         return buffer;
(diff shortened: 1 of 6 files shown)


core/logic/stringutil.cpp
```diff
@@ -420,7 +424,11 @@ cell_t InternalFormat(IPluginContext *pCtx, const cell_t *params, int start)
 	{
 		if (maxlen > sizeof(g_formatbuf))
 		{
-			__copy_buf = g_extrabuf.GetWithSize(maxlen);
+			char *tmpbuff = g_extrabuf.GetWithSize(maxlen);
+			if (!tmpbuff)
+				pCtx->ThrowNativeError("Unable to allocate buffer with a size of \"%u\"", maxlen);
```
peace-maker: You should return here.

Headline: Oh; absolutely, thank you

Headline:
Since dependency iteration has never historically functioned correctly, it is now deprecated.

Extensions have a very specific place alongside SourceMod, and their goal is to extend SourceMod's plugin API and provide an *extension* to SourceMod functionality. There was never really a need for this feature in the first place and this change allows us to further solidify our expectations for what an extension is meant to be.

PR #1025, opened 2019-05-29
Travis CI: Swap gcc-7 for gcc-9

Headline:
GCC 7.0 always failed with macro redefinition errors (beyond our control, anyway), so might as well swap for GCC 9.x

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,8 +31,8 @@ addons:
       - g++-4.9
       - g++-5-multilib
       - g++-5
-      - g++-7-multilib
-      - g++-7
+      - g++-9-multilib
+      - g++-9
     cache:
       directories:
         - ../mysql-5.0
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -31,8 +31,8 @@ addons:
       - g++-4.9
       - g++-5-multilib
       - g++-5
-      - g++-7-multilib
-      - g++-7
+      - g++-9-multilib
+      - g++-9
```
KyleSanderson: hmmm. Looks like SM doesn't actually build with gcc-9...

Headline: hmm, mine built ok, but this is quite sad

Kenzzer: image
Checks passed though?

Headline:
Closed since gcc-9 cannot successfully build sourcemod with all hl2sdks unless further invasive intervention occurs, which is not justifiable at this time

Issue #1006, opened 2019-05-10
L4D1 & 2 using the wrong file path for matchmaking_ds


Psykotikism:
# Help us help you
  - [x] I have checked that my issue [doesn't exist yet](https://github.com/alliedmodders/sourcemod/issues).
  - [x] I have tried my absolute best to reduce the problem-space and have provided the absolute smallest test-case possible.
  - [x] I can always reproduce the issue with the provided description below.

# Environment
  * Operating System version: Win & Lin
  * Game/AppID (with version if applicable): 500 & 550
  * Current SourceMod version: 1.10
  * Current SourceMod snapshot: 6407
  - [x] I have updated SourceMod to the [latest version](https://www.sourcemod.net/downloads.php) and it still happens.
  - [x] I have updated SourceMod to the [latest snapshot](https://www.sourcemod.net/downloads.php?branch=dev) and it still happens.

# Description
Both Left 4 Dead games have their matchmaking_ds files inside {mod folder}/bin, and currently SM is set up to look for that file inside {mod folder}/bin/{platform}.

Maybe you guys could make an exception for L4D1&2 to define PLATFORM_FOLDER as "" for those games? Just a suggestion.

For convenience, the file path is defined here: https://github.com/alliedmodders/sourcemod/blob/master/core/logic_bridge.cpp#L633

# Problematic Code (or Steps to Reproduce)
N/A

# Logs
* N/A

asherkin:
Thanks for the report. Please add some detail on how this is affecting you such as explaining what is not working or including error logs you are getting from it not being found.

Psykotikism:
Here are the logs on my L4D2 server upon start-up. I have also attached the plugin and the gamedata file associated with it so you can test for yourself.
[l4d-changemission.zip](https://github.com/alliedmodders/sourcemod/files/3165449/l4d-changemission.zip)

asherkin:
It looks like this might be wrong for all games on Linux, I suspect the paths here are specific to macOS (or possibly just the client distributions in general).

PR #1028, opened 2019-06-01
Fix matchmaking_ds bin not being found for gamedata (#1006).

psychonic:
As reported in #1006, the matchmaking_ds binary is not found since our path to it is incorrect.

This changes our lookup to just use the binary name instead of including the path - similar to how we find tier0 and/or vstdlib to find symbols in those. Only tested on Windows, but that confirms that the binary is loaded by the time our lookup runs, so it should also be fine for dlopen.

### DIFF
--- a/core/HalfLife2.h
+++ b/core/HalfLife2.h
@@ -84,6 +84,9 @@ using namespace SourceMod;
 #define SOURCE_BIN_EXT ".so"
 #endif
 
+#define FORMAT_SOURCE_BIN_NAME(basename) \
+	(SOURCE_BIN_PREFIX basename SOURCE_BIN_SUFFIX SOURCE_BIN_EXT)
+
 struct DataTableInfo
 {
 	struct SendPropPolicy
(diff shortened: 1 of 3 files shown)


core/HalfLife2.h
```diff
@@ -84,6 +84,9 @@ using namespace SourceMod;
 #define SOURCE_BIN_EXT ".so"
 #endif
 
+#define FORMAT_SOURCE_BIN_NAME(basename) \
```
asherkin: Think I'd prefer things without this macro, they're evil.

psychonic: I thought about it. I prefer the macro over duplicated logic, and wasn't sure how to otherwise consolidate the concatenation logic at compile-time without it.

Headline: a `constexpr` function :p

psychonic: @Headline I still couldn't find a way to do that, other than including a separate header's worth of copy/pasted constexpr abuse to allow for compile-time string concatenation. (Or doing it at runtime).

I'm going to merge as-is, but feel free to improve upon it!

PR #1031, opened 2019-06-02
 Added Event methodmap constructor

JoinedSenses:
Event methodmap missing constructor, unlike other methodmaps.

`{"Event.Event",         sm_CreateEvent},`


plugins/include/events.inc
```diff
@@ -74,6 +74,18 @@ typeset EventHook
 
 methodmap Event < Handle
 {
+
+	// Creates a game event to be fired later.
+	//
+	// The Handle should not be closed via CloseHandle().  It must be closed via 
+	// event.Fire() or event.Cancel().
```
asherkin: This is why it doesn't have one - it isn't newable as it isn't deletable.

PR #1034, opened 2019-06-16
Fix @spec not targeting clients in CSGO or CSS unassigned team.

TianQiBuTian:
In CSGO or CSS exit Choose team. It won't be teamid 1, it's 0.

### DIFF
--- a/extensions/sdktools/extension.cpp
+++ b/extensions/sdktools/extension.cpp
@@ -459,11 +459,10 @@ bool SDKTools::ProcessCommandTarget(cmd_target_info_t *info)
 				continue;
 #if SOURCE_ENGINE == SE_CSS || SOURCE_ENGINE == SE_CSGO
 			if (plinfo->GetTeamIndex() != 0 && plinfo->GetTeamIndex() != 1)
-				continue;
 #else
 			if (plinfo->GetTeamIndex() != 1)
-				continue;
 #endif
+				continue;
 			if (playerhelpers->FilterCommandTarget(pAdmin, player, info->flags) ==
 				COMMAND_TARGET_VALID)
 			{
--- a/extensions/sdktools/extension.cpp
+++ b/extensions/sdktools/extension.cpp
@@ -457,15 +457,21 @@ bool SDKTools::ProcessCommandTarget(cmd_target_info_t *info)
 			IPlayerInfo *plinfo = player->GetPlayerInfo();
 			if (plinfo == NULL)
 				continue;
-			if ((plinfo->GetTeamIndex() == 1
 #if SOURCE_ENGINE == SE_CSS || SOURCE_ENGINE == SE_CSGO
-			 || plinfo->GetTeamIndex() == 0
-#endif
-			) && playerhelpers->FilterCommandTarget(pAdmin, player, info->flags) ==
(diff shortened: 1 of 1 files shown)


extensions/sdktools/extension.cpp
```diff
@@ -457,8 +457,11 @@ bool SDKTools::ProcessCommandTarget(cmd_target_info_t *info)
 			IPlayerInfo *plinfo = player->GetPlayerInfo();
 			if (plinfo == NULL)
 				continue;
-			if (plinfo->GetTeamIndex() == 1 &&
-			    playerhelpers->FilterCommandTarget(pAdmin, player, info->flags) ==
+			if ((plinfo->GetTeamIndex() == 1
+#if SOURCE_ENGINE == SE_CSS || SOURCE_ENGINE == SE_CSGO
+			 || plinfo->GetTeamIndex() == 0
```
Headline: I'm really not a fan of the way the formatting of this is.
Could we just `continue` on bad cases to keep a clearer distinction between the two behaviors?

```cpp
#if SOURCE_ENGINE == SE_CSS || SOURCE_ENGINE == SE_CSGO
    if (badcond)
        continue;
#else
    if (otherbadcond)
        continue;
#endif
```

asherkin:
> Can't really comment on whether or not this change is something we'll take, but this logic can be cleaned up.

I'm definitely not a fan of taking this without more justification (if just for the sake of the process), unassigned is not spec (but I can see why it makes sense), but if this is done it should be for all games not gated on CS.

TianQiBuTian:
I don't know the other game can exit "Choose team". But I can exit in CSS or CSGO. Some players in spectator but they exclude @spec.

KyleSanderson:
Spec has different semantics in cstrike, for instance when I played music automatically unassigned was immune. I didn't use this targeting but it was still treated differently. This is the same as the 'White' team in TF (which was legit, TF has 4 active teams) but I think now fired folk papered over years ago. Anyway, DoD:S is the only outlier I think on this for population, HL2:DM can likely weather the changes.

extensions/sdktools/extension.cpp
```diff
@@ -457,12 +457,21 @@ bool SDKTools::ProcessCommandTarget(cmd_target_info_t *info)
 			IPlayerInfo *plinfo = player->GetPlayerInfo();
 			if (plinfo == NULL)
 				continue;
+#if SOURCE_ENGINE == SE_CSS || SOURCE_ENGINE == SE_CSGO
+			if ((plinfo->GetTeamIndex() == 0 || plinfo->GetTeamIndex() == 1) &&
```
Headline: this is just code duplication, please review my proposed changes from earlier

KyleSanderson:
@asherkin how do you feel about this one? Lets merge or give closure to the issue if we can.

KyleSanderson:
@TianQiBuTian these are actually very different scenarios. I don't think we can transparently handle this situation based on this change (although I agree it may indeed be a gap, and we should have a different target covering this). Thank you very much for trying to improve SM, but at this time I don't think we want this upstream. `@spec` indicates a conscious choice to observe, if you're making ramen and come back post level change you're technically not spectating.

PS: While not relevant to AM, this may provide some humour from a simpler time: http://plaguefest.com/threads/abuse-by-afk-man.14335/

Issue #713, opened 2017-10-24
TF2_IgnitePlayer Broken


fearts:
TF2_IgnitePlayer is no longer working.

fearts:
OS: Windows
Issue: Trying to ignite a player does nothing
Error Logs: None

fearts:
Running a test with a simple plugin (https://pastebin.com/fvf4qBEC) to just sm_igniteplayer I notice I hear the flame sound and I also hear my player being hurt. However, no damage happens and after the sounds at the start nothing else happens.

psychonic:
The default value of -1 for the duration no longer works the same as before with the new internal afterburn implementation. (I'm still not sure why there seem to be other cases where the game itself still calls with -1, working presumably fine there, but fix was to hardcode our documented duration of 10).

MitchDizzle:
Why was this hardcoded to 10.0 seconds? Is the value ignored or can it be set to different elapsed times? Might be useful to just add an argument for the time, which defaults to 10 seconds. Just makes it easier to keep the flame on longer/shorter without creating some kind of timer to control the conditions etc.

Headline:
@MitchDizzle You’re correct, an optional parameter to allow for time would be a great idea. 10.0 is entirely arbitrary

psychonic:
> Why was this hardcoded to 10.0 seconds?

...

>  but fix was to hardcode our documented duration of 10).

> The default value of -1 for the duration no longer works

MitchDizzle:
Was just wondering if it was a decision some where to make it only 10 seconds? If not I think it'd be worthwhile to add in another native to set the duration. I'll see if I can make the pull request but my c++ knowledge is very little.

PR #1038, opened 2019-07-05
Update TF2_IgnitePlayer to support set duration of fire

FortyTwoFortyTwo:
See #713 

This PR adds 3rd param `TF2_IgnitePlayer` to allow set duration of fire, while default value kept as 10.0 so older plugins should not break.

However when i tested this, i found out fire duration is capped to 10.0 seconds, even when tempted to set higher than 10. I believe this is valve's shitty fix on infinite fire duration from some weapons, which is really unfortunate. May still be helpful to some plugins when setting fire less than 10 seconds.

Example code when testing this, setting yourself on fire with duration.
```
#include <sourcemod>
#include <tf2>

public void OnPluginStart()
{
	RegConsoleCmd("sm_fire", Command_Fire);
}

public Action Command_Fire(int iClient, int iArgs)
{
	if (GetCmdArgs() == 0)
	{
		TF2_IgnitePlayer(iClient, iClient);
	}
	else
	{
		char sBuffer[4];
		GetCmdArg(1, sBuffer, sizeof(sBuffer));
		float flDuration = StringToFloat(sBuffer);
		TF2_IgnitePlayer(iClient, iClient, flDuration);
	}
	return Plugin_Handled;
}
```

### DIFF
--- a/extensions/tf2/natives.cpp
+++ b/extensions/tf2/natives.cpp
@@ -92,7 +92,7 @@ cell_t TF2_MakeBleed(IPluginContext *pContext, const cell_t *params)
 	return 1;
 }
 
-// native TF2_Burn(client, target)
+// native TF2_Burn(client, target, duration)
 cell_t TF2_Burn(IPluginContext *pContext, const cell_t *params)
 {
 	static ICallWrapper *pWrapper = NULL;
@@ -129,7 +129,7 @@ cell_t TF2_Burn(IPluginContext *pContext, const cell_t *params)
 	void *obj = (void *)((uint8_t *)pEntity + playerSharedOffset->actual_offset);
 	ArgBuffer<void*, CBaseEntity*, CBaseEntity*, 
 										float> //duration
-										vstk(obj, pTarget, nullptr, 10.0f);
+										vstk(obj, pTarget, nullptr, sp_ctof(params[3]));
 
 	pWrapper->Execute(vstk, nullptr);
 	return 1;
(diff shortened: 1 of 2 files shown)


extensions/tf2/natives.cpp
```diff
@@ -129,7 +129,7 @@ cell_t TF2_Burn(IPluginContext *pContext, const cell_t *params)
 	void *obj = (void *)((uint8_t *)pEntity + playerSharedOffset->actual_offset);
 	ArgBuffer<void*, CBaseEntity*, CBaseEntity*, 
 										float> //duration
-										vstk(obj, pTarget, nullptr, 10.0f);
+										vstk(obj, pTarget, nullptr, sp_ctof(params[3]));
```
asherkin: You need to check the provided argument count `params[0]` before accessing `params[3]`, as older plugins won't pass a 3rd argument and it'll be garbage.

MitchDizzle:
This is a non-passive change since you're not adding a new function, you're editing an existing one which can break compiling plugins that use it.

psychonic:
It won't break compilation of existing code since the new parameter is being added with a default value. Plugins that don't have the new argument specified will implicitly use the default.

MitchDizzle:
My bad, definitely didn't see the default value.

FortyTwoFortyTwo:
oops new to those stuff, hopefully i did this correctly

asherkin:
Thanks for the contribution!

Issue #1041, opened 2019-07-10
Private forwards crash when freed mid-call on specific platforms


nosoop:
### Prerequisites
  - I have checked that my issue [doesn't exist yet](https://github.com/alliedmodders/sourcemod/issues).
  - I have tried my absolute best to reduce the problem-space and have provided the absolute smallest test-case possible.
  - I can always reproduce the issue with the provided description below.

### Environment
  * Operating System version: Debian 10 (Buster. 64-bit)
  * SourceMod version: 1.9.0.6281

### Description

Server crashes if the private forward handle being called is `delete`d during one of the callbacks in `Call_Finish()`.  This is something that [NativeVotes](https://github.com/powerlord/sourcemod-nativevotes/) does to be similar in design to the native `Menu` handle type.  Definitely fixable there, but there probably should be safeguards in place on the SourceMod side.

This occurs on my "production" server instance of Debian 10; unfortunately Steam's content servers aren't cooperating and so I can't create another environment to test in.

[More information here.](https://forums.alliedmods.net/showthread.php?p=2658570)

### Problematic Code (or Steps to Reproduce)
  ```
#pragma semicolon 1
#include <sourcemod>

#pragma newdecls required

public void OnPluginStart() {
	RegAdminCmd("call_freed_forward", CallFreedForward, ADMFLAG_ROOT);
}

public Action CallFreedForward(int client, int argc) {
	Handle fwd = CreateForward(ET_Ignore, Param_Cell);
	AddToForward(fwd, INVALID_HANDLE, ForwardCall);
	
	Call_StartForward(fwd);
	Call_PushCell(fwd);
	Call_Finish();
	
	PrintToServer("Finished call");
}

public void ForwardCall(Handle fwd) {
	// this is similar to what the nativevotes menu does -- it frees the handler on its callback
	delete fwd;
}
  ```
Call `call_freed_forward` through the server console.  As previously noted, this plugin crashes on my instance of Debian 10, and does not crash on Ubuntu 16.04 64-bit.

The offending plugin did not crash on Debian 9, so I would assume this code also doesn't.

### Logs

[Crash occurs here.](https://github.com/alliedmodders/sourcemod/blob/14eaa097cb68fb2313036842e1949d58ab534dc8/core/logic/ForwardSys.cpp#L255)  The call stack is otherwise inconsistent and dependent on what the offending plugin was doing.

asherkin:
Nice job tracking this one down!

[...]

PR #1043, opened 2019-07-16
Prevent plugins from freeing active private forward handle

nosoop:
PR is a fix for #1041.

Because the handle is only read once during `sm_CallStartForward`, and to avoid keeping an owned plugin's handle (where the plugin can delete it prior to finishing the call) and interfering with the behavior of `sm_CallFinish` (by changing the order of `ResetCall()` and `pFunction->Execute()`), I opted to clone it twice, which doesn't seem ideal.

I've marked this as a draft to get opinions.  The unimplemented alternatives I can think of would be to either:
- Hold a copy of the calling plugin's `Handle_t` value and ensure its validity at `sm_CallFinish` (which should be acceptable in theory; if the handle was freed prior to the call being finished ~~I'd imagine it would also cause issues~~ it causes a segfault anyways)
- Free the handle outside of `ResetCall()` when SourceMod is finishing the call (I didn't want to modify `ResetCall` with a bool param if I didn't have to)

### DIFF
--- a/core/logic/smn_functions.cpp
+++ b/core/logic/smn_functions.cpp
@@ -35,6 +35,7 @@
 #include <IForwardSys.h>
 #include <ISourceMod.h>
 #include <amtl/am-autoptr.h>
+#include "AutoHandleRooter.h"
 
 HandleType_t g_GlobalFwdType = 0;
 HandleType_t g_PrivateFwdType = 0;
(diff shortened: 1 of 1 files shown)


KyleSanderson:
I don't have time atm but the ideal fix is to mark MM (handlesys) as protected in weird circumstances like that and error on the second call.

nosoop:
I'm not sure what you mean by MM and marking it as protected; can't find anything in the handlesys-related core / public headers with that terminology.

nosoop:
Closing as #1048 looks better suited for fixing the issue.

PR #1044, opened 2019-07-17
Removal of unneeded underscores.

404UNFca:
Just a nitpicky change.


Headline:
This change doesn’t really *do* anything, and there’s really no need for us to make changes like this.

I suppose this could break somebody, too, if they’re `wyld`

PR #1046, opened 2019-07-18
Workaround CS:GO Stringtable bug

asherkin:
The cause of the infamous "Index error writing string table baseline"
error appears to be a timing issue between the engine's network message
processing, the stringtable encoding, and command processing when adding
stringtable entries in OnConfigsExecuted.

When the first client connects the map is re-loaded which causes a full
refresh, the game's stringtable entries are added at tick 65, the client
connection is registered at tick 66, and stringtable entries added in
OnConfigsExecuted are registered as being added in tick 67. The engine
later calls WriteBaselines with the client's signon tick, and neglects
to encode the SM added entries as it considers them from the future.

To avoid this, always pass INT_MAX as the current tick when encoding
the baseline, so all stringtable entries are included regardless of when
they were added. Tested on both Windows and Linux.


peace-maker:
Are you sure there are no side effects from always passing `INT_MAX` instead of only for the first client?

asherkin:
It’s definitely fine, for later clients their signon tick is already far in the future than the added entries.

KyleSanderson:
Can we get this is cstrike?

asherkin:
> Can we get this in cstrike?

As far as I know CS:GO is the only engine branch with this delta stringtable encoding, all the older games unconditionally send the entire table on signon and never again.

KyleSanderson:
> > Can we get this in cstrike?
> 
> As far as I know CS:GO is the only engine branch with this delta stringtable encoding, all the older games unconditionally send the entire table on signon and never again.

Which is why it should have been in cstrike ;-).

PR #1048, opened 2019-07-21
Handlesys: Introduce concept of exclusive identity handles

KyleSanderson:
This one is likely a doozy (and not even compile tested). This is to fix the freeing during callback problems that plague us to throw an error immediately mid-callback. menusys, functions, timers, anything with a delayed callback can cause us problems. I think this is the best option after sitting on it for a couple days and reviewing this morning least invasive approach we can take.

https://github.com/alliedmodders/sourcemod/issues/1041

### DIFF
--- a/core/logic/HandleSys.cpp
+++ b/core/logic/HandleSys.cpp
@@ -1169,3 +1169,16 @@ void HandleSystem::Dump(const HandleReporter &fn)
 	rep(fn, "-- Approximately %d bytes of memory are in use by Handles.\n", total_size);
 }
 
+HandleError HandleSystem::GetHandleAccess(Handle_t handle, HandleAccess &*pAccess)
+{
+	unsigned int index;
+	QHandle *pHandle;
+	HandleError err;
+	IdentityToken_t *ident = NULL;
+
+	if ((err=GetHandle(handle, ident, &pHandle, &index)) != HandleError_None)
+		return err;
+
+	pAccess = &(pHandle->sec);
+	return err;
+}
--- a/core/logic/HandleSys.cpp
+++ b/core/logic/HandleSys.cpp
@@ -532,6 +532,9 @@ bool HandleSystem::CheckAccess(QHandle *pHandle, HandleAccessRight right, const 
 	/* Check if the owner is allowed */
 	if (access & HANDLE_RESTRICT_OWNER)
 	{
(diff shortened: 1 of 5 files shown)


nosoop:
:eyes:  This looks much more robust.

I'll run it against my test case and report back on whether or not they crash when I'm not on snooze.

dvander:
What problem is this solving?

KyleSanderson:
> What problem is this solving?
Immediate errors during a callback when it's actively in-use. Fyren mentioned we should continue to just paper over but an error here feels a little more appropriate (to me atleast, obviously debatable; the example of creating a timer to free memory or waiting to breakout of the callback isn't ideal).

nosoop:
Apologies for the delay.

Seems like it still crashes at the same spot in `CForward::Execute`; attached SM is built from 155248a in case that'd help if you also want to repro with the environment (only built with TF2).

Also have accelerator submissions: [OXOR-TW5T-FB4A], [5IOG-N6TT-SK3F], [23U6-PPR7-UD7L].

[OXOR-TW5T-FB4A]: https://crash.limetech.org/oxortw5tfb4a
[5IOG-N6TT-SK3F]: https://crash.limetech.org/5iogn6ttsk3f
[23U6-PPR7-UD7L]: https://crash.limetech.org/23u6ppr7ud7l
[sourcemod.tar.gz](https://github.com/alliedmodders/sourcemod/files/3438139/sourcemod.tar.gz)

Did some logging tests; from my understanding it seems like `access_special` isn't set so it's still pulling it from the handle type?

KyleSanderson:
@nosoop so, this is completely on me. I let this one fester and lay dormant (to the point of needing to be re-based). My apologies.

What's your current take on this one? What do you feel would be the best way forward here? Is this still an issue in SM?

nosoop:
I think this approach is fine once the access inheritance issue is ironed out, but I don't have knowledge of the handle system to have an opinion on other approaches.  I'm also mostly indifferent about whether it throws an error or is handled transparently; I can see valid reasons for both.

I think it's still worth resolving on SourceMod's side &mdash; the most recent discussion related to this crash was on August 3 this year.

KyleSanderson:
@TheDS (or @dvander) how do you think we can fix this MM issue in SM? I still think taking a parent clone is the right path but I'm totally open to other solutions.

PR #1050, opened 2019-07-24
Update SourcePawn to 093edef

asherkin:
Includes the following SourcePawn changes:
- PR#339 Fix various compilation problems
- PR#340 Restyle spcomp
- PR#344 Fix uninitialized variable guarding profiling scopes
- PR#346 Fix Travis Emscripten build
- PR#345 Simplify inline exit frames


asherkin:
Whoops, looks like that is why the pthread flags were getting doubled up in alliedmodders/sourcepawn#346

PR #1053, opened 2019-07-29
Check short name for remote extensions

nosoop:
Checks the filename passed in via `LoadExternal` for remote extensions for same-file tests.

This fixes `FindExtensionByFile` not finding an existing remote extension in the case where the filepath is not equal to the filename (and subsequently loading it twice) and allows plugins to successfully load when specifying that name as a dependency.

### DIFF
--- a/core/logic/ExtensionSys.cpp
+++ b/core/logic/ExtensionSys.cpp
@@ -1364,7 +1364,7 @@ bool CLocalExtension::IsSameFile(const char *file)
 
 bool CRemoteExtension::IsSameFile(const char *file)
 {
-	/* :TODO: this could be better, but no one uses this API anyway. */
-	return strcmp(file, m_Path.c_str()) == 0;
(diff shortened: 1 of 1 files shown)


asherkin:
What are you using the remote extension API for? As per the existing comment here we don't really expect there to be any consumers.

This change looks fine - is there any reason to check the path rather than just the short name?

nosoop:
I'm exposing some natives from [RCBot2](https://github.com/nosoop/rcbot2) (branch not pushed yet) while keeping it as a MetaMod:Source plugin.  I'm sure there's other ways to handle it, but the first useful resource I found while working on it was [this wiki entry](https://wiki.alliedmods.net/Writing_Extensions_from_Metamod:Source_Plugins), so I went with that.

I kept the path handling as-is for backwards compatibility since it's already there.

dvander:
Indeed this code is basically untested as we never really supported it. This change looks harmless though so let's take it.

PR #1054, opened 2019-07-29
NPOTB: Enable GCC-9 builds in Travis-CI.

KyleSanderson:
Confirming the claim that GCC9 actually builds a complete SM... :-).


KyleSanderson:
Incomplete.

PR #1056, opened 2019-07-31
Add File.Flush methodmap

hydrogen-mvm:
Here is a snippet from the current TF2 event I'm working on:
```
g_FlatFile.WriteLine("wave,%d,%d,%d,%d", Steam3, TimeStamp, MissionIndex, WaveNumber);
FlushFile(g_FlatFile);
```
I would like to be able to flush the file like this so that I can consistently use methodmaps on the file handle instead of flip flopping between legacy and new API:
```
g_FlatFile.WriteLine("wave,%d,%d,%d,%d", Steam3, TimeStamp, MissionIndex, WaveNumber);
g_FlatFile.Flush();
```
Here is a patch (one for smn_filesystem.cpp and one for files.inc) that adds the Flush method to the file methodmap. Please review and let me know if I accidentally broke something or violated any coding conventions. I don't have a C++ compiler so I wasn't able to test if this will break the buildbots or CI.

Thanks.

### DIFF
--- a/core/logic/smn_filesystem.cpp
+++ b/core/logic/smn_filesystem.cpp
@@ -1195,6 +1195,7 @@ REGISTER_NATIVES(filesystem)
 	{"File.WriteLine",			sm_WriteFileLine},
 	{"File.EndOfFile",			sm_IsEndOfFile},
 	{"File.Seek",				sm_FileSeek},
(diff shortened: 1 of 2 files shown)


psychonic:
Definitely looks like it was an oversight. Thanks!

PR #1058, opened 2019-08-04
Create math class with basic math functions

stickz:
```
int abs = Math.Abs(-1);
int minB = Math.MinBounds(3, 10);
int minV = Math.MinValue(3, 10);
int maxB = Math.MaxBounds(10, 1);
int maxV = Math.MaxValue(10, 1);
float clamp = FMath.Clamp(0.5, 1.8, 3.9);
```

### DIFF
--- a/plugins/include/math.inc
+++ b/plugins/include/math.inc
@@ -1,0 +1,60 @@
+#if defined _math_included
+  #endinput
+#endif
+#define _math_included
+
+methodmap Math 
+{
+	/**
+	 * Converts a negative value into a positive value.
+	 *
+	 * @param number		A number that can be positive or negative.
+	 * @return				Positive number.
+	 */
+	public static any Abs(any number)
+	{
+		return number >= 0 ? number : (number * -1);
+	}
+	
+	/**
+	 * Sets a number to a given min value
+	 * if the value is smaller than min.
+	 * 
+	 * @param value			Value
+	 * @param min			Min Value used as lower border
+	 * @return				Correct value not lower than min
+	 */
+	public static any Min(any value, any min)
+	{
+		return value < min ? min : value;
+	}
+	
+	/**
+	 * Sets a number to a given max value, 
+	 * if the value is greater than max.
+	 * 
+	 * @param value			Value
+	 * @param max			Max Value used as upper border
+	 * @return				Correct value not upper than max
+	 */
+	public static any Max(any value, any max) 
+	{	
+		return value > max ? max : value;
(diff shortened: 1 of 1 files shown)


peace-maker:
You've got a funny definition of `min` and `max` there. Usually `min` returns the smaller of the two arguments and `max` the larger one.

```c
#define MIN(a,b) (((a)<(b))?(a):(b))
#define MAX(a,b) (((a)>(b))?(a):(b))
```

stickz:
This is definition of min and max is more applicable for sourcemod. Code like this was occurring in my plugins constantly, which is why I choose to adopt that definition. smlib also uses this definition.

```
int value = 3;
if (value < 5)
    value = 5;
else if (value > 10)
    value = 10;
```

stickz:
> You've got a funny definition of `min` and `max` there. Usually `min` returns the smaller of the two arguments and `max` the larger one.
> 
> ```c
> #define MIN(a,b) (((a)<(b))?(a):(b))
> #define MAX(a,b) (((a)>(b))?(a):(b))
> ```
Alright, I just updated the math functions. They include both bounds and value now.

KyleSanderson:
You're looking for 'clamp' from C... Anyway, shouldn't it worry you greatly that you're adding these stable math functions where the definitions are changing on a whim.

peace-maker:
Arithmetic with the `any` type will always treat the value as an int, so those functions don't work as expected with floats. There needs to be some kind of warning about that and possibly float versions of those functions as well?

stickz:
Sourcepawn doesn't support overloading in methodmaps. I can't define both an int and float function with the same name. So I created anther methodmap called FMath for floats and set ints for the other. The user will receive a tag mismatch warning, if they use the wrong methodmap.

asherkin:
Thank you for the contribution but I don't think this is something we want to include in SourceMod at this time.

From the comments here and how it has increased in complexity show, it is not as straightforward as it first appears and there are multiple conflicting ideas of what these functions should do. Additionally, I do not think these functions would be discoverable as-implemented and would likely lead to increased developer confusion. [SMLIB](https://github.com/bcserv/smlib/tree/transitional_syntax) is a popular collection of utility stocks and functions if you would like to use these kinds of helpers in your plugins.

Please have a read of [CONTRIBUTING.md](https://github.com/alliedmodders/sourcemod/blob/master/.github/CONTRIBUTING.md#pull-requests) for what we look for in PRs and other contributions to SourceMod.

Issue #1057, opened 2019-08-01
Func ReplaceString() problem with UTF8


PlayBoy31:
Hello,

There is a strange issue with ReplaceString() function

with this code:

> #define STRING_MESSAGE_SIZE 191
> 
> Message is set as : "в доме"
> 
> ReplaceString(Message, STRING_MESSAGE_SIZE, "´", "");

The ReplaceString function detect that there is a "´" in "в доме" whereas it shouldn't

asherkin:
SourceMod's String API only supports standard ASCII by design, see also #978

PlayBoy31:
What can i do ? is there a way to convert all to ascii ?

have SM dev in target to implement full UTF8 support ? in this case when?

KyleSanderson:
The vast majority of games do not support full UTF8 (which is why we haven't, security problems)... There are tricks to get it in there (and SM does have partial coverage through DBI) but it's not on any road-map currently to have it in the language natively.

asherkin:
You can write your own multibyte-safe ReplaceString using the IsCharMB function.

PR #1060, opened 2019-08-08
Allow to escape string even if buffer is not 2*strlen(string)+1.

PlayBoy31:
The problem with the actual code that if the buffer is not size x 2 +1 the function returns false even if it was possible to escape string.

I suggest this implementation to allow this case. First call the basic function mysql_real_escape_string() then test if needed is in range of maxlength. The return of mysql_real_escape_string() return -1 if failure and allow the native to return false.

### DIFF
--- a/extensions/mysql/mysql/MyDatabase.cpp
+++ b/extensions/mysql/mysql/MyDatabase.cpp
@@ -152,24 +152,23 @@ const char *MyDatabase::GetError(int *errCode)
 bool MyDatabase::QuoteString(const char *str, char buffer[], size_t maxlength, size_t *newSize)
 {
 	unsigned long size = static_cast<unsigned long>(strlen(str));
-	unsigned long needed = size * 2 + 1;
-
-	if (maxlength < needed)
-	{
-		if (newSize)
-		{
-			*newSize = (size_t)needed;
-		}
-		return false;
-	}
+	unsigned long needed = 0;
 
 	needed = mysql_real_escape_string(m_mysql, buffer, str, size);
+	
 	if (newSize)
 	{
 		*newSize = (size_t)needed;
 	}
-
-	return true;
+	
+	if (maxlength < needed)
+	{
+		return false;
+	}
+	else
+	{
+		return true;
+	}
 }
 
 bool MyDatabase::DoSimpleQuery(const char *query)


asherkin:
Thank you for the PR, this restriction is intentional to avoid incorrect use of the function and is specified by the upstream MySQL connection documentation.

PlayBoy31:
why close ? My implementation is correct and is safe against upstream MySQL connection

peace-maker:
What's your use case for this?
Supporting invalid buffer sizes will introduce subtle bugs into plugins, where escaping inconsistently fails sometimes if the username is too long.

Why would you ever not pass a long enough buffer to potentially escape every character in the string?

PlayBoy31:
> 
> 
> What's your use case for this?
> Supporting invalid buffer sizes will introduce subtle bugs into plugins, where escaping inconsistently fails sometimes if the username is too long.
> 
> Why would you ever not pass a long enough buffer to potentially escape every character in the string?

The final goal of this function is to store string in a sql field. for example, my field in sql is 128, and the text to encode is 100. in this text no special char to encode, so the size is still the same 100. But the function will return false if it was possible.

Another problem, if i change my buffer on plugin, then use a format function to store to 128 becasue of mysql field limit i can have a problem like this => when doing the format the format function can split a => \" then we will have an SQL error => "incorrect string value: "

peace-maker:
The escaping is only done in the query. It adds backslashes in front of a few characters like `'` to avoid the values from breaking the query structure. Even if every character of your 128 field would need to be escaped, it's only escaped in the query on its way to your database server. The server unescapes the data and only stores 128 characters again.

So just use proper buffer sizes for your queries to avoid truncation.

asherkin:
Awesome, the check did its job - improper user was found immediately instead of being a landmine to discover later.

PlayBoy31:
> 
> 
> The escaping is only done in the query. It adds backslashes in front of a few characters like `'` to avoid the values from breaking the query structure. Even if every character of your 128 field would need to be escaped, it's only escaped in the query on its way to your database server. The server unescapes the data and only stores 128 characters again.
> 
> So just use proper buffer sizes for your queries to avoid truncation.

Here a proof of problem:

// Prepare message
			decl String:Message_Formated[STRING_MESSAGE_SIZE] = ""; // STRING_MESSAGE_SIZE is 191
			new String:TmpMessage[(2*STRING_MESSAGE_SIZE)+1] = ""; 
					
			if (SQL_EscapeString(chats_db, Message, TmpMessage, sizeof(TmpMessage)) == false)
			{
				LogError("[Chats] SQL_EscapeString failure for \"%s\" (%s) with message \"%s\".", g_name[client], g_authid[client], Message);
			}
			else
			{
				/* SQL_EscapeString ok */
				Format(Message_Formated, sizeof(Message_Formated), "%s", TmpMessage);
			}

Message is (131 lenth) : 🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬🐬

=>Incorrect string value: '\xF0\x9F' for column `sourcemod`.`chats`.`message` at row 1.

asherkin:
🐬 is a multi-byte character, each one needs 4 bytes, your `Message_Formated` array is far too small and is cutting a character in the middle. This was discussed in #1057.

PlayBoy31:
> 
> 
> 🐬 is a multi-byte character, each one needs 4 bytes, your `Message_Formated` array is far too small and is cutting a character in the middle. This was discussed in #1057.

Ok, so is it possible to integrate this logic to SM  InternalFormat() in order to protect format function to not cut character in the middle ?

logic:
`new len = strlen(Message);

for (new i = 0; i < len; i++)
			{
				new bytes = IsCharMB(Message[i]);
				
				if (bytes > 1)
				{
					if (len - i < bytes)
					{
						Message[i] = '\0';
						break;
					}
					
					i += (bytes - 1);
				}
			}`

This is a sourcepawn code but need to be converted to C

We can use a optional bool parameter set as false by default, like: bool: safe_character_cut in format() prototype to do this code only when true

PR #1068, opened 2019-09-02
Allow RequestFrameCallback not require data param

FortyTwoFortyTwo:
When i want to call `RequestFrame` without any data param, `RequestFrameCallback` still requires me to add pointless data param.
```
public void OnPluginStart()
{
	RequestFrame(OnFrame);
}

public void OnFrame(int pointless)	//pointless param
{
	//do something
}
```

I want to allow plugin to be compiled without needing data param, making it look bit more nicer when not passing any data.
```
public void OnPluginStart()
{
	RequestFrame(OnFrame);
}

public void OnFrame()	//param not needed as RequestFrame not passing any data
{
	//do something
}
```

### DIFF
--- a/plugins/include/functions.inc
+++ b/plugins/include/functions.inc
@@ -615,7 +615,10 @@ native int FormatNativeString(int out_param,
  *
  * @param data          Data passed to the RequestFrame native.
  */
-typedef RequestFrameCallback = function void (any data);
+typeset RequestFrameCallback {
+	function void ();
+	function void (any data);
+}
 
 /**
  * Creates a single use Next Frame hook.


Headline:
What are you using request frame for that made you want to make this contribution?

FortyTwoFortyTwo:
One simple reason is when i have global variable(s) for stuffs around the code, I may need `RequestFrame` to reset, set or increase by 1 a frame later, without needing any extra params to set.

I could say same thing to timers, and `typeset Timer` already supports data param not required.

KyleSanderson:
haha. The param is still there - it's just 0...

PR #1071, opened 2019-09-02
Avoid losing console messages.

BotoX:
Buffers up to 16k bytes of SVC_Print if buffer would overflow, then sends chunks every frame.
Sends up to 2048 bytes per frame and does not split messages.

Only tested on cstrike, works without any changes on plugins.
On client the order is still messed up sometimes because the client itself fucks up the order (also on console commands like find)

### DIFF
--- a/core/PlayerManager.cpp
+++ b/core/PlayerManager.cpp
@@ -95,9 +95,10 @@ SH_DECL_EXTERN0_void(ConCommand, Dispatch, SH_NOATTRIB, false);
 #endif
 SH_DECL_HOOK2_void(IVEngineServer, ClientPrintf, SH_NOATTRIB, 0, edict_t *, const char *);
 
-static void FrameHook(bool simulating)
+static void PrintfBuffer_FrameAction(void *data)
 {
-	g_Players.OnGameFrame();
+	int client = (int)(intptr_t)data;
+	g_Players.OnPrintfFrameAction(client);
 }
 
 ConCommand *maxplayersCmd = NULL;
@@ -212,12 +213,10 @@ void PlayerManager::OnSourceModAllInitialized()
 		SH_ADD_HOOK(ConCommand, Dispatch, pCmd, SH_STATIC(CmdMaxplayersCallback), true);
 		maxplayersCmd = pCmd;
 	}
-	g_SourceMod.AddGameFrameHook(&FrameHook);
 }
 
 void PlayerManager::OnSourceModShutdown()
 {
-	g_SourceMod.RemoveGameFrameHook(&FrameHook);
 	SH_REMOVE_HOOK(IServerGameClients, ClientConnect, serverClients, SH_MEMBER(this, &PlayerManager::OnClientConnect), false);
 	SH_REMOVE_HOOK(IServerGameClients, ClientConnect, serverClients, SH_MEMBER(this, &PlayerManager::OnClientConnect_Post), true);
 	SH_REMOVE_HOOK(IServerGameClients, ClientPutInServer, serverClients, SH_MEMBER(this, &PlayerManager::OnClientPutInServer), true);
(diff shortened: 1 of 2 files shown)


core/PlayerManager.cpp
```diff
@@ -846,6 +857,146 @@ void PlayerManager::OnClientDisconnect_Post(edict_t *pEntity)
...
+		RETURN_META(MRES_SUPERCEDE);
+		return;
+	}
+
+	SH_CALL(engine, &IVEngineServer::ClientPrintf)(pEdict, szMsg);
+
+	RETURN_META(MRES_SUPERCEDE);
+}
+
+void PlayerManager::OnGameFrame()
```
peace-maker: Instead of this frame hook which is looping all players every frame, while the overflow this prevents is actually rare, you'd rather use `AddFrameAction` per player to reduce the overhead of this to only when it's needed.

core/PlayerManager.cpp
```diff
@@ -846,6 +857,146 @@ void PlayerManager::OnClientDisconnect_Post(edict_t *pEntity)
...
+	INetChannel *pNetChan = static_cast<INetChannel *>(engine->GetPlayerNetInfo(client));
+	if(pNetChan == NULL)
+		return;
+
+	CPrintfBuffer *pPrintfBuffer = &m_Players[client].m_PrintfBuffer;
+
+	size_t nMsgLen = strlen(szMsg);
+	int nNumBitsWritten = pNetChan->GetNumBitsWritten(false);
+
+	// 2048 = SVC_Print buffersize (-1 for \0)
```
peace-maker: Some named constants instead of these magic values would be nice. Are they the same for all engines?

core/PlayerManager.cpp
```diff
@@ -846,6 +857,146 @@ void PlayerManager::OnClientDisconnect_Post(edict_t *pEntity)
...
+	{
+		// Don't send any more messages for this player until the buffer is empty.
+		pPrintfBuffer->m_StopSend = true;
+
+		pPrintfBuffer->Append(szMsg, nMsgLen);
+		RETURN_META(MRES_SUPERCEDE);
+		return;
+	}
+
+	SH_CALL(engine, &IVEngineServer::ClientPrintf)(pEdict, szMsg);
```
peace-maker: You can just return `MRES_IGNORED` instead of manually calling and returning `MRES_SUPERCEDE`.

core/PlayerManager.cpp
BotoX:
```C++
bool SVC_Print::WriteToBuffer( bf_write &buffer )
{
	buffer.WriteUBitLong( GetType(), NETMSG_TYPE_BITS );
	return buffer.WriteString( m_szText ? m_szText : " svc_print NULL" );
}
```
NETMSG_TYPE_BITS is 5 in 2007 code, can't tell for sure if it's 6 in sdk2013 or I did that by trial&error or for safety reasons.
2048 is the size of the buffer in SVC_Print (also 2007 code), -1 for terminating \0 at the end of the string.
This just checks if the string is too big to fit into the current SVC_Print buffer so it doesn't have to be 100% accurate, the 2048 bufsize in SVC_Print is the most important.

core/PlayerManager.h
```diff
@@ -67,6 +67,32 @@ union serial_t
...
 
+class CPrintfBuffer
+{
+public:
+	struct CChunk
+	{
+		void *m_pMem;
+		char *m_pMessage;
+		size_t m_Length;
+		struct CChunk *m_pNext;
```
Headline: Why are we making our own list here, can we use an AMTL container with AString?

core/PlayerManager.cpp
```diff
@@ -846,6 +857,146 @@ void PlayerManager::OnClientDisconnect_Post(edict_t *pEntity)
 	}
 }
 
+/* CPrintfBuffer */
+CPrintfBuffer::CPrintfBuffer(size_t MaxLength)
+{
+	m_pFirstChunk = NULL;
```
Headline: all new code should use nullptr, if you see other `NULL`s around your browsing feel feee to change em too

core/PlayerManager.cpp
```diff
@@ -846,6 +857,146 @@ void PlayerManager::OnClientDisconnect_Post(edict_t *pEntity)
...
+
+bool CPrintfBuffer::Append(const char *pString, size_t Length)
+{
+	if(Length <= 0 || Length >= 1023)
+		return false;
+
+	if(m_Length >= m_MaxLength)
+		return false;
+
+	CChunk *pChunk = new CChunk();
```
Headline: I’d like to avoid us allocating our own memory, by using an amtl container we can avoid `new`

peace-maker:
We should reset the buffer in `CPlayer::Disconnect`.

core/PlayerManager.h
```diff
@@ -152,6 +153,8 @@ class CPlayer : public IGamePlayer
 #if SOURCE_ENGINE == SE_CSGO
 	QueryCvarCookie_t m_LanguageCookie = InvalidQueryCvarCookie;
 #endif
+	ke::LinkedList<ke::AString> m_PrintfBuffer;
```
peace-maker: amtl docs recommend to avoid `LinkedList` and use `Vector` instead.

core/PlayerManager.cpp
```diff
@@ -846,6 +856,83 @@ void PlayerManager::OnClientDisconnect_Post(edict_t *pEntity)
...
+		int nNumBitsWritten = 0;
+#else
+		int nNumBitsWritten = pNetChan->GetNumBitsWritten(false); // SVC_Print uses unreliable netchan
+#endif
+
+		const int NETMSG_TYPE_BITS = 6; // SVC_Print overhead for netmsg type
+		const int SVC_Print_BufferSize = 2048 - 1; // -1 for terminating \0
+
+		// stop if we'd overflow the SVC_Print buffer
+		if((nNumBitsWritten + NETMSG_TYPE_BITS) / 8 + nMsgLen >= SVC_Print_BufferSize)
```
peace-maker: What happens if someone tries to write 2048 characters in one call to ClientPrintf? Is that possible?
The current implementation seems to block console messages forever in that case.
We'll need to send just the allowed amount of bytes per frame instead of trying to send the items as a whole.

BotoX: In SourceMod at least the maximum is 1024. But I guess this should be handled properly in case the engine tries to pass >= 2048

Headline:
I'm cool with code quality, here. Dismissing the rest to others

core/PlayerManager.cpp
```diff
@@ -846,6 +856,83 @@ void PlayerManager::OnClientDisconnect_Post(edict_t *pEntity)
...
+
+		iter = m_Players[client].m_PrintfBuffer.erase(iter);
+	}
+
+	// Buffer is empty, send new messages without enqueuing again.
+	if (iter == m_Players[client].m_PrintfBuffer.end())
+		m_Players[client].m_PrintfStop = false;
+	else
+	{ // buffer not empty yet, continue processing it on the next gameframe.
+		g_pSM->AddFrameAction(PrintfBuffer_FrameAction, (void *)(intptr_t)client);
```
Headline: One more tidbit for windows builds, we're already in core so we can just use SourceModBase directly. 

sub `g_pSM->`  for `g_SourceMod.`

core/PlayerManager.h
BotoX: I don't see the point of using a vector here, both use new to allocate the element.
The vector has to grow sometimes using am_malloc, the list has to am_malloc on every insert to allocate a new list item.
I pop elements from the front, the vector always has to reorder itself after that, the list can just unlink an item.
So the difference is:
append: vector 1 new, maybe grow+move | list: 1 new & 1 (small) am_malloc
popFront: vector reorder all elements | list: just set two pointers

KyleSanderson: You're always going to cache miss on the list, on the vector the processor has a chance in hell of predicting.

Fyren: amtl has a deque type that's basically a ring buffer.

core/PlayerManager.cpp
```diff
@@ -846,6 +855,95 @@ void PlayerManager::OnClientDisconnect_Post(edict_t *pEntity)
...
+	while (!player.m_PrintfBuffer.empty())
+	{
+#if SOURCE_ENGINE == SE_EPISODEONE
+		int nNumBitsWritten = 0;
+#else
+		int nNumBitsWritten = pNetChan->GetNumBitsWritten(false); // SVC_Print uses unreliable netchan
+#endif
+
+		const int NETMSG_TYPE_BITS = 5; // SVC_Print overhead for netmsg type
+		const int SVC_Print_BufferSize = 2048 - 1; // -1 for terminating \0
```
Fyren: These should have a common definition instead of one in each function that uses them.

Fyren:
Has this been tried on something other than BotoX's CSS servers?

core/PlayerManager.cpp
Headline: adding onto fyren, `constexpr int` works great here

BotoX:
I didn't try it anywhere else, but SVC_Print *should* be the same across all games/engines.
I mean really all this does is buffer messages when the unreliable netchan has more than ~2000 bytes in it and send them in the next tick.
There's barely anything else that uses the unreliable netchan, especially not anything as big as SVC_Print.
So it could be less performant in other games if other stuff uses up the unreliable netchan and then console messages would be a bit slower.
However that's still better than outright losing them.

It obviously makes sense to test this on other games before merging, just put 64 bots on a server and type status. (should yield more than 2048 characters)
Without this patch the message will be cut off and you'll get this error:
`Netchannel: failed reading message svc_Print from <serverip>`

KyleSanderson:
I've likely bikeshedded this to death. Just needs testing then should be good to ship.

nosoop:
Verified working on CS:GO and TF2 (Linux) with the PR merged into branch 1.10-dev.  Tested with the following plugin:
```sourcepawn
#pragma semicolon 1
#include <sourcemod>

#pragma newdecls required

public void OnPluginStart() {
	RegAdminCmd("sm_overflowsvcprint", OverflowConsoleBuffer, ADMFLAG_ROOT);
}

public Action OverflowConsoleBuffer(int client, int argc) {
	for (int i; i < 512; i++) {
		PrintToConsole(client, "%04d: %08x %08x %08x %08x %08x %08x %08x %08x",
				i, GetURandomInt(), GetURandomInt(), GetURandomInt(), GetURandomInt(),
				GetURandomInt(), GetURandomInt(), GetURandomInt(), GetURandomInt());
	}
	return Plugin_Handled;
}
```

KyleSanderson:
Thank you very much, sorry for the huge delay.

PR #1072, opened 2019-09-02
cstrike: Fix wrong timeleft calculation

BotoX:
(I actually forgot what exactly the buggy behavior was)

### DIFF
--- a/extensions/cstrike/timeleft.cpp
+++ b/extensions/cstrike/timeleft.cpp
@@ -58,7 +58,17 @@ void TimeLeftEvents::FireGameEvent(IGameEvent *event)
 		if (get_new_timeleft_offset || !round_end_found)
 		{
 			get_new_timeleft_offset = false;
-			timersys->NotifyOfGameStart();
+
+			float flGameStartTime = gpGlobals->curtime;
+			sm_sendprop_info_t info;
+			if (gamehelpers->FindSendPropInfo("CCSGameRulesProxy", "m_flGameStartTime", &info))
+			{
+				uintptr_t gamerules = (uintptr_t)g_pSDKTools->GetGameRules();
+				if (gamerules)
+					flGameStartTime = *(float *)(gamerules + info.actual_offset);
+			}
+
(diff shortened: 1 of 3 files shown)


BotoX:
[8:18 PM] Moltard: "cstrike: Fix wrong timeleft calculation"
wasnt it that the timeleft from the scoreboard and the sourcemod timeleft were different, and the vote happened on the sourcemod timeleft. but final round happened on the scoreboard timeleft
[8:19 PM] Moltard: the difference between the 2 happened after doing a mp_restartgame 1 in maps
[8:24 PM] zaCade: ye
[8:24 PM] zaCade: that was it
[8:25 PM] zaCade: The time in sourcemod wouldnt reset
[8:25 PM] zaCade: while the scoreboard ect would
[8:25 PM] Moltard: and if the difference was too big, the map could end with no nextmap and it was a random nextmap picked
[8:25 PM] zaCade: yep
[8:25 PM] Moltard: nah I think ti's the opposite
[8:25 PM] Moltard: the scoreboard time didnt reset
[8:25 PM] Moltard: but sourcemod reset
[8:25 PM] zaCade: no it does
[8:25 PM] Moltard: on mg it happened so many time
[8:26 PM] Moltard: (or on mg, that was different)
[8:26 PM] zaCade: That was something else

extensions/cstrike/timeleft.cpp
```diff
@@ -58,7 +58,17 @@ void TimeLeftEvents::FireGameEvent(IGameEvent *event)
 		if (get_new_timeleft_offset || !round_end_found)
 		{
 			get_new_timeleft_offset = false;
-			timersys->NotifyOfGameStart();
+
+			float flGameStartTime = gpGlobals->curtime;
+			sm_sendprop_info_t info;
+			if (gamehelpers->FindSendPropInfo("CCSGameRulesProxy", "m_flGameStartTime", &info))
```
KyleSanderson: Can you nest this under gamerules? Doesn't make sense to do the lookup if the variable is only used if gamerules is valid. Not that this is a hot path - just a nit.

PR #1078, opened 2019-09-11
Add OnEntitySpawned to SDKHooks.

BotoX:
As a much much much faster alternative to:
```C++
public void OnEntityCreated(int entity, const char[] classname)
{
	SDKHook(entity, SDKHook_SpawnPost, OnEntitySpawned);
}
```

now simply use:
```C++
public void OnEntitySpawned(int entity, const char[] classname)
```

The first one causes substantial lag/delay on round change due to all entities being recreated (unhook & hook).
Some plugins need to hook certain entities right after they spawn and filter those by user supplied config files which use props like targetname, hammerid, etc.
The entity props are only available after Spawn() and can not be accessed in OnEntityCreated().

I'm finding it weird that this wasn't exported as a forward right away as it's very useful and straightforward to do.

### DIFF
--- a/extensions/sdkhooks/extension.cpp
+++ b/extensions/sdkhooks/extension.cpp
@@ -888,7 +888,6 @@ void SDKHooks::OnEntityCreated(CBaseEntity *pEntity)
 void SDKHooks::OnEntitySpawned(CBaseEntity *pEntity)
 {
 	// Call OnEntitySpawned forward
-	int ref = gamehelpers->EntityToReference(pEntity);
 	int index = gamehelpers->ReferenceToIndex(ref);
 
 	// This can be -1 for player ents before any players have connected
@@ -903,6 +902,7 @@ void SDKHooks::OnEntitySpawned(CBaseEntity *pEntity)
 		return;
 	}
 
+	int ref = gamehelpers->EntityToReference(pEntity);
 	HandleEntitySpawned(pEntity, index, ref);
 }
 
--- a/extensions/sdkhooks/extension.cpp
+++ b/extensions/sdkhooks/extension.cpp
@@ -247,6 +247,7 @@ bool SDKHooks::SDK_OnLoad(char *error, size_t maxlength, bool late)
 	sharesys->AddInterface(myself, &g_Interface);
 	sharesys->AddCapabilityProvider(myself, this, "SDKHook_DmgCustomInOTD");
 	sharesys->AddCapabilityProvider(myself, this, "SDKHook_LogicalEntSupport");
+	sharesys->AddCapabilityProvider(myself, this, "SDKHook_OnEntitySpawned");
 
 	playerhelpers->AddClientListener(&g_Interface);
 	
@@ -359,6 +360,7 @@ void SDKHooks::SDK_OnUnload()
 
 	sharesys->DropCapabilityProvider(myself, this, "SDKHook_DmgCustomInOTD");
 	sharesys->DropCapabilityProvider(myself, this, "SDKHook_LogicalEntSupport");
+	sharesys->DropCapabilityProvider(myself, this, "SDKHook_OnEntitySpawned");
 
 	CUtlVector<IEntityListener *> *entListeners = EntListeners();
 	entListeners->FindAndRemove(this);
--- a/extensions/sdkhooks/extension.cpp
+++ b/extensions/sdkhooks/extension.cpp
@@ -1799,21 +1799,26 @@ bool SDKHooks::Hook_WeaponSwitchPost(CBaseCombatWeapon *pWeapon, int viewmodelin
 void SDKHooks::HandleEntityCreated(CBaseEntity *pEntity, int index, cell_t ref)
 {
 	const char *pName = gamehelpers->GetEntityClassname(pEntity);
-	cell_t bcompatRef = gamehelpers->EntityToBCompatRef(pEntity);
-
+	if (!pName)
+		pName = "";
+	
 	// Send OnEntityCreated to SM listeners
 	SourceHook::List<ISMEntityListener *>::iterator iter;
 	ISMEntityListener *pListener = NULL;
 	for (iter = m_EntListeners.begin(); iter != m_EntListeners.end(); iter++)
 	{
 		pListener = (*iter);
-		pListener->OnEntityCreated(pEntity, pName ? pName : "");
+		pListener->OnEntityCreated(pEntity, pName);
 	}
 
 	// Call OnEntityCreated forward
-	g_pOnEntityCreated->PushCell(bcompatRef);
-	g_pOnEntityCreated->PushString(pName ? pName : "");
-	g_pOnEntityCreated->Execute(NULL);
+	if (g_pOnEntityCreated->GetFunctionCount())
+	{
+		cell_t bcompatRef = gamehelpers->EntityToBCompatRef(pEntity);
+		g_pOnEntityCreated->PushCell(bcompatRef);
+		g_pOnEntityCreated->PushString(pName);
+		g_pOnEntityCreated->Execute(NULL);
+	}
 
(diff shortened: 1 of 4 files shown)


extensions/sdkhooks/extension.cpp
```diff
@@ -1791,6 +1816,26 @@ void SDKHooks::HandleEntityCreated(CBaseEntity *pEntity, int index, cell_t ref)
...
+	SourceHook::List<ISMEntityListener *>::iterator iter;
+	ISMEntityListener *pListener = NULL;
+	for (iter = m_EntListeners.begin(); iter != m_EntListeners.end(); iter++)
+	{
+		pListener = (*iter);
+		pListener->OnEntitySpawned(pEntity, pName ? pName : "");
+	}
+
+	// Call OnEntitySpawned forward
+	g_pOnEntitySpawned->PushCell(bcompatRef);
```
KyleSanderson: GetFunctionCount

extensions/sdkhooks/extension.h
```diff
@@ -78,11 +78,11 @@ enum SDKHookType
 	SDKHook_Reload,
 	SDKHook_ReloadPost,
 	SDKHook_GetMaxHealth,
-	SDKHook_Blocked,
```
KyleSanderson: Can you reset these changes?

BotoX: I actually can't, I tried three different text editors.
That file had mixed line endings.
All the source files in SM use Windows CRLF but those few lines were using UNIX LF

Headline: `git reset --soft HEAD~2`
`git add -p`
then go through and apply only the changes you care about and not these

extensions/sdkhooks/extension.cpp
```diff
@@ -1791,6 +1816,32 @@ void SDKHooks::HandleEntityCreated(CBaseEntity *pEntity, int index, cell_t ref)
...
+		const char *pName = gamehelpers->GetEntityClassname(pEntity);
+		cell_t bcompatRef = gamehelpers->EntityToBCompatRef(pEntity);
+
+		// Send OnEntitySpawned to SM listeners
+		SourceHook::List<ISMEntityListener *>::iterator iter;
+		ISMEntityListener *pListener = NULL;
+		for (iter = m_EntListeners.begin(); iter != m_EntListeners.end(); iter++)
+		{
+			pListener = (*iter);
+			pListener->OnEntitySpawned(pEntity, pName ? pName : "");
```
KyleSanderson: pName should be figured out outside of the loop here.

extensions/sdkhooks/extension.cpp
```diff
@@ -1791,6 +1816,32 @@ void SDKHooks::HandleEntityCreated(CBaseEntity *pEntity, int index, cell_t ref)
...
+void SDKHooks::HandleEntitySpawned(CBaseEntity *pEntity, int index, cell_t ref)
+{
+	if (g_pOnEntitySpawned->GetFunctionCount() || m_EntListeners.size())
+	{
+		const char *pName = gamehelpers->GetEntityClassname(pEntity);
+		cell_t bcompatRef = gamehelpers->EntityToBCompatRef(pEntity);
+
+		// Send OnEntitySpawned to SM listeners
+		SourceHook::List<ISMEntityListener *>::iterator iter;
+		ISMEntityListener *pListener = NULL;
```
KyleSanderson: This definition is temporary - throw it in the loop as well as the iter.

extensions/sdkhooks/extension.cpp
```diff
@@ -1820,23 +1820,23 @@ void SDKHooks::HandleEntitySpawned(CBaseEntity *pEntity, int index, cell_t ref)
 {
 	if (g_pOnEntitySpawned->GetFunctionCount() || m_EntListeners.size())
 	{
-		const char *pName = gamehelpers->GetEntityClassname(pEntity);
 		cell_t bcompatRef = gamehelpers->EntityToBCompatRef(pEntity);
+		const char *pName = gamehelpers->GetEntityClassname(pEntity);
+		if(!pName)
```
KyleSanderson: 👺

Headline:
I’m not too sure about the details involving the game engine in this circumstance, but over all I’m hesitant to want to take this change in-tree. 

Does this lead to taking more sdktools hooks into core? Is the performance issue something that we can fix in sdktools itself?

KyleSanderson:
The performance issue was fixed - it's a DHooks bug.

KyleSanderson:
@BotoX is this actually still used? or are you okay now with that awful dhooks bug being reverted/resolved?

BotoX:
I still use this, but it's definitely not needed any more now that the performance issue is fixed in DHooks.
But I do like this method a lot more instead of doing OnEntityCreated + SDKHook PostSpawn
A lot of times you want an entity spawn hook in a plugin and the engine already offers one for free.
Other than breaking compatibility with compiled extensions I don't see anything wrong with having this in SourceMod.

BotoX:
> The only thing I'm not sure about is getting the classname, that's a cost even for no users and it is just as easy for a plugin to do

The classname is only retrieved if there is at least one plugin/extension which registered the forward/an ISMEntityListener.

dvander:
This broke a handful of plugins which already had a forward with the same name but used a different signature. Assumedly those plugins will now get two invocations of the callback for each spawned entity, so that's also some binary compatibility breakage.

Some solutions:
- A revert would be the best thing, but it might be too late since this landed a year ago.
- Don't invoke the new signature if the old one is hooked. This would require dropping the use of a global forward and iterating through plugins instead.
- Don't invoke the specific spawn hook if the plugin has the global forward.

dvander:
Since those plugins no longer compile, it would also be valid to bump the core.inc version and do a check that way.

asherkin:
@dvander I think the source-level breakage here is fine, it has to be assumed that SM can take the entire `On*` namespace for global forwards - the binary-level breakage is very unfortunate though, I forgot that existing plugins would pick up the new global forward.

It has been a good while, but given #1558 as well I'd be pretty happy to back this out before 1.11 is declared stable. I don't suppose if we did that there is a good way to have plugins trying to use the global forward error rather than it just silently not being called? (That said, I believe most prolific users have run into #1558 by now and reverted to the SDKHook.)

As a future SP change it might be nice to have matched up `forward` declarations be required as part of the global forward binding process - we also occasionally get people who forget to include the file an extension-provided forward is defined in and then they lack type checking. Both of those would require `public` to be reserved for forwards or a new keyword for global forward listeners though (both fairly intrusive changes).

dvander:
SGTM - two suggestions for making dependencies fail after the revert. One is to do something like:

```
stock void OnEntitySpawned() { DO_NOT_CALL_ME; }
```

This would fail if being compiled and fail if being implemented as a public function.

Or,

```
forward void OnEntitySpawned(bool INVALID_FORWARD_DO_NOT_USE);
```

Since the signature changed it wouldn't build anymore.

asherkin:
That works for breaking compilation, but it's mostly the binary side I'm concerned about. There is also a nasty side effect there where we're still breaking the source-level compatibility for plugins coming from 1.10 using OnEntitySpawned as the callback name for the SDKHook, but not giving them any new feature.

I think it's probably best to just revert and shout about it on Discord, and see how that goes down.

dvander:
SGTM.

I agree re: `forward`, this came up in IRC too. But it seems like something that could wreak havoc on compatibility. If anyone has ideas that don't, I'd be happy to entertain them. IMO a new keyword for "forward" (like "event") and then using something like C#'s delegate syntax would be a nice carrot.

nosoop:
I think adding an `SDKHooks_AddEntitySpawnListener(...)` would be a decent compromise; plugins can opt-in with a callback of their choice.

Issue #1080, opened 2019-09-15
Add new native TR_GetHitBox()


Einyux:
# Help us help you
  - [X] I have checked that my issue [doesn't exist yet](https://github.com/alliedmodders/sourcemod/issues).

# Description

In **CGameTrace** there's this member:

```
// NOTE: this member is overloaded.
// If hEnt points at the world entity, then this is the static prop index.
// Otherwise, this is the hitbox index.
int			hitbox;					// box hit by trace in studio
```

https://github.com/ValveSoftware/source-sdk-2013/blob/master/sp/src/public/gametrace.h#L73

It would be nice to have access to this member in SourceMod. 
It can be used when creating decals using TE:

```
"CTEDecal"
{
	"name"		"Entity Decal"
	"index"		"22"

	"SendTable"
	{
		"m_vecOrigin"		"vector"
		"m_vecStart"		"vector"
		"m_nEntity"		"int"
		"m_nHitbox"		"int"      // <=====
		"m_nIndex"		"int"
	}
}
```

-----
If needed, I can propose a PR.
Thanks.

KyleSanderson:
A PR would definitely speed it up

KyleSanderson:
Thank you for your contribution :)

PR #1085, opened 2019-09-17
Add TR_GetHitBoxIndex() native

Einyux:
Related to the issue #1080.

Add **TR_GetHitBoxIndex()** native to retrieve the member "hitbox" from the class "CGameTrace".

https://github.com/ValveSoftware/source-sdk-2013/blob/master/sp/src/public/gametrace.h#L73

### DIFF
--- a/extensions/sdktools/trnatives.cpp
+++ b/extensions/sdktools/trnatives.cpp
@@ -998,6 +998,22 @@ static cell_t smn_TRGetHitGroup(IPluginContext *pContext, const cell_t *params)
 	return tr->hitgroup;
 }
 
+static cell_t smn_TRGetHitBox(IPluginContext *pContext, const cell_t *params)
+{
+	sm_trace_t *tr;
+	HandleError err;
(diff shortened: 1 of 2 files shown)


extensions/sdktools/trnatives.cpp
```diff
@@ -998,6 +998,22 @@ static cell_t smn_TRGetHitGroup(IPluginContext *pContext, const cell_t *params)
...
+static cell_t smn_TRGetHitBoxIndex(IPluginContext *pContext, const cell_t *params)
+{
+	sm_trace_t *tr;
+	HandleError err;
+	HandleSecurity sec(pContext->GetIdentity(), myself->GetIdentity());
+
+	if (params[1] == BAD_HANDLE)
+	{
+		tr = &g_Trace;
+	} else if ((err = handlesys->ReadHandle(params[1], g_TraceHandle, &sec, (void **)&tr)) != HandleError_None) {
```
Headline: This if is malformed

Issue #1086, opened 2019-09-24
"A copy of the JIT License is available in JIT.txt."


DJPlaya:
### Help us help you
  - [x] I have checked that my issue [doesn't exist yet](https://github.com/alliedmodders/sourcemod/issues).

The Sourcemod License says "A copy of the JIT License is available in JIT.txt." but that File dosent exist, and JIT isent a License Type.
Might be a Typo, did you mean the MIT License?

dvander:
In fact, we used to have a separate license for the SourcePawn VM. It was closed-source until SourceMod 1.1.

Anyway, that was a trillion years ago, so feel free to submit a PR removing this ancient text.

DJPlaya:
#Done

PR #1091, opened 2019-09-29
Removed outdated License Info

DJPlaya:
Ref #1086

### DIFF
--- a/licenses/LICENSE.txt
+++ b/licenses/LICENSE.txt
@@ -1,5 +1,5 @@
 SOURCEMOD LICENSE INFORMATION
-VERSION: JUNE-13-2008
+VERSION: 29.9.2019
 -----------------------------
 
 SourceMod is licensed under the GNU General Public License, version 3.  
--- a/licenses/LICENSE.txt
+++ b/licenses/LICENSE.txt
@@ -1,5 +1,5 @@
 SOURCEMOD LICENSE INFORMATION
-VERSION: 29.9.2019
+VERSION: 2019-9-30
 -----------------------------
 
 SourceMod is licensed under the GNU General Public License, version 3.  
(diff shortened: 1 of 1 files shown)


Headline:
@dvander should we bump license version/date or is this fine as is?

DJPlaya:
My mistake, its important to change that too ^^

licenses/LICENSE.txt
```diff
@@ -1,5 +1,5 @@
 SOURCEMOD LICENSE INFORMATION
-VERSION: JUNE-13-2008
+VERSION: 29.9.2019
```
dvander: If you want to bump the version, might as well keep it in the same format (or YYYY.MM.DD which looks nicer)

asherkin:
I can't do a line comment on it as it is outside of the changed lines (WTF GitHub), but the exceptions list includes a mention of the JIT that can be removed as well.

It also looks like the last word got lost at some point, it should read "All trademarks are property of their respective owners in the US and other countries."

licenses/LICENSE.txt
DJPlaya: Ime using the [most often used format](https://en.wikipedia.org/wiki/Date_format_by_country) DMY, or is there a default format for this repo?

DJPlaya:
@asherkin 
The last word was missing before, dident noticed ^^

You mean changing:
"As a special exception, AlliedModders LLC gives you permission to link the code
of this program (as well as its derivative works) to "Half-Life 2," the "Source 
Engine," the "SourcePawn JIT," and any Game MODs that run on software by the 
Valve Corporation. "
to
"As a special exception, AlliedModders LLC gives you permission to link the code
of this program (as well as its derivative works) to "Half-Life 2," the "Source 
Engine and any Game MODs that run on software by the 
Valve Corporation. "
?

licenses/LICENSE.txt
dvander: Well, if you're gonna change the formatting here, YYYY-MM-DD looks better. I think that's ISO standard.

DJPlaya: Ok, dident know SM is using ISO 8601

Headline:
@DJPlaya yes that's what he meant

PR #1097, opened 2019-10-07
Fix admin map chooser not using display names for menu

nosoop:
From the SM Discord:

> 4:52 PM] HiGPS: Is there any plugin that helps with the workshop map names not working properly in TF2, they show ID's and not the map names in the admin menu , but works in the nominations menu

This copies the display name-related lines from the `LoadMapList` function in `basevotes/votemap.sp` to `basecommands/map.sp`.  It's otherwise left as-is &mdash; I'd have no issues changing the styling / hoisting the variable declaration to make it more consistent if desired.

### DIFF
--- a/plugins/basecommands/map.sp
+++ b/plugins/basecommands/map.sp
@@ -159,8 +159,10 @@ int LoadMapList(Menu menu)
 	
 	for (int i = 0; i < map_count; i++)
 	{
+		char displayName[PLATFORM_MAX_PATH];
 		GetArrayString(g_map_array, i, map_name, sizeof(map_name));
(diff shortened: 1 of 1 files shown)


TheByKotik:
i'm think is not good idea.
Maps can have same name but must have unique paths.

nosoop:
That's already an issue with using display names in the votemap menu with PR #354.

Personally, I'm indifferent on disambiguation (how many server ops have maps that share the same exact display name?).  Having a togglable option to use the result of `FindMap` instead of `GetMapDisplayName` might be a compromise worth looking into; @TheByKotik might want to file an issue on it to get a proper discussion going.

PR #1100, opened 2019-10-11
Clientprefs Modernization

Headline:
Hey folks

This is the vision I have for the future of the SourceMod code base, the biggest step this extension takes is that it brings in stdlibc++.

This is a large diff, fairly untested, and more experimentation must be done but I'm proposing this for comments to be given about the direction I went with this, so any quarrels can be discussed earlier rather than later.

:clinking_glasses:

### DIFF
--- a/extensions/clientprefs/AMBuilder
+++ b/extensions/clientprefs/AMBuilder
@@ -7,7 +7,12 @@ for arch in SM.archs:
     os.path.join(SM.mms_root, 'core', 'sourcehook'),
   ]
   if binary.compiler.family == 'gcc' or binary.compiler.family == 'clang':
-    binary.compiler.cxxflags += ['-fno-rtti']
+    binary.compiler.cxxflags += [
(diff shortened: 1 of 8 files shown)


CrazyHackGUT:
Hey!
What about adding a feature to write own storage system for default ClientPrefs?
This is will be really cool, if we have ability implement our own storage which is possible to use in default Client Preferences extension.

We're write something plugin what works with REST API, and backend of this REST API can storage client preferences.

As i see this, clientprefs will have native function with name like `SetClientPreferencesHandler()`, what accepts functions for saving/loading data and creating cookies.

I hope, i was able to convey my thought.

Headline:
@CrazyHackGUT Feel free to create an issue with a little specification about how you might want that to work, and it can be discussed. This, however, is not intended to bring new features in

KyleSanderson:
uh, why not just use -fno-exceptions instead of adding noexcept to every single definition? :S

PR #1102, opened 2019-10-16
TF2: Add new condition from Scream Fortress XI

nosoop:
Scream Fortress XI added condition 128 `TF_COND_HALLOWEEN_HELL_HEAL` ~~and damagecustom 84 `TF_DMG_CUSTOM_GENERIC_BOMB`~~.


psychonic:
Thanks!

Pelipoika:
I think TF_DMG_CUSTOM_GENERIC_BOMB got added in the 1st update but removed in the 2nd, i can't find it in the client.dll anymore

nosoop:
Good catch; looks like they did remove it shortly after the update.  Gonna have to revert that bit.

`CTFGenericBomb::Event_Killed` now sets the custom damage type to `TF_CUSTOM_PUMPKIN_BOMB` only if the map name is "koth_slaughter_event", otherwise it's set to 0.

Issue #1105, opened 2019-10-19
Add new knife to list


wopox1337:
https://github.com/alliedmodders/sourcemod/blob/master/plugins/include/cstrike.inc#L158
```
CSWeapon_KNIFE_CSS = 503
```

wopox1337:
@asherkin can you add new item into SourceMod::CSGO?

Headline:
Added, thanks!

PR #1111, opened 2019-10-31
Add classic knife to CSWeaponID

Bara:
Instead of KNIFE_CSS i would call it WEAPON_CLASSIC, like it [Valve](http://media.steampowered.com/apps/csgo/blog/images/cs20/knife13.png) does.

#1105

### DIFF
--- a/plugins/include/cstrike.inc
+++ b/plugins/include/cstrike.inc
@@ -156,6 +156,7 @@ enum CSWeaponID
 	CSWeapon_BUMPMINE = 85,
 	CSWeapon_MAX_WEAPONS_NO_KNIFES, // Max without the knife item defs, useful when treating all knives as a regular knife.
 	CSWeapon_BAYONET = 500,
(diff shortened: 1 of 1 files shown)


sneak-it:
But Valve does not call it that, it is weapon_knife_css:

Image

We should be consistent with the rest of the knife names, no? The rest of the entries utilize the items manifest names, I don't think we should change this now.

@Bara @Headline @asherkin

Headline:
it’s now a breaking change to change the name, but I don’t think it’s a big deal

Bara:
Breaking change from 2 days old change in a pretty new dev branch/version? 🤔

The official name is "Classic Knife" (csgo_english.txt -> "SFUI_WPNHUD_KnifeCSS") and for internal stuff (files, classname, ...) "knife_css".

Headline:
if you updated and started using it, changing the name would break scripts, so yes unfortunately 

should be fine as-is, but if there’s a real push for this to be changed something can be worked out

PR #1113, opened 2019-11-01
Fix builds on newer gcc version

Headline:
Adds a few `-Wno-` warning supressors in places where hl2sdk code affects us, and then fixed a portion of our code that was throwing the error.

Newer clang versions are fine, needs alliedmodders/metamod-source#63 to merge first


KyleSanderson:
NAK. CI needs to be added for this, use: https://github.com/alliedmodders/sourcemod/pull/1054

PR #1114, opened 2019-11-01
Add slot define for healthshot/shield and tablet

Bara:
I'm still not sure about the name "Util", but it's a more general name  (Valve call this "boost" in his "items_game.txt") and we don't know if valve adds more items/weapons to this slot.

Let me know, if anyone has other suggestions about this name.


psychonic:
Valve calls that one the boosts slot (as you mention), and the slot after it (12), the utility slot (where the tablet is). It probably makes sense to just follow their naming.

Bara:
@psychonic you're right 👍  I changed it and added define for tablet slot.

PR #1116, opened 2019-11-03
Add windows supported SDKs to powershell checkout-deps

JoinedSenses:
.sh version has this bit:

```
if [ $ismac -eq 0 ]; then
  # Add these SDKs for Windows or Linux
  sdks+=( orangebox blade episode1 bms )

  # Add more SDKs for Windows only
  if [ $iswin -eq 1 ]; then
    sdks+=( darkm swarm bgt eye contagion )
  fi
fi
```
Added these to the ps1 SDK list. If there is a better approach, please let me know.

### DIFF
--- a/tools/checkout-deps.ps1
+++ b/tools/checkout-deps.ps1
@@ -8,6 +8,15 @@
 [CmdletBinding()]
 param(
     [string[]]$SDKs = @(
+        'orangebox',
+        'blade',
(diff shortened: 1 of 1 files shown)


JoinedSenses:
Missing sql stuff at the moment

asherkin:
Thanks!

PR #1117, opened 2019-11-04
Add new custom flags

komashchenko:
Sometimes 6 custom flags are not enough to limit access to necessary functions, and using groups to restrict one or two functions is not justified

### DIFF
--- a/public/IAdminSystem.h
+++ b/public/IAdminSystem.h
@@ -35,7 +35,7 @@
 #include <IShareSys.h>
 
 #define SMINTERFACE_ADMINSYS_NAME		"IAdminSys"
-#define SMINTERFACE_ADMINSYS_VERSION	9
+#define SMINTERFACE_ADMINSYS_VERSION	8
 
 /**
  * @file IAdminSystem.h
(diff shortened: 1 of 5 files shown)


Septembers:
I agree with you

sydro263:
That's cool!

asherkin:
👎 from me. Admin groups are the correct thing to use here and give you many, many more options if you want full control, without being that much more complicated.

These unnamed admin flags just end up creating a confusing soup of permissions that is impossible to debug what is granting access to what.

public/IAdminSystem.h
```diff
@@ -35,7 +35,7 @@
 #include <IShareSys.h>
 
 #define SMINTERFACE_ADMINSYS_NAME		"IAdminSys"
-#define SMINTERFACE_ADMINSYS_VERSION	8
+#define SMINTERFACE_ADMINSYS_VERSION	9
```
asherkin: The interface version shouldn't need bumping here I don't think, none of the changes here are in interop.

KyleSanderson:
yeah sorry @komashchenko this isn't going to result in a net benefit for the community to land as groups are the way to accomplish this. If groups don't cover your use case lets improve that scenario. I do understand there are plugins out there that don't fully support groups, but adding support to them is trivial and will help people further down the road (maybe we can even add a wrapper?).

PR #1118, opened 2019-11-05
Raised limits for csgo HintText and SayText protobuffs

GAMMACASE:
Raised limits for SayText and HintText protobuffs and also enabled SayText as a chat message in csgo. Was compiled and tested on win, that pr will allow a lot more characters in chat messages and hints and that will allow use of much more colored messages (in chat) and more html tags (in hints).

Not sure how known that thing is, but big thanks to [@Rostu13](https://github.com/Rostu13), as he told me about that.


Bara:
>  use of much more colored messages (in chat)

Sounds interesting.

PR #1123, opened 2019-11-16
use tf2 ignition if playing that game

Naleksuh:


peace-maker:
Why? What's the difference? If you really want this to be merged, you should consider giving a description.

Naleksuh:
> Why? What's the difference? If you really want this to be merged, you should consider giving a description.

it uses the tf2 native when running tf2, which works a lot better than igniteentity imo. it doesn't change anything for other games.

psychonic:
Outside of the functionality changes, you need to check both engine and gamedir (or just appid, if we expose that) if you want to detect a specific game. Just detecting the engine has the possibility for breakage down the road.

Naleksuh:
i'm not sure this code will be possible to use, because now it appears that the tf2.ext is required to load for funcommands to work.

anyone knows a way around this?

SlidyBat:
> i'm not sure this code will be possible to use, because now it appears that the tf2.ext is required to load for funcommands to work, breaking funcommands on any game besides tf2
> 
> anyone knows a way around this?

undef `REQUIRE_EXTENSIONS`

plugins/funcommands.sp
```diff
@@ -37,7 +37,7 @@
 #include <sdktools>
 #undef REQUIRE_PLUGIN
 #include <adminmenu>
-
+#include <tf2>
```
KyleSanderson: jannik called this out a couple months ago - can you make these static engine checks to not impact performance substantially by checking upon every invocation?

Naleksuh: huh? no, tf2 is included to use TF2_IgnitePlayer

KyleSanderson:
@Naleksuh any update on this one?

Naleksuh:
Okay, so I didn't get an email for the "any update" until this was closed. Can it be reopened?

PR #1124, opened 2019-11-17
Allow custom sort natives and callback use any types and optional data

FortyTwoFortyTwo:
3 main things done with `SortCustom1D`, `SortCustom2D`, and `ArrayList.SortCustom` natives:

- Allow `any` variable to custom sort rather than limited to `int`. Meaning floats, enums, methodmap etc can be used to pass and custom sort.
- Replace optional `Handle hndl` pass to `any data`. I don't see anywhere on why it should be limited to handles, so allow `any` without needing to create new handles. Also updated core to use `cell_t` data, doesn't make any difference but may well then do it.
- Use `ArrayList` instead of `Handle` in `SortFuncADTArray` callback. `Handle` variable can still be used from old plugins.

Only thing missing is `any` for regular 1D sorting. I'm not sure if it a good idea to allow `any` in `SortIntegers` native to sort enums, but i left it untouched.
`SortADTArrayCustom` also left untouched because we have `ArrayList.SortCustom` traditional syntax instead.

Hopefully i didn't miss anything that would break old plugins, but please do let me know if i did break one.
Compiled with this, no warnings and tested.

```
public void OnPluginStart()
{
	// New sortings
	
	int Data = 322;
	Action Array1D[3] = {Plugin_Handled, Plugin_Continue, Plugin_Stop};
	SortCustom1D(Array1D, sizeof(Array1D), Sort1D, Data);
	
	float Array2D[3][2] = {{3.7, 1.1}, {7.4, -4.7}, {1.0, 2.3}};
	SortCustom2D(Array2D, sizeof(Array2D), Sort2D);
	
	ArrayList ArrayADT = new ArrayList();
	ArrayADT.Push(2);
	ArrayADT.Push(7);
	ArrayADT.Push(3);
	ArrayADT.SortCustom(SortADT);
	delete ArrayADT;
	
	
	// Old sortings
	
	int OldArray1D[3] = {5, 1, 3};
	SortCustom1D(OldArray1D, sizeof(OldArray1D), OldSort1D);
	
	Handle Trie = CreateTrie();
	int OldArray2D[3][2] = {{2, 4}, {7, 1}, {2, 6}};
	SortCustom2D(OldArray2D, sizeof(OldArray2D), OldSort2D, Trie);
	CloseHandle(Trie);
	
	Handle OldArrayADT = CreateArray();
	PushArrayCell(OldArrayADT, 1);
	PushArrayCell(OldArrayADT, 8);
	PushArrayCell(OldArrayADT, 2);
	SortADTArrayCustom(OldArrayADT, OldSortADT);
	CloseHandle(OldArrayADT);
}

public int Sort1D(Action elem1, Action elem2, const Action[] array, any data)
{
	PrintToServer("Sort1D - %d %d - data %d", elem1, elem2, data);
	return 0;
}

public int Sort2D(float[] elem1, float[] elem2, const float[][] array)
{
	PrintToServer("Sort2D - %.2f %.2f", elem1[0], elem2[0]);
	return 0;
}

public int SortADT(int index1, int index2, ArrayList array)
{
	PrintToServer("SortADT - %d %d", array.Get(index1), array.Get(index2));
	return 0;
}

public int OldSort1D(int elem1, int elem2, const int[] array, Handle hndl)
{
	PrintToServer("OldSort1D - %d %d", elem1, elem2);
	return 0;
}

public int OldSort2D(int[] elem1, int[] elem2, const int[][] array, Handle hndl)
{
	PrintToServer("OldSort2D - %d %d - hndl %x", elem1[0], elem2[0], hndl);
	return 0;
}

public int OldSortADT(int index1, int index2, Handle array, Handle hndl)
{
	PrintToServer("OldSortADT - %d %d", GetArrayCell(array, index1), GetArrayCell(array, index2));
	return 0;
}
```

### DIFF
--- a/plugins/include/sorting.inc
+++ b/plugins/include/sorting.inc
@@ -90,12 +90,16 @@ native void SortStrings(char[][] array, int array_size, SortOrder order = Sort_A
  * @param elem1         First element to compare.
  * @param elem2         Second element to compare.
  * @param array         Array that is being sorted (order is undefined).
- * @param hndl          Handle optionally passed in while sorting.
+ * @param data          Data optionally passed in while sorting.
  * @return              -1 if first should go before second
  *                      0 if first is equal to second
  *                      1 if first should go after second
  */
-typedef SortFunc1D = function int (int elem1, int elem2, const int[] array, Handle hndl);
+typeset SortFunc1D
+{
+	function int (any elem1, any elem2, const any[] array);
+	function int (any elem1, any elem2, const any[] array, any data);
+}
 
 /**
  * Sorts a custom 1D array.  You must pass in a comparison function.
@@ -103,9 +107,9 @@ typedef SortFunc1D = function int (int elem1, int elem2, const int[] array, Hand
  * @param array         Array to sort.
  * @param array_size    Size of the array to sort.
  * @param sortfunc      Sort function.
- * @param hndl          Optional Handle to pass through the comparison calls.
+ * @param data          Optional data to pass through the comparison calls.
  */
-native void SortCustom1D(int[] array, int array_size, SortFunc1D sortfunc, Handle hndl=INVALID_HANDLE);
+native void SortCustom1D(any[] array, int array_size, SortFunc1D sortfunc, any data=0);
 
 /**
  * Sort comparison function for 2D array elements (sub-arrays).
@@ -114,15 +118,15 @@ native void SortCustom1D(int[] array, int array_size, SortFunc1D sortfunc, Handl
  * @param elem1         First array to compare.
  * @param elem2         Second array to compare.
  * @param array         Array that is being sorted (order is undefined).
- * @param hndl          Handle optionally passed in while sorting.
+ * @param data          Data optionally passed in while sorting.
  * @return              -1 if first should go before second
  *                      0 if first is equal to second
  *                      1 if first should go after second
  */
 typeset SortFunc2D
 {
-	function int (int[] elem1, int[] elem2, const int[][] array, Handle hndl);
-	function int (char[] elem1, char[] elem2, const char[][] array, Handle hndl);
+	function int (any[] elem1, any[] elem2, const any[][] array);
+	function int (any[] elem1, any[] elem2, const any[][] array, any data);
 };
 
 /**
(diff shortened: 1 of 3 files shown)


plugins/include/sorting.inc
```diff
@@ -114,15 +118,15 @@ native void SortCustom1D(int[] array, int array_size, SortFunc1D sortfunc, Handl
...
- * @param hndl          Handle optionally passed in while sorting.
+ * @param data          Data optionally passed in while sorting.
  * @return              -1 if first should go before second
  *                      0 if first is equal to second
  *                      1 if first should go after second
  */
 typeset SortFunc2D
 {
-	function int (int[] elem1, int[] elem2, const int[][] array, Handle hndl);
-	function int (char[] elem1, char[] elem2, const char[][] array, Handle hndl);
```
Scags: Bit late to this, but this edit will break bcompat. You cannot coerce `char[]` to `any[]`, so any old plugins that treated it as such will not compile.

FortyTwoFortyTwo: There currently an issue about coerce `char[]` to `any[]` https://github.com/alliedmodders/sourcepawn/issues/447. I could either wait until that fixes, or readd `char[]` to typeset, i'm not sure which is the best route to go.

asherkin: > There currently an issue about coerce `char[]` to `any[]` [alliedmodders/sourcepawn#447](https://github.com/alliedmodders/sourcepawn/issues/447). I could either wait until that fixes, or readd `char[]` to typeset, i'm not sure which is the best route to go.

That issue is unrelated, I don't think `char[]` will ever be allowed to convert to `any[]` again - they're very different so having both in the same typeset might give the wrong behaviour. Not sure what the best option is.

FortyTwoFortyTwo: Right, `char[]` actually compiles with `any[]` in callback function, though it may indeed give wrong behavior. However `SortCustom2D` native already have `any[][]` param that cant be used for `char[][]`, so how did `char[]` in callback even get here? Think it best to just ditch `char[]`.

FortyTwoFortyTwo:
Updated as `int` to `any` change was merged from #2146. This PR now only deals with optional `any` data and replacing `Handle` to `ArrayList` in `SortFuncADTArray`.

Using `char` for `any` in `SortFunc2D` also does seems to no longer work for latest SP version, so I've bring the char callback back in.

FortyTwoFortyTwo:
Egh pull request reverted, don't even know myself if it's worth for me to update this to bring back the `int` to `any` change...

FortyTwoFortyTwo:
Updated branch to latest, restored `int` to `any` request change as reverted #2146 was never restored back.
My test example at original post still compiles as normal for both 1.12 and 1.13.
Still not giving up on my 5 years old PR!

PR #1135, opened 2019-11-30
Speculative fix for MySQL crashes

asherkin:
https://crash.limetech.org/stats/dbi.mysql.ext.%25/my_real_read
https://crash.limetech.org/stats/dbi.mysql.ext.%25/net_real_write

Both of these are caused by the VIO ptr ending up as null in the middle of reading/writing to a connection - I can't find any indication of a fix for this made to MySQL, so don't think it is a bug fix we're missing, but there are some musings around the internet that it could be caused by improper thread-safety initialisation.

`my_init` (what we had here) is called internally by `mysql_library_init` but I think would have still led to an automatic `mysql_library_init` call the first time `mysql_init` was called (which we can do on a thread in case of threaded connections), which is exactly the thread-safety issue called out by the MySQL docs, so hopefully doing things properly here will help.

### DIFF
--- a/extensions/mysql/extension.cpp
+++ b/extensions/mysql/extension.cpp
@@ -46,9 +46,12 @@ SMEXT_LINK(&g_MySqlDBI);
 
 bool DBI_MySQL::SDK_OnLoad(char *error, size_t maxlength, bool late)
 {
+	if (mysql_library_init(0, NULL, NULL) != 0) {
+		smutils->Format(error, maxlength, "Could not initialize MySQL client library");
+		return false;	
+	}
+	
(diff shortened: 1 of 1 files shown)


rumblefrog:
Woah, we didn't have init check prior :open_mouth:

asherkin:
> Woah, we didn't have init check prior 😮

`my_init` is void, and as far as I can tell this can't actually fail for the client library we're using, just for the embedded server if you pass options that are not supported.

rumblefrog:
I always thought we had mysql_library_init, as it was one of the first things in 23.7.3.2 for writing threaded clients. TIL.

PR #1147, opened 2019-12-23
Mistake in CanAdminTarget()

CrazyHackGUT:
Looks like a simple mistake. My SRCDS instance is got a infinite cycle and killed by WatchDog.


KyleSanderson:
Thanks man.

PR #1152, opened 2019-12-30
Fix empty api descriptions in sdhooks.inc

Bara:
As example for the [new-api](http://sourcemod.net/new-api/sdkhooks/SDKHookCB) site: https://i.imgur.com/ZyLtnrr.png

This change should be more readable for new coders/programmers they use this site.


Headline:
This should certainly help newer folks, thanks!

PR #1153, opened 2020-01-07
WriteBaselines gamedata update after latest csgo update

GAMMACASE:


asherkin:
Thanks for the PR. These sigs seem a bit short and are likely to match a different function rather than failing to match in the future, most of what's being matched against here is just the vprof boilerplate common to many functions. A change that early in the function (for both platforms) is often indicative of the params being changed, did you check they were the same? Has this been tested?

GAMMACASE:
Didn't notice any param changes, and also win sig don't seems to be broken, so I can restore it to what it was before, it is linux one that broke, due to vprof functions are moved into a separate function.

GAMMACASE:
Alright, made them bigger, is that okay?

PR #1156, opened 2020-01-11
Fix paramater typo

ErikMinekus:


plugins/include/functions.inc
```diff
@@ -123,7 +123,7 @@ methodmap GlobalForward < Handle {
 	// @param type              Execution type to be used.
 	// @param ...               Variable number of parameter types (up to 32).
 	// @return                  Handle to new global forward.
-	// @error                   More than 32 paramater types passed.
+	// @error                   More than 32 parameter types passed.
```
Headline: We could probably improve the English here further. `parameter types` feels misleading as to what actually triggers the error.

`Upon more than 32 parameters passed`

ErikMinekus: `parameter types` refers to the number of `ParamType` parameters though, it doesn't include the `ExecType` parameter.

Headline: ah yeah, I see. This looks fine

PR #1158, opened 2020-01-11
Fix GetSteamAccountID validation opt-out

### LINKED ISSUE #1157: GetSteamAccountID(client, false) returns 0 when GetClientAuthId(client, AuthId_Steam3, steamid, sizeof(steamid), false); returns a non-zero number
As explained in the title, I am setting verify to false for both functions. But GetSteamAccountID returns 0 for unverified clients while GetClientAuthId returns a non zero value.

> liljimy100<3297><STEAM_ID_PENDING><> GetSteamAccountID was 0. GetClientAuthId [U:1:999357900]

bottiger1:
pass validated parameter to GetSteamId. Fixes #1157.

### DIFF
--- a/core/PlayerManager.cpp
+++ b/core/PlayerManager.cpp
@@ -2238,7 +2238,7 @@ unsigned int CPlayer::GetSteamAccountID(bool validated)
 {
 	if (!IsFakeClient() && (!validated || IsAuthStringValidated()))
 	{
-		const CSteamID &id = GetSteamId();
+		const CSteamID &id = GetSteamId(validated);
 		if (id.IsValid())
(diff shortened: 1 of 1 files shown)


asherkin:
Have you tested that this change does actually fix the issue?

bottiger1:
There doesn't seem to be any issues with it, but I am not sure how I am supposed to reproduce the Steam servers being down.

Kenzzer:
If you have a local server, unplug your Ethernet cable/Disable wifi.

bottiger1:
@asherkin I tested this yesterday during steam maintenance and it appears to work

PR #1161, opened 2020-01-20
fix incorrect command reply in nominations

thorgot:
When a map is nominated and the max nominations have been reached, the wrong text is returned if the map hasn't been nominated yet. Instead of printing the "Max Nominations" text, it prints out the "Map Already Nominated" text instead.

### DIFF
--- a/plugins/nominations.sp
+++ b/plugins/nominations.sp
@@ -338,7 +338,7 @@ void AttemptNominate(int client, const char[] map, int size)
 		}
 		else
 		{
-			ReplyToCommand(client, "[SM] %t", "Map Already Nominated");
+			ReplyToCommand(client, "[SM] %t", "Max Nominations");
 		}
 		
 		return;	


Headline:
Thanks for contributing!

Looking over this, I'm seeing an explicit check for maps that have already been nominated. I believe something more nefarious is going on than a simple logic issue. The part you have edited, from what I've observed, is only possible to be hit when the map list is full, indicating to me that the original message is correct

https://github.com/alliedmodders/sourcemod/blob/de1a186c479d9208896afb4d031184f3901391bd/plugins/nominations.sp#L325

thorgot:
> Thanks for contributing!
> 
> Looking over this, I'm seeing an explicit check for maps that have already been nominated. I believe something more nefarious is going on than a simple logic issue. The part you have edited, from what I've observed, is only possible to be hit when the map list is full, indicating to me that the original message is correct
> 
> https://github.com/alliedmodders/sourcemod/blob/de1a186c479d9208896afb4d031184f3901391bd/plugins/nominations.sp#L323-L326

`MAPSTATUS_DISABLED` is the status flag when the map is disabled in the nominate menu. The block you linked to is reached when the map has been nominated already, is the current map, or is in the exclude list (which is the last X maps). A map which has not been nominated, has not been played recently, and is not the current map will skip that entire `if` block and reach the line I changed.

Headline:
The maps are given `MAPSTATUS_DISABLED` unconditionally when they're successfully nominated as shown here

https://github.com/alliedmodders/sourcemod/blob/de1a186c479d9208896afb4d031184f3901391bd/plugins/nominations.sp#L347-L354

PR #1164, opened 2020-01-27
Fix documentation on DataPack methods

Deathreus:
Also fixup a couple descriptions in the methodmap

### DIFF
--- a/plugins/include/datapack.inc
+++ b/plugins/include/datapack.inc
@@ -58,6 +58,16 @@ methodmap DataPack < Handle
 	// @param val           Float to add.
 	// @param insert        Determines whether mid-pack writes will insert instead of overwrite.
 	public native void WriteFloat(float val, bool insert = false);
+	
+	// Packs a vector into a data pack.
+	//
+	// @param val		Float array to add.
+	// @param insert	Determines whether mid-pack writes will insert instead of overwrite.
+	public void WriteVector(const float val[3], bool insert = false)
+	{
+		for(int x=0; x<3; ++x)
+			this.WriteFloat(val[x]);
+	}
 
 	// Packs a string into a data pack.
 	//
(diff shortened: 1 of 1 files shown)


plugins/include/datapack.inc
```diff
@@ -58,6 +58,16 @@ methodmap DataPack < Handle
...
 	public native void WriteFloat(float val, bool insert = false);
+	
+	// Packs a vector into a data pack.
+	//
+	// @param val		Float array to add.
+	// @param insert	Determines whether mid-pack writes will insert instead of overwrite.
+	public void WriteVector(const float val[3], bool insert = false)
+	{
+		for(int x=0; x<3; ++x)
+			this.WriteFloat(val[x]);
```
rumblefrog: Seems like you forgot to pass insert to WriteFloat

plugins/include/datapack.inc
```diff
@@ -135,6 +154,19 @@ native void WritePackCell(Handle pack, any cell);
...
 native void WritePackFloat(Handle pack, float val);
 
+/**
+ * Packs a vector into a data pack.
+ *
+ * @param pack		Handle to the data pack.
+ * @param val			Float array to add.
+ * @error			Invalid handle.
+ */
+stock void WritePackVector(Handle pack, const float val[3])
```
Headline: We don't need to add oldstyle decl for new features

plugins/include/datapack.inc
```diff
@@ -171,6 +203,19 @@ native any ReadPackCell(Handle pack);
...
 native float ReadPackFloat(Handle pack);
 
+/**
+ * Reads a vector from a data pack.
+ *
+ * @param pack		Handle to the data pack.
+ * @param buffer		Destination array.
+ * @error			Invalid handle, or bounds error.
+ */
+stock void ReadPackVector(Handle pack, float buffer[3])
```
Headline: ditto

FortyTwoFortyTwo:
Would it be better to have `WriteCellArray`, `WriteFloatArray`, `ReadCellArray` and `ReadFloatArray` instead? Allowing to pass any array size rather than limited to 3, write/read vectors can be done using `WriteFloatArray`/`ReadFloatArray` with size 3.

Deathreus:
> Would it be better to have `WriteCellArray`, `WriteFloatArray`, `ReadCellArray` and `ReadFloatArray` instead? Allowing to pass any array size rather than limited to 3, write/read vectors can be done using `WriteFloatArray`/`ReadFloatArray` with size 3.

Actually, yeah, that would be far better, I'll commit that instead at the native level

Headline:
After conversation with @Deathreus we'll be leaving the vector changes for later and take the documentation ones for now.

peace-maker:
What happened with the submodules? Revert those changes before merging!

Headline:
@peace-maker thank you for keeping an eye out, but I’m aware.

PR #1167, opened 2020-02-01
Add default values to TeleportEntity

Scags:
Seems appropriate to me.
Now you'd be able to do something like:
```c#
float ang[3] = {-90.0, 0.0, 0.0};
TeleportEntity(client, .angles = ang);
```
With less taps on the keyboard.

### DIFF
--- a/plugins/include/sdktools_functions.inc
+++ b/plugins/include/sdktools_functions.inc
@@ -96,7 +96,7 @@ native void ExtinguishEntity(int entity);
  * @param velocity      New velocity, or NULL_VECTOR for no change.
  * @error               Invalid entity or client not in game, or lack of mod support.
  */
-native void TeleportEntity(int entity, const float origin[3], const float angles[3], const float velocity[3]);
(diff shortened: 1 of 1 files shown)


Headline:
I'm not sure if this is something we should consider promoting for the future of sourcepawn. @dvander can we get your input on this syntax and if it's something we're keeping?

Scags:
Regardless of whether or not named parameter operators are going to be kept, would that be a factor in implementing this pr? I think that defaulted null vectors would be beneficial, with tossing '.' around being a bonus.

Headline:
I believe so, I can’t think off the top of my head where else in the scripting api we’d promote using this syntax. I like the idea, but I want to make sure that it’s something we can move forward with

Thanks!

dvander:
Sorry for the late reply - yes, I'm a big fan of this syntax. It greatly improves readability. And it's worth the implementation cost, which is small.

KyleSanderson:
Indeed - the more places where we can make scripting easier on people the better off we will be in the long run. Big proponent of sane, default values.

Issue #1170, opened 2020-02-02
[TF2] PrecacheScriptSound not precaching MvM sounds


naydef:
# Help us help you
  - [x] I have checked that my issue [doesn't exist yet](https://github.com/alliedmodders/sourcemod/issues).
  - [x] I have tried my absolute best to reduce the problem-space and have provided the absolute smallest test-case possible.
  - [x] I can always reproduce the issue with the provided description below.

# Environment
  * Operating System version: Windows
  * Game/AppID (with version if applicable): Team Fortress 2
  * Current SourceMod version: Sourcemod 1.10.0
  * Current SourceMod snapshot: 6425
  * Current Metamod: Source snapshot: 1.10.7-dev
  - [ ] I have updated SourceMod to the [latest version](https://www.sourcemod.net/downloads.php) and it still happens.
  - [ ] I have updated SourceMod to the [latest snapshot](https://www.sourcemod.net/downloads.php?branch=dev) and it still happens.
  - [ ] I have updated SourceMM to the [latest snapshot](https://sourcemm.net/downloads.php?branch=dev) and it still happens.

# Description
[PrecacheScriptSound](https://sm.alliedmods.net/new-api/sdktools_sound/PrecacheScriptSound) is not precaching the sound files. Using PrecacheSound and directly using the sound path from the soundscript file(including sound characters) works. Error messages about non-precached sound files are printed to console, too. I've made a test code. After typing sm_sound_precache# and then sm_sound_play, I found that only sm_sound_test_precache2 allows sm_sound_play play the sound.
Also PrecacheScriptSound returns true
The server is:
running Freak Fortress 2
testing is done on non-mvm map

# Problematic Code (or Steps to Reproduce)
  ```
#include <sourcemod>
#include <sdktools>

public void OnPluginStart()
{
	RegConsoleCmd("sm_sound_test_precache1", Command_TestSoundPrecache1);
	RegConsoleCmd("sm_sound_test_precache2", Command_TestSoundPrecache2);
	RegConsoleCmd("sm_sound_test_precache3", Command_TestSoundPrecache3);
	RegConsoleCmd("sm_sound_play1", Command_TestSoundPlay1);
}

//Not working
public Action Command_TestSoundPrecache1(int client, int args)
{
	if(PrecacheScriptSound("MVM.Warning"))
	{
		ReplyToCommand(client, "Precache successful!");
	}
	return Plugin_Handled;
}

//Working
[...]

PR #1172, opened 2020-02-03
Fix issue #1170

naydef:
After doing some testing with a debugger, I noticed that `engsound->IsSoundPrecached(waveName)` returns `true`, even though the sound is not precached. Because I don't have the patience to compile Sourcemod, I replaced the jnz instruction with nops and now sounds are successfully precached.
For this reason, I removed `IsSoundPrecached` from `PrecacheSoundScript` and `GetGameSoundParameters`

I think it'll be a good idea to add a note to [IsSoundPrecached](https://sm.alliedmods.net/new-api/halflife/IsSoundPrecached) warning that this function may not work correctly(always returning `true`)

Fixes issue #1170

### DIFF
--- a/extensions/sdktools/vsound.cpp
+++ b/extensions/sdktools/vsound.cpp
@@ -677,11 +677,7 @@ bool InternalPrecacheScriptSound(const char *soundname)
 	for (int wave = 0; wave < waveCount; wave++)
 	{
 		const char* waveName = soundemitterbase->GetWaveName(internal->GetSoundNames()[wave].symbol);
-		// return true even if we precache no new wavs
-		if (!engsound->IsSoundPrecached(waveName))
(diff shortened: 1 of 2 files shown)


extensions/sdktools/vsound.cpp
```diff
@@ -1589,11 +1585,7 @@ static cell_t smn_GetGameSoundParams(IPluginContext *pContext, const cell_t *par
 
 	pContext->StringToLocal(params[6], params[7], soundParams.soundname);
 
-	// Precache the sound we're returning
```
asherkin: The comment got lost.

naydef:
I'm not sure what else is required here.

naydef:
Alright!

PR #1174, opened 2020-02-09
Remove specific CSGO check for 'cl_language'

SM9CC:
'cl_language' is a hidden ConVar in CSGO however it can be queried reliably and represents an accurate value.

### DIFF
--- a/core/PlayerManager.cpp
+++ b/core/PlayerManager.cpp
@@ -503,9 +503,6 @@ bool PlayerManager::OnClientConnect(edict_t *pEntity, const char *pszName, const
 	/* Get the client's language */
 	if (m_QueryLang)
 	{
-#if SOURCE_ENGINE == SE_CSGO
-		pPlayer->m_LangId = translator->GetServerLanguage();
-#else
 		const char *name;
 		if (!pPlayer->IsFakeClient() && (name=engine->GetClientConVarValue(client, "cl_language")))
 		{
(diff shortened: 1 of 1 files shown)


psychonic:
Has this been tested?

At least at one point, the ConVar did exist and was hidden, but still could not be accessed with IVEngineServer::GetClientConVarValue because it no longer had the FCVAR_USERINFO flag. (Just that part can be tested most easily by seeing if the GetClientInfo native can read the language)

sneak-it:
As of the latest CS:GO Update (2/3):

> – Game language can now be controlled in the game properties Language tab.

https://blog.counter-strike.net/index.php/2020/02/28378/

So perhaps this changed.

SM9CC:
Yes, it was tested by somebody in discord: https://i.imgur.com/8PBXRRR.png

psychonic:
> Yes, it was tested by somebody in discord: https://i.imgur.com/8PBXRRR.png

That only mentions it working with QueryClientConVar, not GetClientInfo. The former is an asynchronous looking that asks the client for a response with the value. The latter is fully serverside, looking up for the cache of the client-replicated userinfo ConVars.

SM9CC:
ok, I shall test it now.

SM9CC:
Sorry for jumping the gun here, I made a bad assumption, I shall try and come up with a better fix that makes use of a QueryClientConVar to cache the value.

PR #1175, opened 2020-02-09
Regex Little Changes

TheByKotik:
* Prevented to triple and double call `strlen`.
* More informative message on `if (offset >= len)`.
* Add missing parametr in navite `MatchRegex`.

P.S. I'm have little exp in c++, sorry if something wrong.

### DIFF
--- a/extensions/regex/CRegEx.cpp
+++ b/extensions/regex/CRegEx.cpp
@@ -94,7 +94,7 @@ int RegEx::Compile(const char *pattern, int iFlags)
 	return 1;
 }
 
-int RegEx::Match(const char *str, unsigned int offset)
+int RegEx::Match(const char *const str, size_t len, const size_t offset)
 {
 	int rc = 0;
 
@@ -104,10 +104,8 @@ int RegEx::Match(const char *str, unsigned int offset)
 	this->ClearMatch();
 
 	//save str
-	subject = new char[strlen(str)+1];
+	subject = new char[++len];
 	strcpy(subject, str);
-
-	unsigned int len = strlen(subject);
 
 	rc = pcre_exec(re, NULL, subject, len, offset, 0, mMatches[0].mVector, MAX_CAPTURES);
 
@@ -138,11 +136,11 @@ int RegEx::MatchAll(const char *str)
 	this->ClearMatch();
 
 	//save str
-	subject = new char[strlen(str) + 1];
+	size_t len = strlen(subject)+1;
+	subject = new char[len];
 	strcpy(subject, str);
 
-	unsigned int offset = 0;
-	unsigned int len = strlen(subject);
+	size_t offset = 0;
 	unsigned int matches = 0;
 
 	while (matches < MAX_MATCHES && offset < len && (rc = pcre_exec(re, 0, subject, len, offset, 0, mMatches[matches].mVector, MAX_CAPTURES)) >= 0)
--- a/extensions/regex/CRegEx.cpp
+++ b/extensions/regex/CRegEx.cpp
@@ -104,9 +104,9 @@ int RegEx::Match(const char *const str, const size_t offset)
 	this->ClearMatch();
 
 	//save str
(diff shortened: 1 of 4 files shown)


extensions/regex/CRegEx.cpp
```diff
@@ -138,11 +136,11 @@ int RegEx::MatchAll(const char *str)
 	this->ClearMatch();
 
 	//save str
-	subject = new char[strlen(str) + 1];
+	size_t len = strlen(subject)+1;
+	subject = new char[len];
 	strcpy(subject, str);
```
Headline: while we’re cleaning up around here, this should be `ke::SafeStrcpy`

plugins/include/regex.inc
```diff
@@ -199,12 +199,13 @@ native Regex CompileRegex(const char[] pattern, int flags = 0, char[] error="",
  * @param regex         Regex Handle from CompileRegex()
  * @param str           The string to check.
  * @param ret           Error code, if applicable.
+ * @param offset        Offset in the string to start searching from. MatchOffset returns the offset of the match.
```
Headline: The bit about MatchOffset can be redacted, this is already documented above

extensions/regex/CRegEx.cpp
TheByKotik: I just want to clarify, did I understand correctly, should this look like this?
```
size_t len = strlen( str );
subject = new char[len +1];
ke::Safestrcpy(subject, len, str);```

Comment (Headline) [reply extensions/regex/CRegEx.cpp]: since this is just a copy, this could actually just be a

```cpp
subject = strdup(str);
```

extensions/regex/extension.cpp
```diff
@@ -114,35 +114,33 @@ static cell_t MatchRegex(IPluginContext *pCtx, const cell_t *params)
...
 	{
 		pCtx->ThrowNativeError("Regex data not found\n");
-
 		return 0;
 	}
 
+	size_t offset = 0;
+	if (params[0] >= 4)
+	{
+		offset = static_cast<size_t>( params[4] );
```
Headline: Padding here is inconsistent with our convention throughout the project, you can just remove the padding

extensions/regex/CRegEx.cpp
```diff
@@ -132,17 +129,16 @@ int RegEx::MatchAll(const char *str)
...
+	if (mFree || re == nullptr)
 		return -1;
 
 	this->ClearMatch();
 
 	//save str
-	subject = new char[strlen(str) + 1];
-	strcpy(subject, str);
+	subject = strdup( str );
+	size_t len = strlen( subject );
```
Headline: ditto padding on the above lines

extensions/regex/CRegEx.cpp
```diff
@@ -94,22 +94,19 @@ int RegEx::Compile(const char *pattern, int iFlags)
...
 
 	//save str
-	subject = new char[strlen(str)+1];
-	strcpy(subject, str);
+	subject = strdup( str );
 
-	unsigned int len = strlen(subject);
-
-	rc = pcre_exec(re, NULL, subject, len, offset, 0, mMatches[0].mVector, MAX_CAPTURES);
+	rc = pcre_exec(re, nullptr, subject, strlen( subject ), offset, 0, mMatches[0].mVector, MAX_CAPTURES);
```
Headline: More padding, check line 107 as well

TheByKotik: Done. 👌
Sorry for that.

Headline:
Has this been tested?

TheByKotik:
> Has this been tested?

Windows build only with [test.sp](https://gist.github.com/TheByKotik/bb26b99d780e5f3adde6870490afa075).

KyleSanderson:
Not really a bug but a feature request. Nice hidden API exposure there buddy :-) thanks for the PR and resolving this.

PR #1179, opened 2020-02-12
Update PrepSDKCall_SetFromConf to use GameData

xerox8521:
This Updates PrepSDKCall_SetFromConf to use GameData instead of just Handle.
Updates the description to include Signature.
Updates _source_ parameter description to include Addresses.


asherkin:
Thank you for the PR, unfortunately this is a breaking change. Generally non-methodmap natives still take in `Handle` rather than the specific types - as while the specific types coerce to `Handle` automatically, the inverse is not true.

PR #1183, opened 2020-02-21
Add option for persistent previous map storage

stickz:
Adds option to persistently store the previous map vote history in a text file. The previous map ArrayList is repopulated after the configuration files are executed for the first time.

### DIFF
--- a/plugins/mapchooser.sp
+++ b/plugins/mapchooser.sp
@@ -219,7 +219,7 @@ public void OnConfigsExecuted()
 		}
 	}
 	
-	// Recall previous maps from a text file, if persistency is enabled
+	/* First-load previous maps from a text file when persistency is enabled. */
 	static bool g_FirstConfigExec = true;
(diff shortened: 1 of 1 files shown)


plugins/mapchooser.sp
```diff
@@ -105,6 +106,8 @@ int g_winCount[MAXTEAMS];
 #define VOTE_EXTEND "##extend##"
 #define VOTE_DONTCHANGE "##dontchange##"
 
+#define MAPCHOOSER_TXT "data/mapchooser_history.txt"
```
KyleSanderson: can you make this a string const as opposed to a define? reason is so it's easier to debug (and with a specific iteration of the language this was deprecated).

stickz: I noticed other plugins have hard coded this when calling BuildPath, so I did the same. It also doesn't need to be in the global scope because it's only used in 1 place. GetTextFilePath()

PR #1185, opened 2020-02-21
Add stock AssertStrEq to testing include

nosoop:
Adds a new stock function `AssertStrEq` to assert equality between two strings.

It's a minor convenience over `AssertTrue` since it prints the expected string without having to include it in the `text` argument.

### DIFF
--- a/plugins/include/testing.inc
+++ b/plugins/include/testing.inc
@@ -79,3 +79,17 @@ stock void AssertTrue(const char[] text, bool value)
 		ThrowError("test %d (%s in %s) failed", TestNumber, text, TestContext);
 	}
 }
+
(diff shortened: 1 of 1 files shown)


nosoop:
Sure; this came out of a desire to ensure strings were correct in the test suite for the [TF2 Econ Data compatibility shim](https://github.com/nosoop/SM-TFEconDataCompat), which checks to see that both the compatibility shim and the original TF2ItemsInfo natives function similarly.

I'd also like to make use of it in tests for the [double extension](https://github.com/XutaxKamay/double_ext) if this gets accepted.

KyleSanderson:
Yeah, you were right @Headline - thanks for bringing that up. thanks @nosoop for the contribution.

PR #1186, opened 2020-02-21
mapchooser: Set minimum number of maps to include in vote

stickz:
Allows user to have a consistent minimum number of maps in the vote, regardless how many previous maps are excluded.

### DIFF
--- a/plugins/mapchooser.sp
+++ b/plugins/mapchooser.sp
@@ -975,8 +975,8 @@ void CreateNextVote()
 	GetCurrentMap(map, sizeof(map));
 	RemoveStringFromArray(tempMaps, map);
 	
-	if (	g_Cvar_ExcludeMaps.IntValue && tempMaps.Length > g_Cvar_ExcludeMaps.IntValue
-		g_Cvar_IncludeMinMaps.IntValue && tempMaps.Length > g_Cvar_IncludeMinMaps.IntValue)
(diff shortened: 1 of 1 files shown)


plugins/mapchooser.sp
```diff
@@ -973,16 +975,24 @@ void CreateNextVote()
 	GetCurrentMap(map, sizeof(map));
 	RemoveStringFromArray(tempMaps, map);
 	
-	if (g_Cvar_ExcludeMaps.IntValue && tempMaps.Length > g_Cvar_ExcludeMaps.IntValue)
+	if (	g_Cvar_ExcludeMaps.IntValue && tempMaps.Length > g_Cvar_ExcludeMaps.IntValue &&
```
KyleSanderson: what's up with the tab?

stickz:
@KyleSanderson When testing my changes, I found a bug with the mapchooser. If "sm_mapvote_exclude" was set higher than the number of maps on the server, it would not exclude any previous maps. This bug also prevented a minimum option count from being set.

I created a new exclude algorithm to resolve this bug and make my changes work. I tested all 3 conditions inside the new while loop thoroughly, to ensure they work as intended.

stickz:
I forgot to mention. This new algorithm also fixes anther bug, where if you change "sm_mapvote_exclude" it may not take effect immediately. 

This is because the `g_OldMapList` just gets the first array list index erased on map end, if the array size is greater than the value of "sm_mapvote_exclude". Anther value is pushed at the same time.

The number of maps excluded is counted each time now. It stops when it reaches the value of "sm_mapvote_exclude".

PR #1187, opened 2020-02-22
Beacon changes (funcommands)

MartLegion:
Some changes to the beacon fun function.

Changes to funcommands.sp / beacon.sp / funcommands.games.txt:

- Created beacon color config through reading gamedata file (funcommands.games.txt).
- Added game specific color settings for "Team 2" and "Team 3" on L4D/L4D2. (They should be inverse as e.g: CSGO)
- Added different color for Team 4 (hidden team on L4D2).
- Added color config for unknown teams (teams that are not in: 1,2,3,4).
- Loading external beacon color (first beacon) from gamedata.
- Updated "if else" to "switch case" statement.
- Added "teamBeaconColor" var to set the beacon color based on the client's team, instead of calling TE_SetupBeamRingPoint on each point.
- Moved TE_SetupBeamRingPoint calls more near each other.


plugins/funcommands.sp
```diff
@@ -297,3 +340,17 @@ void AddTranslatedMenuItem(Menu menu, const char[] opt, const char[] phrase, int
 	Format(buffer, sizeof(buffer), "%T", phrase, client);
 	menu.AddItem(opt, buffer);
 }
+
+int[] GetColor(const char[] buffer)
```
asherkin: Very minor thing, but could you rename this to `ParseColor` as that is a more accurate description of what it does.

plugins/funcommands.sp
```diff
@@ -297,3 +340,17 @@ void AddTranslatedMenuItem(Menu menu, const char[] opt, const char[] phrase, int
...
+	ExplodeString(buffer, ",", sColor, sizeof(sColor), sizeof(sColor[]));
+	
+	int iColor[4];
+	iColor[0] = StringToInt(sColor[0]);
+	iColor[1] = StringToInt(sColor[1]);
+	iColor[2] = StringToInt(sColor[2]);
+	iColor[3] = StringToInt(sColor[3]);
+
+	return iColor;
+}
```
asherkin: Another small one, the file should have an extra newline at the end.

plugins/funcommands.sp
MartLegion: Done.

plugins/funcommands.sp
MartLegion: Done.

MartLegion:
Some var colors aren't being used now, like "greencolor" and "redcolor" so the build will fail.
I don't know-how is safe to remove this vars, any suggestion?

asherkin:
> Some var colors aren't being used now, like "greencolor" and "redcolor" so the build will fail.
> I don't know-how is safe to remove this vars, any suggestion?

I'd go with just removing the unused ones.

asherkin:
Looks like the changes from the previous commit got reverted.

MartLegion:
Yeah thanks! Repaired.

asherkin:
Thanks!

PR #1188, opened 2020-02-25
Fix wording of IsServerProcessing comment.

JoinedSenses:
Comment wording is incoherent.


JoinedSenses:
I added in context for hibernation; however, some folks in the SM discord suggest that timers will continue to work now in hibernation. Can anyone verify?

PR #1193, opened 2020-02-27
Update MersenneTwister.h

wanted241:
Update MersenneTwister.h

### DIFF
--- a/core/logic/MersenneTwister.h
+++ b/core/logic/MersenneTwister.h
@@ -1,14 +1,15 @@
 // MersenneTwister.h
 // Mersenne Twister random number generator -- a C++ class MTRand
 // Based on code by Makoto Matsumoto, Takuji Nishimura, and Shawn Cokus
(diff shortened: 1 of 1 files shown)


Headline:
What’s the motivation for this patch?

wanted241:
@Headline I by myself always prefer fresh and new products over the older ones. This is fresh and new. (Can't say it's very very new, but it's just fresher than what SM using right now.)

wanted241:
And better, I guess.

wanted241:
I think this should improve Randomize in Source Mod, since the code is only gets better.

KyleSanderson:
Yeah, unfortunately at this time despite there being a couple (unverified) sources online for a 1.1 update there's no inclination for us to change our rng as there's no documented defect for this one. Nothing on you, please don't let this be a preclusion to you contributing in the future. At this time as there's no documented issues with this rls of Mersenne Twister so we're going to stick it out.

PR #1196, opened 2020-02-28
mapchooser: Configure previous map storage

stickz:
**This pull request requires PR #1186 to be merged first.**

Configures how many previous maps to store. In the event `sm_mapvote_exclude` is increased, changes may take effect instantly, due to having a record of previous maps played.


KyleSanderson:
NAK. Why isn't `sm_maphistory_size` being used?

PR #1197, opened 2020-03-01
mapchooser: Fix previous map exclusion bug

stickz:
This commit fixes a bug where if the value of `sm_mapvote_exclude` is reduced, the change may not take effect for an extended period of time. The size of the old map list is not updated when this happens.

### DIFF
--- a/plugins/mapchooser.sp
+++ b/plugins/mapchooser.sp
@@ -260,7 +260,7 @@ public void OnMapEnd()
 	RemoveStringFromArray(g_OldMapList, map);
 	g_OldMapList.PushString(map);
 				
-	if (g_OldMapList.Length > g_Cvar_ExcludeMaps.IntValue)
+	while (g_OldMapList.Length > g_Cvar_ExcludeMaps.IntValue)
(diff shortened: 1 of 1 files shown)


KyleSanderson:
Interesting catch, thanks!

plugins/mapchooser.sp
```diff
@@ -260,7 +260,7 @@ public void OnMapEnd()
 	RemoveStringFromArray(g_OldMapList, map);
 	g_OldMapList.PushString(map);
 				
-	if (g_OldMapList.Length > g_Cvar_ExcludeMaps.IntValue)
+	while (g_OldMapList.Length > g_Cvar_ExcludeMaps.IntValue)
 	{
 		g_OldMapList.Erase(0);
 	}
```
TheByKotik, line 266: Maybe better do something like this?
```
if ( g_OldMapList.Length > g_Cvar_ExcludeMaps.IntValue )
{
	g_OldMapList.Resize( Cvar_ExcludeMaps.IntValue );
}	
```

PR #1201, opened 2020-03-04
Add an option to build against no SDKs

asherkin:
Quite often I find myself only needing to build the SDK-independent code, which we have more and more of these days.

### DIFF
--- a/AMBuildScript
+++ b/AMBuildScript
@@ -143,6 +143,7 @@ class SMConfig(object):
 
   def detectSDKs(self):
     sdk_list = builder.options.sdks.split(',')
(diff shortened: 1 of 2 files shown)


KyleSanderson:
Isn't this accomplished by building with `-s present`? if it isn't lets fix that instead?

asherkin:
> Isn't this accomplished by building with `-s present`? if it isn't lets fix that instead?

No, that (sanely, I think) still errors if you have 0 available. But that doesn't solve the problem I want to solve, which is having all the SDKs checked out but not wanting to use any of them to build.

PR #1213, opened 2020-03-14
sm_cookies displaying number before cookie name

MartLegion:
This is a PR related to #977

*Added a cookie number before cookie name.
*Removed unnecessary '\0'.
*Removed some unnecessary front whitespaces.
*Reordered some lines.

Special thanks to "eyal282" which had the idea and did most of the job.
Changes based on the "sm_help" command.

### DIFF
--- a/plugins/clientprefs.sp
+++ b/plugins/clientprefs.sp
@@ -66,14 +66,12 @@ public Action Command_Cookie(int client, int args)
 		/* Show list of cookies */
 		Handle iter = GetCookieIterator();
 		
+		CookieAccess access;
 		char name[30];
-		name[0] = '\0';
 		char description[255];
-		description[0] = '\0';
+		int i;
 		
 		PrintToConsole(client, "%t:", "Cookie List");
-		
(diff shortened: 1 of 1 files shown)


MartLegion:
Some notes:

I know that the previous PR #977 does almost the same but it is opened for a long time, so I did some fixes from my branch to be able to submit here and maybe, close the previous one.

In case what I did is not OK, just please close this PR and I will be sorry.

The only difference here from the eyal version is that I didn't add the part that shows up the cookie description when setting its value, IMO it didn't fit but I can add in case someone thinks that it is an interesting feature.

I was also based on sm_help plugin, so another difference is that in sm_help it shows empty descriptions as "No description available", I didn't add because I would have to change the translation file or even "links" this KV translation to the adminhelp.phrases.txt file.

I can also make the cookie "count" pad based on its size, like is doing on "sm plugins list" now, but I would have to make two loops to get the size first, so I thought it would be better not to, but in case it is interesting, again, I can do.

Thanks

PR #1217, opened 2020-03-21
Prevent clients from spamming global chat using sm_nominate

Douile:
Using the nominate command it is possible to spam global chat by alternating between two maps e.g.
`bind mwheeldown "sm_nominate de_dust2;sm_nominate de_mirage;"`
By replying to successful nominations instead of putting in global this issue is prevented as clients can only spam themselves.

### DIFF
--- a/plugins/nominations.sp
+++ b/plugins/nominations.sp
@@ -351,7 +351,7 @@ void AttemptNominate(int client, const char[] map, int size)
 	char name[MAX_NAME_LENGTH];
 	GetClientName(client, name, sizeof(name));
 
-	PrintToChatAll("[SM] %t", "Map Nominated", name, displayName);
+	ReplyToCommand(client, "[SM] %t", "Map Nominated", name, displayName);
 	
 	return;
 }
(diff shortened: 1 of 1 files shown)


asherkin:
How do you feel about broadcasting it if the result is not `Nominate_Replaced`? I think it is quite important to advertise the nominations as it lets people know they *can* nominate maps, but yeah it doesn't make sense to let people constantly do it like this.

Douile:
@asherkin I do think it would be useful to see nominations globally. The only other way I see to fix is by adding some kind of rate limiting to the command.

plugins/nominations.sp
```diff
@@ -351,7 +351,7 @@ void AttemptNominate(int client, const char[] map, int size)
 	char name[MAX_NAME_LENGTH];
 	GetClientName(client, name, sizeof(name));
 
-	PrintToChatAll("[SM] %t", "Map Nominated", name, displayName);
+	ReplyToCommand(client, "[SM] %t", "Map Nominated", name, displayName);
```
asherkin:
```suggestion
        if (result == Nominate_Added) {
	    PrintToChatAll("[SM] %t", "Map Nominated", name, displayName);
	} else {
	    ReplyToCommand(client, "[SM] %t", "Map Nominated", name, displayName);
	}
```

I think this should work - it'll only allow a client to print globally once per nomination cycle.

Naleksuh:
@asherkin You can do something similar with votes and sm_revote - should this be changed similarly?

PR #1219, opened 2020-03-26
Add an Array variant to DataPack's Read/Write methods

Deathreus:
I have not pen-tested this at all and only did a minimal functionality check with [this plugin](https://www.dropbox.com/s/uk3cn79lbnyv0f2/vip_locations.sp?dl=0) for proof of concept to handling vectors

With the pandemic going on I've been getting all my projects on the back burner out of the way, this being one of them

### DIFF
--- a/core/logic/CDataPack.cpp
+++ b/core/logic/CDataPack.cpp
@@ -122,31 +122,25 @@ void CDataPack::PackString(const char *string)
 	elements.insert(position++, val);
 }
 
-void CDataPack::PackCellArray(cell_t const *vals, size_t count)
+void CDataPack::PackCellArray(cell_t const *vals, cell_t count)
 {
 	InternalPack val;
 	val.type = CDataPackType::CellArray;
-	
-	if(count < 1)
-		return;
 
 	val.pData.aval = new cell_t [count + 1];
 	memcpy(&val.pData.aval[1], vals, sizeof(cell_t) * (count + 1));
-	val.pData.aval[0] = (cell_t)count;
+	val.pData.aval[0] = count;
 	elements.insert(position++, val);
 }
 
-void CDataPack::PackFloatArray(cell_t const *vals, size_t count)
+void CDataPack::PackFloatArray(cell_t const *vals, cell_t count)
 {
 	InternalPack val;
 	val.type = CDataPackType::FloatArray;
-	
-	if(count < 1)
-		return;
 
 	val.pData.aval = new cell_t [count + 1];
 	memcpy(&val.pData.aval[1], vals, sizeof(cell_t) * (count + 1));
-	val.pData.aval[0] = (cell_t)count;
+	val.pData.aval[0] = count;
 	elements.insert(position++, val);
 }
 
@@ -215,14 +209,18 @@ const char *CDataPack::ReadString(size_t *len) const
 	return val.chars();
 }
 
-cell_t *CDataPack::ReadCellArray(size_t *size) const
+cell_t *CDataPack::ReadCellArray(cell_t *size) const
 {
-	cell_t *ptr = nullptr;
 	if (!IsReadable() || elements[position].type != CDataPackType::CellArray)
-		return ptr;
+	{
+		if(size)
+			*size = 0;
+		
+		return nullptr;
+	}
 
 	cell_t *val = elements[position].pData.aval;
-	ptr = &(val[1]);
+	cell_t *ptr = &(val[1]);
 	++position;
 
 	if (size)
@@ -231,14 +229,18 @@ cell_t *CDataPack::ReadCellArray(size_t *size) const
 	return ptr;
 }
 
-cell_t *CDataPack::ReadFloatArray(size_t *size) const
+cell_t *CDataPack::ReadFloatArray(cell_t *size) const
 {
-	cell_t *ptr = nullptr;
 	if (!IsReadable() || elements[position].type != CDataPackType::FloatArray)
-		return ptr;
+	{
+		if(size)
+			*size = 0;
+		
+		return nullptr;
+	}
 
 	cell_t *val = elements[position].pData.aval;
-	ptr = &(val[1]);
+	cell_t *ptr = &(val[1]);
 	++position;
 
 	if (size)
--- a/core/logic/CDataPack.cpp
+++ b/core/logic/CDataPack.cpp
@@ -130,8 +130,8 @@ void CDataPack::PackCellArray(cell_t const *vals, size_t count)
 	if(count < 1)
 		return;
 
-	val.pData.aval = new cell_t [sizeof(cell_t) * count + sizeof(count)];
-	memcpy(&val.pData.aval[1], vals, sizeof(cell_t) * count + sizeof(count));
+	val.pData.aval = new cell_t [count + 1];
+	memcpy(&val.pData.aval[1], vals, sizeof(cell_t) * (count + 1));
 	val.pData.aval[0] = (cell_t)count;
 	elements.insert(position++, val);
 }
@@ -144,8 +144,8 @@ void CDataPack::PackFloatArray(cell_t const *vals, size_t count)
 	if(count < 1)
 		return;
 
-	val.pData.aval = new cell_t [sizeof(cell_t) * count + sizeof(count)];
-	memcpy(&val.pData.aval[1], vals, sizeof(cell_t) * count + sizeof(count));
+	val.pData.aval = new cell_t [count + 1];
+	memcpy(&val.pData.aval[1], vals, sizeof(cell_t) * (count + 1));
 	val.pData.aval[0] = (cell_t)count;
 	elements.insert(position++, val);
 }
--- a/core/logic/CDataPack.cpp
+++ b/core/logic/CDataPack.cpp
@@ -122,6 +122,36 @@ void CDataPack::PackString(const char *string)
 	elements.insert(position++, val);
 }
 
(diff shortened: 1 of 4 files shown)


Deathreus:
So actual arrays instead, I see, I'll look into it

Deathreus:
@asherkin 
How does that look?

core/logic/CDataPack.h
```diff
@@ -90,6 +92,19 @@ class CDataPack
 	 */
 	float ReadFloat() const;
 
+	/**
+	 * @brief Reads an array of values from the data stream.
+	 *
```
Headline, line 97: Missing @param

core/logic/CDataPack.h
```diff
@@ -90,6 +92,19 @@ class CDataPack
...
 
+	/**
+	 * @brief Reads an array of values from the data stream.
+	 *
+	 * @return			A cell array read from the current position.
+	 */
+	cell_t *ReadCellArray(size_t *len) const;
+	/**
+	 * @brief Reads an array of values from the data stream.
+	 *
```
Headline, line 103: ditto

core/logic/CDataPack.cpp
```diff
@@ -187,6 +215,38 @@ const char *CDataPack::ReadString(size_t *len) const
...
+	ptr = &(val[1]);
+	++position;
+
+	if (size)
+		*size = val[0];
+
+	return ptr;
+}
+
+cell_t *CDataPack::ReadFloatArray(size_t *size) const
```
Headline, line 234: can we accept `size_t&` here? I dont see why it'd need to be a pointer

core/logic/CDataPack.cpp
```diff
@@ -122,6 +122,34 @@ void CDataPack::PackString(const char *string)
...
+
+void CDataPack::PackFloatArray(cell_t const *vals, size_t count)
+{
+	InternalPack val;
+	val.type = CDataPackType::FloatArray;
+	
+	if(count < 1)
+		return;
+
+	val.pData.aval = new cell_t [sizeof(cell_t) * count + sizeof(count)];
```
Headline, line 147: hm; now that we're using new `sizeof(cell_t)` can be removed. This can just be `new cell_t[count+1]`.

core/logic/CDataPack.cpp
```diff
@@ -122,6 +122,34 @@ void CDataPack::PackString(const char *string)
...
+	memcpy(&val.pData.aval[1], vals, sizeof(cell_t) * count + sizeof(count));
```
Headline, line 148: with the above in mind, we can just copy `sizeof(cell_t)*(count+1)` bytes

core/logic/smn_datapacks.cpp
```diff
@@ -166,6 +166,72 @@ static cell_t smn_WritePackString(IPluginContext *pContext, const cell_t *params
...
+	sec.pIdentity = g_pCoreIdent;
+
+	if ((herr = handlesys->ReadHandle(hndl, g_DataPackType, &sec, (void **)&pDataPack))
+		!= HandleError_None)
+	{
+		pContext->ReportError("Invalid data pack handle %x (error %d).", hndl, herr);
+		return 0;
+	}
+
+	bool insert = (params[0] >= 4) ? params[4] : false;
```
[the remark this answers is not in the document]
Deathreus: I thought it wouldn't be if you don't provide a argument for a defaulted parameter

plugins/include/datapack.inc
```diff
@@ -92,6 +106,18 @@ methodmap DataPack < Handle
 	// @return              Function pointer.
 	public native Function ReadFunction();
 
+	// Reads a cell array from a data pack.
+	//
+	// @param buffer        Destination buffer.
+	// @param count         Maximum length of output buffer.
+	public native void ReadCellArray(any[] buffer, int count);
+
+	// Reads a cell array from a data pack.
```
peace-maker, line 115: Reads a float array!

core/logic/smn_datapacks.cpp
```diff
@@ -312,6 +372,110 @@ static cell_t smn_ReadPackFunction(IPluginContext *pContext, const cell_t *param
...
+	}
+
+	if (pDataPack->GetCurrentType() != CDataPackType::CellArray)
+	{
+		pContext->ReportError("Invalid data pack type (got %d / expected %d).", pDataPack->GetCurrentType(), CDataPackType::CellArray);
+		return 0;
+	}
+
+	size_t packCount = 0;
+	cell_t *pData = pDataPack->ReadFloatArray(&packCount);
```
asherkin, line 406: ReadCellArray?

plugins/include/datapack.inc
```diff
@@ -71,6 +71,20 @@ methodmap DataPack < Handle
 	// @param insert        Determines whether mid-pack writes will insert instead of overwrite.
 	public native void WriteFunction(Function fktptr, bool insert = false);
 
+	// Packs a cell array into a data pack.
```
asherkin, line 74: “an array of cells” for all of these reads better

core/logic/CDataPack.cpp
```diff
@@ -187,6 +215,38 @@ const char *CDataPack::ReadString(size_t *len) const
 	return val.chars();
 }
 
+cell_t *CDataPack::ReadCellArray(size_t *size) const
+{
+	cell_t *ptr = nullptr;
+	if (!IsReadable() || elements[position].type != CDataPackType::CellArray)
+		return ptr;
```
asherkin, line 222: This took a 2nd look to understand - I would prefer it if the `ptr` variable was declared where it’s actually used down below and this returned `nullptr` directly.

core/logic/CDataPack.cpp
```diff
@@ -122,6 +122,34 @@ void CDataPack::PackString(const char *string)
 	elements.insert(position++, val);
 }
 
+void CDataPack::PackCellArray(cell_t const *vals, size_t count)
+{
+	InternalPack val;
+	val.type = CDataPackType::CellArray;
+	
+	if(count < 1)
```
asherkin, line 130: This is failing silently but there is nothing on the path to here from a plugin to reject it, so would be an error on reading.

I think we should just allow 0-length arrays - the code here will work fine with this check removed, and it’ll just need some checks removed in the read path as well by the looks of it.

core/logic/CDataPack.cpp
```diff
@@ -122,6 +122,34 @@ void CDataPack::PackString(const char *string)
...
+{
+	InternalPack val;
+	val.type = CDataPackType::CellArray;
+	
+	if(count < 1)
+		return;
+
+	val.pData.aval = new cell_t [count + 1];
+	memcpy(&val.pData.aval[1], vals, sizeof(cell_t) * (count + 1));
+	val.pData.aval[0] = (cell_t)count;
```
asherkin, line 135: Rather than casting it, why not just take a `cell_t` as the size? It’s what the SP side is passing anyway :)

core/logic/smn_datapacks.cpp
```diff
@@ -166,6 +166,66 @@ static cell_t smn_WritePackString(IPluginContext *pContext, const cell_t *params
 	return 1;
 }
 
+static cell_t smn_WritePackCellArray(IPluginContext *pContext, const cell_t *params)
+{
+	Handle_t hndl = static_cast<Handle_t>(params[1]);
+	size_t count = params[3];
```
asherkin, line 172: Add new things with C-style declaring at the start isn’t super desired, would be good to move all these `count` variables to where they’re used. Generally the preference is to try and read the `params` array in order where it makes sense and wouldn’t involve creating temporaries for the sake of it, as it makes understanding the logic coming from SP easier.

core/logic/smn_datapacks.cpp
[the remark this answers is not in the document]
asherkin: It’ll be there as long as the argument is declared on the SP side - where we need to check the length is when a new param is added to an existing native.

core/logic/smn_datapacks.cpp
Deathreus: I may as well change the other natives to match that while I'm at it

core/logic/smn_datapacks.cpp
Deathreus: Whoops

Deathreus:
Is there anything missing still?

core/logic/CDataPack.cpp
```diff
@@ -237,6 +299,13 @@ bool CDataPack::RemoveItem(size_t pos)
 			delete elements[pos].pData.sval;
 			break;
 		}
+
+		case CDataPackType::CellArray:
+		case CDataPackType::FloatArray:
+		{
+			delete elements[pos].pData.aval;
```
Ne3tCode, line 306: I think there must be `delete []` 🤔

Headline: Since it's POD, it doesn't really matter too much. If you'd like to open a PR to fix this that would be appreciated

Ne3tCode: > Since it's POD, it doesn't really matter too much.

Relying on your "knowledge" of how this works is a bad idea. AFAIK, mixing `new/new[]` with `delete[]/delete` is undefined behavior, this is all implementation specific and dependent on the underlying memory manager.

No, I'm not going to open PR.

Headline: Thanks!

dvander: It's poor form but doesn't matter here. The correct thing to do is to remove this code entirely and implement a destructor for the container.

PR #1221, opened 2020-03-31
Change unnecessary int[]'s to any[]

TotallyMehis:
Nothing should stop you from using eg. a float array on these natives.

It gets kinda annoying to use `view_as<int>()`

### DIFF
--- a/plugins/include/entity.inc
+++ b/plugins/include/entity.inc
@@ -715,7 +715,7 @@ native int GetEntPropArraySize(int entity, PropType type, const char[] prop);
  * @param dataSize      Size of each value in bytes (1, 2, or 4).
  * @error               Invalid entity or offset out of reasonable bounds.
  */
-stock void GetEntDataArray(int entity, int offset, int[] array, int arraySize, int dataSize=4)
+stock void GetEntDataArray(int entity, int offset, any[] array, int arraySize, int dataSize=4)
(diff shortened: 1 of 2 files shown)


JoinedSenses:
Has this been tested with enum structs?

asherkin:
Any thoughts on how this does/should behave when `dataSize`/`size` is not equal to 4 and a float array is passed?

> Has this been tested with enum structs?

They don't have a stable in-memory representation so I don't think it really matters, but there shouldn't be any technical problems with these and them.

TotallyMehis:
> Any thoughts on how this does/should behave when `dataSize`/`size` is not equal to 4 and a float array is passed?

In that case, yea, it would make sense to have a separate func for floats like `GetEntDataArrayFloat` that doesn't allow the data size to be changed.

I guess this PR depends on the trajectory and philosophy of SP. Personally, I want to give plugin authors all the power. Want to pass an enum struct or float array and do some weird magic with the data size? Go ahead.

KyleSanderson:
@asherkin I think this is cool. The natives are already pretty unsupported ({G,S}etEntData in the sense that they're completely unsafe - opening this up to what was intended (arbitrary sizes) will likely make things more clear.

With this being said, if we macro'd this that would be the way to make it more robust.

PR #1224, opened 2020-04-07
Added some new string manipulation stocks

milutinke:
Added the following stocks:
- IsStringNumber
- IsStringFloat
- StringToLowerCase
- StringToUpperCase
- IndexOfChar
- LastIndexOfChar
- CopyC

Tested:
Test plugin source link: https://pastebin.com/hqfSZNxY
Test results: https://i.imgur.com/4CawNKq.png


plugins/include/string.inc
```diff
@@ -557,3 +557,149 @@ stock int ImplodeStrings(const char[][] strings, int numStrings, const char[] jo
...
 	return total;
 }
+
+/**
+ * Checks if the given string is an integer (Supports negative values)
+ *
+ * @param szString     Input string
+ * @return             Boolean
+ */
+stock bool IsStringNumber(const char[] szString)
```
dvander, line 567: Does need a temporary string? I think it would work to skip past spaces and then call StringToIntExt() and check the return value.

plugins/include/string.inc
```diff
@@ -557,3 +557,149 @@ stock int ImplodeStrings(const char[][] strings, int numStrings, const char[] jo
...
+ * Converts the given string to lower case
+ *
+ * @param szString     Input string for conversion and also the output
+ * @return             void
+ */
+stock void StringToLowerCase(char[] szInput) 
+{
+	int iIterator = 0;
+	int iLength = strlen(szInput);
+	char[] szOutput = new char[iLength + 1];
```
dvander, line 596: Is a temporary needed here, given that the string is modified in place?

plugins/include/string.inc
```diff
@@ -557,3 +557,149 @@ stock int ImplodeStrings(const char[][] strings, int numStrings, const char[] jo
...
+ * Converts the given string to upper case
+ *
+ * @param szString     Input string for conversion and also the output
+ * @return             void
+ */
+stock void StringToUpperCase(char[] szInput) 
+{
+	int iIterator = 0;
+	int iLength = strlen(szInput);
+	char[] szOutput = new char[iLength + 1];
```
dvander, line 620: Same comment as above.

plugins/include/string.inc
```diff
@@ -557,3 +557,149 @@ stock int ImplodeStrings(const char[][] strings, int numStrings, const char[] jo
...
+ * @return             Boolean
+ */
+stock bool IsStringFloat(const char[] szString) 
+{
+	char cCharacter;
+	int iIterator;
+	int iP;
+	int iLength = strlen(szString);
+	
+	while (0 < szString[iIterator] <= 255) {
```
dvander, line 647: It will be much easier to write this as a wrapper around StringToFloatEx(), which will also make sure the parsing is consistent.

plugins/include/string.inc
milutinke: I never heard of StringToIntExt and I cannot find it in the documentation.
The reason I used a temporary value is to avoid modifying the original string when trimming.

dvander: Sorry, StringToIntEx. You don't need to modify the string if you skip spaces with a counter.

milutinke:
I have simplified the StringToUpper and StringToLower case.
Regarding the FindCharInString, I did not know about that function, I can change IndexOfChar and LastIndexOfChar to wrap the FindCharInString.
IsStringNumber and IsStringFloat do not work intended when they wrap StringToInt and StringToFloat. Results of the test with those: https://i.imgur.com/4l1JzGd.png

peace-maker:
That's why dvander suggested using the `Ex` variants of those natives. They return the number of characters consumed for the number, so checking if the whole string was consumed does what you want.

```
stock bool IsStringInteger(const char[] str, int nBase=10)
{
	int result;
	return StringToIntEx(str, result, nBase) == strlen(str);
}

stock bool IsStringFloat(const char[] str)
{
	float result;
	return StringToFloatEx(str, result) == strlen(str);
}
```

dvander:
If you want to be very strict, you can do:

```
stock bool IsStringFloat(const char[] str)
{
	float result;
	return StringToFloatEx(str, result) == strlen(str) && !IsStringInt(str);
}
```

milutinke:
Changed the code to apply your advices.

Changed IsStringNumber and IsStringFloat to wrap StringToIntEx and StringToFloatEx.
Changed IndexOfChar to wrap FindCharInString.
Added CopyC

KyleSanderson:
@milutinke I've made some adjustments that I thought were appropriate. What do you think?

KyleSanderson:
@milutinke any update on this?

KyleSanderson (approved):

milutinke:
> @milutinke any update on this?
Sorry for replying this late, I've not received a notification about this for some reason back then.
CopyC does not exist in Source Mod, at least to my knowledge.

Alienmario:
> If the goal is to add more string stocks for common stuff, I feel like borrowing a few more ideas from python would make sense. Such as IsStringAlpha/Numeric/AlNum, IsStringASCII, IsStringPrintable, StringEndsWith. There's already char versions of some of those, so making string equivalents doesn't seem unreasonable.

+1 I've missed StringEndsWith in almost every other plugin I made.

Alienmario:
Sorry to bump, can we get some update on this? I feel like converting String case is a very fundamental operation and it would be nice to have SM ship with it.

peace-maker:
Would adding wrappers for other available stock functions really help or make the API more confusing?

I've reviewed this again and am in favor of adding more string utility stocks as suggested, but am hesitant on adding wrappers like `IndexOfChar` -> `FindCharInString`. Having multiple functions doing the same thing doesn't help readability of code. But `FindCharInString` is a mouthful and not the best choice of name for what it does after all.

milutinke:
> Would adding wrappers for other available stock functions really help or make the API more confusing?
> 
> I've reviewed this again and am in favor of adding more string utility stocks as suggested, but am hesitant on adding wrappers like `IndexOfChar` -> `FindCharInString`. Having multiple functions doing the same thing doesn't help readability of code. But `FindCharInString` is a mouthful and not the best choice of name for what it does after all.

People who come from other programming languages are used to IndexOfChar name, so it would not hurt having it in my opinion.
But yeah, I am fine without it being added too.

Alienmario:
@sirdigbot Maybe you should make a PR after all, as this one appears cursed.
I can help providing some stocks I've been using.

milutinke:
For some reason I did not get notifications until now (all of them) for this PR even I'm subscribed.
Since I've stopped doing any CS:GO related stuff years ago, I'll close the PR, thus a new one can be opened with the proposed changes.

Alienmario:
> Sure, go ahead. I'll take a look. Though I already have a massive list of proposed stocks from the last time I had a go at a PR.

```
stock bool StrEndsWith(const char[] str, const char[] suffix, bool caseSensitive = true)
{
	return StrEndsWithEx(str, strlen(str), suffix, caseSensitive);
}

stock bool StrEndsWithEx(const char[] str, int realLen, const char[] suffix, bool caseSensitive = true)
{
	int suffixLen = strlen(suffix);
	return (realLen >= suffixLen && StrEqual(str[realLen - suffixLen], suffix, caseSensitive));
}

// Non-ex func already used by arraylist
stock int FindStringInArrayEx(const char[][] array, int len, const char[] str, bool caseSensitive = true)
{
	for (int i = 0; i < len; i++)
	{
		if (StrEqual(str, array[i], caseSensitive))
		{
			return i;
		}
	}
	return -1;
}

// Non-ex funcs already used by arraylist
stock int FindValueInArrayEx(const any[] array, int len, const any val)
{
	for (int i = 0; i < len; i++)
	{
		if (array[i] == val)
		{
			return i;
		}
	}
	return -1;
}

stock int FindCharInArray(const any[] array, int len, const int val, bool caseSensitive = true)
{
	int i = FindValueInArrayEx(array, len, val);
	if (!caseSensitive && i == -1)
	{
		int val2 = IsCharLower(val) ? CharToUpper(val) : CharToLower(val);
		if (val2 != val)
		{
			i = FindValueInArrayEx(array, len, val2);
		}
	}
	return i;
}

// Like SplitString, but more efficient?
stock int SplitByChar(const char[] szSource, char c, char[] szBuffer, int nSize)
{
	int i = 0;
	while (szSource[i] != c)
	{
		if (szSource[i] == '\0')
		{
			return -1;
		}
		i++;
	}
	i++;
	strcopy(szBuffer, i < nSize ? i : nSize, szSource);
	return i;
}
```
If you pick any, I can clean them up and add docs. I especially like StrEndsWithEx as it lets me check for multiple file extensions without calculating string length repeatedly.

Additionally I think FindCharInString stock can be optimized (in forward mode at least) by removing strlen at the beginning and checking for end of string inside the loop.

Alienmario:
> StrEndsWithEx I feel is too specific and your use case is better replaced by regex.

Here is my use case https://github.com/Alienmario/smartdm-redux/blob/main/scripting/include/smartdm_redux.inc#L110
I fell like regex is a overkill for such a simple function.
The gains may be smaller than expected, but I like having the option.
Keep in mind strlen computation time grows with length of string.

PR #1231, opened 2020-04-14
fix(docs::menus) - fixed docgen picking up wrong or no offset for symbols

rumblefrog:
This commit fixes the documentation for MenuAction entries to prevent the docgen from picking up none or wrong offset for the entries.

Previously, `MenuAction_End` docs were assigned to `MenuAction_VoteEnd`, leaving `_End` offset empty


asherkin:
While we've moved away from it, this comment style should be valid so this is really a bug in docgen not consuming the entire comment. The `<` character means the doccomment should be attached to the preceding documentable item, rather than the next.

PR #1232, opened 2020-04-14
Make GetStringTableData native binary-safe

### LINKED ISSUE #1230: GetStringTableData uses string conversion function but can contain raw bytes
# Help us help you

# Environment
  * Game/AppID (with version if applicable): CS:GO
  * Current SourceMod version: 1.10.0.6.6460
  - [ ] I have updated SourceMod to the [latest version](https://www.sourcemod.net/downloads.php) and it still happens.
  - [ ] I have updated SourceMod to the [latest snapshot](https://www.sourcemod.net/downloads.php?branch=dev) and it still happens.

# Description
Some network string tables inside the engine send raw data from meory, not strings (e. g. the `userinfo` table as seen [here](https://github.com/VSES/SourceEngine2007/blob/43a5c90a5ada1e69ca044595383be67f40b33c61/se2007/engine/baseserver.cpp#L777)). Sadly, SourceMod uses a bytes-to-string conversion function inside `GetStringTableData`, so everything after `\0` is truncated.

Problem: https://github.com/alliedmodders/sourcemod/blob/aae71612731eaa6771067a31f3a7ad34bdb0df28/extensions/sdktools/vstringtable.cpp#L207

(Same goes true for `SetStringTableData`).

You can validate by reading any `userinfo` entry with `GetStringTableData` and compare the return value (which is the length of string characters read instead of bytes read) with `GetStringTableDataLength` (which returns the correct size).

# Possible solution
Either create an additional function to read raw data or allow to specify a flag to the function.

thewavelength:
Replaced StringToLocalUTF8 with LocalToString and memcpy to make this binary-safe. Discussed this with @asherkin on Discord.

This probably won't hurt as it looks like the function was originally intended to be binary safe.

resolves #1230

### DIFF
--- a/extensions/sdktools/vstringtable.cpp
+++ b/extensions/sdktools/vstringtable.cpp
@@ -202,8 +202,8 @@ static cell_t GetStringTableData(IPluginContext *pContext, const cell_t *params)
 	char *addr;
 	pContext->LocalToString(params[3], &addr);
 
-	size_t maxBytes = params[4];
-	size_t numBytes = MIN(datalen, maxBytes);
+	int maxBytes = params[4];
+	size_t numBytes = MIN(maxBytes, datalen);
 
 	if (userdata)
 	{
--- a/extensions/sdktools/vstringtable.cpp
+++ b/extensions/sdktools/vstringtable.cpp
@@ -212,7 +212,6 @@ static cell_t GetStringTableData(IPluginContext *pContext, const cell_t *params)
 	else if (datalen > 0)
 	{
 		addr[0] = '\0';
-		datalen = 1;
 	}
 
 	return datalen;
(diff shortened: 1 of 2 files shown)


extensions/sdktools/vstringtable.cpp
```diff
@@ -199,14 +199,22 @@ static cell_t GetStringTableData(IPluginContext *pContext, const cell_t *params)
 	}
 
 	userdata = (const char *)pTable->GetStringUserData(stringidx, &datalen);
-	if (!userdata)
+
+	char *addr;
+	pContext->LocalToString(params[3], &addr);
+
+	datalen = params[4];
```
asherkin, line 206: You don't want to do this:
* `datalen` is the length of the real data the engine has.
* `params[4]` is the length of the buffer in the plugin.

So the `MAX()` before was the right, it just had to use `datalen` rather than `numBytes`.

Headline:
Do we have to worry about people relying on the return value to SetStringTableData being the bytes written or are we fine to change that behavior?

asherkin:
> Do we have to worry about people relying on the return value to SetStringTableData being the bytes written or are we fine to change that behavior?

It hasn't changed - as usual the docs were wrong.

@thewavelength as it only returns not-1 when throwing an error, it should actually be `void` with no `@return` in the docs.

Issue #1233, opened 2020-04-17
m_AllocSize corrupted on allocation failure


Kailo97:
I'm investigated this crush https://crash.limetech.org/7luvujw7t3ni
And here is 100% sure, what here is problem, because we must to get "error spam" in log, but not crush.
In console we also see this error in some moments before crush (probably in previous frame)
```
L 04/17/2020 - 21:30:29: [SM] Exception reported: Failed to grow array
L 04/17/2020 - 21:30:29: [SM] Blaming: SurfTimer.smx
L 04/17/2020 - 21:30:29: [SM] Call stack trace:
L 04/17/2020 - 21:30:29: [SM]   [0] PushArrayArray
L 04/17/2020 - 21:30:29: [SM]   [1] Line 649, surftimer/replay/replay.sp::RecordReplay
L 04/17/2020 - 21:30:29: [SM]   [2] Line 950, surftimer/hooks/hooks.sp::OnPlayerRunCmd
```
https://github.com/alliedmodders/sourcemod/blob/0ac78758eb8254aed212acb44a0a9d0ad35bb0d2/core/logic/CellArray.h#L210-L241
So, i see here very important bug to fix:
m_AllocSize changed before we do allocation and confirm success or failure.
Steps:
```
State 1:
m_AllocSize = N
m_Size = N
```
Frame M:
We tried to push (PushArrayArray -> CellArray::push
(sub call to CellArray::GrowIfNeeded with count = 1)
`m_Size + count > m_AllocSize` N + 1 > N -> dobule m_AllocSize
realloc failed and return NULL
Get "Failed to grow array" error
```
State 2:
m_AllocSize = N * 2
m_Size = N
```
**We failed to alloc, but change allocation size!!!**
Frame M+1:
So, we have also to push
Grow will think what we have enough space to fit array, but space not accually allocated.
https://github.com/alliedmodders/sourcemod/blob/0ac78758eb8254aed212acb44a0a9d0ad35bb0d2/core/logic/CellArray.h#L212-L216
So, memcopy will try to write, and here is get out of bounds.

This bug also will be on memalloc, if we failed to alloc. m_Data will be NULL, but m_AllocSize not zero for access at next time.

covertxd:
This issue was resolved in the SurfTimer once the memory leak was fixed. The issue occurred when SRCDS was at 4GB memory usage which resulted in an OOM error.

PR #1235, opened 2020-04-21
Fix crash when ArrayList runs out of memory #1233

peace-maker:
The allocation size was still updated to the bigger size even if memory allocation failed. Trying to write to the supposedly available new space would overflow the heap and crash.

### DIFF
--- a/core/logic/CellArray.h
+++ b/core/logic/CellArray.h
@@ -35,6 +35,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <ICellArray.h>
+#include <amtl/am-bits.h>
 
 extern HandleType_t htCellArray;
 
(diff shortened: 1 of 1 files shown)


core/logic/CellArray.h
```diff
@@ -214,30 +214,32 @@ class CellArray : public ICellArray
...
 		/* Set a base allocation size of 8 items */
-		if (!m_AllocSize)
+		if (!newAllocSize)
 		{
-			m_AllocSize = 8;
+			newAllocSize = 8;
 		}
 		/* If it's not enough, keep doubling */
-		while (m_Size + count > m_AllocSize)
+		while (m_Size + count > newAllocSize)
```
dvander, line 224: This (old) code is missing overflow checks as well, but I'll leave it to you whether or not you want to fix them.

core/logic/CellArray.h
```diff
@@ -214,30 +214,32 @@ class CellArray : public ICellArray
...
-			if (!data)  // allocation failure
-			{
-				return false;
-			}
-
-			m_Data = data;
+			data = static_cast<cell_t*>(realloc(m_Data, sizeof(cell_t) * m_BlockSize * newAllocSize));
 		} else {
-			m_Data = static_cast<cell_t*>(malloc(sizeof(cell_t) * m_BlockSize * m_AllocSize));
+			data = static_cast<cell_t*>(malloc(sizeof(cell_t) * m_BlockSize * newAllocSize));
```
dvander, line 234: realloc(NULL, size) is the same as malloc(NULL, size), btw.

peace-maker:
Good points, I've included the overflow checks from `ke::Vector`.

PR #969, opened 2019-03-17
Move appveyor in-tree and convert to MSVC2015.

dvander:

PR #1236, opened 2020-04-21
Always use hl2sdk-proxy-repo in checkout-deps

peace-maker:
There were two separate issues with the checkout-deps scripts on windows and linux.

Windows didn't use the hl2sdk repository mirror "hl2sdk-proxy-repo" when cloning the different engine branches into their own folders like `hl2sdk-csgo`, `hl2sdk-css`, ... but cloned from github.com every time. When rerunning the checkout-deps.ps1 script later it would fail to find the "hl2sdk-proxy-repo" folder and fail to update the sdks.

Linux did use the local mirror repository for the initial clone of the different game sdk branches, but used github.com for later updates of the game sdks instead of using the local mirror.

Fix both scripts to always use the local mirror.

Looks like the AppVeyor config has to be updated too if it doesn't use the checkout-deps.ps1 script. #969

### DIFF
--- a/tools/checkout-deps.ps1
+++ b/tools/checkout-deps.ps1
@@ -56,13 +56,13 @@ Function Checkout-Repo
         Set-Location $Name
         If ($Origin)
         {
-            & git remote set-url origin ..\$Repo
+            & git remote set-url origin ..\$Repo 2>&1 | Write-Host
         }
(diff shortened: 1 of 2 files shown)


dvander:
That seems reasonable.

PR #1241, opened 2020-04-23
CStrike Ext: Add native to retrieve loadout slot of weapon

roflmuffin:
Some events, mainly the "dm_bonus_weapon_start" event, pass the loadout slot of the weapon that is now the bonus weapon, rather than an item index or a weapon id.

We are already creating a map of this in the cstrike extension, this native simply returns another member from that mapped value.

Let me know if I have missed anything, do I also need to update the .inc file in the tests folder?

### DIFF
--- a/extensions/cstrike/natives.cpp
+++ b/extensions/cstrike/natives.cpp
@@ -947,6 +947,21 @@ static cell_t CS_WeaponIDToItemDefIndex(IPluginContext *pContext, const cell_t *
 #endif
 }
 
+static cell_t CS_WeaponIDToLoadoutSlot(IPluginContext *pContext, const cell_t *params)
+{
+#if SOURCE_ENGINE == SE_CSGO
+	WeaponIDMap::Result res = g_mapWeaponIDToDefIdx.find((SMCSWeapon)params[1]);
+
+	if (!res.found())
+		return  pContext->ThrowNativeError("Invalid weapon id passed.");
+
+	return res->value.m_iLoadoutSlot;
+#else
+	return pContext->ThrowNativeError("CS_WeaponIDToLoadoutSlot is not supported on this game");
+#endif
+}
+
+
 sp_nativeinfo_t g_CSNatives[] = 
 {
 	{"CS_RespawnPlayer",			CS_RespawnPlayer}, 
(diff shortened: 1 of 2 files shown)


plugins/include/cstrike.inc
```diff
@@ -442,6 +442,16 @@ native CSWeaponID CS_ItemDefIndexToID(int iDefIndex);
  */
 native int CS_WeaponIDToItemDefIndex(CSWeaponID id);
 
+/**
+ * Returns the loadout slot based on the CSWeaponID.
+ *
```
Headline, line 447: We should probably mention it's only useful for CSGO here

extensions/cstrike/natives.cpp
```diff
@@ -947,6 +947,21 @@ static cell_t CS_WeaponIDToItemDefIndex(IPluginContext *pContext, const cell_t *
...
+{
+#if SOURCE_ENGINE == SE_CSGO
+	WeaponIDMap::Result res = g_mapWeaponIDToDefIdx.find((SMCSWeapon)params[1]);
+
+	if (!res.found())
+		return  pContext->ThrowNativeError("Invalid weapon id passed.");
+
+	return res->value.m_iLoadoutSlot;
+#else
+	return pContext->ThrowNativeError("CS_WeaponIDToLoadoutSlot is not supported on this game");
```
KyleSanderson, line 960: why?

KyleSanderson:
@roflmuffin are you okay to continue working on this one?

roflmuffin:
I'm sorry, I'm not familiar enough with this repo, I initially created this PR as a launching off point. For the commit I basically just copied the existing CS_WeaponIDToItemDefIndex and replaced the value it was returning, but I'm not sure what changes you are requesting, sorry.

KyleSanderson:
https://github.com/alliedmodders/sourcemod/pull/1241#issuecomment-670333916

You're likely going to need the game to test unfortunately. My apologies for the delay on the response, if you'd like to connect with me offline (or though here) I'd be happy to try and help.

peace-maker:
I think `g_mapWeaponIDToDefIdx` is only initialized for CSGO.
@KyleSanderson do you think it's best to add a hardcoded list of loadoutslots to return for the limited set of weapons available in CSS?

peace-maker (approved):
Lets get this into the tree for CS:GO already. Adding the same info for CS:S can be done in a later step.

KyleSanderson (approved):

Headline (approved):

PR #1245, opened 2020-04-26
GeoIP2 extension to sourcemod

Accelerator74:
I updated the geoip extension for sourcemod using my geoip2 extension.


KyleSanderson:
@Accelerator74 can you resolve the merge conflicts.
@psychonic do you know the latest on this stuff? I know it's been a minute since we looked at this stuff but you're way more present with the maxmind stuff than I am.

Accelerator74:
Added automatic search for the database file, because the mmdb format is also used by [DB-IP](https://db-ip.com/db) (so that you don't have to rename database files).

KyleSanderson:
@peace-maker this looks okay to me but I haven't looked at the upstream source. WDYT?

extensions/geoip/extension.cpp
```diff
@@ -96,51 +154,229 @@ inline void StripPort(char *ip)
...
+{
+	char *ip;
+	char *lang;
+	pCtx->LocalToString(params[1], &ip);
+	pCtx->LocalToString(params[4], &lang);
+	StripPort(ip);
+
+	if (strlen(lang) == 0) strcpy(lang, "en");
+
+	const char *path[] = {"country", "names", lang, NULL};
```
peace-maker, line 248: How does libmaxminddb handle path lookups of invalid languages? Does it fall back to "en" on it's own?

If it just doesn't return anything for invalid languages, this should check if the supplied language is supported in the `metadata.languages` array and fall back to "en" if it's not.
The same needs to be done for all other natives supporting that language parameter.

.gitmodules
```diff
@@ -1,6 +1,6 @@
-[submodule "public/amtl"]
```
peace-maker, line 1: No need to touch this file.

extensions/geoip/extension.cpp
```diff
@@ -96,51 +154,229 @@ inline void StripPort(char *ip)
...
+	std::string str = lookupString(ip, path);
+	const char *ccode = str.c_str();
+
+	pCtx->StringToLocal(params[2], params[3], ccode);
```
peace-maker, line 252: StringToLocalUTF8

Same applies to all other natives.

extensions/geoip/extension.cpp
```diff
@@ -96,51 +154,229 @@ inline void StripPort(char *ip)
...
+
+static cell_t sm_Geoip_CountryEx(IPluginContext *pCtx, const cell_t *params)
+{
+	char *ip;
+	char *lang;
+	pCtx->LocalToString(params[1], &ip);
+	pCtx->LocalToString(params[4], &lang);
+	StripPort(ip);
+
+	if (strlen(lang) == 0) strcpy(lang, "en");
```
peace-maker, line 246: It's not allowed to write to data returned by `LocalToString`. This can cause a buffer overflow in the VM.

plugins/include/geoip.inc
```diff
@@ -69,6 +87,93 @@ native bool GeoipCode3(const char[] ip, char ccode[4]);
  */
 native bool GeoipCountry(const char[] ip, char[] name, int maxlength);
 
+/**
+ * Gets the full country name.
+ *
+ * @param ip			Ip to determine the country code.
+ * @param name			Destination string buffer to store the country name.
+ * @param maxlength			Maximum length of output string buffer.
+ * @param lang			Which language to the output of result the geoip2 lookup.
```
peace-maker, line 96: How about: `The desired language of the output as a language code.`
Same for all other natives with that parameter.

peace-maker (changes requested):
This is cool stuff! Some comments inline.

Also is there an actual use case to supply the desired output language as a string like `"en"` to the natives?
I only see it being used to translate the country or whatever to a client's language to display it nicely. So you would always do. 
```
char sClientLang[8];
GetLanguageInfo(GetClientLanguage(client), sClientLang, sizeof(sClientLang), "", 0);
GeoipCountryEx(ip, ..., sClientLang);
```

Why not pass the target client index instead and lookup the client's language directly in the extension?

extensions/geoip/extension.cpp
Accelerator74: Returns emptiness. In the native, there is a check for an empty result :)

For example:
```
#include <geoip>

public OnPluginStart()
{
	char temp[64];
	if (GeoipCountry("8.8.8.8", temp, sizeof(temp), "ch"))
	{
		PrintToServer("true %s", temp);
	}
	else
	{
		PrintToServer("false %s", temp);
	}
}
```
Result `false`

Is it still worth checking languages?)

extensions/geoip/extension.cpp
```diff
@@ -29,44 +29,102 @@
...
 	g_pShareSys->RegisterLibrary(myself, "GeoIP");
-	g_pSM->LogMessage(myself, "GeoIP database info: %s", GeoIP_database_info(gi));
+
+	char date[40];
+	const time_t epoch = (const time_t)mmdb.metadata.build_epoch;
+	strftime(date, 40, "%F %T UTC", gmtime(&epoch));
+
+	g_pSM->LogMessage(myself, "GeoIP2 database loaded: %s (%s)", mmdb.metadata.database_type, date);
+
+	char buf[64];
```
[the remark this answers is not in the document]
Accelerator74: This simply cannot be if the database was able to load)

The buffer should be sufficient.

The maxmind database contains: `de en es fr ja pt-BR ru zh-CN` (buffer size 29 characters)
In DBIP: `en de es fr ja pt-BR ru zh-CN fa ko` (buffer size 35 characters)

.gitmodules
Accelerator74: This came from the geoip2-wip branch, from psychonic, when the libmaxminddb library was loaded there.

extensions/geoip/extension.cpp
Accelerator74: I don't quite understand what kind of fix is needed here :)

Accelerator74:
> Why not pass the target client index instead and lookup the client's language directly in the extension?

For example, I don’t want some countries to be displayed in the player’s language and I want to choose which one language to display a message about geo data by IP address.

extensions/geoip/extension.cpp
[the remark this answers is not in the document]
peace-maker: The [docs ](http://maxmind.github.io/libmaxminddb/index.html) say

> There is no requirement that the database metadata include languages or descriptions, so the count for these parts of the metadata can be zero. All of the other MMDB_metadata_s fields should be populated.

So we should handle that case for the future.

Cool, the buffer size is fine I guess 👍

plugins/include/geoip.inc
```diff
@@ -69,6 +91,89 @@ native bool GeoipCode3(const char[] ip, char ccode[4]);
...
+ *
+ * @param lat1       The first IP latitude.
+ * @param lon1       The first IP longitude.
+ * @param lat2       The second IP latitude.
+ * @param lon2       The second IP longitude.
+ * @param system     The system of measurement, 0 = Metric(kilometers) or 1 = English(miles).
+ *
+ * @return           The distance as result in specified system of measurement.
+ */
+native float GeoipDistance(float lat1, float lon1, float lat2, float lon2, system = SYSTEM_METRIC);
```
peace-maker, line 175: Whops, `system` needs a type here.

extensions/geoip/extension.cpp
peace-maker: Hm, I see. Is this the most intuitive and useful behavior though? Most likely you really want to display/store the data when you use one of the Geoip* natives, so if lookup for a language fails, you'd have to manually fall back to some other language. Every plugin would have to implement that logic, so wouldn't it be useful to do it straight in the natives to save some typing?

Instead of checking metadata.languages, which apparently might be empty, we could just see if the result is empty and try again with the server's language. This would mimic the fallback behavior of the general translation system in SourceMod. `Desired language -> server language -> english`

extensions/geoip/extension.cpp
peace-maker: Hm, since the natives have a default value of `"en"` already, maybe just removing that line is enough? Together with the fallback to the server language, this'll behave the same.

extensions/geoip/extension.cpp
Accelerator74: Nevertheless, it seems to me it will be strange if the plugin tries, for example, to find a country name on Japanese language, which is not in the database, and instead of an empty result it will display information in English. For me this is not logical)

The main task of the extension is, after all, the definition of geo data, and the choice of language, it seems to me, is only an additional option. Whether to use it or not, everyone must decide. The functionality of the sourcemod allows to determine the language of the player and who needs it, he will use it.

English remains the nationwide language, in which we actually display information by default :)

Accelerator74: >Instead of checking metadata.languages, which apparently might be empty, we could just see if the result is empty and try again with the server's language.

Besides, if we check not by the array of languages (`mmdb.metadata.languages.names`), we will have to access the database 2 more times (if suddenly the server language is also not found in the database). Yes, this is not essential, but why complicate the code so much? :)

extensions/geoip/extension.cpp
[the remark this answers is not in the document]
Accelerator74: Well, so far I have never seen in 2 years that the list of languages is not displayed :)

extensions/geoip/extension.cpp
Accelerator74: Ok, i added language check :)

peace-maker:
Thank you! The fallback to the server language looks good.

I'll wait for a comment from other maintainers about the `const char[] lang` vs. `int client` parameter question to determine which language to use as proposed in my previous comment. Deliberately choosing a different language than the player's selected one to display a message to them sounds like a very "niche" use case. And you could still use `SetClientLanguage` briefly to force it to something else..

Accelerator74:
Keep in mind that there are already some plugins that use this natives, because the geoip2 extension was released in 2018 :) But we can just add additional natives that will require the `client index`, not `char [] lang`.

Accelerator74:
Nevertheless, I made the definition of the language by the client index :) I see colleagues from the amxmodx team did the same :)

P.S.: By the way, yes, most of the code was borrowed from them, but it seems that everything is one team :)

KyleSanderson:
@peace-maker do you need anything from me to help move this one forward?

peace-maker:
@KyleSanderson No, this looks good. I didn't notice the changes after my last comment, so this is fine now from my side. If you don't have any strong feelings towards selecting the country language by string, 🚢 
There are a couple of nits left like using `.c_str()` and then calculating the length again using `strlen()` while we could just use the length of the `std::string` and indentation issues in the include doc comments.

But above all this requires changes to the packaging scripts, since we can't ship the old `GeoIp.dat` with this anymore. We can't include the latest database either due to licensing requirements as you know. We can merge this and do a follow up to fix the packaging and documentation afterwards though, since that's not something Accelerator74 can do.

plugins/include/geoip.inc
```diff
@@ -46,7 +62,7 @@
  *
  * @param ip            Ip to determine the country code.
  * @param ccode         Destination string buffer to store the code.
- * @return              True on success, false if no country found.
+ * @return				True on success, false otherwise.
```
peace-maker, line 65: Please use spaces instead of tabs for text alignment. Same for all documentation in this file!

extensions/geoip/extension.cpp
```diff
@@ -96,51 +158,263 @@ inline void StripPort(char *ip)
...
-	ccode = GeoIP_country_code_by_addr(gi, ip);
+	const char *path[] = {"country", "iso_code", NULL};
+	std::string str = lookupString(ip, path);
+	const char *ccode = str.c_str();
 
-	pCtx->StringToLocal(params[2], 3, ccode ? ccode : "");
+	pCtx->StringToLocalUTF8(params[2], 3, ccode, NULL);
 
-	return ccode ? 1 : 0;
+	return (strlen(ccode) != 0) ? 1 : 0;
```
peace-maker, line 170: No need to call `strlen`, since we can just use `str.length()`. This pattern appears in nearly all other native callbacks as well.

peace-maker (approved):

extensions/geoip/extension.cpp
Accelerator74: Hopefully ok now) I left `strlen` for RegionCode only. It's just that this function contains "hard" code :octocat:

PR #1246, opened 2020-04-26
Added VoiceHook functions

Accelerator74:
Added hooks when a player uses a microphone:
```
/**
 * Called when a client is speaking.
 *
 * @param client        The client index
 */
forward void OnClientSpeaking(int client);

/**
 * Called once a client speaking end.
 *
 * @param client        The client index
 */
forward void OnClientSpeakingEnd(int client);
```

And native function:
```
/**
 * Retrieves if the certain player is speaking.
 *
 * @param client        The client index
 * @return              True if player is speaking, false otherwise.
 */
native bool IsClientSpeaking(int client);
```


JoinedSenses:
Is there any other useful data that can be passed to this forward? otherwise, it seems not very useful.

Accelerator74:
> Is there any other useful data that can be passed to this forward? otherwise, it seems not very useful with only client index

Some people need a talking player hook event.

PR #1247, opened 2020-04-26
Added VoiceHook functions

Accelerator74:
Added hooks when a player uses a microphone:
```
/**
 * Called when a client is speaking.
 *
 * @param client        The client index
 */
forward void OnClientSpeaking(int client);

/**
 * Called once a client speaking end.
 *
 * @param client        The client index
 */
forward void OnClientSpeakingEnd(int client);
```

And native function:
```
/**
 * Retrieves if the certain player is speaking.
 *
 * @param client        The client index
 * @return              True if player is speaking, false otherwise.
 */
native bool IsClientSpeaking(int client);
```

Sorry for double request. My fails :(


JoinedSenses:
Adding comment here since other was deleted. Is there any other info that can be passed to this forward making it more useful besides just client?

Accelerator74:
> Is there any other info that can be passed to this forward making it more useful besides just client?

No, but there are plugins that use these hooks.

CrazyHackGUT:
Why support only CS:GO? I think, this PR should not be merged before you don't add support to another possible games.

Accelerator74:
It's supported for all games. For csgo used ClientVoice hook, for others - ProccesVoiceData.

OSCAR-WOS:
This is a feature I would fully take advantage off if implemented

extensions/sdktools/extension.cpp
```diff
@@ -57,6 +57,9 @@ SH_DECL_HOOK6(IServerGameDLL, LevelInit, SH_NOATTRIB, false, bool, const char *,
 #if SOURCE_ENGINE == SE_CSS || SOURCE_ENGINE == SE_CSGO
 SH_DECL_HOOK1_void_vafmt(IVEngineServer, ClientCommand, SH_NOATTRIB, 0, edict_t *);
 #endif
+#if SOURCE_ENGINE == SE_CSGO
```
KyleSanderson, line 60: this is incomplete. Swarm is supported + other engines.

extensions/sdktools/voice.cpp
```diff
@@ -214,6 +236,28 @@ void SDKTools::OnClientDisconnecting(int client)
...
+	IClient *pClient = (IClient *)((intptr_t)(META_IFACEPTR(IClient)) - 4);
+	if (pClient != NULL)
+	{
+		int client = pClient->GetPlayerSlot() + 1;
+
+		g_fSpeakingTime[client] = gpGlobals->curtime;
+
+		if (g_hTimerSpeaking[client] == NULL)
+		{
+			g_hTimerSpeaking[client] = timersys->CreateTimer(this, 0.3f, (void *)(intptr_t)client, TIMER_FLAG_REPEAT);
```
KyleSanderson, line 251: TIMER_FLAG_NO_MAPCHANGE?

extensions/sdktools/voice.cpp
```diff
@@ -165,8 +168,27 @@ bool SDKTools::OnSetClientListening(int iReceiver, int iSender, bool bListen)
...
 void SDKTools::OnClientDisconnecting(int client)
 {
+#if !defined CLIENTVOICE_HOOK_SUPPORT
+	IClient *pClient = iserver->GetClient(client-1);
+	if (pClient != NULL)
+	{
+		SH_REMOVE_HOOK(IClientMessageHandler, ProcessVoiceData, (IClientMessageHandler *)((intptr_t)(pClient) + 4), SH_MEMBER(this, &SDKTools::ProcessVoiceData), true);
+	}
+#endif
+
```
KyleSanderson, line 191: In this same hook you need to invalidate g_hTimerSpeaking.

Accelerator74: I did not kill the timer to that in any case hook worked even after disconnecting the client (OnClientSpeakingEnd). But if need to kill the timer, I will rewrite the code.
0.3 was taken specifically so as not to produce timers. This delay is also used in this plugin https://forums.alliedmods.net/showthread.php?t=245384 Of course, can use less delay, but this makes little sense, as it seems to me :)

Accelerator74: I did not kill the timer to that in any case hook worked even after disconnecting the client (OnClientSpeakingEnd). But if need to kill the timer, I will rewrite the code.
0.3 was taken specifically so as not to produce timers. Of course, can use less delay, but this makes little sense, as it seems to me :)

Accelerator74:
> Where'd you get 0.3f from? I remember a peer saying this on direct frames said this wasn't enough so I'd like to understand where the metric came from.

0.3 was taken so as not to create many timers when called. This delay is also used in franc1sco's VoiceAnnounceEX plugin. Of course, can use less delay, but this makes little sense, as it seems to me :)

KyleSanderson:
> > Where'd you get 0.3f from? I remember a peer saying this on direct frames said this wasn't enough so I'd like to understand where the metric came from.
> 
> 0.3 was taken so as not to create constantly new timers when called and the OnClientSpeakingEnd hook is called correctly. This delay is also used in franc1sco's VoiceAnnounceEX plugin.

Yeah - this one in particular was really buggy with a clean imp... if there's issues we need to remember that this value is too low.

extensions/sdktools/voice.cpp
```diff
@@ -165,8 +168,33 @@ bool SDKTools::OnSetClientListening(int iReceiver, int iSender, bool bListen)
 	RETURN_META_VALUE(MRES_IGNORED, bListen);
 }
 
+#if !defined CLIENTVOICE_HOOK_SUPPORT
+void SDKTools::OnClientConnected(int client)
+{
+	IClient *pClient = iserver->GetClient(client-1);
+	if (pClient != NULL)
+	{
+		SH_ADD_HOOK(IClientMessageHandler, ProcessVoiceData, (IClientMessageHandler *)((intptr_t)(pClient) + 4), SH_MEMBER(this, &SDKTools::ProcessVoiceData), true);
```
KyleSanderson, line 177: CVTableHook

Accelerator74: I hope everything is true)

extensions/sdktools/hooks.cpp
```diff
@@ -502,6 +553,16 @@ void CHookManager::OnPluginLoaded(IPlugin *plugin)
 			}
 		}
 	}
+	
+#if !defined CLIENTVOICE_HOOK_SUPPORT
```
KyleSanderson, line 557: This needs to be wrapped in the same way as FILE_used, otherwise you're invoking this callback for every single plugin that loads, which can be hundreds or thousands.

extensions/sdktools/hooks.cpp
```diff
@@ -538,6 +599,15 @@ void CHookManager::OnPluginUnloaded(IPlugin *plugin)
 		m_netChannelHooks.clear();
 		FILE_used = false;
 	}
+	
+#if !defined CLIENTVOICE_HOOK_SUPPORT
+	for (size_t i = 0; i < m_netProcessVoiceData.length(); ++i)
```
KyleSanderson, line 604: This is bugged, one plugin unloading will unhook everything.

Accelerator74:
> There's a merge conflict
I don’t understand what’s wrong?)
```
Conflicting files
extensions/sdktools/hooks.h
```

asherkin:
#1283 changed these hook lists from `ke::Vector` to `std::vector`. You'll want to `git checkout master` then `git pull` to update your local master branch, then `git checkout voicehook` and `git rebase master` to re-apply your changes on top of master - you'll then need to change the new `ke::Vector` you've added to a `std::vector` following the example of the other ones in that PR. Hopefully there should be no conflicts while updating, git should be able to figure it all out. If you have any trouble give a shout and someone can rebase it up for you.

extensions/sdktools/voice.cpp
```diff
@@ -355,6 +360,31 @@ static cell_t IsClientMuted(IPluginContext *pContext, const cell_t *params)
...
+	{
+		return pContext->ThrowNativeError("Client %d is not connected", params[1]);
+	}
+	else if (!player->IsInGame())
+	{
+		return pContext->ThrowNativeError("Client %d is not in-game", params[1]);
+	}
+	else if (player->IsFakeClient())
+	{
+		return pContext->ThrowNativeError("Client %d is a bot", params[1]);
```
asherkin, line 382: I think it would be best to just return `false` for fakeclients - alternatively this needs including in the `@error` docblock. Same for the `IsInGame` check possibly?

extensions/sdktools/extension.h
```diff
@@ -61,6 +61,12 @@
 #include <itoolentity.h>
 #endif
 
+#if SOURCE_ENGINE == SE_ALIENSWARM || SOURCE_ENGINE == SE_PORTAL2 || SOURCE_ENGINE == SE_INSURGENCY || SOURCE_ENGINE == SE_DOI || SOURCE_ENGINE == SE_BLADE || SOURCE_ENGINE == SE_CSGO
+#define CLIENTVOICE_HOOK_SUPPORT
+#else
+#include <inetmsghandler.h>
```
asherkin, line 67: This appears to be the first time this header is being used in SourceMod - that's generally a good sign something should be in gamedata instead (especially as we don't care about the param or return value, so it can just be forward declared.

Happy to take this as-is using the SDK, but have you checked that the SDKs are correct on the engines that use this branch?

extensions/sdktools/voice.cpp
```diff
@@ -167,6 +167,11 @@ bool SDKTools::OnSetClientListening(int iReceiver, int iSender, bool bListen)
 
 void SDKTools::OnClientDisconnecting(int client)
 {
+	if (g_hTimerSpeaking[client])
+	{
+		timersys->KillTimer(g_hTimerSpeaking[client]);
```
asherkin, line 172: Shouldn't this be setting `g_hTimerSpeaking[client]` to `nullptr`?

Otherwise if this happened close to map change the start condition wont be true and the forward will never fire again for that index.

extensions/sdktools/extension.cpp
```diff
@@ -532,6 +557,44 @@ void SDKTools::OnSendClientCommand(edict_t *pPlayer, const char *szFormat)
...
+		m_OnClientSpeakingEnd->Execute();
+
+		return Pl_Stop;
+	}
+	return Pl_Continue;
+}
+
+void SDKTools::OnTimerEnd(ITimer *pTimer, void *pData)
+{
+	g_hTimerSpeaking[(int)(intptr_t)pData] = NULL;
```
asherkin, line 575: `nullptr` is preferred over `NULL` for new code - applies to all of them.

extensions/sdktools/hooks.cpp
```diff
@@ -459,6 +493,29 @@ bool CHookManager::SendFile(const char *filename, unsigned int transferID)
...
+	IClient *pClient = (IClient *)((intptr_t)(META_IFACEPTR(IClient)) - 4);
+	if (pClient != NULL)
+	{
+		int client = pClient->GetPlayerSlot() + 1;
+
+		g_fSpeakingTime[client] = gpGlobals->curtime;
+
+		if (g_hTimerSpeaking[client] == NULL)
+		{
+			g_hTimerSpeaking[client] = timersys->CreateTimer(&g_SdkTools, 0.3f, (void *)(intptr_t)client, TIMER_FLAG_REPEAT|TIMER_FLAG_NO_MAPCHANGE);
```
asherkin, line 508: `TIMER_FLAG_NO_MAPCHANGE` looks unnecessary and the timers are being cleaned up on disconnect - it's not really safe to use when storing the handle anyway.

extensions/sdktools/extension.cpp
```diff
@@ -532,6 +557,44 @@ void SDKTools::OnSendClientCommand(edict_t *pPlayer, const char *szFormat)
...
+}
+
+#if defined CLIENTVOICE_HOOK_SUPPORT
+void SDKTools::OnClientVoice(edict_t *pPlayer)
+{
+	if (pPlayer)
+	{
```
asherkin, line 582: Early return is preferred over indentation of the entire body.

extensions/sdktools/voice.cpp
Accelerator74: When calling KillTimer it seems like the OnTimerEnd function is called, which already assigns `nullptr to the parameter :)
https://github.com/alliedmodders/sourcemod/blob/master/core/TimerSys.cpp#L393

extensions/sdktools/extension.h
Accelerator74: I don’t quite understand where exactly this needs to be moved :)

I went through all the sdk to check for the existence of a `ClientVoice` call. Where it is absent, `ProcessVoiceData` is called. Everything must work correctly.

extensions/sdktools/extension.cpp
```diff
@@ -532,6 +556,43 @@ void SDKTools::OnSendClientCommand(edict_t *pPlayer, const char *szFormat)
...
+	if (!pPlayer)
+	{
+		return;
+	}
+
+	int client = IndexOfEdict(pPlayer);
+
+	if (g_hTimerSpeaking[client])
+	{
+		timersys->KillTimer(g_hTimerSpeaking[client]);
```
Accelerator74, line 586: Maybe it’s better to use delete instead KillTimer, what do you think of this?

extensions/sdktools/voice.cpp
asherkin: Yep - all good then, I’m not very familiar with the C++ timer API. 

Lots of old code will be using `NULL`.

extensions/sdktools/extension.cpp
asherkin: Na, KillTimer is correct here.

extensions/sdktools/voice.cpp
```diff
@@ -355,6 +360,31 @@ static cell_t IsClientMuted(IPluginContext *pContext, const cell_t *params)
...
+	}
+	else if (!player->IsConnected())
+	{
+		return pContext->ThrowNativeError("Client %d is not connected", params[1]);
+	}
+	else if (!player->IsInGame())
+	{
+		return false;
+	}
+	else if (player->IsFakeClient())
```
KyleSanderson, line 380: Ah, fake clients can actually speak. Can you remove this check?

KyleSanderson:
@asherkin what's the skinny on this one?

asherkin:
Sorry for the delay cycling back to this, things have been super crazy with moving.

This is all looking good to me other than one issue, which is that the native is a no-op if there are no active consumers of the hook. I think we should just pull this in without the native to get the hook in-tree, as plugins can easily re-build the native behaviour using the hook and it can always be discussed separately if super common.

Was going to make that change myself before pulling it (and get someone else to do a final review), but it has been too long already so leaving this comment with the reasoning if you or KyleS want to make that change before I get a chance (but I'll make sure to review and pull if that happens.)

PR #1249, opened 2020-04-26
Update basevotes

Accelerator74:
When a player leaves during a voteban, he will be banned anyway. Also added a cvar with a ban time setting.

### DIFF
--- a/plugins/basevotes.sp
+++ b/plugins/basevotes.sp
@@ -109,7 +109,7 @@ public void OnPluginStart()
 	g_Cvar_Limits[0] = CreateConVar("sm_vote_map", "0.60", "percent required for successful map vote.", 0, true, 0.05, true, 1.0);
 	g_Cvar_Limits[1] = CreateConVar("sm_vote_kick", "0.60", "percent required for successful kick vote.", 0, true, 0.05, true, 1.0);	
 	g_Cvar_Limits[2] = CreateConVar("sm_vote_ban", "0.60", "percent required for successful ban vote.", 0, true, 0.05, true, 1.0);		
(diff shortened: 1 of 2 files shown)


crashzk:
This could be done in the Mute, Gag & Silence polls. I believe you must have the same problem right? If the player leaves the game during the vote, he is not punished.

asherkin:
> This could be done in the Mute, Gag & Silence polls. I believe you must have the same problem right? If the player leaves the game during the vote, he is not punished.

A player could re-join after any of those to have the effect removed anyway - that is where `voteban` is different.

plugins/basevotes.sp
```diff
@@ -108,6 +109,7 @@ public void OnPluginStart()
 	g_Cvar_Limits[0] = CreateConVar("sm_vote_map", "0.60", "percent required for successful map vote.", 0, true, 0.05, true, 1.0);
 	g_Cvar_Limits[1] = CreateConVar("sm_vote_kick", "0.60", "percent required for successful kick vote.", 0, true, 0.05, true, 1.0);	
 	g_Cvar_Limits[2] = CreateConVar("sm_vote_ban", "0.60", "percent required for successful ban vote.", 0, true, 0.05, true, 1.0);		
+	g_Cvar_Voteban = CreateConVar("sm_voteban_time", "30", "time ban in minutes for successful ban vote.", 0, true, 0.0);
```
Headline, line 112: this can probably be better described as "length of ban in minutes"

Accelerator74: For me it works at normally :)

Accelerator74:
> Just one small comment inline
> 
> Has this been tested?

For me it works at normally :)

PR #1250, opened 2020-04-26
Change int to any with Address Functions

Scags:
Since getting from and storing to addresses gives no regard to the type you throw at it, it makes sense to change these to `any`.

Would help with having to do some obnoxious tagging at the moment:
```cs
// Before
public void OnPluginStart()
{
	Address foo;
	float bar = 12.0;
	StoreToAddress(foo, view_as< int >(bar), NumberType_Int32);
	g_flFooBar = view_as< float >(LoadFromAddress(foo, NumberType_Int32));
}
// After
public void OnPluginStart()
{
	Address foo;
	float bar = 12.0;
	StoreToAddress(foo, bar, NumberType_Int32);
	g_flFooBar = LoadFromAddress(foo, NumberType_Int32);
}
```

### DIFF
--- a/plugins/include/sourcemod.inc
+++ b/plugins/include/sourcemod.inc
@@ -676,6 +676,7 @@ enum Address
  *
  * @param addr          Address to a memory location.
  * @param size          How many bytes should be read.
+ *                      If loading a floating-point value, use NumberType_Int32.
  * @return              The value that is stored at that address.
  */
 native any LoadFromAddress(Address addr, NumberType size);
@@ -686,6 +687,7 @@ native any LoadFromAddress(Address addr, NumberType size);
  * @param addr          Address to a memory location.
  * @param data          Value to store at the address.
  * @param size          How many bytes should be written.
+ *                      If storing a floating-point value, use NumberType_Int32.
  */
 native void StoreToAddress(Address addr, any data, NumberType size);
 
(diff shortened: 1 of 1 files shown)


nosoop:
There's one pitfall with this change on `LoadFromAddress`:

```
Address pFloat; // address of a float value
float flSomeMathResult = LoadFromAddress(pFloat, NumberType_Int32) + 1.0; 
```
The addition operator will always treat the left operand as an int, in either case you'd have to explicitly reinterpret it as a float but the current documentation makes the integer behavior clear.

I think having a separate `LoadFromAddressFloat` and / or a compiler warning on implicit `any` conversion in general would help there.

Scags:
Fair point, but there's two things here:

1: People who are even trying to do this kind of stuff would (should) be aware of how float operations work.
2: There are already SourceMod natives that perform the same action in the same manner (like `GetArrayCell/ArrayList.Get`).

I think compiler warnings on implicit `any` -> `float` are a bit overkill, though. On the other hand, a `#pragma` to warn against it on certain functions would be a cool feature.

dvander:
I can't think of any reason not to do this, but it should be clarified that they're 32-bit loads if that's not done already.

Scags:
Certainly. Added notes concerning bit size with the latest commit.

PR #1251, opened 2020-04-27
Fixed memory leak

hlstriker:
When a pack was cleared or destroyed the String and Raw types could cause memory leaks. This happens when "position" is sitting at the end of the vector and can never get past the "if (pos >= elements.length())" statement. This means there is a memory leak in any plugin that clears/destroys a pack with strings and doesn't set the position to length-1 or less beforehand.

### DIFF
--- a/core/logic/CDataPack.cpp
+++ b/core/logic/CDataPack.cpp
@@ -65,12 +65,13 @@ CDataPack::Free(CDataPack *pack)
 
 void CDataPack::Initialize()
 {
+	position = 0;
+	
 	do
 	{
 	} while (this->RemoveItem());
 
 	elements.clear();
-	position = 0;
 }
 
 void CDataPack::ResetSize()
@@ -214,6 +215,7 @@ bool CDataPack::RemoveItem(size_t pos)
 	{
 		pos = position;
 	}
+	
 	if (pos >= elements.length())
 	{
 		return false;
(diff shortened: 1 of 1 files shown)


core/logic/CDataPack.cpp
```diff
@@ -214,6 +215,7 @@ bool CDataPack::RemoveItem(size_t pos)
 	{
 		pos = position;
 	}
+	
 	if (pos >= elements.length())
```
KyleSanderson, line 219: This is where I think the real fix is. should just be pos > elements.length(), no?

Headline: pos is just our index; without >= `elements[pos].type` below this can throw

hlstriker: Last night I thought maybe that was the fix as well but after looking at all the edge cases you do want the original statement there.

KyleSanderson:
@hlstriker can I get a r? on that additional delete operator for the array? I think after that we should be good here.

dvander:
This fix looks very brittle. Really, the RemoveItem loop should not even be necessary. I would either:

(1) Remove the CDataPack cache, since it almost certainly useless, or
(2) Forbid copy construction on CDataPack::InternalPack, then add a move constructor and destructor.

hlstriker:
@KyleSanderson nice catch, seems good.

@dvander, I agree it's not the end all fix. I'd like to implement your suggestions but it's a bit beyond myself and my available time right now.

Issue #1166, opened 2020-01-29
ConVar handles leak


dragokas:
# Help us help you
  - [x] I have checked that my issue [doesn't exist yet](https://github.com/alliedmodders/sourcemod/issues).
  - [x] I have tried my absolute best to reduce the problem-space and have provided the absolute smallest test-case possible.
  - [x] I can always reproduce the issue with the provided description below.

# Environment
  * Operating System version: Debian 9 / Windows
  * Game/AppID (with version if applicable): 500, 550
  * Current SourceMod version: 1.10.6460
  * Current SourceMod snapshot: 1.10.6460
  * Current Metamod: Source snapshot:
  - [x] I have updated SourceMod to the [latest version](https://www.sourcemod.net/downloads.php) and it still happens.
  - [x] I have updated SourceMod to the [latest snapshot](https://www.sourcemod.net/downloads.php?branch=dev) and it still happens.
  - [x] I have updated SourceMM to the [latest snapshot](https://sourcemm.net/downloads.php?branch=dev) and it still happens.

# Problematic Code (or Steps to Reproduce)
  ```
	// Test
	ConVar t = CreateConVar("TTTTT", "0.1");
	PrintToServer("T0 %d %d", t, FindConVar("TTTTT")); // SAME
	PrintToServer("T1 %d", FindConVar("tTTTT")); // ALWAYS DIFFERENT
	PrintToServer("T2 %d", FindConVar("TtTTT")); // ALWAYS DIFFERENT
	PrintToServer("T3 %d", FindConVar("TTtTT")); // ALWAYS DIFFERENT

	// Real cvar registered as: Smoker_escape_range
	PrintToServer("A %d", FindConVar("smoker_escape_range")); // ALWAYS DIFFERENT
	PrintToServer("B %d", FindConVar("smoker_escape_range")); // ALWAYS DIFFERENT
	PrintToServer("C %d", FindConVar("smoker_escape_range")); // ALWAYS DIFFERENT
	PrintToServer("D %d", FindConVar("Smoker_escape_range")); // ALWAYS SAME
	PrintToServer("E %d", FindConVar("Smoker_escape_range")); // ALWAYS SAME
	PrintToServer("F %d", FindConVar("Smoker_escape_range")); // ALWAYS SAME
  ```

# Description
Most people write cvar names to server.cfg in all lower case. The same is true for 3rd party plugins. 
This causes ConVar leaks on each map start for some specific cvars. Example: 
[...]

PR #1256, opened 2020-05-01
Fix unnecessary ConCommand cache misses

wanted241:
Fixes a rare crash caused by GetCommandFlags() if there are some commands registered by plugins (Reg*Cmd()) which containing capitals. The same command ending up in the cache multiple times would explain the bug. (https://crash.limetech.org/stats/dc3d4e0530c48bb491d6b448c5a4d8a3)

### DIFF
--- a/core/concmd_cleaner.cpp
+++ b/core/concmd_cleaner.cpp
@@ -108,20 +108,20 @@ public:
 			listener = listener->next;
 		}
 
-        while (iter != tracked_bases.end())
-        {
-            if ((*iter)->pBase == pBase)
-            {
-                pInfo = (*iter);
-                iter = tracked_bases.erase(iter);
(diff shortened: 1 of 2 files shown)


wanted241:
This is the same problem as it was in https://github.com/alliedmodders/sourcemod/issues/1166, but related to ConCommand rather than ConVar.

GamerXCZ:
Any plans with this changes move to stable build?

asherkin:
It won't be backported to 1.10, but 1.11 is very usable and we're hard at work towards getting it promoted.

peace-maker:
You can always compile the plugins using deprecated syntax with an older compiler until they're updated. The .smx created by any older compiler will load and run fine on newer runtimes. The forum remembers which compiler version was used to create the plugin in the release post and uses the correct version when clicking "Get Plugin". I know this isn't optimal, but at least you don't HAVE to touch a plugin's code just to use it.

PR #1262, opened 2020-05-06
Fix timelimit not correct for Black Mesa

PlayBoy31:
Black Mesa is particular and use timelimit in seconds instead of minutes

### DIFF
--- a/core/TimerSys.cpp
+++ b/core/TimerSys.cpp
@@ -69,6 +69,13 @@ class DefaultMapTimer :
 	public IConVarChangeListener
 {
 public:
+
+	#if SOURCE_ENGINE == SE_BMS
+    static constexpr int kMapTimeScaleFactor = 60;
+    #else
+    static constexpr int kMapTimeScaleFactor = 1;
+    #endif
+	
 	DefaultMapTimer()
 	{
 		m_bInUse = false;
@@ -81,11 +88,7 @@ public:
 
 	int GetMapTimeLimit()
 	{
-		#if SOURCE_ENGINE == SE_BMS
-		return (mp_timelimit->GetInt() / 60);
-		#else
-		return mp_timelimit->GetInt();
-		#endif
+		return (mp_timelimit->GetInt() / kMapTimeScaleFactor);
 	}
 
 	void SetMapTimerStatus(bool enabled)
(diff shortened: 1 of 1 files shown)


core/TimerSys.cpp
```diff
@@ -107,7 +111,7 @@ class DefaultMapTimer :
 
 		extra_time /= 60;
 
-		mp_timelimit->SetValue(mp_timelimit->GetInt() + extra_time);
+		mp_timelimit->SetValue(GetMapTimeLimit() + extra_time);
```
peace-maker, line 114: So is it in seconds or minutes now? You're setting the convar to minutes here. Probably need `* 60`?

PlayBoy31: Hummm, in fact we must stay in seconds here and convert only extratime if mp_timelimit is in minutes

peace-maker:
`#if` preprocessor conditions usually aren't indented in this codebase. Looks good otherwise.

core/TimerSys.cpp
```diff
@@ -81,7 +81,11 @@ class DefaultMapTimer :
 
 	int GetMapTimeLimit()
 	{
+		#if SOURCE_ENGINE == SE_BMS
```
dvander, line 84: Instead, can we have a single declaration at the top:

```
    #if SOURCE_ENGINE == SE_BMS
    static constexpr int kMapTimeScaleFactor = 60;
    #else
    static constexpr int kMapTimeScaleFactor = 1;
    #endif  
```

Then this becomes:

    return mp_timelimit->GetInt() / kMapTimeScaleFactor;

core/TimerSys.cpp
```diff
@@ -105,7 +109,9 @@ class DefaultMapTimer :
 			return;
 		}
 
+		#if SOURCE_ENGINE != SE_BMS
 		extra_time /= 60;
+		#endif
```
dvander, line 114: And this becomes:

    extra_time /= (60 / kMapTimeScaleFactor);

PlayBoy31: Nice, its more generic, like it. I set a new proposal

core/TimerSys.cpp
PlayBoy31: approved

core/TimerSys.cpp
```diff
@@ -69,6 +69,13 @@ class DefaultMapTimer :
 	public IConVarChangeListener
 {
 public:
+
+	#if SOURCE_ENGINE == SE_BMS
```
dvander, line 73: Thanks, this looks much cleaner! As Peace-Maker noted, #if should be in the first column, preprocessor directives technically are not supposed to be indented. Though, this works if you'd like:

    #     if SOURCE_ENGINE == SE_BMS

PlayBoy31: Ho sorry forget that ! no problem i correct

dvander:
Looks great, thanks!

PR #1264, opened 2020-05-08
Increase custom flags possibilities to max

PlayBoy31:
This request comes from admins (and me too) that need more than 6 customs flags for their servers.

Generally its used for extra features. With this change we can exploit all flags possibilities.

### DIFF
--- a/configs/admin_levels.cfg
+++ b/configs/admin_levels.cfg
@@ -37,6 +37,11 @@ Levels
 		"custom4"		"r"
 		"custom5"		"s"
 		"custom6"		"t"
+		"custom7"		"u"
+		"custom8"		"v"
+		"custom9"		"w"
+		"custom10"		"x"
+		"custom11"		"y"
 		
 		/**
 		 * Root is a magic access flag that grants all permissions.
--- a/core/logic/AdminCache.cpp
+++ b/core/logic/AdminCache.cpp
@@ -73,11 +73,10 @@ public:
 		if (!Parse())
 		{
 			memcpy(g_FlagLetters, g_DefaultFlags, sizeof(AdminFlag) * 26);
-			for (unsigned int i=0; i<20; i++)
+			for (unsigned int i=0; i<26; i++)
 			{
 				g_FlagCharSet[i] = true;
 			}
-			g_FlagCharSet[25] = true;
 		}
 	}
 private:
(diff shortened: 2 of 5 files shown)


asherkin:
Admin groups let you allow/deny access to specific commands, which is an infinitely scalable system which should be used instead of flags for complex admin setups.

PlayBoy31:
Depends of needs. Sometimes its more interresting to allow a flag to a func instead of admin groups. This update is safe and exploit capacities of flags

ambaca:
CheckCommandAccess
https://sm.alliedmods.net/new-api/console/CheckCommandAccess

Not need recompile plugin everytime, not need rely on flag & bits, not need create cvars with adminflags (which admin can change in game and bypass)

CheckCommandAccess and Admin Groups are powerful admin system.

PlayBoy31:
> 
> 
> CheckCommandAccess
> https://sm.alliedmods.net/new-api/console/CheckCommandAccess
> 
> Not need recompile plugin everytime, not need rely on flag & bits, not need create cvars with adminflags (which admin can change in game and bypass)
> 
> CheckCommandAccess and Admin Groups are powerful admin system.

I don't think its related to the purpose of the pull request.

What i have done is only expose full possibility of what it has been designed for flags system. Its has no impact, safe, and permit to use the full potential of it. Please review and accept it

ErikMinekus:
Either way it's a duplicate of #1117.

PlayBoy31:
> Either way it's a duplicate of #1117, which hasn't been accepted for the same reason.

Didnt see it. It proove that people want this.

ErikMinekus:
Just because people want something doesn't mean it's the right thing to do.

SourceMod has had the same number of flags since 2008, so if the SM team wanted to add more, I'm sure they would have done it already. Instead they believe `CheckCommandAccess` is the way forward.

PlayBoy31:
Just to be clear. Its not an evolution with new implementation etc.

Its expose the full potential. Its not because you think its not your right way that all must accept this, i think we are in  democracy. Dont be mad, just open eyes of what is the need, what people want and how people want to use this.

Yes SM devs are pretty good indeed and its not a reason to always trust gods:) for example yesterday i fix an error they have done since 2017.

Code looks fine and was tested. Im surpised that there is so much debat for a little thing

dvander:
Indeed plugins should use ACLs for access that is not easily bucketed into flags. The custom stuff is just copy-pasted stuff from AMX Mod, which itself was basically copied from Admin-Mod. About 20 years ago.

dvander:
Put another way, the flag system was intended as a convenience for people porting AMX Mod setups from a time period where CS 1.6 to CS:S transitions were still a thing. And that system itself was based on a Admin-Mod where only a few kinds of commands we even allowed. Admin-Mod had strict policies on how gameplay could be altered, so it made sense that commands would bucket into only a few categories.

The "custom" stuff is intended for private plugins. It's not really maintainable for public ones, since it's impossible to know whether two totally different plugins fall into the same bucket.

So, take advantage of the ACL (CheckCommandAccess) system. If there's some way we can improve it or make it easier, let's do that instead.

PlayBoy31:
Ok lets do this

PR #1265, opened 2020-05-12
Throw error when invalid address passed to SDKCall raw ptr

FortyTwoFortyTwo:
`LoadFromAddress` and `StoreToAddress` throws error when `null` or below `VALID_MINIMUM_MEMORY_ADDRESS` is passed, but `SDKCall` with `this` as raw address does not have one. Throw error and prevent crash from doing so.


KyleSanderson:
Great idea, thanks for the contribution 😸

dragokas:
Some questions, related to "ThisPtr address cannot be null", if you don't mind:
https://forums.alliedmods.net/showthread.php?t=329123

PR #1272, opened 2020-05-17
Update AMTL; replace AutoPtr/UniquePtr with STL.

dvander:


core/ConsoleDetours.cpp
```diff
@@ -307,7 +307,7 @@ bool ConsoleDetours::AddListener(IPluginFunction *fun, const char *command)
 	}
 	else
 	{
-		ke::AutoPtr<char[]> str(UTIL_ToLowerCase(command));
+		std::unique_ptr<char[]> str(UTIL_ToLowerCase(command));
```
Headline, line 310: `std::string` would probably make more sense here, huh?

core/ConsoleDetours.cpp
```diff
@@ -329,7 +329,7 @@ bool ConsoleDetours::RemoveListener(IPluginFunction *fun, const char *command)
...
```
Headline, line 332: ditto

core/ConsoleDetours.cpp
dvander: Agreed, though I'm still working through what changes are permissible and haven't gotten to std::string yet, and I'd like to keep these as behavior-preserving as possible.

PR #1276, opened 2020-05-20
Replace AString with std::string.

dvander:


ramonberrutti:
@dvander can you explain why the changes?

asherkin:
The reasoning is on the AMTL PR: https://github.com/alliedmodders/amtl/pull/97

dvander:
FYI there will be a few more PRs like this. I'm trying to pare down AMTL so it's not duplicating STL. Many of our containers predate C++11 and have no value add these days. So it's not worth maintaining them. It's annoying for people to have to learn two different conventions.

PR #1280, opened 2020-05-28
Change bootstrap.pl and startbuild.pl to also take CXX.

Fyren:


KyleSanderson:
Impressive that we've had this issue for this long. This destroyed me on a couple distributions of Linux for sure when building from source. Good catch!

dvander:
This does not affect building from source, it changes buildbot parameters only.

PR #1283, opened 2020-05-31
Replace ke::Vector with std::vector.

dvander:


core/MenuStyle_Base.cpp
```diff
@@ -661,16 +661,16 @@ bool CBaseMenu::InsertItem(unsigned int position, const char *info, const ItemDr
...
 }
 
 bool CBaseMenu::RemoveItem(unsigned int position)
 {
-	if (position >= m_items.length())
+	if (position >= m_items.size())
 		return false;
 
-	m_items.remove(position);
+	ke::RemoveAt(&m_items, position);
```
Headline, line 673: Any reason we’re not doing

`m_items.erase(m_items.begin()+ position)`

core/logic/CDataPack.cpp
```diff
@@ -87,7 +87,7 @@ size_t CDataPack::CreateMemory(size_t size, void **addr)
 	val.type = CDataPackType::Raw;
 	val.pData.vval = new uint8_t[size + sizeof(size)];
 	reinterpret_cast<size_t *>(val.pData.vval)[0] = size;
-	elements.insert(position, val);
+	ke::InsertAt(&elements, position, val);
```
Headline, line 90: std::vector::insert?

core/logic/DatabaseConfBuilder.h
```diff
@@ -90,7 +90,7 @@ class ConfDbInfoList : public ke::Vector<ConfDbInfo *>
 		m_DefDriver = std::string(input);
 	}
 	void ReleaseMembers() {
-		for (size_t i = 0; i < this->length(); i++) {
+		for (size_t i = 0; i < this->size(); i++) {
```
Headline, line 93: range based for

dvander: To minimize friction (and time spent writing these patches), I tried to make these changes as syntax-preserving as I could. Improvements like range-based for loops can be done as needed. The argument for RemoveAt/InsertAt is weaker though so I'll revert those.

PR #1295, opened 2020-06-18
Add basic Insurgency support on Windows x64

peace-maker:
This allows SourceMod to load on x64 Insurgency. There are still a lot of variable truncation warnings that have to be dealt with.

Requires https://github.com/alliedmodders/sourcepawn/pull/499


peace-maker:
SourceMod has to be compiled against MM:S 1.11 on win64 due to required changes in a SourceHook header.
https://github.com/alliedmodders/metamod-source/blob/f17a958c6647f98b3761bd5f73162b3e69957420/core/sourcehook/sh_memfuncinfo.h#L128-L132

Otherwise vtable hooks won't work.

extensions/mysql/msvc15hack.c
```diff
@@ -1,79 +1,37 @@
...
+	
+	if (*assembly == 0x83 && *(assembly + 1) == 0xC0 && (*(assembly + 2) == 0x20 || *(assembly + 2) == 0x40))
 	{
-		return NULL;
-	}
-
-	{
-		unsigned char const * assembly = (unsigned char const *)(s.AddrReturn.Offset);
-
-		if (*assembly == 0x83 && *(assembly + 1) == 0xC0 && (*(assembly + 2) == 0x20 || *(assembly + 2) == 0x40))
```
KyleSanderson, line 58: holy crap this file is fiddly haha. impressed (and expected 😉) you made it though 😸

core/logic/ExtensionSys.cpp
```diff
@@ -79,7 +79,7 @@ CLocalExtension::CLocalExtension(const char *filename, bool bRequired)
 		g_pSM->BuildPath(Path_SM,
 			path,
 			PLATFORM_MAX_PATH,
-			"extensions/%s." PLATFORM_LIB_EXT,
+			"extensions/" PLATFORM_ARCH_FOLDER "%s." PLATFORM_LIB_EXT,
```
KyleSanderson, line 82: hmmm. This is my only aversion to taking this. Do we really want a magic folder in extensions where subarch extensions live? what do you think about extensions_win64/coolthing.dll vs extensions/win64/coolthing.dll ? If this is the convention elsewhere that's cool, but this feels like a big change that I may have forgotten.

peace-maker: `PLATFORM_ARCH_FOLDER` is just `x64` or empty on x86, not `win64` or `win32`.
https://github.com/alliedmodders/sourcemod/blob/ce1a4dcac0c42e3aee3de9af1a26b3823b273208/public/sm_platform.h#L123

So all 64bit extensions live in `extensions/x64/coolthing.dll/so/dylib` which is the same we already do for stuff in `sourcemod/bin/x64/sourcemod.logic.dll` and the game does for e.g. `bin/x64/engine.dll`. Seems like an oversight, because the package script already puts the files in `extensions/x64`, they just weren't loaded from there.

KyleSanderson: Good catch, yeah I saw the definitions and missed the actual name. @TheDS are you okay with this?

KyleSanderson:
Thanks @peace-maker 😄

PR #1299, opened 2020-06-26
Add string_t SetEntPropString support for ep1 (fixes #1287)

### LINKED ISSUE #1287: SetEntPropString - string_t support for source engine < 2007
# Help us help you
  - [ ] I have tried my absolute best to reduce the problem-space and have provided the absolute smallest test-case possible.
  - [ ] I can always reproduce the issue with the provided description below.

# Environment
  * Game/AppID (with version if applicable): Any game engine < SE_ORANGEBOX
  * Current SourceMod version: Latest 1.11
  * Current SourceMod snapshot: 6562
  * Current Metamod: Source snapshot: 
  - [ ] I have updated SourceMM to the [latest snapshot](https://sourcemm.net/downloads.php?branch=dev) and it still happens.

# Description

Tried to use SetEntPropString on a custom-made entity, and I've got an error telling me that string_t is not supported by this game.

I look at the source code of Fortress Forever at first, as this game runs on an older engine, and it seems that the struct exists so normally should be supported. However, after a bit more investitagion, by looking here https://cs.alliedmods.net/sourcemod/source/core/smn_entities.cpp#2384 , it seems that SetEntPropString will throw an error if engine < orange box. It would be nice to have this feature added in the latest SourceMod versions, both 1.10 stable and 1.11 dev .

psychonic:
Quick and dirty fix for #1287

### DIFF
--- a/core/HalfLife2.cpp
+++ b/core/HalfLife2.cpp
@@ -1367,8 +1367,36 @@ bool CHalfLife2::IsMapValid(const char *map)
 	return FindMap(map) != SMFindMapResult::NotFound;
 }
 
-// TODO: Add ep1 support for this. (No IServerTools available there)
-#if SOURCE_ENGINE >= SE_ORANGEBOX
+#if SOURCE_ENGINE == SE_EPISODEONE
+class VKeyValuesSS_Helper {};
+static bool VKeyValuesSS(CBaseEntity* pThisPtr, const char *pszKey, const char *pszValue, int offset)
+{
+	void** this_ptr = *reinterpret_cast<void***>(&pThisPtr);
+	void** vtable = *reinterpret_cast<void***>(pThisPtr);
+	void* vfunc = vtable[offset];
+
+	union
+	{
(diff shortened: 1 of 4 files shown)


DarthMan:
> Nice! The changes here look good, but I'm pretty sure this'll break the Dark Messiah build - which has gotten lost from the `SE_` range here (ep1 is 1, dm is 2, and ob is 3).
> 
> Guessing this code'll work fine on DM as well, so it probably just needs to be `SOURCE_ENGINE < SE_ORANGEBOX`?

I guess it will. Anyway, my priority now is just Fortress Forever. So it's ok, but sure, if it's gonna work fine for DM too, u could just change it to work for source engine < orangebox.

psychonic:
> Nice! The changes here look good, but I'm pretty sure this'll break the Dark Messiah build - which has gotten lost from the `SE_` range here (ep1 is 1, dm is 2, and ob is 3).
> 
> Guessing this code'll work fine on DM as well, so it probably just needs to be `SOURCE_ENGINE < SE_ORANGEBOX`?

Good catch. Thanks

PR #1306, opened 2020-07-09
Properly document admin_levels.cfg regarding rcon

eyal282:
Not enough server owners know what rcon does and "Remote console" is possibly the worst way to document root's brother.

### DIFF
--- a/configs/admin_levels.cfg
+++ b/configs/admin_levels.cfg
@@ -23,7 +23,19 @@ Levels
 		"chat"			"j"			//Special chat privileges
 		"vote"			"k"			//Voting
 		"password"		"l"			//Password the server
-		"rcon"			"m"			//Remote console
+		
+		/*
+		 * Rcon allows executing commands as the server itself.
+		 * Rcon administrators can get the rcon password,
+		 * amybody who has the rcon password can control the server as if he is a rcon admin,
+		 * and also do this without being online in the server
+		 * This should only be given to trusted administrators.  
+		 * Since the server never obeys immunity, rcon users can target anyone regardless of immunity,
+		 * however, they themselves are not automatically immune.
+		 * Example: "sm_rcon sm_ban BAILOPAN" will ban BAILOPAN even if he has a higher immunity than me.
+		 */
+		"rcon"			"m"			
+		
 		"cheats"		"n"			//Change sv_cheats and related commands
 		
 		/**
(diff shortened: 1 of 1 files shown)


configs/admin_levels.cfg
```diff
@@ -23,7 +23,19 @@ Levels
 		"chat"			"j"			//Special chat privileges
 		"vote"			"k"			//Voting
 		"password"		"l"			//Password the server
-		"rcon"			"m"			//Remote console
+		
+		/*
+		 * Rcon allows executing commands as the server itself.
+		 * Rcon administrators can get the rcon password,
+		 * anybody who has the rcon password can control the server as if he is a rcon admin,
```
Headline, line 30: This line contains “he”, we should use a non-gendered pronoun

Overall though I believe this paragraph is long and too explanatory. New server owners should just see a little warning indication that giving this flag may be dangerous. I think it’s best if we keep the comment inline as it was, but add a short warning among the lines of “Allows users to execute commands as the server, this can be a dangerous permission to grant”

eyal282: What is a non gendered pronoun?

asherkin: > What is a non gendered pronoun?

"they"

As was discussed on Discord, simply "Execute commands as the server." seems best here (that said I don't personally think the existing description is that lacking). The documentation on the Wiki is the place to expand on these if wanted.

DarkDeviL: > New server owners should just see a little warning indication that giving this flag may be dangerous.

If so, it should be given for anything that poses a potential risk.  Flag h (cvars) can be used to alter severe settings as well. 

So maybe move all these kind of flags into their own separate category, to signal "potential risk - only for trusted administrators"?

> I think it’s best if we keep the comment inline as it was,

That would most likely be overlooked by many, but on the other hand, if the readers won't pay enough attention to what's provided for them, there's only one to blame for that.

configs/admin_levels.cfg
```diff
@@ -23,7 +23,14 @@ Levels
 		"chat"			"j"			//Special chat privileges
 		"vote"			"k"			//Voting
 		"password"		"l"			//Password the server
-		"rcon"			"m"			//Remote console
+		
+		/*
+		 * Rcon allows executing commands as the server itself.
+		 * Admins wtih rcon can execute most commands restricted to their flags, even root.
```
DarkDeviL, line 29: "wtih" could eventually be changed to "with"

PR #1318, opened 2020-07-24
Possibility to only display current admin (self) in list for Custom Admin Menu 

PlayBoy31:
This update allow:
+ to filter only the admin who is in the menu to see only this name (nice when we want to offer features to admin but we dont want to allow him to apply to others players, only him self)
+ not a big list of players when not need

Nota: "type" has been updated to add the case "self" => wiki https://wiki.alliedmods.net/Custom_Admin_Menu_(SourceMod) must be updated.

### DIFF
--- a/plugins/adminmenu/dynamicmenu.sp
+++ b/plugins/adminmenu/dynamicmenu.sp
@@ -49,6 +49,7 @@ enum SubMenu_Type
 	SubMenu_Group,
 	SubMenu_GroupPlayer,
 	SubMenu_Player,
+	SubMenu_Self,
 	SubMenu_MapCycle,
 	SubMenu_List,
 	SubMenu_OnOff	
@@ -205,6 +206,10 @@ void BuildDynamicMenu()
 				{			
 					submenuInput.type = SubMenu_Player;
 				}
+				else if (StrContains(inputBuffer, "self") != -1)
+				{			
+					submenuInput.type = SubMenu_Self;
+				}
 				else if (StrEqual(inputBuffer,"onoff"))
 				{
 					submenuInput.type = SubMenu_OnOff;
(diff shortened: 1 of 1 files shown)


asherkin:
Showing a menu with a single option seems excessive for this - you should already be able to just target `@me` in your configuration as long as they can execute the command themselves.

```
"Kick Myself"
{
	"cmd"		"sm_kick @me"
	"admin"		"sm_kick"
	"execute"	"player"
}
```

PlayBoy31:
> 
> 
> Showing a menu with a single option seems excessive for this - you should already be able to just target `@me` in your configuration as long as they can execute the command themselves.
> 
> ```
> "Kick Myself"
> {
> 	"cmd"		"sm_kick @me"
> 	"admin"		"sm_kick"
> 	"execute"	"player"
> }
> ```

Tried and not working

here an example, i have a cmd: sm_my_test_cmd player value1 value2

> 		"Move boost"
> 		{
> 			"cmd"			"sm_my_test_cmd @me #2 #3"
> 			"execute"		"player"
> 			"2"
> 			{
> 				"type"		"list"
> 				"1"		"0.0"
> 				"1."		"Value x0.0"
> 				"2"		"0.5"
> 				"2."		"Value x0.5"
> 				"3"		"1.0"
> 				"3."		"Value x1.0"
> 			}
> 			"3"
> 			{
> 				"type"		"list"
> 				"1"		"0"
> 				"1."		"Action 1"
> 				"2"		"1"
> 				"2."		"Action 2"
> 			}
> 		}

PlayBoy31:
> 
> 
> Your parameters are misnumbered, they need to start from 1 (regardless of the actual command string.)

Ho ok, I understand the logic

This solution is good for me. Thanks for pointing this.

PlayBoy31:
I close the pull request or you want to integrate it?

asherkin:
I don't think we want to take this change in, glad we could get your problem solved though!

PR #1319, opened 2020-07-24
Correct SDKHooks_TakeDamage native

PerfectLaugh:
https://github.com/alliedmodders/hl2sdk/pull/77

### DIFF
--- a/extensions/sdkhooks/takedamageinfohack.h
+++ b/extensions/sdkhooks/takedamageinfohack.h
@@ -59,7 +59,12 @@ class CTakeDamageInfoHack : public CTakeDamageInfo
 {
 public:
 	CTakeDamageInfoHack( CBaseEntity *pInflictor, CBaseEntity *pAttacker, float flDamage, int bitsDamageType, CBaseEntity *pWeapon, Vector vecDamageForce, Vector vecDamagePosition );
+#if SOURCE_ENGINE == SE_CSGO
+	inline int GetAttacker() const { return m_CSGOAttacker.m_hHndl.IsValid() ? m_CSGOAttacker.m_hHndl.GetEntryIndex() : -1; }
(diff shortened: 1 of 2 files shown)


extensions/sdkhooks/takedamageinfohack.h
```diff
@@ -59,7 +59,12 @@ class CTakeDamageInfoHack : public CTakeDamageInfo
 {
 public:
 	CTakeDamageInfoHack( CBaseEntity *pInflictor, CBaseEntity *pAttacker, float flDamage, int bitsDamageType, CBaseEntity *pWeapon, Vector vecDamageForce, Vector vecDamagePosition );
+#if SOURCE_ENGINE == SE_CSGO
+	inline int GetAttacker() const { return m_CSGOAttacker.m_hHndl.IsValid() ? m_CSGOAttacker.m_hHndl.GetEntryIndex() : -1; }
+	void SetAttacker(CBaseEntity *pAttacker);
+#else
 	inline int GetAttacker() const { return m_hAttacker.IsValid() ? m_hAttacker.GetEntryIndex() : -1; }
```
asherkin, line 66: Just to clean up the implementation a bit, could we get an `inline void SetAttacker(CBaseEntity *pAttacker) { m_hAttacker = pAttacker;  }` in here, so `CTakeDamageInfoHack::CTakeDamageInfoHack` can just call `SetAttacker` with no game checks?

PerfectLaugh: I think the game is just simply ignoring the old m_hAttacker handle even with m_bNeedInit is true. So I think the answer is no.
image

PerfectLaugh: Ahh sorry I was confused by my shitty English ability.
Modified as requested

PR #1322, opened 2020-07-25
Fix velocities assignment of SDKHooks_TakeDamage

PerfectLaugh:
The bug make SDKHooks_TakeDamage's velocity arguments cannot be passed to constructed CTakeDamageInfoHack, causing no damage velocity generated.


KyleSanderson:
@psychonic what do you think about this one? I didn't go back to mercurial to look at how it happened but it looks like an innocent enough issue. My only concern is knockback plugins rely on these values, fixing them will obviously change behaviour. @ErikMinekus any thoughts on this one?

PR #1325, opened 2020-07-28
Updated 'checkout-deps.ps1' to check for 'setuptools' prior to instal…

WildCard65:
…ling 'AMBUild'.


WildCard65:
I meant to create a draft.

PR #1330, opened 2020-08-02
Add offset-reading capability for networked CUtlVector instances

nosoop:
This was initially discussed in the SourceMod Discord; PR provides a way to get the actual offset of networked class members of type `CUtlVector` such as `CBaseFlex::m_AnimOverlay` and `CTFPlayer::m_hMyWearables`.

Tested working in TF2 with the following snippet:
```
public void OnPluginStart() {
	int offs = FindSendPropInfo("CTFPlayer", "m_hMyWearables");
	PrintToServer("CTFPlayer::m_hMyWearables = %04x", offs);
	
	offs = FindSendPropInfo("CTFPlayer", "m_ConditionData");
	PrintToServer("CTFPlayer::m_ConditionData = %04x", offs);
}
```

Not pull-ready yet; need recommendations on code style and how to deal with the `CSendPropExtra_UtlVector` gutted from `public/dt_utlvector_send.cpp`.

Heuristic per following message (the `lengthproxy` also holds the extradata, so I opted to use the parent table instead):
> [4:19 PM] asherkin: telling that a prop is an encoded utlvector seems like the hardest part, and it looks like a simple heuristic would be that it's a table with an offset of 0, and the first child entry's name starts with `lengthprop`

### DIFF
--- a/core/HalfLife2.cpp
+++ b/core/HalfLife2.cpp
@@ -340,7 +340,7 @@ bool UTIL_FindInSendTable(SendTable *pTable,
 		if (pname && strcmp(name, pname) == 0)
 		{
 			// get true offset of CUtlVector
-			if (prop->GetOffset() == 0 && pInnerTable && pInnerTable->GetNumProps() && strstr(pInnerTable->GetProp(0)->GetName(), "lengthproxy") != nullptr)
+			if (prop->GetOffset() == 0 && pInnerTable && pInnerTable->GetNumProps() && strncmp(pInnerTable->GetProp(0)->GetName(), "lengthproxy", 11) == 0)
 			{
 				SendProp *pLengthProxy = pInnerTable->GetProp(0);
 				HACK_CSendPropExtra_UtlVector *pExtra = (HACK_CSendPropExtra_UtlVector*) pLengthProxy->GetExtraData();
--- a/core/HalfLife2.cpp
+++ b/core/HalfLife2.cpp
@@ -340,7 +340,7 @@ bool UTIL_FindInSendTable(SendTable *pTable,
 		if (pname && strcmp(name, pname) == 0)
 		{
 			// get true offset of CUtlVector
-			if (prop->GetOffset() == 0 && pInnerTable && pInnerTable->GetNumProps() && strncmp(pInnerTable->GetProp(0)->GetName(), "lengthproxy", 11) == 0)
+			if (prop->GetOffset() == 0 && pInnerTable && pInnerTable->GetNumProps() && strcmp(pInnerTable->GetProp(0)->GetName(), "lengthproxy") == 0)
 			{
 				SendProp *pLengthProxy = pInnerTable->GetProp(0);
 				HACK_CSendPropExtra_UtlVector *pExtra = (HACK_CSendPropExtra_UtlVector*) pLengthProxy->GetExtraData();
--- a/core/HalfLife2.cpp
+++ b/core/HalfLife2.cpp
@@ -313,6 +313,16 @@ IChangeInfoAccessor *CBaseEdict::GetChangeAccessor()
 }
 #endif
 
+struct HACK_CSendPropExtra_UtlVector
+{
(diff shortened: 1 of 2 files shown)


core/HalfLife2.cpp
```diff
@@ -326,15 +336,25 @@ bool UTIL_FindInSendTable(SendTable *pTable,
 	{
 		prop = pTable->GetProp(i);
 		pname = prop->GetName();
+		SendTable *pInnerTable = prop->GetDataTable();
 		if (pname && strcmp(name, pname) == 0)
 		{
+			// get true offset of CUtlVector
+			if (prop->GetOffset() == 0 && pInnerTable && pInnerTable->GetNumProps() && strncmp(pInnerTable->GetProp(0)->GetName(), "lengthproxy", 11) == 0)
```
asherkin, line 343:
```suggestion
			if (prop->GetOffset() == 0 && pInnerTable && pInnerTable->GetNumProps() && strcmp(pInnerTable->GetProp(0)->GetName(), "lengthproxy") == 0)
```

It's matching the full string, so probably just want `strcmp` here?

nosoop:
A disassembly of `SendPropUtlVector` suggests that `CSendPropExtra_UtlVector`'s size and assigned member orders match between CS:GO and TF2, so I'd imagine it's fine.

`FindSendPropInfo` built against this PR reports that `CBaseFlex::m_AnimOverlay` is at 0x4EC for CS:GO, which seems right since it shows up in `CBaseAnimatingOverlay::StudioFrameAdvance`, and TF2 reports the offset found in its own function.

I'll go ahead and set this up for review since it appears to clear both games.

psychonic:
Do we even need the whole struct? It looks like only one member is used. If there's any chance that this could different across engine variables, why not just put its offset in gamedata?

nosoop:
I don't have an opinion for / against putting it in gamedata, but I'd imagine we could also use `CSendPropExtra_UtlVector::m_ElementStride` in the future should there be a desire to access `CUtlVector` elements via natives; the other four members would remain unused.

No problems switching it over to gamedata if that's the consensus.

asherkin:
Using gamedata makes sense :+1:

core/HalfLife2.cpp
```diff
@@ -326,15 +336,25 @@ bool UTIL_FindInSendTable(SendTable *pTable,
 	{
 		prop = pTable->GetProp(i);
 		pname = prop->GetName();
+		SendTable *pInnerTable = prop->GetDataTable();
 		if (pname && strcmp(name, pname) == 0)
 		{
+			// get true offset of CUtlVector
+			if (prop->GetOffset() == 0 && pInnerTable && pInnerTable->GetNumProps() && strcmp(pInnerTable->GetProp(0)->GetName(), "lengthproxy") == 0)
```
KyleSanderson, line 343: Careful, the name needs to be checked here for validity.

asherkin: Checked how? The count checks that prop 0 exists, and a prop must have a name.

KyleSanderson: interesting 😄 why do we check if the name is valid directly above this on line 340? It's been in the tree since 2008 😉

nosoop:
Moved it over to offsets as requested.  I'm not sure where in `gamedata/core.games/` the offset should go in; at the moment I've tentatively put the following in `common.games` (currently uncommitted):
```
"SendPropExtraVecOffset"
{
	"windows"		"16"
	"linux"			"16"
}
```
Let me know if I should add it into common.games or all the engine ones.

asherkin:
`common.games` makes sense to me - even stuff that's just the same for a handful of engines lives in there currently. I'd probably be inclined to name it after the struct/member to make it more obvious when updating it later / adding the stride - just `CSendPropExtra_UtlVector::m_Offset`?

nosoop:
Done.  `Class::m_Member`'s my go-to for gamedata naming; just didn't want to stray from existing conventions here.

KyleSanderson:
Nice work on this one - didn't want to nit you on this one @nosoop so made a tiny commit and landed. Thanks for the contribution!

PR #1333, opened 2020-08-09
mapchooser: Set minimum number maps to include in vote

stickz:
This commit allows the user to have a consistent minimum number of maps in the vote, regardless how many previous maps are excluded. It also improves the handling of the situation, when the number of previous maps to exclude is set higher than the number of available maps. Instead of excluding no previous maps, it will exclude until the minimum include count is reached.

### DIFF
--- a/plugins/mapchooser.sp
+++ b/plugins/mapchooser.sp
@@ -62,7 +62,8 @@ ConVar g_Cvar_ExtendTimeStep;
 ConVar g_Cvar_ExtendRoundStep;
 ConVar g_Cvar_ExtendFragStep;
 ConVar g_Cvar_ExcludeMaps;
-ConVar g_Cvar_IncludeMaps;
+ConVar g_Cvar_IncludeMinMaps;
+ConVar g_Cvar_IncludeMaxMaps;
 ConVar g_Cvar_PersistentMaps;
 ConVar g_Cvar_NoVoteMode;
 ConVar g_Cvar_Extend;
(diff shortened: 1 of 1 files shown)


plugins/mapchooser.sp
```diff
@@ -993,16 +995,20 @@ void CreateNextVote()
...
-		for (int i = 0; i < g_OldMapList.Length; i++)
+		if (tempMaps.Length <= g_Cvar_IncludeMinMaps.IntValue)
 		{
-			g_OldMapList.GetString(i, map, sizeof(map));
-			RemoveStringFromArray(tempMaps, map);
+			// Exit if we hit the minimum option count
+			break;
 		}
+		
+		g_OldMapList.GetString(i, map, sizeof(map));
```
KyleSanderson, line 1007: This is not correct. There's no reason to triple scan the array.

KyleSanderson (changes requested):
The other quantity checks seem okay, but fundamentally this is dangerous and not correct. We should have indexes that need to be removed if needed and this provides the expected fast path for basic programming.

stickz:
There are 2 ArrayList classes being worked with here. I don't see a "fast path" to just remove an index.

`g_OldMapList` is an array ordered from least to most recently played. 
`tempMaps` is a working array from `g_MapList` which is a random order.

I think it's required to search for the string that needs to be removed in `tempMaps`.

The code gets the string at index i from `g_OldMapList`. Then gets the index of the string in `tempMaps`. Afterwards, it erases the index from `tempMaps` if the string was found in the ArrayList.

PR #1334, opened 2020-08-13
Add native IsLateLoadTime

c0rp3n:
Adding this to reduce the amount of code required in a plugin to perform an action if the plugin was loaded late as this can be a general problem for handling a plugin when it is reloaded.
This saves having to store the late load state your self from the `AskPluginLoad2` forward.

Currently:
```sourcepawn
bool g_bLateLoaded = false;

public APLRes AskPluginLoad2(Handle plugin, bool late, char[] error, int err_max)
{
    g_bLateLoaded = late;
}

public void OnPluginStart()
{
    if (g_bLateLoaded)
    {
        // Do stuff on late load
    }
}
```

After this addition:
```sourcepawn
public void OnPluginStart()
{
    if (IsLateLoadTime())
    {
        // Do stuff on late load
    }
}
```

### DIFF
--- a/core/logic/smn_core.cpp
+++ b/core/logic/smn_core.cpp
@@ -34,6 +34,7 @@
 #include <stdlib.h>
 #include "common_logic.h"
 #include "Logger.h"
+#include "PluginSys.h"
 
 #include <ISourceMod.h>
 #include <ITranslator.h>
@@ -302,6 +303,11 @@ static cell_t GetPluginFilename(IPluginContext *pContext, const cell_t *params)
 	pContext->StringToLocalUTF8(params[2], params[3], pPlugin->GetFilename(), NULL);
 
 	return 1;
+}
+
+static cell_t IsLateLoadTime(IPluginContext *pContext, const cell_t *params)
+{
+	return g_PluginSys.IsLateLoadTime() ? 1 : 0;
 }
 
 static cell_t IsPluginDebugging(IPluginContext *pContext, const cell_t *params)
@@ -956,6 +962,7 @@ REGISTER_NATIVES(coreNatives)
 	{"ReadPlugin",				ReadPlugin},
 	{"GetPluginStatus",			GetPluginStatus},
 	{"GetPluginFilename",		GetPluginFilename},
+	{"IsLateLoadTime",			IsLateLoadTime},
 	{"IsPluginDebugging",		IsPluginDebugging},
 	{"GetPluginInfo",			GetPluginInfo},
 	{"SetFailState",			SetFailState},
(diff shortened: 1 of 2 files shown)


Headline:
Can't you just check if `MaxClients` is non-zero?

c0rp3n:
I'm not sure (not something I'd have thought to try) MaxClients could be 0 if no one is on the server? but if not this could just be replaced with a stock but without the stock the code wouldnt be clear as to what you are doing, and this is just exposing a variable that is already accesible but just means you do not have to cache it.
```sourcepawn
stock bool IsLateLoadTime()
{
    return MaxClient > 0;
}
```

c0rp3n:
After re-reading the code I would say that checking if `MaxClients > 0` is only useful in the case that you want to loop say `OnClientPutInServer` but say if you want to do a native call or something it shant be correct all of the time.

psychonic:
What is the use case for needing to know that a plugin is loaded late?

It does seem that most often that is check just as an assumption that there are players, or that it is mid-game or mid-round. None of those would necessarily be true and could still be checked explicitly.

c0rp3n:
Yes the most common use case is just checking if there are clients on the server and if so calling say `OnClientPutInServer` and in this case it is more to make the intent of the code clearer.
Other uses for knowing if it is late load time are for registering with other plugins, though this can be achived by checking if that other plugin is ready if they have some form of ready native.

```sourcepawn
public void OnPluginStart()
{
    for (int i = 1; i < MaxClients; ++i) // There are clients so must be late load time
    {
         // Do stuff
    }

    if (Store_IsReady()) // In this case it acts as our IsLateLoad
    {
        // Register some items
    }
}
```

Just would make the code easier to interpret if this is all wrapped in `IsLateLoad` essentially is why I want it.
Not because it is required as it can be worked around.

KyleSanderson:
> Can't you just check if `MaxClients` is non-zero?

no man 🤣 this is not a stable way to check. I seem to recall psychonic almost fixed that behaviour one night many many many many moons ago.

@c0rp3n I appreciate the PR but I don't think we want this. We took a hard stance on natives like this in general a decade ago (use the forwards to validate your state as to avoid races), checking this every time removes the callback methodology we've embraced with the scripting language. It's kind of like having a native IsFileSystemReady and invoking that on every call, the better pattern is to stall until the filesystemready forward is called, and then hook from there.

I completely appreciate where you're coming from on this one, as a matter of fact stuff like this I would have loved, but from a pure throughput / sanity perspective relying on the active forwards / callbacks are the better path. I'm going to leave this open if another peer feels differently, but adding natives like this goes against the current patterns that have held true since the beginning of the project.

c0rp3n:
Yup, I understand, I don't need this for any of my own current code as we have ready forwards and natives to handle this.
It was more knowing the variable exists but is awkwardly exposed and wanting to make the code easier to understand for others as to why we loop clients in onpluginstart etc.

Thanks for the reply.

PR #1336, opened 2020-08-16
Update to AMBuild 2.2.

dvander:
This is a pretty big diff because SourceMod had lots of multi-arch
workarounds that can now go away. I've also changed 'x64' to 'x86_64' in
many places since this is how AMBuild normalizes it, and it's far too
late to pick the shorter string, so we might as well suck it up.

The --target-archs parameter has been replaced with --targets. It works
the same way.

The default behavior for SDK inclusion is now "present" instead of
"all" since this lowers the burden of storing many SDKs. Official builds
will still be made with --sdks=all.

Multi-arch builds on Windows now work. However, they will not work on our buildbot yet, we'll need to change how the environment is set up to make that work.

### DIFF
--- a/AMBuildScript
+++ b/AMBuildScript
@@ -1,5 +1,6 @@
 # vim: set sts=2 ts=8 sw=2 tw=99 et ft=python:
 import os, sys
+import traceback
 
 class SDK(object):
   def __init__(self, sdk, ext, aDef, name, platform, dir):
@@ -19,20 +20,20 @@ class SDK(object):
     else:
       self.platformSpec = platform
 
-  def shouldBuild(self, target, archs):
-    if target.platform not in self.platformSpec:
-      return False
-    if not len([i for i in self.platformSpec[target.platform] if i in archs]):
-      return False
-    return True
+  def shouldBuild(self, targets):
+    for cxx in targets:
+      if cxx.target.platform in self.platformSpec:
+        if cxx.target.arch in self.platformSpec[cxx.target.platform]:
+          return True
+    return False
 
 WinOnly = ['windows']
 WinLinux = ['windows', 'linux']
 WinLinuxMac = ['windows', 'linux', 'mac']
 CSGO = {
   'windows': ['x86'],
-  'linux': ['x86', 'x64'],
-  'mac': ['x64']
+  'linux': ['x86', 'x86_64'],
+  'mac': ['x86_64']
 }
 
 PossibleSDKs = {
@@ -80,34 +81,19 @@ def ResolveEnvPath(env, folder):
 def Normalize(path):
   return os.path.abspath(os.path.normpath(path))
   
-def SetArchFlags(compiler, arch, platform):
+def SetArchFlags(compiler):
   if compiler.behavior == 'gcc':
-    if arch == 'x86':
-      compiler.cflags += ['-m32']
-      compiler.linkflags += ['-m32']
-      if platform == 'mac':
-        compiler.linkflags += ['-arch', 'i386']
-    elif arch == 'x64':
-      compiler.cflags += ['-m64', '-fPIC']
-      compiler.linkflags += ['-m64']
-      if platform == 'mac':
-        compiler.linkflags += ['-arch', 'x86_64']
+    if compiler.target.arch == 'x86_64':
+      compiler.cflags += ['-fPIC']
   elif compiler.like('msvc'):
-    if builder.target.arch == 'x86':
-      compiler.linkflags += ['/MACHINE:X86']
-    elif builder.target.arch == 'x64':
-      compiler.linkflags += ['/MACHINE:X64']
-    if arch == 'x64':
+    if compiler.target.arch == 'x86_64':
       compiler.defines += ['WIN64']
-
-def AppendArchSuffix(binary, name, arch):
-  if arch == 'x64':
-    binary.localFolder = name + '.x64'
 
 class SMConfig(object):
   def __init__(self):
     self.sdks = {}
     self.binaries = []
+    self.spvm = []
     self.extensions = []
     self.generated_headers = None
     self.mms_root = None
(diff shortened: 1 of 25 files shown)


configure.py
```diff
@@ -12,31 +12,34 @@
...
-parser.options.add_option('--enable-debug', action='store_const', const='1', dest='debug',
+parser.options.add_argument('--enable-debug', action='store_const', const='1', dest='debug',
                        help='Enable debugging symbols')
-parser.options.add_option('--enable-optimize', action='store_const', const='1', dest='opt',
+parser.options.add_argument('--enable-optimize', action='store_const', const='1', dest='opt',
                        help='Enable optimization')
-parser.options.add_option('--no-mysql', action='store_false', default=True, dest='hasMySql',
+parser.options.add_argument('--no-mysql', action='store_false', default=True, dest='hasMySql',
                        help='Disable building MySQL extension')
-parser.options.add_option('-s', '--sdks', default='all', dest='sdks',
```
KyleSanderson, line 35: This was an old (well-debated even) bug from mercurial days where I wanted `present` to be in here like the change suggests, but for buildbot purposes @TheDS and @psychonic both wanted `all` to be in here. We went back and forth on this one, but `all` won out because this was the primary shipping mechanism. I don't know if we have a stance on this now nearly a decade later, but just wanted to bring this up as this was seen as a "regression" at the time.

core/sourcemod.cpp
```diff
@@ -182,10 +182,16 @@ bool SourceModBase::InitializeSourceMod(char *error, size_t maxlength, bool late
 	/* There will always be a path by this point, since it was force-set above. */
 	m_GotBasePath = true;
 
+#if defined PLATFORM_X86
```
KyleSanderson, line 185: good catch

configure.py
```diff
@@ -12,31 +12,34 @@
 		sys.stderr.write('http://www.alliedmods.net/ambuild\n')
 	sys.exit(1)
 
-def make_objdir_name(p):
-    return 'obj-' + util.Platform() + '-' + p.target_arch
+# Hack to show a decent upgrade message, which wasn't done until 2.2.
+ambuild_version = getattr(run, 'CURRENT_API', '2.1')
+if ambuild_version.startswith('2.1'):
```
KyleSanderson, line 17: This likely is not the right spot to bring this up, but is AMB2.2 compatible with AMB2.1 projects?

configure.py
```diff
@@ -12,31 +12,34 @@
...
+parser.options.add_argument('--enable-debug', action='store_const', const='1', dest='debug',
                        help='Enable debugging symbols')
-parser.options.add_option('--enable-optimize', action='store_const', const='1', dest='opt',
+parser.options.add_argument('--enable-optimize', action='store_const', const='1', dest='opt',
                        help='Enable optimization')
-parser.options.add_option('--no-mysql', action='store_false', default=True, dest='hasMySql',
+parser.options.add_argument('--no-mysql', action='store_false', default=True, dest='hasMySql',
                        help='Disable building MySQL extension')
-parser.options.add_option('-s', '--sdks', default='all', dest='sdks',
+parser.options.add_argument('-s', '--sdks', default='present', dest='sdks',
```
KyleSanderson, line 36: This was an old (well-debated even) bug from mercurial days where I wanted present to be in here like the change suggests, but for buildbot purposes @TheDS and @psychonic both wanted all to be in here. We went back and forth on this one, but all won out because this was the primary shipping mechanism. I don't know if we have a stance on this now nearly a decade later, but just wanted to bring this up as this was seen as a "regression" at the time.

configure.py
KyleSanderson: hit the wrong button - sorry for the double highlight folks.

configure.py
dvander: Installing AMBuild 2.2 gets you the complete API set (1.0, 2.0. 2.1, 2.2, etc). But projects using the 2.2 API will not configure on 2.1.

KyleSanderson: Got it. thank you.

configure.py
dvander: The best argument against it is that it can lead to surprising behavior: if you want to compile for CSGO, and you're missing CSGO, it'll just be missing from the build. That's pretty normal behavior as far as configuration scripts go. I don't have a strong opinion and am happy to flip this back if there's disagreement.

KyleSanderson: Yeah, I fought this battle and lost in like 2011. I think it needs to be flipped back.

configure.py
```diff
@@ -12,31 +12,34 @@
...
+parser.options.add_argument('--mysql-path', type=str, dest='mysql_path', default=None,
 		                   help='Path to MySQL 5')
-parser.options.add_option('--mysql64-path', type=str, dest='mysql64_path', default=None,
+parser.options.add_argument('--mysql64-path', type=str, dest='mysql64_path', default=None,
 		                   help='Path to 64-bit MySQL 5')
-parser.options.add_option('--mms-path', type=str, dest='mms_path', default=None,
+parser.options.add_argument('--mms-path', type=str, dest='mms_path', default=None,
                        help='Path to Metamod:Source')
-parser.options.add_option('--enable-debug', action='store_const', const='1', dest='debug',
+parser.options.add_argument('--enable-debug', action='store_const', const='1', dest='debug',
```
WildCard65, line 30: This option and "--enable-optimize" should both really be mutually exclusive (especially on MSVC), argparse does have a method to create mutually exclusive arguments.

dvander: They're not mutually exclusive, eg if you want assertions and debug info in an optimized build.

dvander:
The latest iteration will print a warning message for each SDK it didn't find. I'll check this in today as soon as the Windows builder is working again.

AMBuildScript
```diff
@@ -168,6 +180,9 @@ class SMConfig(object):
       raise Exception('No SDKs were found that build on {0}-{1}, nothing to do.'.format(
         builder.target.platform, builder.target.arch))
```
ErikMinekus, line 181: `builder.target` no longer exists.

AMBuildScript
```diff
@@ -191,32 +206,31 @@ class SMConfig(object):
...
           raise Exception('Could not find a path to 64-bit MySQL!')
-        self.mysql_root['x64'] = Normalize(self.mysql_root['x64'])
+        self.mysql_root['x86_64'] = Normalize(self.mysql_root['x86_64'])
 
   def configure(self):
     builder.AddConfigureFile('pushbuild.txt')
     
-    if not set(self.archs).issubset(['x86', 'x64']):
+    if not set(self.target_archs).issubset(['x86', 'x86_64']):
       raise Exception('Unknown target architecture: {0}'.format(builder.target.arch))
```
ErikMinekus, line 228: `builder.target` no longer exists.

PR #1338, opened 2020-08-23
Update SourcePawn.

dvander:
This removes calls to CreateFakeNative.


ambaca:
Could this update cause errors ?
https://forums.alliedmods.net/showthread.php?t=327705

Something to do native errors
[SM] Exception reported: Not enough space on the heap
[SM] Exception reported: Not enough space on the stack

PR #1342, opened 2020-08-26
Remove OnEntitySpawned C++ listener

asherkin:
This is causing crashes with existing extensions using entity listeners.

Currently no one is asking for a C++ hook here, if it is desired it
needs to be done with proper versioning of the listener interface, which
is going to be a little bit more complicated without an initial hard
break.

https://forums.alliedmods.net/showpost.php?p=2715337&postcount=789
https://discordapp.com/channels/335290997317697536/335290997317697536/748101258186850334

This is a partial revert of #1078.

### DIFF
--- a/extensions/sdkhooks/extension.cpp
+++ b/extensions/sdkhooks/extension.cpp
@@ -1831,18 +1831,11 @@ void SDKHooks::HandleEntityCreated(CBaseEntity *pEntity, int index, cell_t ref)
 
 void SDKHooks::HandleEntitySpawned(CBaseEntity *pEntity, int index, cell_t ref)
 {
-	if (g_pOnEntitySpawned->GetFunctionCount() || m_EntListeners.size())
+	if (g_pOnEntitySpawned->GetFunctionCount())
 	{
 		const char *pName = gamehelpers->GetEntityClassname(pEntity);
 		if (!pName)
 			pName = "";
-
-		// Send OnEntitySpawned to SM listeners
-		for (SourceHook::List<ISMEntityListener *>::iterator iter = m_EntListeners.begin(); iter != m_EntListeners.end(); iter++)
-		{
-			ISMEntityListener *pListener = (*iter);
-			pListener->OnEntitySpawned(pEntity, pName);
-		}
 
 		// Call OnEntitySpawned forward
 		if (g_pOnEntitySpawned->GetFunctionCount())
--- a/public/extensions/ISDKHooks.h
+++ b/public/extensions/ISDKHooks.h
@@ -36,7 +36,7 @@
 #include <IShareSys.h>
 
 #define SMINTERFACE_SDKHOOKS_NAME		"ISDKHooks"
-#define SMINTERFACE_SDKHOOKS_VERSION	2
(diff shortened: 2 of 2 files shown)


psychonic (approved):
😐

BotoX:
https://github.com/alliedmodders/sourcemod/pull/1078#issuecomment-656010295
>Other than breaking compatibility with compiled extensions I don't see anything wrong with having this in SourceMod.

Previous major SM releases broke compatibility with extensions before too.
Actually the entire PR was caused by me recompiling latest master of dhooks for SM 1.11, which had issues that the precompiled version on AM didn't have.

Why is the versioning so useless is the better question?

asherkin:
The default interface versioning policy assumes only backwards compatible changes will be made (like adding new functions to the end of interface vtables) so allows older versions to be considered compatible. It can be overridden for different behaviour like the hard break this change needed - but if anyone cares for it there are better backwards-compatible options so we don’t need a break at all.

KyleSanderson:
@asherkin so this was added to the end of the vtable for this iface(!). I've asked a number of times for a stack trace / call graph and it additionally looks like from the sources provided one was not given, ever 😭 .

I'm okay with removing this as you asked for the feature 😄 but I'd still like to know exactly why this was backed out rather than "it was crashing". This is absolute pants that one extension was crashing and we've done zero investigation into why. At this point I'm thinking the extension is simply broken because the version was bumped.

asherkin:
> this was added to the end of the vtable for this iface(!)

It’s not the versioned interface vtable - it’s the entity listener vtable it was added to, which is called in other extensions by SDKHooks. So taking DHooks for example, builds without this change only have 2 functions in this vtable, so calling into the new 3rd function (which happened whether ISDKHooks v1 or v2 was actually requested) was out of bounds.

Normally for these listener interfaces we’ll version them independently, which allows the implementor to check what version was compiled against by the consumer and only call functions that were present in that version (you can see this in the main extension interface, or in any of the new ones recently added to SourcePawn).

The non-breaking path forward for this is to rename the existing listener interface and add/remove functions to be e.g. appended with “V1”. Then to implement a new interface (with the existing name) that adds a version-returning function to the start of the vtable and the new listener to the end, and new corresponding add/remove functions to the end of the ISDKHooks interface (with a version bump). That will allow existing compiled extensions to continue working using the existing listener interface, and as they’re re-compiled be transparently updated to the new versioned one, which can have additional callbacks easily added safely in the future. Internally, SDKHooks will either need separate lists or a tagged union to keep track of things.

Alternatively we just re-add it as-is but add the missing logic to mark v2 as a breaking change (which can ride into 1.11), but I’m loathe to do that due to the user pain of the unavoidable 1.10 breaking interface changes that affected DHooks, when we have a painless path forward.

> I'd still like to know exactly why this was backed out rather than "it was crashing". This is absolute pants that one extension was crashing and we've done zero investigation into why.

Crashing is bad, investigation was done, reasoning for (probably temporarily) removing this rather than fixing is in the PR description.

The important part of the original PR was the forward for plugins, which was unaffected by this change - it wasn’t “backed out” in the traditional sense.

PR #1347, opened 2020-09-01
Add native GetOS

komashchenko:
This will help you find out which OS the plugin is running on, and will save plugins from crutches.
https://github.com/GAMMACASE/MomSurfFix/blob/4037f73f53566859edc9361f02b512a98f74ce4e/addons/sourcemod/gamedata/momsurffix2.games.txt#L68
https://github.com/qubka/Zombie-Plague/blob/7c55343164216a26879d4c78291e53259c33627e/gamedata/plugin.zombieplague.txt#L54
https://github.com/quasemago/CSGO_WeaponStickers/blob/41282cac31dda0fd455c0efb656e1150cd842b3f/addons/sourcemod/gamedata/csgo_weaponstickers.games.txt#L7

### DIFF
--- a/core/logic/smn_core.cpp
+++ b/core/logic/smn_core.cpp
@@ -946,6 +946,42 @@ static cell_t LogStackTrace(IPluginContext *pContext, const cell_t *params)
 	return 0;
 }
 
+enum OS
+{
+	OS_Unknown,
+	OS_Windows_x86,
+	OS_Windows_x86_64,
+	OS_Linux_x86,
+	OS_Linux_x86_64,
+	OS_Apple_x86,
+	OS_Apple_x86_64
+};
+
+static cell_t GetOS(IPluginContext *pContext, const cell_t *params)
+{
+#if defined PLATFORM_WINDOWS
+#ifdef PLATFORM_X86
+	return OS_Windows_x86;
+#else
+	return OS_Windows_x86_64;
+#endif
+#elif defined PLATFORM_LINUX
+#ifdef PLATFORM_X86
+	return OS_Linux_x86;
+#else
+	return OS_Linux_x86_64;
+#endif
+#elif defined PLATFORM_APPLE
+#ifdef PLATFORM_X86
+	return OS_Apple_x86;
+#else
+	return OS_Apple_x86_64;
+#endif
+#else
+	return OS_Unknown;
+#endif
+}
+
 REGISTER_NATIVES(coreNatives)
 {
 	{"ThrowError",				ThrowError},
@@ -977,6 +1013,7 @@ REGISTER_NATIVES(coreNatives)
 	{"IsNullVector",			IsNullVector},
 	{"IsNullString",			IsNullString},
 	{"LogStackTrace",           LogStackTrace},
+	{"GetOS",                   GetOS},
 	
 	{"FrameIterator.FrameIterator",				FrameIterator_Create},
 	{"FrameIterator.Next",						FrameIterator_Next},
--- a/plugins/include/sourcemod.inc
+++ b/plugins/include/sourcemod.inc
@@ -691,6 +691,27 @@ native any LoadFromAddress(Address addr, NumberType size);
  */
 native void StoreToAddress(Address addr, any data, NumberType size);
 
+/**
+ * OS statuses.
+ */
+enum OS
+{
+	OS_Unknown,
+	OS_Windows_x86,
+	OS_Windows_x86_64,
+	OS_Linux_x86,
+	OS_Linux_x86_64,
+	OS_Apple_x86,
+	OS_Apple_x86_64
+};
+
+/**
+ * Returns the operating system the plugin is running.
+ *
+ * @return              OS status.
+ */
+native OS GetOS();
+
 methodmap FrameIterator < Handle {
 	// Creates a stack frame iterator to build your own stack traces.
 	// @return              New handle to a FrameIterator.


Headline:
We're gonna need some more justification for taking this patch. The server OS should be no factor for plugins.

komashchenko:
@Headline Can you guarantee me that one and the same function in CS: GO when called via SDKCall will have the same number of parameters and the same type of call?

Wend4r:
@Headline, for example Valve in CS:GO on Windows build on O2 optimisation which forces some functions to inline and cut off the of arguments.

psychonic:
That just means that gamedata should have a way of handling the call differentiation, not that plugins should need to know what OS they’re running on.

GAMMACASE:
There's some functions that are missing on one platform and are available on another, which leads to having this OSType thing in gamedata as you need to know what OS is currently used to use correct methods. As an example I can provide my plugin, where on linux some game functions are missing and thus requiring me to recreate them, here https://github.com/GAMMACASE/ShowPlayerClips/blob/master/addons/sourcemod/scripting/showplayerclips.sp#L555 and here you can see same function but on windows, where it's just a plain sdkcall https://github.com/GAMMACASE/ShowPlayerClips/blob/master/addons/sourcemod/scripting/showplayerclips.sp#L468-L473

komashchenko:
@psychonic Let's say the entire SDKCall call is built through gamedata as in DHooks, but then how can I find out in the plugin how many parameters I need to transfer if there are 3 of them on Windows and 2 on Linux?

KyleSanderson:
> @psychonic Let's say the entire SDKCall call is built through gamedata as in DHooks, but then how can I find out in the plugin how many parameters I need to transfer if there are 3 of them on Windows and 2 on Linux?

Gamedata. Have a fields form this then switch based on if it's 2 or 3 params. 

Anyway, yeah, I totally understand the motivation but the idea is all plugins should be cross platform. If this is not the case lets work on improving those areas of the product.

GAMMACASE:
Won't this be an improvement tho? You still have an ability to get the current OS via gamedata, so this will not change much or add something absolutely new to the current system, it'll just make life easier to ones who is in need of this and are using the gamedata way, cuz there's nothing else they can do about it. And exactly this thing will make plugins cross platform as you would be able to use correct methods for all platforms, as otherwise you can't, as you can't really tell when to use what methods, if you don't take in account the gamedata way of determining the OS.

komashchenko:
@KyleSanderson How do you want to solve such a case using gamedata, I do not understand
L `SDKCall(hCall, a1, a2);`
W `SDKCall(hCall, a1, LoadFromAddress(view_as<Address>(a1 + 0x54), NumberType_Int32), a2);`

KyleSanderson:
` + 0x54` should beyond obviously be in gamedata @komashchenko - hard coding that has always been completely unacceptable. Additionally for the param issues, that should also be completely solvable in gamedata.

SlidyBat:
The issue here is the different number of params per platform, which I agree isn't currently solvable via gamedata (at least not in a very meaningful way). The actual SP code needs to change based on the platform, since you can't configure SDKCalls in gamedata.

Your suggested solution:
> Gamedata. Have a fields form this then switch based on if it's 2 or 3 params.

is just a rewording of the OS. Linux is 2, Windows is 3. Not sure how having it called "field" instead of "os" makes the code any better. The code would still have to be different per platform so you don't really get any of the advantages of gamedata.

Even if you could configure SDKCalls completely in gamedata, that doesn't address GAMMACASEs issue of having to either SDKCall or recreate function in SP depending on platform.

komashchenko:
By creating SDKCall, you are already questioning the fact that the code should be platform independent.
A function call on Linux can consist of two sub-functions, and on Windows, due to optimization, it will be just one function, some may not be available on one of the platforms at all.
For some reason, you yourself preferred to use asm inserts instead of adding support for non-standard calls with registers to bintools.

KyleSanderson:
> Your suggested solution:
> 
> > Gamedata. Have a fields form this then switch based on if it's 2 or 3 params.
> 
> is just a rewording of the OS. Linux is 2, Windows is 3. Not sure how having it called "fields" instead of "os" makes the code any better. The code would still have to be different per platform so you don't really get any of the advantages of gamedata.

No, it absolutely is not (but I understand why you may be confused).

The Windows path being "optimized" can be passed back to the default Linux option with this method when the Windows path is destroyed. This absolutely happens and is a great example with the raw implementation example that was provided above. Yes, I agree that LTO is causing damage here to folks. I fully agree this is incredibly painful and people looking to use library functions should have better.  However, when the library changes because of a single line of code changing and LTO takes another direction, you just won't have the same fallback opportunity when this signature changes and this is going to cause people to grab new binaries. Having a "OS" specifier here specifically causes damage for a previously trivial change.

>  you yourself preferred to use asm inserts instead of adding support for non-standard calls with registers to bintools

I haven't the foggiest what you're on about. I've never written any assembly outside of Zilog Z80 stuff (and some ARM stuff outside of AM) but that predated GA of Arduino in my ecosystem. Can we try to keep this one technical? I'm not trying to block you guys, just trying to keep the ecosystem healthy if we can. If there's an actual technically legitimate reason to land this I'm your advocate, but until that happens I see this doing far more harm than good.

SlidyBat:
> The Windows path being "optimized" can be passed back to the default Linux option with this method when the Windows path is destroyed.

Although this can happen, I've rarely (if ever) seen it. Even if it did, you now have dead/unnecessary code & gamedata for handling the LTO which you'd likely want to remove anyways. Not to mention functions that just inherently have different params/behavior due to `#ifdef` and not LTO.

LTO changes rarely have easy to predict/modify changes like signatures/offsets do, and almost always require code changes. So with the current gamedata functionality, I don't see it as a good fit. Seems silly to me to shoehorn it into gamedata.

The way I see it if you have some kind of switch/platform-dependent code in SourcePawn, then you've already lost the benefits of the gamedata. Mixing both platform-dependent code and gamedata just wastes time and makes things more confusing.

I also don't see the "harm" that adding this native would do. While it certainly can be misused, so can a lot of other things. Should trust that developers are competent enough to use it effectively.

asherkin:
The use of gamedata to discover the OS has clearly gained some legs (along with more low-level stuff in plugins in general) so I quite like the idea of making this easier, but there are definitely some tradeoffs that make it not clear cut.

The big one is definitely that this isn't something most plugins need to know, and plugins that do need to know it are almost certainly already using other gamedata - so adding an extra key in is flexible and basically free. I like that plugins can pick their own constants and what they care about in their own gamedata - and have the choice whether they want to be generic about the OS or specialise on each "thing" they care about.

As implemented I wouldn't want to take this as-is - particularly due to merging the concerts of OS and architecture, which is very much solving the problems of people who want this for low-level usage and isn't as useful for plugins wanting to e.g. run the correct command with System2. This design choice, along with the varied other comments on this PR, is why we implore people to open an issue for discussion before writing code.

I think it would be an improvement to SM to include natives for getting the server's OS and architecture (as separate enums) however. To bikeshed the naming in advance, I don't like "OS" in the native name as it isn't very searchable and isn't camel-case, perhaps just System or Platform (GetPlatformType, GetPlatformArch?). We can all agree that it'd be nice for gamedata to be flexible to handle this (and we definitely have the technical capability in IBinTools), but there is code that needs the OS for other reasons and we can accommodate both at the same time - hopefully without plugins becoming needlessly OS-specific, as there are many that are needlessly game-specific due to spurious checks.

PR #1349, opened 2020-09-04
Fix "Command Group" override type admin flags

Sikarii:
Fixes "Command Group" override type getting set wrong flags upon the creation of the command group override.

One way to test this broken behaviour is to:
1. Create an admin command with ADMFLAG_ROOT admin flags initially and assign it under a command group `MyTestGroup`.
    - `RegAdminCmd("sm_testcommand", Command_Test, ADMFLAG_ROOT, "Description", "MyTestGroup");`
2. Create a command group override within `admin_overrides.cfg` - `"@MyTestGroup" ""`.
3. Make sure you do not have ROOT permissions, boot up the server and observe that you cannot access "sm_testcommand".

### DIFF
--- a/core/ConCmdManager.cpp
+++ b/core/ConCmdManager.cpp
@@ -496,7 +496,7 @@ void ConCmdManager::UpdateAdminCmdFlags(const char *cmd, OverrideType type, Flag
 		for (PluginHookList::iterator iter = group->hooks.begin(); iter != group->hooks.end(); iter++)
 		{
 			CmdHook *hook = *iter;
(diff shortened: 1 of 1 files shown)


asherkin (approved):
Thanks for the fix - I love how this one survived the ages.

It must be almost 10 years since I also tried to setup command group overrides and just went "well that doesn't work" - so great job digging into it and fixing them!

ambaca:
I have tried to test "Command group" one whole day for nothing :D dam bug.

PR #1357, opened 2020-09-30
Kill CDataPack cache

Headline:
Killing CDataPack's cache has been something on my list for a while - I'm not exactly sure when this became so easy (probably once IDataPack died), but here we are.

Technically, before this patch, we'd hold on to datapacks forever and their memory was never freed, we'd just save them for reuse. I do not think we're concerned anymore with the cost of allocating and deallocating this memory as-needed, so I'm just removing some bloat here.

### DIFF
--- a/core/logic/CDataPack.cpp
+++ b/core/logic/CDataPack.cpp
@@ -46,25 +46,6 @@ CDataPack::~CDataPack()
 	Initialize();
 }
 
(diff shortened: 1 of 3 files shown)


KyleSanderson (approved):

KyleSanderson:
Indeed. CPUs over the past decade + have come a long way with caching. This was a great feature at the time, but garbage collecting like this now per type is not really necessary (and may actually increase fragmentation / misses).

PR #1359, opened 2020-09-30
NPOTB: Fix no SDK target exception

### LINKED ISSUE #1343: AMBuildScript still references builder.target
@dvander the following lines still reference `builder.target`, which no longer exists:

~~https://github.com/alliedmodders/sourcemod/blob/master/AMBuildScript#L181~~ fixed in #1373
https://github.com/alliedmodders/sourcemod/blob/master/AMBuildScript#L228

Headline:
Another small regression from #1336 

Fixes #1343

I wasn't sure exactly how to format this, or how to get the platform from inside of this context - so instead of displaying platform we just show what build targets couldn't be found. I figure ignoring the platform here is fine

### DIFF
--- a/AMBuildScript
+++ b/AMBuildScript
@@ -224,7 +224,7 @@ class SMConfig(object):
     builder.AddConfigureFile('pushbuild.txt')
     
     if not set(self.target_archs).issubset(['x86', 'x86_64']):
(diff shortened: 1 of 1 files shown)


KyleSanderson:
@dvander any thoughts on this one? It's a really small PR but I honestly don't understand this stuff anymore (at this time).

peace-maker (approved):

PR #1360, opened 2020-09-30
Fix sm_dump_netprops_xml

Scags:
Fixes some quirks with `sm_dump_netprops_xml`. I ran into some problems parsing this a few months back.

Adds an main tree.

Netprops can commonly use `"` in their names (`"healing_array"` in TF2 for example), but xml supports using both single and double quotes in attribute values, so now it opts to put in single quotes.

Edit:
I've provided a [sample netprop dump for TF2](https://hatebin.com/oofbfrcmcu) along with a [basic Python script](https://hatebin.com/oofbfrcmcu) so you can compare the difference between them if you'd like. Script will fail with current command functionality.

### DIFF
--- a/extensions/sdktools/vhelpers.cpp
+++ b/extensions/sdktools/vhelpers.cpp
@@ -412,12 +412,12 @@ void UTIL_DrawSendTable_XML(FILE *fp, SendTable *pTable, int space_count)
 	SendTable *pOtherTable;
 	SendProp *pProp;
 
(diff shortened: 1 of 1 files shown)


Headline (approved):
Looks good! Thanks!

I think it was understood that the xml being generated was bad, so changes to this area are very welcome. Hopefully this papers over any issues anyone was having with this xml output.

PR #1362, opened 2020-10-08
Improve logging on map/generic votes

PlayBoy31:
Logging action is useful for servers’ owners with large number of admins. Actually, map votes and generic votes are not very verbose on what have been done. This request tries to improve the logging of those votes by adding maps names and question with answers in the log.

### DIFF
--- a/plugins/basevotes.sp
+++ b/plugins/basevotes.sp
@@ -51,6 +51,8 @@ public Plugin myinfo =
 #define VOTE_NO "###no###"
 #define VOTE_YES "###yes###"
 
+#define GENERIC_COUNT 5
+
 Menu g_hVoteMenu = null;
 
 ConVar g_Cvar_Limits[3] = {null, ...};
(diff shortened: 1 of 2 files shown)


KyleSanderson:
Thank you for the PR!

I've had donation for admin for well over a decade, however logging results from votes from clients hasn't been on the list. I'm really conflicted on this one, but I won't close this one out for a month+ until my peers have had a chance to review it as well. I simply can't see a case where I'd care what little Jimmy voted about what big Playboy31 started the vote about. I'd perhaps care about about what privileged commands Playboy31 is running (readas: the vote, and others), but then I'd add a CommandListener and see it all as part of the standard API.

If visibility is what you're after I can try to help you, but for now I don't see this helping core SM.

plugins/basevotes.sp
```diff
@@ -182,12 +184,14 @@ public Action Command_Vote(int client, int args)
...
 	GetCmdArgString(text, sizeof(text));
 
-	char answers[5][64];
+	char answers[GENERIC_COUNT][64];
 	int answerCount;	
 	int len = BreakString(text, g_voteArg, sizeof(g_voteArg));
 	int pos = len;
 	
-	while (args > 1 && pos != -1 && answerCount < 5)
+	char answers_list[GENERIC_COUNT * 64] = "";
```
peace-maker, line 192: The buffer size is too small if all answers actually use the whole 64 characters since you're adding quotes and spaces ` "%s"`. Same for the map vote even though a truncation is much more unlikely there.

The ` = ""` initialization is unnecessary btw.

peace-maker (changes requested):
This doesn't add logging for every single client's answer on a vote, but enhances the log to include the actual topic of the vote. So instead of only logging "PlayBoy31 initiated a generic vote." it logs the question and the pool of answers to choose from as well like "PlayBoy31 initiated a generic vote (question: XXX, answers: YY ZZ)".

Thank you, looks good. One thing inline.

plugins/basevotes.sp
PlayBoy31: Yes thanks for the feedback, should be ok now

peace-maker (approved):

Headline (approved):

KyleSanderson:
Ah! My bad, thanks for that. Not sure what I was thinking

PlayBoy31:
Thanks for the review all and for the approval

PR #1364, opened 2020-10-11
Improvements on !admin menu flags

menesakdeniz:
Lets say we have override the sm_unmute command and changed it to ADMFLAG_CUSTOM1.
Then create an admin, we gived our admin ADMFLAG_Chat flag, admin can't use sm_unmute command cause it doesnt have access to this command.
But if admin go into "!admin" menu then, he will able to run sm_unmute on "player command" menus

this vulnerability also affects mute,gag and silence commands

### DIFF
--- a/plugins/basecomm/gag.sp
+++ b/plugins/basecomm/gag.sp
@@ -53,39 +53,57 @@ void DisplayGagTypesMenu(int client)
 
 	if (!playerstate[target].isMuted)
 	{
-		AddTranslatedMenuItem(menu, "0", "Mute Player", client, (CheckCommandAccess(client, "sm_mute", ADMFLAG_CHAT, false) ? ITEMDRAW_DEFAULT : ITEMDRAW_DISABLED));
+		if(CheckCommandAccess(client, "sm_mute", ADMFLAG_CHAT, false))
+		{
+			AddTranslatedMenuItem(menu, "0", "Mute Player", client);
+		}
 	}
 	else
 	{
-		AddTranslatedMenuItem(menu, "1", "UnMute Player", client, (CheckCommandAccess(client, "sm_unmute", ADMFLAG_CHAT, false) ? ITEMDRAW_DEFAULT : ITEMDRAW_DISABLED));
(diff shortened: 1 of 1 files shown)


peace-maker:
Nice catch. I'm not sure if showing the disabled options is useful though. You should just not add the menu option at all if the user doesn't have access to it.

menesakdeniz:
i made this way because if the admin wants to see if user gagged or not. It can be changed easily

peace-maker:
Seeing who is muted is part of the permission in my opinion. If you can't use sm_mute you shouldn't see anything about it anywhere.

menesakdeniz:
Well, I changed it according to your opinion

plugins/basecomm/gag.sp
```diff
@@ -85,7 +103,7 @@ void AddTranslatedMenuItem(Menu menu, const char[] opt, const char[] phrase, int
 {
 	char buffer[128];
 	Format(buffer, sizeof(buffer), "%T", phrase, client);
-	menu.AddItem(opt, buffer);
+	menu.AddItem(opt, buffer, ITEMDRAW_DEFAULT);
```
peace-maker, line 106: That change isn't necessary anymore

plugins/basecomm/gag.sp
```diff
@@ -156,6 +174,7 @@ public int MenuHandler_GagPlayer(Menu menu, MenuAction action, int param1, int p
 	}
 }
 
+
```
peace-maker, line 177: Spurious whitespace changes

peace-maker (changes requested):
Thank you, this looks good except for the nits! 👍

plugins/basecomm/gag.sp
menesakdeniz: this line is deleted

plugins/basecomm/gag.sp
menesakdeniz: deleted unnecessary argument

peace-maker:
Thank you!

psychonic:
Was this already fixed in master? (I don’t have a PC handy but saw merged to 1.10-dev)

PR #1368, opened 2020-10-16
Fix spcomp64 filename for Windows

ErikMinekus:
Currently the 64-bit compiler for Windows is named `spcomp.exe64`, instead of `spcomp64.exe`.


peace-maker (approved):

peace-maker:
Thanks!

PR #1369, opened 2020-10-18
[ZPS] Gamedata update for sdktools/sdkhooks

Silenci0:
Hello,

ZPS has recently been updated to version 3.1 which brought about changes to the gamedata files. Here is what was updated:

- Updated offsets for sdkools/sdkhooks (Windows and Linux)
- Updated "FindEntityByClassname" windows signature for sdktools.

I ran a few tests with several plugins to make sure of the changes and, so far, everything seems to be working. If there are any questions/concerns, please let me know.

### DIFF
--- a/gamedata/sdkhooks.games/game.zpanic.txt
+++ b/gamedata/sdkhooks.games/game.zpanic.txt
@@ -1,24 +1,23 @@
 "Games"
 {
-	/* Zombic Panic Source 3.0.7 (Hotfix-3) */
 	"zps"
 	{
 		"Offsets"
 		{
 			"EndTouch"
 			{
-				"windows"	"113"
-				"linux"		"114"
+				"windows"	"110"
+				"linux"		"111"
 			}
 			"FireBullets"
 			{
-				"windows"	"125"
-				"linux"		"126"
+				"windows"	"123"
+				"linux"		"125"
 			}
 			"OnTakeDamage"
 			{
-				"windows"	"74"
-				"linux"		"75"
+				"windows"	"72"
+				"linux"		"73"
 			}
 			"PreThink"
 			{
@@ -32,23 +31,23 @@
 			}
 			"SetTransmit"
 			{
-				"windows"	"26"
-				"linux"		"27"
-			}
-			"ShouldCollide"
-			{
-				"windows"	"22"
-				"linux"		"23"
-			}
-			"Spawn"
-			{
 				"windows"	"28"
 				"linux"		"29"
 			}
+			"ShouldCollide"
+			{
+				"windows"	"16"
+				"linux"		"17"
+			}
+			"Spawn"
+			{
+				"windows"	"30"
+				"linux"		"31"
+			}
 			"StartTouch"
 			{
-				"windows"	"111"
-				"linux"		"112"
+				"windows"	"108"
+				"linux"		"109"
 			}
 			"Think"
 			{
@@ -57,8 +56,8 @@
 			}
 			"Touch"
 			{
-				"windows"	"112"
-				"linux"		"113"
+				"windows"	"109"
+				"linux"		"110"
 			}
 			"TraceAttack"
 			{
@@ -67,13 +66,13 @@
 			}
 			"Use"
 			{
-				"windows"	"110"
-				"linux"		"111"
+				"windows"	"107"
+				"linux"		"108"
 			}
 			"VPhysicsUpdate"
 			{
-				"windows"	"170"
-				"linux"		"171"
+				"windows"	"171"
+				"linux"		"172"
 			}
 			"Weapon_CanSwitchTo"
 			{
@@ -82,23 +81,23 @@
 			}
 			"Weapon_CanUse"
 			{
-				"windows"	"274"
-				"linux"		"275"
-			}
-			"Weapon_Drop"
-			{
-				"windows"	"277"
-				"linux"		"278"
-			}
-			"Weapon_Equip"
-			{
 				"windows"	"275"
 				"linux"		"276"
 			}
-			"Weapon_Switch"
+			"Weapon_Drop"
 			{
 				"windows"	"278"
 				"linux"		"279"
+			}
+			"Weapon_Equip"
+			{
+				"windows"	"276"
+				"linux"		"277"
+			}
+			"Weapon_Switch"
+			{
+				"windows"	"279"
+				"linux"		"280"
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


gamedata/sdkhooks.games/game.zpanic.txt
```diff
@@ -1,24 +1,23 @@
...
-				"linux"		"114"
+				"windows"	"110"
+				"linux"		"111"
 			}
 			"FireBullets"
 			{
-				"windows"	"125"
-				"linux"		"126"
+				"windows"	"123"
+				"linux"		"125"
```
psychonic, line 15: Can you double-check this one? It would be odd for those to be apart by two for just this offset.

gamedata/sdktools.games/game.zpanic.txt
```diff
@@ -43,56 +43,56 @@
...
 		}
 		"Signatures"
 		{
 			"FindEntityByClassname"
 			{
 				"library"	"server"
-				"windows"	"\x55\x8B\xEC\x53\x56\x57\x8B\xF9\x8B\x4D\x2A\x85\xC9\x74\x2A\x8B\x01\xFF\x2A\x2A\x8B\x2A\x81\xE6\x2A\x2A\x2A\x2A\x46\x03\xF6\x8B\x34\x2A\xEB\x2A\x8B\x2A\x2A\x2A\x2A\x2A\x85\xF6\x74\x2A\x8B\x5D\x2A\x8B\x3E\x85\xFF\x75\x2A\x68\x2A\x2A\x2A\x2A\xFF\x2A\x2A\x2A\x2A\x83\xC4\x2A\xEB\x2A\x39"
+				"windows"	"\x55\x8B\xEC\x53\x56\x57\x8B\xF9\x8B\x4D\x2A\x85\xC9\x74\x2A\x8B\x01\xFF\x2A\x2A\x8B\x2A\x81\xE6\x2A\x2A\x2A\x2A\x46\x03\xF6\x8B\x34\x2A\xEB\x2A\x8B\x2A\x2A\x2A\x2A\x2A\x85\xF6\x74\x2A\x8B\x5D\x2A\x8B\x3E\x85\xFF\x75\x2A\x68\x2A\x2A\x2A\x2A\xFF\x2A\x2A\x2A\x2A\x2A\x83\xC4\x2A\xEB\x2A\x39"
 				"linux"		"@_ZN17CGlobalEntityList21FindEntityByClassnameEP11CBaseEntityPKc"
 			}
```
psychonic, line 97: FindEntityByClassname can probably be removed from the file altogether. I believe that ZPS likely supports the latest IServerTool interface version.

Same for the CBaseTempEntity signature and the s_pTempEntities signature and offset.

Silenci0:
Sorry for the late reply, but thank you for letting me know about those changes. 

I double-checked the offsets for FireBullets and there were two functions:

- FireBullets(FireBulletsInfo_t const&) Offsets: 124 L, 124 W
- FireBullets(FireBulletsInfo_t const&, int&) Offsets: 125 L, 123 W 

I'm not entirely sure why there are two of them now (both come from CBaseEntity), but after comparing the functions to older server binaries for ZPS, I believe the function without int& is the one that should be correct. I have updated the offsets accordingly.

As for the signatures, since ZPS now runs on Source 2013 now, it does use the IServerTool interface. I removed the signatures/offsets and, just to be sure, I did a few tests without them and it looks like its working correctly. On a side note, I'm very glad that there are less signatures to look for now.

Everything should be updated now, but if there is anything else, let me know.

psychonic:
> I double-checked the offsets for FireBullets and there were two functions:
> 
>     * FireBullets(FireBulletsInfo_t const&) Offsets: 124 L, 124 W
> 
>     * FireBullets(FireBulletsInfo_t const&, int&) Offsets: 125 L, 123 W

Ah, that makes sense. With multiple function overloads in the same vtable, MSVC reverses the order. (CommitSuicide has the same effect).

> I believe the function without int& is the one that should be correct. I have updated the offsets accordingly.

I believe you're correct.

Thanks for updating these!

psychonic (approved):

JoelTroch:
> I double-checked the offsets for FireBullets and there were two functions:
> 
>     * FireBullets(FireBulletsInfo_t const&) Offsets: 124 L, 124 W
> 
>     * FireBullets(FireBulletsInfo_t const&, int&) Offsets: 125 L, 123 W
> 
> 
> I'm not entirely sure why there are two of them now (both come from CBaseEntity), but after comparing the functions to older server binaries for ZPS, I believe the function without int& is the one that should be correct. I have updated the offsets accordingly.

The extra `int &` is the client index to ignore when tracing the hit, there are 2 scenarios for this:

- It allows a Survivor to shoot a Zombie through another Survivor.
- Melee weapons now behave the same way as L4D2 when it comes to swinging them (the so called "melee arcs"). It prevents the victim from getting hit more than twice from the same swing.

`CBaseEntity::FireBullets( FireBulletsInfo_t const & )` is basically a redirection to `CBaseEntity::FireBullets( FireBulletsInfo_t const &, int & )` with the same value for `FireBulletsInfo_t` and a value of "0" for the `int &`. We kept the "SDK 2013 MP version" to limit breaking the existing stuff as much as we can.

For plugins authors that wish to call (not capture) `FireBullets`: I strongly suggest you use the `CZP_Player` class instead because this one does handle the lag compensation stuff, `CBaseEntity` does not.

Silenci0:
Thank you @JoelTroch  , that is very good to know. 

I also noticed CZP_Player when I looked into offsets. Going forward, would it be right to assume that this class should be relied on more for other functions (besides FireBullets) relating to player entities than CHL2MP_Player/CBaseEntity?

JoelTroch:
> 
> 
> Thank you @JoelTroch , that is very good to know.
> 
> I also noticed CZP_Player when I looked into offsets. Going forward, would it be right to assume that this class should be relied on more for other functions (besides FireBullets) relating to player entities than CHL2MP_Player/CBaseEntity?

`CZP_Player` is the player class since ZPS v3.1. `CHL2MP_Player` is no longer present and thus should be avoided.

asherkin:
> For plugins authors that wish to call (not capture) `FireBullets`: I strongly suggest you use the `CZP_Player` class instead because this one does handle the lag compensation stuff, `CBaseEntity` does not.

> I also noticed CZP_Player when I looked into offsets. Going forward, would it be right to assume that this class should be relied on more for other functions (besides FireBullets) relating to player entities than CHL2MP_Player/CBaseEntity?

Thanks to the magic of vtables, when you're hooking/calling things on specific entities you're always getting the class that entity is implemented as in the game. Nothing special needs doing to achieve this with virtual hooks / calls.

PR #1371, opened 2020-10-22
Updated OnConfigsExecuted() note

87andrewh:


peace-maker:
What's the reason behind this change? `OnConfigsExecuted` isn't the only callback that's fired on late load.

87andrewh:
Oh, I suppose there's a conflict between clarity and consistency. Should I modify all the callbacks that are fired on a late load?

KyleSanderson:
> Oh, I suppose there's a conflict between clarity and consistency. Should I modify all the callbacks that are fired on a late load?
> 
> The benefit is quite simple. People won't have to ask or wonder whether it's fired on a late load.

This is a tough one. While we should maybe have documentation saying that things are fired regardless of plugin load time, if they aren't we should file them as bugs and fix the actual issue. AskPluginLoad(2), OnPluginStart, OnMapStart, OnConfigsExecuted, etc are all called as part of the load process. But if clients are already connecting, we don't call OnClientConnect+ because they're not actively connecting. For a late plugin though, the level and similar are loading. At this point in time it's a losing proposition to document every case, but level loading as a rule of thumb is covered with late loads which is well discussed (and honestly, expected).

PR #1374, opened 2020-10-24
Fixed unauthorized menu usages

menesakdeniz:
Let's say we have kick flag but don't have admin menu flag.
When we type !kick, menu opens with a back button, when we click back button, we go parent menu without admin menu flag


peace-maker (approved):
Thanks! This is a pretty special case of allowing access to sub commands while not giving access to the overarching main menu, but this change might still help some server operators.

Headline (approved):
That’s for the contribution! Looks great.

PR #1377, opened 2020-10-28
Cleanly remove all hooks on extension unload

peace-maker:
Allows to e.g. reload sdktools during runtime without crashing. Useful for fast development cycles.
Tiny steps towards getting `ForceRestartAfterUpdate` to work reliably.

https://github.com/alliedmodders/sourcemod/blob/master/extensions/updater/extension.cpp#L209-L220

### DIFF
--- a/extensions/sdktools/hooks.cpp
+++ b/extensions/sdktools/hooks.cpp
@@ -124,6 +124,58 @@ void CHookManager::Initialize()
 
 void CHookManager::Shutdown()
 {
+	if (basefilesystemPatch)
+	{
+		SH_RELEASE_CALLCLASS(basefilesystemPatch);
+		basefilesystemPatch = NULL;
+	}
+
+	if (PRCH_used)
+	{
+		for (size_t i = 0; i < m_runUserCmdHooks.size(); ++i)
+		{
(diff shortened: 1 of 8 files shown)


extensions/sdktools/hooks.cpp
```diff
@@ -124,6 +124,58 @@ void CHookManager::Initialize()
 
 void CHookManager::Shutdown()
 {
+	if (basefilesystemPatch)
+	{
+		SH_RELEASE_CALLCLASS(basefilesystemPatch);
+		basefilesystemPatch = NULL;
+	}
+
+	if (PRCH_used)
```
KyleSanderson, line 133: While technically correct - do we need these branches?

KyleSanderson (approved):
Rest LGTM.

extensions/sdktools/hooks.cpp
peace-maker: Probably not, I blindly copy & pasted the cleanup from `OnPluginUnloaded` :D

peace-maker: I'll leave this as is, since it's cleaner to only call those functions if it's necassary. We have those guard variables, so why not use them.

PR #1379, opened 2020-10-30
Update cookie funcs to return Cookie methodmap

JoinedSenses:
Just like the behavior of functions like CreateArray which return ArrayList, cookie functions should return Cookie.

### DIFF
--- a/plugins/include/clientprefs.inc
+++ b/plugins/include/clientprefs.inc
@@ -192,7 +192,7 @@ native Cookie RegClientCookie(const char[] name, const char[] description, Cooki
  * no longer needed.
  *
  * @param name          Name of cookie to find.
- * @return              A handle to the cookie if it is found. INVALID_HANDLE otherwise.
+ * @return              A handle to the cookie if it is found. Null otherwise.
  */
 native Cookie FindClientCookie(const char[] name);
 
--- a/plugins/include/clientprefs.inc
+++ b/plugins/include/clientprefs.inc
@@ -119,7 +119,7 @@ methodmap Cookie < Handle {
 	// no longer needed.
 	//
 	// @param name          Name of cookie to find.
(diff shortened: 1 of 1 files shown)


plugins/include/clientprefs.inc
```diff
@@ -119,7 +119,7 @@ methodmap Cookie < Handle {
 	// no longer needed.
 	//
 	// @param name          Name of cookie to find.
-	// @return              A handle to the cookie if it is found. INVALID_HANDLE otherwise.
+	// @return              A handle to the cookie if it is found. Null otherwise.
```
peace-maker, line 122:
```suggestion
	// @return              A handle to the cookie if it is found, null otherwise.

```

plugins/include/clientprefs.inc
```diff
@@ -192,9 +192,9 @@ native Handle RegClientCookie(const char[] name, const char[] description, Cooki
  * no longer needed.
  *
  * @param name          Name of cookie to find.
- * @return              A handle to the cookie if it is found. INVALID_HANDLE otherwise.
+ * @return              A handle to the cookie if it is found. Null otherwise.
```
peace-maker, line 195:
```suggestion
 * @return              A handle to the cookie if it is found, null otherwise.

```

peace-maker (changes requested):
Since the keyword is lowercase, documenting it as `Null` might confuse users when they copy it from the docs? `INVALID_HANDLE` had unambiguous casing before 😄

JoinedSenses:
> Since the keyword is lowercase, documenting it as `Null` might confuse users when they copy it from the docs? `INVALID_HANDLE` had unambiguous casing before 😄

¯\_(ツ)_/¯

peace-maker (approved):
Thanks!

Headline (approved):
Looks great, thanks!

PR #1384, opened 2020-11-26
Execute ConVarQueryFinished if client disconnects

Deathreus:
So people passing handles to the extra data can close them, adds a new return value to check if this happened

### DIFF
--- a/core/ConVarManager.cpp
+++ b/core/ConVarManager.cpp
@@ -236,19 +236,18 @@ void ConVarManager::OnPluginUnloaded(IPlugin *plugin)
 
 void ConVarManager::OnClientDisconnected(int client)
 {
-	IPluginFunction *pCallback = NULL;
 	/* Remove convar queries for this client that haven't returned results yet */
 	for (List<ConVarQuery>::iterator iter = m_ConVarQueries.begin(); iter != m_ConVarQueries.end();)
 	{
 		ConVarQuery &query = (*iter);
 		if (query.client == client)
 		{
-			pCallback = query.pCallback;
+			IPluginFunction *pCallback = query.pCallback;
 			if (pCallback)
 			{
 				cell_t ret;
 
-				pCallback->PushCell(InvalidQueryCvarCookie);
+				pCallback->PushCell(0);
 				pCallback->PushCell(client);
 				pCallback->PushCell(eQueryCvarValueStatus_Cancelled);
 				pCallback->PushString("");
--- a/core/ConVarManager.cpp
+++ b/core/ConVarManager.cpp
@@ -43,6 +43,10 @@ const ParamType CONVARCHANGE_PARAMS[] = {Param_Cell, Param_String, Param_String}
 typedef List<const ConVar *> ConVarList;
 NameHashSet<ConVarInfo *, ConVarInfo::ConVarPolicy> convar_cache;
 
+enum {
+	eQueryCvarValueStatus_Cancelled = -1,
+};
+
 class ConVarReentrancyGuard
 {
 	ConVar *cvar;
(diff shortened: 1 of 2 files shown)


core/ConVarManager.cpp
```diff
@@ -232,12 +236,26 @@ void ConVarManager::OnPluginUnloaded(IPlugin *plugin)
...
+			pCallback = query.pCallback;
+			if (pCallback)
+			{
+				cell_t ret;
+
+				pCallback->PushCell(InvalidQueryCvarCookie);
+				pCallback->PushCell(client);
+				pCallback->PushCell(eQueryCvarValueStatus_Cancelled);
+				pCallback->PushString("\0");
+				pCallback->PushString("\0");
```
asherkin, line 255: Any reason for the double-nul in these strings here?

```suggestion
				pCallback->PushString("");
				pCallback->PushString("");
```

asherkin (commented):
This all looks sane to me. I'll leave it open for a bit in case any one else has any comments on the use of `-1` (I saw a bit on Discord this morning.) I think the only real risk here is SMAC's convar checks, so as long as that's not going to start banning people when getting the value here I'm happy!

psychonic:
It might be worth exposing a feature capability so that plugins that want or need to require this can know if it would ever be called.

core/ConVarManager.cpp
Deathreus: No particular reason, just thought it was a convention, I'll change it

core/ConVarManager.cpp
```diff
@@ -232,12 +236,26 @@ void ConVarManager::OnPluginUnloaded(IPlugin *plugin)
...
 		if (query.client == client)
 		{
+			pCallback = query.pCallback;
+			if (pCallback)
+			{
+				cell_t ret;
+
+				pCallback->PushCell(InvalidQueryCvarCookie);
+				pCallback->PushCell(client);
+				pCallback->PushCell(eQueryCvarValueStatus_Cancelled);
```
KyleSanderson, line 253: This should be QUERYCOOKIE_FAILED equivalent. The Client is already disconnected, and the query failed.

core/ConVarManager.cpp
```diff
@@ -232,12 +236,26 @@ void ConVarManager::OnPluginUnloaded(IPlugin *plugin)
...
 	{
 		ConVarQuery &query = (*iter);
 		if (query.client == client)
 		{
+			pCallback = query.pCallback;
+			if (pCallback)
+			{
+				cell_t ret;
+
+				pCallback->PushCell(InvalidQueryCvarCookie);
```
KyleSanderson, line 251: This feels broken. QUERYCOOKIE_FAILED is 0 and InvalidQueryCvarCookie is -1.

core/ConVarManager.cpp
```diff
@@ -232,12 +236,26 @@ void ConVarManager::OnPluginUnloaded(IPlugin *plugin)
 
 void ConVarManager::OnClientDisconnected(int client)
 {
+	IPluginFunction *pCallback = NULL;
```
KyleSanderson, line 239: This should be in the loop - we're not using the post-result in this scope.

KyleSanderson (changes requested):
This doesn't feel quite right - see comments please.

core/ConVarManager.cpp
asherkin: The naming here was discussed before the PR was made, `_Cancelled` was preferred as there was no failure, and it lets us extend this result to the other cancellation cases in the future (such as the plugin unloading) if desired.

core/ConVarManager.cpp
Deathreus: You right

core/ConVarManager.cpp
Deathreus: The value is the number it is due to this reusing a SDK enum for the values, and the name makes more sense in context and lines up to said enum

plugins/include/convars.inc
```diff
@@ -49,6 +49,7 @@ enum ConVarBounds
  */
 enum ConVarQueryResult
 {
+	ConVarQuery_Cancelled = -1,			//< Client disconnected during query */
```
peace-maker, line 52: The other comments used spaces for alignment instead of tabs.

peace-maker:
What happened here?

Deathreus:
> What happened here?

Oopsie, pulled into here on accident, let me see if I can clean it up

core/ConVarManager.cpp
```diff
@@ -238,6 +242,19 @@ void ConVarManager::OnClientDisconnected(int client)
 		ConVarQuery &query = (*iter);
 		if (query.client == client)
 		{
+			IPluginFunction *pCallback = query.pCallback;
+			if (pCallback)
+			{
+				cell_t ret;
+
+				pCallback->PushCell(0);
```
KyleSanderson, line 250: Where's this value?

KyleSanderson (changes requested):
I agree we should handle this case, but the only avenue the client (readas: script caller) has to reconcile this is to stay on the QueryCookie. When we return 0 that still leaves the initial issue present when there's multiple inflight cases. It's like you're in prison, have two people with the name Wyatt, and get two callbacks for Wyatt but don't know who's clothes are who. You can certainly guess, but you're going to be wrong half the time (and more with the more callbacks outstanding).

core/ConVarManager.cpp
asherkin: Good catch, it looks like it should be available as `query.cookie`.

Deathreus: This is a good point, but I'm not sure how the cookie is needed? I added the value but a more clear explanation rather than an analogy would be appreciated, thanks

peace-maker (approved):

KyleSanderson (approved):

PR #1388, opened 2020-11-28
adminhelp.sp -> Switch to the new CommandIterator()

iNilo:
`ReadCommandIterator` 
->
```
CmdIter.GetName(Name, sizeof(Name));
Flags = CmdIter.Flags;
CmdIter.GetDescription(Desc, sizeof(Desc));
```

### DIFF
--- a/plugins/adminhelp.sp
+++ b/plugins/adminhelp.sp
@@ -119,7 +119,6 @@ public Action HelpCmd(int client, int args)
 			for (i=0; CmdIter.Next() && i<EndCmd; )
 			{
 				CmdIter.GetName(Name, sizeof(Name));
-				CmdIter.GetDescription(Desc, sizeof(Desc));
 
 				if (CheckCommandAccess(client, Name, CmdIter.Flags))
 				{
@@ -160,8 +159,6 @@ public Action HelpCmd(int client, int args)
 		/* Test if there are more commands available */
 		if (CmdIter.Next())
 		{
-			CmdIter.GetName(Name, sizeof(Name));
-
 			PrintToConsole(client, "%t", "Type sm_help to see more", PageNum+1);
 		}
 	}
--- a/plugins/adminhelp.sp
+++ b/plugins/adminhelp.sp
@@ -86,7 +86,6 @@ public Action HelpCmd(int client, int args)
 	char Name[64];
 	char Desc[255];
 	char NoDesc[128];
-	int Flags;
 	CommandIterator CmdIter = new CommandIterator();
 
 	FormatEx(NoDesc, sizeof(NoDesc), "%T", "No description available", client);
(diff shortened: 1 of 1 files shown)


plugins/adminhelp.sp
```diff
@@ -150,9 +162,15 @@ public Action HelpCmd(int client, int args)
...
 
 		/* Test if there are more commands available */
-		if (ReadCommandIterator(CmdIter, Name, sizeof(Name), Flags, Desc, sizeof(Desc)) && CheckCommandAccess(client, Name, Flags))
+		if (CmdIter.Next())
 		{
-			PrintToConsole(client, "%t", "Type sm_help to see more", PageNum+1);
+			CmdIter.GetName(Name, sizeof(Name));
+			Flags = CmdIter.Flags;
+
+			if(CheckCommandAccess(client, Name, Flags))
```
peace-maker, line 170: This check prevents the next page from showing up based on if the first item on the next page isn't available to the user. There might be more commands that are available, so the check should stay gone.

plugins/adminhelp.sp
```diff
@@ -87,15 +87,19 @@ public Action HelpCmd(int client, int args)
...
 	FormatEx(NoDesc, sizeof(NoDesc), "%T", "No description available", client);
 
 	if (DoSearch)
 	{
 		int i = 1;
-		while (ReadCommandIterator(CmdIter, Name, sizeof(Name), Flags, Desc, sizeof(Desc)))
+		while (CmdIter.Next())
 		{
+			CmdIter.GetName(Name, sizeof(Name));
+			Flags = CmdIter.Flags;
```
peace-maker, line 100: You can get rid of the `Flags` variable and use `CmdIter.Flags` directly.

peace-maker (changes requested):
Thanks for the cleanup! A few things inline.

plugins/adminhelp.sp
```diff
@@ -165,12 +161,8 @@ public Action HelpCmd(int client, int args)
 		if (CmdIter.Next())
 		{
 			CmdIter.GetName(Name, sizeof(Name));
```
peace-maker, line 163: This doesn't appear to be necessary anymore.

plugins/adminhelp.sp
```diff
@@ -86,17 +86,19 @@ public Action HelpCmd(int client, int args)
...
 	char NoDesc[128];
-	int Flags;
-	Handle CmdIter = GetCommandIterator();
+	CommandIterator CmdIter = new CommandIterator();
 
 	FormatEx(NoDesc, sizeof(NoDesc), "%T", "No description available", client);
 
 	if (DoSearch)
 	{
 		int i = 1;
```
KyleSanderson, line 95: @peace-maker do all of our command indexes start at 1? These aren't cookies or actual indicies, but what the client can see.

KyleSanderson: This should also respect COMMANDS_PER_PAGE.

KyleSanderson (changes requested):
This should respect COMMANDS_PER_PAGE on help. This isn't on you, but there's existing tech debt here that we can hopefully improve.

KyleSanderson:
@iNilo is there anything we can do to help you with moving this one forward?

plugins/adminhelp.sp
peace-maker: It think that index is just for display and fine to start at 1 since it's not used anywhere.
Paginating the `sm_search` output would require changes to the translation files and seems out of scope for this PR.

peace-maker (approved):

KyleSanderson (approved):

PR #1390, opened 2020-11-29
Add new trie native: ContainsKey()

Einyux:
Nothing fancy here, just a new native to check if a key exists in a trie.
It helps to avoid to create a dummy variable when using GetTrieXXX() to check a key existence.

Native name can be changed if it's not to your liking :-D.

### DIFF
--- a/core/logic/smn_adt_trie.cpp
+++ b/core/logic/smn_adt_trie.cpp
@@ -350,6 +350,27 @@ static cell_t SetTrieString(IPluginContext *pContext, const cell_t *params)
 	return 1;
 }
 
+static cell_t ExistInTrie(IPluginContext *pContext, const cell_t *params)
+{
+	CellTrie *pTrie;
+	HandleError err;
+	HandleSecurity sec = HandleSecurity(pContext->GetIdentity(), g_pCoreIdent);
+
+	Handle_t hndl = params[1];
+
+	if ((err = handlesys->ReadHandle(hndl, htCellTrie, &sec, (void **)&pTrie)) != HandleError_None)
+	{
+		return pContext->ThrowNativeError("Invalid Handle %x (error %d)", hndl, err);
+	}
+
+	char *key;
+	pContext->LocalToString(params[2], &key);
+
+	StringHashMap<Entry>::Result r = pTrie->map.find(key);
+
+	return r.found() ? 1 : 0;
+}
+
 static cell_t RemoveFromTrie(IPluginContext *pContext, const cell_t *params)
 {
 	CellTrie *pTrie;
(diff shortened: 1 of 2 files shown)


core/logic/smn_adt_trie.cpp
```diff
@@ -693,6 +714,7 @@ REGISTER_NATIVES(trieNatives)
 	{"GetTrieArray",			GetTrieArray},
 	{"GetTrieString",			GetTrieString},
 	{"GetTrieValue",			GetTrieValue},
+	{"ExistInTrie",				ExistInTrie},
```
Headline, line 717: Feel free to remove this binding - old syntax can stay in the past

plugins/include/adt_trie.inc
```diff
@@ -255,6 +261,16 @@ native bool GetTrieArray(Handle map, const char[] key, any[] array, int max_size
...
 
+/**
+ * Checks if a key entry exists in a Map.
+ *
+ * @param map           Map Handle.
+ * @param key           Key string.
+ * @return              True if the key entry has been found, else false.
+ * @error               Invalid Handle.
+ */
+native bool ExistInTrie(Handle map, const char[] key);
```
Headline, line 272: ditto

plugins/include/adt_trie.inc
```diff
@@ -114,6 +114,12 @@ methodmap StringMap < Handle
 	//                   as a value or array (not a string).
 	public native bool GetString(const char[] key, char[] value, int max_size, int &size=0);
 
+	// Checks if a key entry exists in a Map.
+	//
+	// @param key        Key string.
+	// @return           True if the key entry has been found, else false.
+	public native bool Exist(const char[] key);
```
Headline, line 121: I think `HasKey` is a better name that is easier to understand

Headline (changes requested):
Just a few nits, otherwise I'm okay with taking this.

core/logic/smn_adt_trie.cpp
Einyux: Ok, I will remove it.

plugins/include/adt_trie.inc
Einyux: Same

plugins/include/adt_trie.inc
Einyux: Browsing though Internet I found some other names:

public native bool Contain(const char[] key); // From C++ 20
public native bool ContainKey(const char[] key); // From Java

Einyux: For the moment, I updated it to "ContainsKey", similar to C#.
https://docs.microsoft.com/en-gb/dotnet/api/system.collections.generic.idictionary-2.containskey?view=net-5.0

Headline (approved):

asherkin (approved):
LGTM, thanks - could bikeshed the name forever but I think `ContainsKey` is nice and clear.

PR #1396, opened 2020-12-05
Fix error description

PlayBoy31:
As coded in https://github.com/alliedmodders/sourcemod/blob/aae71612731eaa6771067a31f3a7ad34bdb0df28/core/logic/smn_players.cpp#L1405, its a "client not in game", not "client not connected".


peace-maker (approved):

peace-maker:
Thank you!

PR #1397, opened 2020-12-05
Take into account the returned value in SDKHook_Think callbacks

Einyux:
To disable entirely the call of "**EntityX::Think()**", using the property "**m_nNextThinkTick**" is enough.
But sometimes I want to use the "**EntityX::Think()**" to do something else, i.e. to replace it with my own code.

In SDKHook_Think callbacks, the returned value need to be taken into account to do this.

### DIFF
--- a/extensions/sdkhooks/extension.cpp
+++ b/extensions/sdkhooks/extension.cpp
@@ -1458,7 +1458,11 @@ void SDKHooks::Hook_StartTouchPost(CBaseEntity *pOther)
 
 void SDKHooks::Hook_Think()
 {
-	Call(META_IFACEPTR(CBaseEntity), SDKHook_Think);
+	cell_t result = Call(META_IFACEPTR(CBaseEntity), SDKHook_Think);
+
+	if(result >= Pl_Handled)
+		RETURN_META(MRES_SUPERCEDE);
+
 	RETURN_META(MRES_IGNORED);
 }
 
--- a/plugins/include/sdkhooks.inc
+++ b/plugins/include/sdkhooks.inc
@@ -237,11 +237,12 @@ typeset SDKHookCB
 	function void (int client);
 
 	// Spawn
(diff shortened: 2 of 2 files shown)


Einyux:
It must be noticed that SDKHook_Think doesn't seem to be the only case.

- SDKHook_PreThink
- SDKHook_PostThink
- SDKHook_VPhysicsUpdate

have the same issue.

It should be let to the user the possibility to block any game call in pre-hooks (**At his own risk**...), or else put in the documentation that for these cases the return value is discarded (For security purposes...).

KyleSanderson:
@psychonic 
image this seems fine to me outside of the crunched style. WDYT?

KyleSanderson:
@asherkin do you know the implications of what happens when an entity stops thinking by chance?

asherkin:
> @asherkin do you know the implications of what happens when an entity stops thinking by chance?

I do not, but it'll be entity-specific and should generally be fine.

> or else put in the documentation that for these cases the return value is discarded.

It is in the documentation,

```
typeset SDKHookCB
{
	// PreThink/Post
	// PostThink/Post
	function void (int client);

	// Spawn
	function Action (int entity);

	// GroundEntChanged
	// SpawnPost
	// Think/Post
	// VPhysicsUpdate/Post
	function void (int entity);
```

That'll need updating for this change as well.

Einyux:
@asherkin Indeed, I missed the callback prototype. Thinking about it now, but will it not break the compatibility with some plugins (And are we okay with this)...?

By the way, what's your opinion about doing the same thing for the three following hooks:
* SDKHook_PreThink
* SDKHook_PostThink
* SDKHook_VPhysicsUpdate

Einyux:
I updated **sdkhooks.inc**. Let's see for other hooks later.

KyleSanderson (approved):

Issue #1173, opened 2020-02-06
MySQL charset not restored upon reconnection


CrazyHackGUT:
Sometimes we need load some plugins before all anothers for priority in forward calling.
Currently, SM doesn't provide any method for changing load priority.

dvander:
Plugins should be agnostic of load order, otherwise, you could get circular dependencies.

Headline:
We already should be handling this internally, what's the issue you're hitting specifically?

CrazyHackGUT:
> what's the issue you're hitting specifically?

I have some plugins who works with MySQL. This plugins doesn't supports `utf8mb4`, so i wrote for myself usage small plugin who remove all special characters from nicknames.
When plugin is uploaded on server as `ClientNameFixer.smx`, some plugins (which are displayed first before my) builds query with special characters such as smilies or any another dumbest symbols who is not be saved without charset `utf8mb4`.
When i rename my plugin into `1.smx` and reload server, my plugin displayed as first in plugin list, and this issue is disappears.

dvander:
There is fundamentally no way to solve this problem, since two plugins could ask to be "first". Your best bet is to fix the plugins not supporting utf8.

CrazyHackGUT:
I will be happy to do it, seriously.
But... I have problem with MySQL driver.

[When MySQL driver losts a connection and reconnects](https://github.com/alliedmodders/sourcemod/blob/5611ec54a21c3045cc1680b954631c6ca049c768/extensions/mysql/mysql/MyDriver.cpp#L110-L114), it does not restore the previously set encoding. The only solution I found to deal with this is to set the encoding before each query, what creates freezes (`SQL_SetCharset` is synchoronous).
Not for all databases i have access to MySQL server configuration for changing default character set.

dvander:
Can we fix this at the DBI level by detecting lost connections?On Feb 6, 2020 7:20 PM, Kruzya <notifications@github.com> wrote:I will be happy to do it, seriously.
But... I have problem with MySQL driver.
When MySQL driver losts a connection and reconnects, it does not restore the previously set encoding. The only solution I found to deal with this is to set the encoding before each query, what creates freezes (SQL_SetCharset is synchoronous).
Not for all databases i have access to MySQL server configuration for changing default character set.

CrazyHackGUT:
> Can we fix this at the DBI level by detecting lost connections?

[...]

PR #1399, opened 2020-12-07
Add OnAdminMenuTitleFormat() forward

CrazyHackGUT:
This PR implements a simple forward for extending admin menu title without requiring a modify plugin.
This forward allows you add any custom content in admin title. For example, primary admin group:
```sourcepawn
#include <sourcemod>
#include <adminmenu>

public void OnPluginStart()
{
    LoadTranslations("adminmenu_groupname.phrases");
}

public void OnAdminMenuTitleFormat(int client, char[] buffer, int maxlength)
{
    AdminId admin = GetUserAdmin(client);
    if (admin == INVALID_ADMIN_ID)
    {
        return; // wut?
    }

    if (admin.GroupCount == 0)
    {
        return; // user doesn't have any admin group.
    }

    // We're suggest first admin group as required for displaying.
    char admingroup_name[64];
    admin.GetGroup(0, admingroup_name, sizeof(admingroup_name));
    Format(buffer, maxlength, "%s\n%T", buffer, "Current Admin Group", client, admingroup_name);
}
```
Example result:
adminmenu_groupname

### DIFF
--- a/plugins/adminmenu.sp
+++ b/plugins/adminmenu.sp
@@ -50,6 +50,7 @@ public Plugin myinfo =
 /* Forwards */
 GlobalForward hOnAdminMenuReady;
 GlobalForward hOnAdminMenuCreated;
+GlobalForward hOnAdminMenuTitleFormat;
 
 /* Menus */
 TopMenu hAdminMenu;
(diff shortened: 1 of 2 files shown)


KyleSanderson:
@CrazyHackGUT while I understand and appreciate where this is coming from, there's no way for this to be properly stacked with a number of plugins listening on the hook (readas: forward) and I'm just seeing massive misuse. A good example is in the screen shot where you've provided Russian and English, this is not a desirable state for an end consumer. If there was Japanese or Korean (perhaps even both!) in there that would offer little value to the end user. This one is tough because I've been there for sure, but this isn't going to be used properly outside of a single plugin author, and that's why I'm closing this one.

I am not the gatekeeper on this, and if there's a more equitable solution I really want to see it to solve the issue because I totally understand what you're trying to do.

CrazyHackGUT:
> there's no way for this to be properly stacked with a number of plugins listening on the hook (readas: forward)

I created a separate issue about setting manually plugin order load some time ago and now it renamed. #1173

> A good example is in the screen shot where you've provided Russian and English, this is not a desirable state for an end consumer.

This is just a PoC. I don't added a russian phrase in my test plugin.
Client in forward is passed, so you can use the right phrase version.

PR #1400, opened 2020-12-07
Implement forward OnNextMapChanged

### LINKED ISSUE #1395: Implement OnNextMapChanged
Can someone implement a forward when the nextmap has been changed ?

Like a OnNextMapChanged() ?

This will be very usefull.

Thanks

PlayBoy31:
Implement forward OnNextMapChanged in order to be notified on plugins when the next map has been changed internally by SetNextMap() native or by cvar change "sm_nextmap". Resolves #1395

### DIFF
--- a/core/NextMap.cpp
+++ b/core/NextMap.cpp
@@ -75,6 +75,10 @@ void NextMapManager::OnSourceModAllInitialized_Post()
 		SH_ADD_HOOK(ConCommand, Dispatch, pCmd, SH_STATIC(CmdChangeLevelCallback), false);
 		changeLevelCmd = pCmd;
 	}
+	
+	m_pOnNextMapChanged = forwardsys->CreateForward("OnNextMapChanged", ET_Ignore, 0, NULL);
+	
+	g_ConVarManager.AddConVarChangeListener("sm_nextmap", this);
 }
 
 void NextMapManager::OnSourceModShutdown()
@@ -98,6 +102,10 @@ void NextMapManager::OnSourceModShutdown()
 		delete (MapChangeData *)*iter;
 		iter = m_mapHistory.erase(iter);
 	}
+	
+	forwardsys->ReleaseForward(m_pOnNextMapChanged);
+	
+	g_ConVarManager.RemoveConVarChangeListener("sm_nextmap", this);
 }
 
 const char *NextMapManager::GetNextMap()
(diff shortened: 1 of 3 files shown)


asherkin:
Thanks for the PR.

What does this add over using a convar change hook for `sm_nextmap` directly?

PlayBoy31:
> 
> 
> Thanks for the PR.
> 
> What does this add over using a convar change hook for `sm_nextmap` directly?

If we dont hook `sm_nextmap` convar change, we will not be notified on the change when the next map is set manually like sm_nextmap a_map

And if the nextmap has been set by SetNextMap() function, we need to detect the change directly on the code on bool NextMapManager::SetNextMap(const char *map)

asherkin:
The existing `NextMapManager::SetNextMap` function sets the `sm_nextmap` cvar, is that set not triggering change hooks?

PlayBoy31:
> 
> 
> The existing `NextMapManager::SetNextMap` function sets the `sm_nextmap` cvar, is that set not triggering change hooks?

I test and i will reply

PlayBoy31:
> 
> 
> The existing `NextMapManager::SetNextMap` function sets the `sm_nextmap` cvar, is that set not triggering change hooks?

Yes indeed there was a double notification, thanks for pointing this. Seems ok now

peace-maker:
> Yes indeed there was a double notification, thanks for pointing this. Seems ok now

So the same functionality is already available using a change hook on `sm_nextmap`? What is this PR adding exactly?

PlayBoy31:
Hello, this PR add a simple way for all to be notified when the nextmap has been changed instead to hook everytime sm_nextmap everywhere on all plugins that need to use it. Its the same logic of the OnMapTimeLeftChanged Forward.

PlayBoy31:
> 
> 
> > Yes indeed there was a double notification, thanks for pointing this. Seems ok now
> 
> So the same functionality is already available using a change hook on `sm_nextmap`? What is this PR adding exactly?

Hello, this PR add a simple way for all to be notified when the nextmap has been changed instead to hook everytime sm_nextmap everywhere on all plugins that need to use it. Its the same logic of the OnMapTimeLeftChanged Forward.

peace-maker:
Hm, `OnMapTimeLeftChanged` doesn't have a simple convar counterpart you can hook, but fires on multiple occasions. I don't think adding such a forward is necassary if all it does is saving to type one line to add a convar hook. It only notifies plugins about changes to what SourceMod thinks is the next map, which might get confusing when the game has its own map handling in parallel.

If changing the next map using `NextMapManager::SetNextMap` wouldn't be observable by a plugin, I'd be down to merge this, but it looks like it is.

PlayBoy31:
Hello. I understand the point, what we do?

asherkin:
I'm going to close this, based on the discussion it seems best for plugins to use a change hook on `sm_nextmap` themselves.

PR #1401, opened 2020-12-07
[CS:GO] Remove control chars from gamedata

c0rp3n:
Someone copied some control chars into the gamedata when updating signatures.


peace-maker (approved):
Someone usually writing right to left?
https://www.fileformat.info/info/unicode/char/202c/index.htm

peace-maker:
Thank you!

PR #1406, opened 2020-12-18
Fix detour HandleCommand_Buy_Internal

komashchenko:


xSL0W:
tested it and it works on linux

asherkin (approved):
Thanks!

PR #1409, opened 2020-12-20
Add sm_dump_datamaps_xml

Scags:
There's an `sm_dump_netprops_xml` but not an `sm_dump_datamaps_xml` and since I went a bit nuts parsing a plain text file, I thought it would be useful to extend it into a parsable format.

This also adds the element offset for sub-class tables since I needed them for my exact purpose.

### DIFF
--- a/extensions/sdktools/vhelpers.cpp
+++ b/extensions/sdktools/vhelpers.cpp
@@ -940,3 +940,154 @@ CON_COMMAND(sm_dump_datamaps, "Dumps the data map list as a text file")
 	fclose(fp);
 
 }
+
+void UTIL_DrawDataTable_XML(FILE *fp, datamap_t *pMap, int level)
+{
+	char spaces[255];
+
+	for (int i = 0; i < level; i++)
+	{
+		spaces[i] = ' ';
+	}
+
+	spaces[level] = '\0';
+
+	const char *externalname;
+	char *flags;
+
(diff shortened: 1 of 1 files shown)


extensions/sdktools/vhelpers.cpp
```diff
@@ -940,3 +940,154 @@ CON_COMMAND(sm_dump_datamaps, "Dumps the data map list as a text file")
...
+
+CON_COMMAND(sm_dump_datamaps_xml, "Dumps the data map list as an XML file")
+{
+#if SOURCE_ENGINE <= SE_DARKMESSIAH
+	CCommand args;
+#endif
+
+	if (args.ArgC() < 2)
+	{
+		META_CONPRINT("Usage: sm_dump_datamaps <file>\n");
```
Headline, line 1002: sm_dump_datamaps_xml here

Headline, line 1002: sm_dump_datamaps_xml here

Headline (changes requested):
Looks good!

Just that one tiny bit I believe

extensions/sdktools/vhelpers.cpp
Scags: Gotcha, funny how I managed to catch that on the other print 🙃.

Headline (approved):

PR #1410, opened 2020-12-20
Add explicit size adjustments to ArrayList

01Pollux:
quite similar to c++'s std vector

### DIFF
--- a/core/logic/CellArray.h
+++ b/core/logic/CellArray.h
@@ -207,6 +207,43 @@ public:
 		return m_AllocSize * m_BlockSize * sizeof(cell_t);
 	}
 
+	size_t capacity()
+	{
+		return m_AllocSize;
+	}
+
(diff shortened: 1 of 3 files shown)


Headline:
Hello - thanks for submitting a patch!

What's the motivation for adding these new natives?

01Pollux:
it's kind of efficient to reserve at least n elements before inserting rather than doing (multiple) reallocations
as for "ShrinkToFit", i just thought it would be nice to reduce number of unused elements/mem usage of an array

KyleSanderson:
This is begging for scripts to start calling GC equivalents on an array whenever they cull elements. I really appreciate the thought behind this, but it's asking for memory management in a language that has no built-in memory. At this time I don't think this is going to be used correctly (we went through this as a community with new vs decl - with decl having decent gains but was not used correctly in a number of cases causing crashes).

PR #1412, opened 2020-12-27
Expose SourcePawn JIT metadata config

asherkin:
Adds a new option to core.cfg to configure the debug/profiling metadata written by the JIT.

Depends on alliedmodders/sourcepawn#533.

### DIFF
--- a/core/sourcemod.cpp
+++ b/core/sourcemod.cpp
@@ -68,6 +68,7 @@ IGameConfig *g_pGameConf = NULL;
 bool g_Loaded = false;
 bool sm_show_debug_spew = false;
 bool sm_disable_jit = false;
+int jit_metadata_flags = JIT_DEBUG_DELETE_ON_EXIT | JIT_DEBUG_PERF_BASIC;
 SMGlobalClass *SMGlobalClass::head = nullptr;
 
 #ifdef PLATFORM_WINDOWS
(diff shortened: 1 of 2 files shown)


KyleSanderson (commented):
Style is all over the map on this one if you can tighten it up a bit.

core/sourcemod.cpp
```diff
@@ -130,8 +131,35 @@ ConfigResult SourceModBase::OnSourceModConfigChanged(const char *key,
...
+			jit_metadata_flags = JIT_DEBUG_DELETE_ON_EXIT | JIT_DEBUG_PERF_BASIC;
+		}
+		else if (strcasecmp(value, "perf") == 0) {
+			jit_metadata_flags = JIT_DEBUG_PERF_BASIC;
+		}
+		else if (strcasecmp(value, "jitdump") == 0) {
+			jit_metadata_flags = JIT_DEBUG_PERF_BASIC | JIT_DEBUG_PERF_JITDUMP;
+		}
+		else {
+			return ConfigResult_Reject;
```
peace-maker, line 157: Would be nice to put the allowed values into the `error` message here.

peace-maker (changes requested):
Is it fine to change the flags during runtime using `sm config JITMetadata jitdump`?

asherkin:
> Is it fine to change the flags during runtime using `sm config JITMetadata jitdump`?

It isn't going to cause any harm, but it may or may not have any effect depending on the current state of things.

peace-maker (approved):

PR #1416, opened 2020-12-30
Update note about non-existing SQL_MoreResults

Impact123:
`SQL_MoreResults` does not exist. The note should refer to `SQL_FetchMoreResults` instead. This was found by Naleksuh and Addie: https://discord.com/channels/335290997317697536/335290997317697536/793685290509139988

### DIFF
--- a/plugins/include/dbi.inc
+++ b/plugins/include/dbi.inc
@@ -797,7 +797,7 @@ native bool SQL_FieldNameToNum(Handle query, const char[] name, int &field);
  * Fetches a row from the current result set.  This must be 
  * successfully called before any results are fetched.
  *
- * If this function fails, SQL_MoreResults() can be used to
+ * If this function fails, SQL_FetchMoreResults() can be used to
  * tell if there was an error or the result set is finished.
  * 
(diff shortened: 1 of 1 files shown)


Naleksuh (approved):
You stole this from me

ErikMinekus:
Pretty sure it should be `SQL_MoreRows`. `SQL_FetchMoreResults` is for fetching multiple result sets from a single query.

plugins/include/dbi.inc
```diff
@@ -162,7 +162,7 @@ methodmap DBResultSet < Handle
 	// Fetches a row from the current result set.  This must be 
 	// successfully called before any results are fetched.
 	//
-	// If this function fails, _MoreResults can be used to
+	// If this function fails, _MoreRows can be used to
```
asherkin, line 165: I think the underscore here was accidentally left over when porting the documentation to the methodmap, and shouldn't be there.

```suggestion
	// If this function fails, MoreRows can be used to
```

asherkin (approved):
LGTM, but it'd be good to get one additional minor change before merging.

PR #1422, opened 2021-01-30
Seed rand only once

Deathreus:
Light testing applied


Deathreus:
See [this](https://hatebin.com/jzqjayshrw) for why it is needed

Headline (approved):
Quite surprised this wasn't discovered earlier - thanks for the patch!

peace-maker:
I'm not sure that `srand` call in core is affecting `rand` calls in `logic`? Might work that way though.

asherkin:
> I'm not sure that `srand` call in core is affecting `rand` calls in `logic`? Might work that way though.

The random seed is thread-local state, so it's shared across all modules in a process.

PR #1424, opened 2021-02-01
Prefer python3 over python(2) in checkout-deps

peace-maker:
We require Python 3 and don't want to accidentally use Python 2 if both are available.

Trying to fix travis builds.


peace-maker:
This all stems from get-pip.py using f-strings python 3 syntax.

Headline (approved):

PR #1428, opened 2021-02-12
fix issues when loading randomcycle from disabled folder

sapphonie:
currently, when loading randomcycle, it runs `Timer_RandomizeNextmap` twice, leading to odd behavior where the map gets changed twice after timelimit is reached. This fixes that issue by only allowing the timer to be run once per map.

### DIFF
--- a/plugins/randomcycle.sp
+++ b/plugins/randomcycle.sp
@@ -4,7 +4,7 @@
  * SourceMod Random Map Cycle Plugin
  * Randomly picks a map from the mapcycle.
  *
- * SourceMod (C)2004-2014 AlliedModders LLC.  All rights reserved.
+ * SourceMod (C)2004-2021 AlliedModders LLC.  All rights reserved.
  * =============================================================================
  *
  * This program is free software; you can redistribute it and/or modify it under
@@ -50,6 +50,7 @@ ConVar g_Cvar_ExcludeMaps;
 ArrayList g_MapList = null;
 ArrayList g_OldMapList = null;
 int g_mapListSerial = -1;
+Handle g_MapTimer = null;
 
 public void OnPluginStart()
 {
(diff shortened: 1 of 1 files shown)


peace-maker:
So `OnConfigsExecuted` is executed multiple times when a plugin is loaded late?

KyleSanderson:
> So `OnConfigsExecuted` is executed multiple times when a plugin is loaded late?

If the map changes twice from our data model there can be multiples of these timers running because this isn't killed with fire when the level ends. To fix this one I think we just need TIMER_FLAG_NOMAPCHANGE and nuke the global as it isn't needed and welcomes glitches.

There was an ancient issue with OnConfigsExecuted firing multiple times, and that could indeed be the issue here (if you want to step through with a debugger that would be cool). However, I'd suspect there's multiple map plugins loaded at the moment, so the server is changing levels twice (which also happened to me, over 10 years ago).

I do believe the current PR resolves the issue, but I don't believe it's the correct fix.

plugins/randomcycle.sp
```diff
@@ -76,7 +77,10 @@ public void OnConfigsExecuted()
 		}
 	}
 	
-	CreateTimer(5.0, Timer_RandomizeNextmap); // Small delay to give Nextmap time to complete OnMapStart()
```
peace-maker, line 79: As KyleS pointed out, the cleaner fix for this issue is setting the `TIMER_FLAG_NO_MAPCHANGE` flag here instead of letting the timer run twice.

peace-maker (changes requested):
Thank you for bringing this up! Please consider the following change if you have the time!

plugins/randomcycle.sp
sapphonie: fixed(?)

plugins/randomcycle.sp
```diff
@@ -76,7 +76,7 @@ public void OnConfigsExecuted()
 		}
 	}
 	
-	CreateTimer(5.0, Timer_RandomizeNextmap); // Small delay to give Nextmap time to complete OnMapStart()
+	CreateTimer(5.0, Timer_RandomizeNextmap, TIMER_FLAG_NO_MAPCHANGE); // Small delay to give Nextmap time to complete OnMapStart()
```
KyleSanderson, line 79: This should be in the flags field, not the data field.

KyleSanderson (changes requested):
The flag is in the wrong arg.

plugins/randomcycle.sp
sapphonie: lmfao oops

KyleSanderson (approved):

peace-maker (approved):

Headline (approved):

PR #1437, opened 2021-03-07
Harden plugin loading path requirements

peace-maker:
Restrict loading of plugins to the `sourcemod/plugins` folder and require the `.smx` file extension.

Symlinks inside the `plugins` folder are fine. Apply the same path restriction to extensions as well.

This behavior was abused as part of justCTF 2020 in the PainterHell challenge by cypis. Thank you!

### DIFF
--- a/core/logic/PluginSys.cpp
+++ b/core/logic/PluginSys.cpp
@@ -974,6 +974,20 @@ IPlugin *CPluginManager::LoadPlugin(const char *path, bool debug, PluginType typ
 	LoadRes res;
 
 	*wasloaded = false;
+
+	if (strstr(path, "..") != NULL)
+	{
+		ke::SafeStrcpy(error, maxlength, "Cannot load plugins outside the \"plugins\" folder");
+		return NULL;
+	}
+
+	const char *ext = libsys->GetFileExtension(path);
+	if (!ext || strcmp(ext, "smx") != 0)
+	{
+		ke::SafeStrcpy(error, maxlength, "Plugin files must have the \".smx\" file extension");
+		return NULL;
+	}
+
 	if ((res=LoadPlugin(&pl, path, true, PluginType_MapUpdated)) == LoadRes_Failure)
 	{
 		ke::SafeStrcpy(error, maxlength, pl->GetErrorMsg());
--- a/core/logic/ExtensionSys.cpp
+++ b/core/logic/ExtensionSys.cpp
@@ -631,6 +631,12 @@ IExtension *CExtensionManager::FindExtensionByName(const char *ext)
 
 IExtension *CExtensionManager::LoadExtension(const char *file, char *error, size_t maxlength)
 {
+	if (strstr(file, "..") != NULL)
+	{
+		ke::SafeStrcpy(error, maxlength, "Cannot load extensions outside the \"extensions\" folder.");
+		return NULL;
+	}
+
 	/* Remove platform extension if it's there. Compat hack. */
 	const char *ext = libsys->GetFileExtension(file);
 	if (strcmp(ext, PLATFORM_LIB_EXT) == 0)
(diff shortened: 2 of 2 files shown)


core/logic/PluginSys.cpp
```diff
@@ -974,6 +974,20 @@ IPlugin *CPluginManager::LoadPlugin(const char *path, bool debug, PluginType typ
...
 	*wasloaded = false;
+
+	if (strstr(path, "..") != NULL)
+	{
+		ke::SafeStrcpy(error, maxlength, "Cannot load plugins outside the \"plugins\" folder");
+		return NULL;
+	}
+
+	const char *ext = libsys->GetFileExtension(path);
+	if (!ext || strcmp(ext, "smx") != 0)
```
KyleSanderson, line 985: Can you add this ptr check to the extension loading code as well (if we're hardening this).

KyleSanderson (commented):
Just one nit in code you didn't touch, but is a crashy bug.

core/logic/PluginSys.cpp
peace-maker: There are no paths that don't enforce an extension to this function, but I'll add the check for good measure.

KyleSanderson: The function is directly callable through IExtensionManager isn't it?

KyleSanderson (approved):

core/logic/PluginSys.cpp
peace-maker: You're right.

naydef:
I'm afraid forcing plugins to have .smx extension will break almost all Freak Fortress 2 subplugins, because of [this line](https://github.com/50DKP/FF2-Official/blob/stable/addons/sourcemod/scripting/include/freak_fortress_2_subplugin.inc#L7)

Working around this is doable i believe, just letting you know about that.

dvander:
Can you explain what this code is trying to do by using a ".ff2" extension?

We believe restricting plugin loading to a single extension is a security improvement, since it'll be harder to load maliciously crafted files. If we were to instead make an extension whitelist, it'd be harder to coordinate that list between GSPs and server operators. ".smx" has been our standard for well over a decade now.

naydef:
Well, it's a fact that Freak Fortress 2 will only load subplugins with .ff2 extension(that's fixable). Plugins identify themselves and other plugins using this_plugin_name variable, populated by GetThisPluginName(), which contains the plugin path with the extension stripped, for ex. _addons/sourcemod/plugins/freaks/default_abilities_. What this variable will contain now is _addons/sourcemod/plugins/freaks/default_abilities.smx_. Honestly, modifying Freak Fortress 2 in a way not breaking subplugins after this change is probably doable, but I don't know whether others(subplugins) have relied on the fact that ff2 subplugins have .ff2 extension.

KyleSanderson:
> Well, it's a fact that Freak Fortress 2 will only load subplugins with .ff2 extension(that's fixable). Plugins identify themselves and other plugins using this_plugin_name variable, populated by GetThisPluginName(), which contains the plugin path with the extension stripped, for ex. _addons/sourcemod/plugins/freaks/default_abilities_. What this variable will contain now is _addons/sourcemod/plugins/freaks/default_abilities.smx_. Honestly, modifying Freak Fortress 2 in a way not breaking subplugins after this change is probably doable, but I don't know whether others(subplugins) have relied on the fact that ff2 subplugins have .ff2 extension.

Totally understood. We're presently doing the exact same thing, I use subfolders to break out requirements (but still use the .smx extension, with symbolic links - which still works with this change). The great thing about this is SM already handles this case built-in, so I think the impact to ff2 here is quite small (and this is a huge quality of life improvement where other files are now ignored, and you won't get owned by arbitrary data in other folders). Valve already has a deny-list built into many games, where smx is covered but I don't believe ff2 is so this should help you there as well.

Making changes like this a decade+ later always hurts someone, but in this case it's a rename of the files impacted. While this isn't my change, if you would like I'd be more than happy to help with reviews in ff2 or similar. If you change to a prefix, as an example, you likely won't get hit by other changes if there's glitches (readas: exploits) found in the future. I also believe the path is visible from GetPluginFilename (so in your case you could likely lean on the `freaks` folder if this is the standard for your distribution of software).

PR #1438, opened 2021-03-08
Update HandleSys.cpp

Wend4r:
Fix return a trash error value on success from HandleSystem::CreateType

### DIFF
--- a/core/logic/HandleSys.cpp
+++ b/core/logic/HandleSys.cpp
@@ -241,6 +241,11 @@ HandleType_t HandleSystem::CreateType(const char *name,
 		pType->children = 0;
 	}
 
+	if (err)
+	{
+		*err = HandleError_None;
+	}
+
 	return index;
 }
 


dvander:
Was this causing a problem anywhere? We typically don't fill success outparams on failure, or failure outparams on success.

Wend4r:
Yes, I created an uninitialized variable thinking that the final handle type creation status would be written there, but there, when successful, I displayed non-standard for HandleError value. I had to look at the contents of the function to understand what needs to be initialized with HandleError_None. It is doesn't look good to do this every time before creating a handle type.

dvander:
You should use the return value of CreateType to determine whether it succeeded or not.

Wend4r:
It would be more readable to get an `enum` value to check for it (for me). Looks at the fact that the HandleType index is returned, it is not very correct to check it for the value zero, instead of -1, for example. I can see that it starts with 1, but I would be more used to seeing -1 as invalid value as it is done in ArrayList.

dvander:
I'm not really against this, but unconditionally filling outparams is not a SourceMod convention. If you're changing the convention for one method in HandleSys, you should change it for all of them for consistency. And as an API change it'd deserve an API number increase in IHandleSys.h. That way, extensions depending on the new behavior won't load on older builds.

KyleSanderson:
Yeah - I agree with dvander on this one. Filling the (optional) error value if there's no error isn't something we do much of elsewhere. Adding the branch and similar when you can test for the failure in the return is the present pattern, and doing the if (!funcsuceed()) { fmt.Printf(err) } is far more common.

PR #1441, opened 2021-03-11
Fix crash when creating threads with Thread_AutoRelease

peace-maker:
Setting the Thread_AutoRelease flag (default when using IThreader::MakeThread) caused a use-after-free after running the thread body.

This fixes Accelerator on SourceMod 1.11 🥳


Headline (approved):
🎉

KyleSanderson (approved):

KyleSanderson:
Nice one!

PR #1444, opened 2021-03-16
Add CloneTrie native

JoinedSenses:
Implementation already exists and is used for StringMap. This PR adds CloneTrie and creates the native link. Also minor reordering and formatting of REGISTER_NATIVES(trieNatives)


peace-maker:
Thanks for the PR, but new features are intentionally only added to the methodmap version if one exists to encourage writing cleaner code.

Issue #1432, opened 2021-02-26
Introduce ConVar.GetDescription() method


dragokas:
Hi,

Can you add please an ability to return "Description" string field by ConVar handle?

It is an argument, filled in CreateConVar() function, and also available via FindFirstConCommand enumerator.

Will be good to easily access everything in one place.
Thank you.

Wend4r:
I also want to add to this request getting the minimum and maximum ConVar values
```
ConVar.HasMinValue.get
ConVar.HasMaxValue.get
ConVar.MinValue.get
ConVar.MaxValue.get
```
I think it will be useful if the plugin would control the value of ConVar through the user (the outer shell) to give him an error and etc..

asherkin:
> I also want to add to this request getting the minimum and maximum ConVar values

Please do not hijack other issues.

What you're after is already implemented as the [GetBounds](https://sourcemod.dev/#/convars/methodmap.ConVar/function.GetBounds) / [SetBounds](https://sourcemod.dev/#/convars/methodmap.ConVar/function.SetBounds) functions.

PR #1449, opened 2021-03-18
Add ConVar.GetDescription() method

Wend4r:
Issue #1432

### DIFF
--- a/core/smn_console.cpp
+++ b/core/smn_console.cpp
@@ -1484,7 +1484,6 @@ REGISTER_NATIVES(consoleNatives)
 	{"SetConVarFlags",		sm_SetConVarFlags},
 	{"ResetConVar",			sm_ResetConVar},
 	{"GetConVarName",		sm_GetConVarName},
-	{"GetConVarDescription",sm_GetConVarDescription},
 	{"GetConVarBounds",		sm_GetConVarBounds},
 	{"SetConVarBounds",		sm_SetConVarBounds},
 	{"QueryClientConVar",	sm_QueryClientConVar},
--- a/core/smn_console.cpp
+++ b/core/smn_console.cpp
@@ -650,6 +650,23 @@ static cell_t sm_GetConVarName(IPluginContext *pContext, const cell_t *params)
 	return 1;
 }
 
+static cell_t sm_GetConVarDescription(IPluginContext *pContext, const cell_t *params)
+{
+	Handle_t hndl = static_cast<Handle_t>(params[1]);
+	HandleError err;
+	ConVar *pConVar;
+
+	if ((err=g_ConVarManager.ReadConVarHandle(hndl, &pConVar))
+		!= HandleError_None)
+	{
+		return pContext->ThrowNativeError("Invalid convar handle %x (error %d)", hndl, err);
+	}
+
+	pContext->StringToLocalUTF8(params[2], params[3], pConVar->GetHelpText(), NULL);
+
+	return 1;
+}
+
 static bool s_QueryAlreadyWarned = false;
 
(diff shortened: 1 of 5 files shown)


Wend4r:
Works
image

plugins/include/convars.inc
```diff
@@ -446,6 +452,16 @@ native void SetConVarBounds(Handle convar, ConVarBounds type, bool set, float va
...
 
+/**
+ * Retrieves the description of a console variable.
+ *
+ * @param convar        Handle to the convar.
+ * @param buffer        Buffer to store the description of the convar.
+ * @param maxlength     Maximum length of string buffer.
+ * @error               Invalid or corrupt Handle.
+ */
+native void GetConVarDescription(Handle convar, char[] buffer, int maxlength);
```
Headline, line 463: I understand adding this for the sake of completion, but we don't add old style functions for new features. Rest looks good to me, though! 😄

Headline (approved):

peace-maker:
We need to figure out how to handle changes to translation phrases before merging this!

Wend4r:
> We need to figure out how to handle changes to translation phrases before merging this!
I followed the standards set by `basecommands`
https://github.com/alliedmodders/sourcemod/blob/ad485069a837f602bdeeeb50f9e02452b3860ecd/plugins/basecommands.sp#L65-L66
Phrases from ConVars are found in these 2 files.
I think this already needs to be fixed with a separate solution.

peace-maker:
Yes, sorry, this is nothing wrong with your changes. Just like #597 changes to the translation files have to be reflected in the [translator tool](https://sm.alliedmods.net/translator/) used to gather all the different languages. That update isn't automated yet and the [process of translation](https://wiki.alliedmods.net/Translating_SourceMod) hasn't been revisited since SourceMod switched to a rolling release cycle afaik.

asherkin:
I think it would be best to remove the basecommands changes so we can get the new API merged.

core/smn_console.cpp
```diff
@@ -650,6 +650,23 @@ static cell_t sm_GetConVarName(IPluginContext *pContext, const cell_t *params)
...
+	HandleError err;
+	ConVar *pConVar;
+
+	if ((err=g_ConVarManager.ReadConVarHandle(hndl, &pConVar))
+		!= HandleError_None)
+	{
+		return pContext->ThrowNativeError("Invalid convar handle %x (error %d)", hndl, err);
+	}
+
+	pContext->StringToLocalUTF8(params[2], params[3], pConVar->GetHelpText(), NULL);
```
peace-maker, line 665: Can `GetHelpText` ever return NULL?

peace-maker (changes requested):
The non-methodmap native needs to go, but otherwise this looks great!

core/smn_console.cpp
Wend4r: It will return a pointer with an empty string if there is no description

Wend4r: https://github.com/alliedmodders/hl2sdk/blob/csgo/tier1/convar.cpp#L163

Wend4r:
When there was an active change with `basecommands` it returns an empty string aka NULL_STRING from `ConVar.GetDescription()` from ConVars without a description.

core/smn_console.cpp
peace-maker: Cool, let's assume this is the case for all games and sdks

peace-maker (approved):

peace-maker:
`NULL_STRING` and an empty string aren't the same, but I guess you're talking about an empty string, so this is fine.

PR #1450, opened 2021-03-20
NormalSHook erroring on non-IsInGame clients

rtldg:
`IEngineSound::EmitSound` seems to be called with clients in the filter that might not be `IsInGame()` which can throw an error when using a `NormalSHook` and returning `Plugin_Changed` since all the clients will will be checked if `IsInGame()`.

Using `CBroadcastRecipientFilter` would add all players than can be grabbed with `UTIL_PlayerByIndex` and is probably cause of `EmitSound` having clients that are not `IsInGame()`.

This PR has been changed to remove clients from the `players` array that are not `IsInGame()`. Also a check for the callback-provided `size` variable has been added

<details>
<summary>previously</summary>
Removing the `IsInGame()` check should be fine since `EmitSound` eventually uses `CGameClient::IsActive()` to only send to active clients.

`EmitSound` -> `EmitSoundInternal` -> `SV_StartSound` ->`sv.BroadcastSound`
```
CGameClient *pClient = Client( index - 1 );

// client must be fully connect to hear sounds
if ( !pClient->IsActive() )
{
	continue;
}

pClient->SendSound( sound, filter.IsReliable() );
```

`CGameClient` -> `CBaseClient`
`virtual	bool	IsActive( void ) const { return m_nSignonState == SIGNONSTATE_FULL; };`
</details>

### DIFF
--- a/extensions/sdktools/vsound.cpp
+++ b/extensions/sdktools/vsound.cpp
@@ -388,6 +388,17 @@ void SoundHooks::OnEmitSound(IRecipientFilter &filter, int iEntIndex, int iChann
 			}
 		case Pl_Changed:
 			{
+				if (size < 0 || size > SM_ARRAYSIZE(players))
+				{
+					pFunc->GetParentContext()->BlamePluginError(pFunc, "Callback-provided size %d is invalid", size);
+
+#if SOURCE_ENGINE >= SE_PORTAL2
+					RETURN_META_VALUE(MRES_IGNORED, -1);
+#else
+					return;
+#endif
+				}
+
 				/* Client validation */
 				for (int i = 0; i < size; i++)
 				{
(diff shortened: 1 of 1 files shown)


Fyren:
> Removing the IsInGame() check should be fine since EmitSound eventually uses CGameClient::IsActive() to only send to active clients.

Did you check if this is true for all engine versions SM supports?  Did you actually test it on any?

rtldg:
> > Removing the IsInGame() check should be fine since EmitSound eventually uses CGameClient::IsActive() to only send to active clients.
> 
> Did you check if this is true for all engine versions SM supports? Did you actually test it on any?

It looks like it has always checked if the client is active in both older and newer engine versions.
Also it was a CSGO server that was throwing these sourcemod errors about players not being ingame after returning `Plugin_Changed` in NormalSHook.

KyleSanderson:
I really don't remember why this was added, I think I hit a filter or something that crashed on a mid-connecting client. That's the reason this one hasn't been merged yet, I know it's painful but the only path to merge this one would be to test every game presently.

rtldg:
Alternatively, how about removing the client from the `players` array if not `IsInGame()`?

KyleSanderson:
> Alternatively, how about removing the client from the `players` array if not `IsInGame()`?

I think that's a fine idea

rtldg:
I force pushed a commit to remove clients from the `players` array if not `IsInGame` and also to check if the `size` variable is valid.

asherkin (approved):
Thanks, I think this looks sane to me, but I'd like a 2nd ack for merge.

KyleSanderson (approved):

PR #1451, opened 2021-03-21
Make logger respect configured datetime format

Mikusch:
Replaces the hardcoded datetime format `%m/%d/%Y - %H:%M:%S` in logging functions with the value of `sm_datetime_format`.

### DIFF
--- a/core/logic/Logger.cpp
+++ b/core/logic/Logger.cpp
@@ -39,6 +39,7 @@
 #include <bridge/include/CoreProvider.h>
 
 Logger g_Logger;
+extern ConVar* g_datetime_format;
 
 ConfigResult Logger::OnSourceModConfigChanged(const char *key, 
 									  const char *value, 
@@ -153,9 +154,10 @@ void Logger::LogToOpenFileEx(FILE *fp, const char *msg, va_list ap)
 	ke::SafeVsprintf(buffer, sizeof(buffer), msg, ap);
 
 	char date[32];
+	const char* fmt = bridge->GetCvarString(g_datetime_format);
 	time_t t = g_pSM->GetAdjustedTime();
 	tm *curtime = localtime(&t);
-	strftime(date, sizeof(date), "%m/%d/%Y - %H:%M:%S", curtime);
+	strftime(date, sizeof(date), fmt, curtime);
 
 	fprintf(fp, "L %s: %s\n", date, buffer);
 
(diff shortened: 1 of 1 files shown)


asherkin:
The logging format is designed to be compatible with the built-in SRCDS log format.

This is likely to break tooling so I'm against it, what's your use case here?

Mikusch:
> 
> 
> The logging format is designed to be compatible with the built-in SRCDS log format.
> 
> This is likely to break tooling so I'm against it, what's your use case here?

Right now it is impossible to change the datetime format used internally by a few functions in the public API such as `LogToFile` or `LogMessage`, which seems like an oversight to me.

Our use case revolves around us trying to use a single datetime format across all of our log files created by SM. It is (in my opinion) a bit misleading and disappointing that the `sm_datetime_format` setting exists but does not change the formatting anywhere except when dumping handles or when using `FormatTime`.

I understand the concerns about this change breaking tooling but I don't think this is an issue considering the default value for `sm_datetime_format` is already the SRCDS log format.

Wend4r:
This format is also specified in the `source engine` logs, and as far as I know, it cannot be changed, except for patches.

Wend4r:
```c
//========= Copyright © 1996-2005, Valve Corporation, All rights reserved. ============//
//=============================================================================//

	tm today;
	Plat_GetLocalTime( &today );

	Q_snprintf( string, sizeof( string ), "L %02i/%02i/%04i - %02i:%02i:%02i: %s",
		today.tm_mon+1, today.tm_mday, 1900 + today.tm_year,
		today.tm_hour, today.tm_min, today.tm_sec, text );

	// Echo to server console
	if ( sv_logecho.GetInt() ) 
	{
		ConMsg( "%s", string );
	}
```

dvander:
Logging to custom files and stuff I don't really have an opinion on. Anything going to the srcds log stream should be in the native format though (and that standard is over twenty years old).

What asherkin meant by "break tooling" is that the cvar is documented as only affecting FormatTime(). For users who changed the cvar with that intent, they'll now have log files in a completely different format. You'd need a new cvar.

Mikusch:
Would it be satisfactory to introduce a new cvar `sm_log_datetime_format` (with the srcds format as its default value) and keep the hardcoded srcds format for the call to `ConPrint`, while respecting the format string from the cvar for writing to log files?

KyleSanderson:
At this time I don't think we want this @Mikusch (as-is, atleast). The reason is the tooling / formatting as expected above. Folks using regex as an example would be hit by this out of the blue, using the native format is the supported path for this. I know it's a little more work but using your own log function in a plugin (you could put it in an inc if you'd like) is the avenue to take this one in my opinion.

Issue #1447, opened 2021-03-17
MTU incorrect size


Wend4r:
Why do you sets MTU to 5000?
https://github.com/alliedmodders/sourcemod/blob/ae485c311541bb0fa609e5a9a33e0fdf1595fd56/extensions/cstrike/rulesfix.cpp#L85
Fixed MTU value 1200 means that the packet is not broken and you patch it what leads to CS:GO engine problems with A2S (Looking at the magical source code).

```
	// Basic UDP send. Packets can't be bigger than 1200 bytes (your typical MTU size). Can be lost, or arrive out of order (rare).
	// The sending API does have some knowledge of the underlying connection, so if there is no NAT-traversal accomplished or
	// there is a recognized adjustment happening on the connection, the packet will be batched until the connection is open again.
```

I have spam in server console when my server is listening through A2S with `host_rules_show 1`
```
[NET] Cannot send 2886-byte packet to x.x.x.x:x.  MTU is 1200.  ff ff ff ff 45
[NET] Cannot send 2886-byte packet to x.x.x.x:x.  MTU is 1200.  ff ff ff ff 45
[NET] Cannot send 2886-byte packet to x.x.x.x:x.  MTU is 1200.  ff ff ff ff 45
[NET] Cannot send 2886-byte packet to x.x.x.x:x.  MTU is 1200.  ff ff ff ff 45
[NET] Cannot send 2886-byte packet to x.x.x.x:x.  MTU is 1200.  ff ff ff ff 45
[NET] Cannot send 2886-byte packet to x.x.x.x:x.  MTU is 1200.  ff ff ff ff 45
[NET] Cannot send 2886-byte packet to x.x.x.x:x.  MTU is 1200.  ff ff ff ff 45
[NET] Cannot send 2886-byte packet to x.x.x.x:x.  MTU is 1200.  ff ff ff ff 45
[NET] Cannot send 2886-byte packet to x.x.x.x:x.  MTU is 1200.  ff ff ff ff 45
```
(I replace `x` instead of the real IP and port the who listening)

with new game.cstrike.ext.2.csgo.so (#614) :
```
sm exts info 4
 File: game.cstrike.ext.2.csgo.so
 Loaded: Yes (version 1.11.0.6666)
 Name: CS Tools (CS extended functionality)
 Author: AlliedModders LLC (http://www.sourcemod.net/)
 Binary info: API version 8 (compiled Mar 15 2021 23:51:29)
[...]

PR #1459, opened 2021-03-30
A2S_Rules fix: Only change host_rules_show if patching was successful

peace-maker:
If lookup of the patch location failed, don't turn on responding to A2S_Rules queries. Otherwise we'll see the log getting spammed with too large packet sizes again. #1447


KyleSanderson (approved):
We should likely make the MTU value part of gamedata as opposed to being a hard coded magic value but that didn't come in with this.

Issue #890, opened 2018-09-27
Add global forward OnPluginUnload


R1KO:
How do you look at adding this?
Maybe I can do it myself and send you a pull request

https://forums.alliedmods.net/showpost.php?p=2091765&postcount=21

Headline:
Thanks for creating an issue!

Can you use `OnLibraryRemoved` instead? Our current dependency system is undefined in terms of load/unload order, but using libraries is much preferred.

KyleSanderson:
I have an entire IPC system that absolutely needed this function 7 years ago... What I ended up doing was a hack in OnPluginEnd in every single plugin to let the memory management plugin know that the user was disappearing. There's another hack that you can do which is a private forward per plugin, and if GetFunctionCount returns 0 you know that user is gone.

Sorry I don't have a better answer for you.

PR #948, opened 2019-01-12
Adding a forward that lets you know when another plugin is unloaded

OkyHp:
This may allow the kernel plugin to delete the data of the module that is being unloaded to avoid problems the next time the module is loaded.

PR #1462, opened 2021-04-03
Add OnNotifyPluginUnloaded() forward

Wend4r:
Issue #890 

I wrote this Pull Request in considerations of the handles principles that a plugin must necessarily know when the handle closes or close manually if it is cloned.

### DIFF
--- a/core/logic/PluginSys.cpp
+++ b/core/logic/PluginSys.cpp
@@ -1505,6 +1505,9 @@ void CPluginManager::Purge(CPlugin *plugin)
 			(*iter)->OnPluginUnloaded(plugin);
 	}
 
+	m_pOnNotifyPluginUnloaded->PushCell(plugin->GetMyHandle());
+	m_pOnNotifyPluginUnloaded->Execute(NULL);
+
 	plugin->DropEverything();
 
 	for (ListenerIter iter(m_listeners); !iter.done(); iter.next())
@@ -1587,6 +1590,7 @@ void CPluginManager::OnSourceModAllInitialized()
 
 	m_pOnLibraryAdded = forwardsys->CreateForward("OnLibraryAdded", ET_Ignore, 1, NULL, Param_String);
 	m_pOnLibraryRemoved = forwardsys->CreateForward("OnLibraryRemoved", ET_Ignore, 1, NULL, Param_String);
+	m_pOnNotifyPluginUnloaded = forwardsys->CreateForward("OnNotifyPluginUnloaded", ET_Ignore, 1, NULL, Param_Cell);
 }
 
 void CPluginManager::OnSourceModShutdown()
@@ -1601,6 +1605,7 @@ void CPluginManager::OnSourceModShutdown()
 
 	forwardsys->ReleaseForward(m_pOnLibraryAdded);
 	forwardsys->ReleaseForward(m_pOnLibraryRemoved);
+	forwardsys->ReleaseForward(m_pOnNotifyPluginUnloaded);
 }
 
 ConfigResult CPluginManager::OnSourceModConfigChanged(const char *key,
--- a/core/logic/PluginSys.cpp
+++ b/core/logic/PluginSys.cpp
@@ -1499,14 +1499,14 @@ void CPluginManager::Purge(CPlugin *plugin)
 	if (plugin->GetStatus() == Plugin_Running)
 		plugin->Call_OnPluginEnd();
 
+	m_pOnNotifyPluginUnloaded->PushCell(plugin->GetMyHandle());
+	m_pOnNotifyPluginUnloaded->Execute(NULL);
+
 	// Notify listeners of unloading.
 	if (plugin->EnteredSecondPass()) {
 		for (ListenerIter iter(m_listeners); !iter.done(); iter.next())
 			(*iter)->OnPluginUnloaded(plugin);
 	}
-
-	m_pOnNotifyPluginUnloaded->PushCell(plugin->GetMyHandle());
-	m_pOnNotifyPluginUnloaded->Execute(NULL);
 
 	plugin->DropEverything();
 
--- a/core/logic/PluginSys.cpp
+++ b/core/logic/PluginSys.cpp
@@ -1505,6 +1505,8 @@ void CPluginManager::Purge(CPlugin *plugin)
 			(*iter)->OnPluginUnloaded(plugin);
 	}
 
+	CallUnloadHooks(plugin);
+
 	plugin->DropEverything();
 
 	for (ListenerIter iter(m_listeners); !iter.done(); iter.next())
@@ -1601,6 +1603,19 @@ void CPluginManager::OnSourceModShutdown()
 
 	forwardsys->ReleaseForward(m_pOnLibraryAdded);
 	forwardsys->ReleaseForward(m_pOnLibraryRemoved);
+
+	{
+		std::list<PluginUnloadHookInfo>::iterator iter = m_PluginUnloadHooks.begin();
+	
+		while (iter!=m_PluginUnloadHooks.end())
+		{
+			forwardsys->ReleaseForward(iter->pForward);
+
+			iter++;
+		}
+
+		m_PluginUnloadHooks.clear();
+	}
 }
 
 ConfigResult CPluginManager::OnSourceModConfigChanged(const char *key,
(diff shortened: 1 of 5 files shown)


Wend4r:
Code to check

`pull1462.sp`
```c
#pragma semicolon 1
#pragma newdecls required

#include <sourcemod>

#define MAX_Cycle_MESSAGE_LENGTH 1024

enum struct CycleMessageInfo
{
	Handle plugin;
	char message[MAX_Cycle_MESSAGE_LENGTH];

	void SendMessage()
	{
		PrintToServer("%s", this.message);
	}
}

ArrayList g_CycleMessageList;

public APLRes AskPluginLoad2(Handle myself, bool late, char[] error, int err_max)
{
	CreateNative("SendCycleMessage", Native_SendCycleMessage);
}

any Native_SendCycleMessage(Handle plugin, int numParams)
{
	CycleMessageInfo info;

	info.plugin = plugin;

	LogMessage("Detected %X plugin", plugin);

	GetNativeString(1, info.message, sizeof(CycleMessageInfo::message));
	g_CycleMessageList.PushArray(info);

	HookPluginUnload(plugin, OnPluginUnloaded);
}

public void OnPluginStart()
{
	g_CycleMessageList = new ArrayList(sizeof(CycleMessageInfo));

	CreateTimer(5.0, OnPrintMessageTimer, _, TIMER_REPEAT);
}

Action OnPrintMessageTimer(Handle timer)
{
	CycleMessageInfo info;

	for(int i = 0, length = g_CycleMessageList.Length; i != length; i++)
	{
		g_CycleMessageList.GetArray(i, info, sizeof(info));
		info.SendMessage();
	}
}

void OnPluginUnloaded(Handle plugin)
{
	LogMessage("Plugin %X is unloaded", plugin);

	int index = g_CycleMessageList.FindValue(plugin, CycleMessageInfo::plugin);

	g_CycleMessageList.Erase(index);
}
```

`pull1462_module.sp`
```c
#pragma semicolon 1
#pragma newdecls required

#include <sourcemod>

native void SendCycleMessage(const char[] message);

public void OnPluginStart()
{
	SendCycleMessage("Hello!");
}
```

Result:
image

KyleSanderson:
@dvander we've been all over the map on this one over the years - what's your take? I need it for memory management in another plugin but that usecase is very different.

plugins/include/sourcemod.inc
```diff
@@ -319,6 +326,26 @@ native bool GetPluginInfo(Handle plugin, PluginInfo info, char[] buffer, int max
...
+
+/**
+ * Removes a hook for when a plugin unloaded.
+ * 
+ * @param plugin        Plugin Handle to unhook.
+ * @param callback      An PluginUnloaded function pointer.
+ * 
+ * @return              True on success, false otherwise.
+ */
+native bool UnhookPluginUnload(Handle plugin, PluginUnloaded callback);
```
dvander, line 347: Do you need the Unhook method? The unload event only happens once per plugin. With a use case like yours, either the controller plugin unloads (and no longer cares about removing the hook), or the target plugin unloads, and the hook fires.

The scenarios where you really care about removing the hook are pretty limited, and it would simplify this change quite a bit.

core/logic/PluginSys.h
```diff
@@ -425,6 +425,12 @@ class CPluginManager :
 
 	bool LibraryExists(const char *lib);
 
+	bool HookPluginUnload(IPlugin *pPluginToHook, IPluginFunction *pHookFunction);
+
+	bool UnhookPluginUnload(IPlugin *pPluginToUnhook, IPluginFunction *pHookFunction);
```
dvander, line 430: No need for spaces between these methods.

core/logic/PluginSys.cpp
```diff
@@ -2202,6 +2217,72 @@ bool CPluginManager::LibraryExists(const char *lib)
...
+	std::list<PluginUnloadHookInfo>::iterator iter = m_PluginUnloadHooks.begin();
+
+	IChangeableForward *pForwardToCall;
+
+	while (iter!=m_PluginUnloadHooks.end())
+	{
+		// There can only be one unique
+		if(iter->pPluginTarget == pTarget)
+		{
+			pForwardToCall = iter->pForward;
```
dvander, line 2271: Move the declaration above to here, eg:

    IChangeableForward *pForwardToCall = iter->pForward;

core/logic/PluginSys.cpp
```diff
@@ -2202,6 +2217,72 @@ bool CPluginManager::LibraryExists(const char *lib)
...
+
+			pForwardToCall->PushCell((cell_t)pTarget->GetMyHandle());
+			pForwardToCall->Execute();
+
+			pForwardToCall->Cancel();
+			m_PluginUnloadHooks.erase(iter);
```
dvander, line 2277: Calling [Unh|H]ookPluginUnload in this callback will crash.

dvander (changes requested):
The idea behind this seems good to me, but there are some complexity and stability issues here.

I think you could fix them by removing PluginUnloadHookInfo. If you put the IChangeableForward directly on the CPlugin object, this patch becomes a lot simpler, and you won't have the stability issues (IChangeableForward handles the complexity of list iteration for you).

plugins/include/sourcemod.inc
Wend4r: The plugin can call natives, which can be execute unhook.
Similar to the concept of allocate an object and closing it.

dvander: Does it matter, though? A global forward - something like - "OnNotifyPluginUnloaded" - would be very simple to implement and would require only a bit of filtering on your end. And if you can do that filtering, you don't need the Unhook method.

Wend4r:
Works!

[source.zip](https://github.com/alliedmodders/sourcemod/files/6534833/source.zip)
```
sm plugins unload pull1462
[SM] Plugin pull1462.smx unloaded successfully.
sm plugins load pull1462
[SM] Loaded plugin pull1462.smx successfully.
sm plugins load pull1462_module
L 05/24/2021 - 23:31:18: [pull1462.smx] Detected 27800CC plugin
[SM] Loaded plugin pull1462_module.smx successfully.
Hello!
Hello!
Hello!
sm plugins unload pull1462_module
L 05/24/2021 - 23:31:34: [pull1462.smx] Destroyed 27800CC plugin
[SM] Plugin pull1462_module.smx unloaded successfully.
```

core/logic/PluginSys.cpp
```diff
@@ -1505,6 +1505,9 @@ void CPluginManager::Purge(CPlugin *plugin)
 			(*iter)->OnPluginUnloaded(plugin);
 	}
 
+	m_pOnNotifyPluginUnloaded->PushCell(plugin->GetMyHandle());
+	m_pOnNotifyPluginUnloaded->Execute(NULL);
```
dvander, line 1509: Can you move this to above line 1502 (eg before OnPluginUnloaded callbacks fire?) That ordering will be a tiny bit safer.

Once "OnPluginUnloaded" has been called, we don't want any new stuff getting accidentally registered on the plugin, since otherwise it might leak or crash.

dvander (changes requested):
Thanks! This is *much* simpler. Just one small comment.

KyleSanderson:
This looks slightly bugged (the forward says Unloaded, but the plugin is still loaded). I think we may need two forwards here, one for OnPluginUnloading, and a final for OnPluginUnloaded after everything has released resources.

dvander (approved):

dvander:
I think this is fine and we can merge it.

Once "OnPluginEnd" has been called, it is for all intents and purposes unloaded, and the only thing left to do is cleanup. Besides not having a clear use case, what you're asking for has a catch-22: if you release the resources, you don't have the handle anymore, so how can you notify?

KyleSanderson:
> Once "OnPluginEnd" has been called, it is for all intents and purposes unloaded, and the only thing left to do is cleanup. 

For my specific usecase Forwards are destroyed in OnPluginUnloaded, so we need a final call for OnPluginDestroyed (or the equivalent of it) in pawn. Wendr is looking for the same behaviour it looks like here, because they're using the plugin handle as a value to be found in an array. When you'd use OnPluginUnloading is if you need the pubinfo from the plugin, or to access resources still allocated which would match the OnClientDisconnecting vs OnClientDisconnected behaviour, no?

Wend4r:
> This looks slightly bugged (the forward says Unloaded, but the plugin is still loaded). I think we may need two forwards here, one for OnPluginUnloading, and a final for OnPluginUnloaded after everything has released resources.

In my case, this forward is needed to clean up the data attaches to the native calling handler at the SourcePawn level. I do not know in which cases an additional forward for a fully cleaned plugin from extensions.

dvander:
"For my specific usecase Forwards are destroyed in OnPluginUnloaded" - This doesn't explain why such a thing would be needed, and indeed there is nothing within SourceMod that needs this distinction internally.

KyleSanderson:
Thanks for bearing with us team, this has now landed (if there are other issues related to this please let them know).

https://github.com/alliedmodders/sourcemod/pull/948
https://github.com/alliedmodders/sourcemod/issues/1482
https://github.com/alliedmodders/sourcemod/issues/1481

dragokas:
That commit does not cover OnPluginLoad(), which is now closed: https://github.com/alliedmodders/sourcemod/issues/1482

PR #1463, opened 2021-04-03
Add defines for Cookie name and description max length

Natanel-Shitrit:
I added the following defines because there is almost no information about the max lengths of the cookie's name and description.

The only way place i found out the max lengths was inside the `CREATE TABLE` query inside the extension module.
(https://github.com/alliedmodders/sourcemod/blob/master/extensions/clientprefs/extension.cpp#L227-L334)

This way it will be easier to know the max lengths and also help avoiding a problem that happens when registering a cookie with a name longer than the max length.

### DIFF
--- a/plugins/include/clientprefs.inc
+++ b/plugins/include/clientprefs.inc
@@ -74,6 +74,9 @@ enum CookieMenuAction
 	CookieMenuAction_SelectOption = 1
 };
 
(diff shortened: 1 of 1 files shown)


KyleSanderson (approved):
nice one - this should be a nice quality of life improvement.

PR #1465, opened 2021-04-04
Correct documentation in adt_trie.inc

yourmnbbn:
Handle map parameter is not needed in this methodmap function.


Headline (approved):
Thanks! Good catch

PR #1469, opened 2021-04-10
Update CheckCommandAccess to not perform override lookup on empty string

JoinedSenses:
Command name shouldn't be empty when performing a lookup. This PR will skip the lookup if string is empty and fallback to default flag.

If there might be a better way to approach this, please let me know.

### DIFF
--- a/core/logic/AdminCache.cpp
+++ b/core/logic/AdminCache.cpp
@@ -2053,7 +2053,7 @@ bool AdminCache::CheckAdminCommandAccess(AdminId adm, const char *cmd, FlagBits 
 				/* First get group-level override */
 				override = GetGroupCommandOverride(gid, cmd, Override_CommandGroup, &rule);
 				/* Now get the specific command override */
-				if (!override && GetGroupCommandOverride(gid, cmd, Override_Command, &rule))
+				if (GetGroupCommandOverride(gid, cmd, Override_Command, &rule))
 				{
 					override = true;
 				}
--- a/core/logic/AdminCache.cpp
+++ b/core/logic/AdminCache.cpp
@@ -2038,32 +2038,36 @@ bool AdminCache::CheckAdminCommandAccess(AdminId adm, const char *cmd, FlagBits 
 			return true;
 		}
 
-		/* Check for overrides
-		* :TODO: is it worth optimizing this?
-		*/
-		unsigned int groups = GetAdminGroupCount(adm);
-		GroupId gid;
-		OverrideRule rule;
-		bool override = false;
-		for (unsigned int i = 0; i<groups; i++)
+		if (cmd[0] != '\0')
 		{
-			gid = GetAdminGroup(adm, i, NULL);
-			/* First get group-level override */
-			override = GetGroupCommandOverride(gid, cmd, Override_CommandGroup, &rule);
-			/* Now get the specific command override */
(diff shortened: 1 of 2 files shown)


dvander:
Why would the input be empty?

core/logic/AdminCache.cpp
```diff
@@ -2038,32 +2038,36 @@ bool AdminCache::CheckAdminCommandAccess(AdminId adm, const char *cmd, FlagBits
...
-				override = true;
-			}
-			if (override)
-			{
-				if (rule == Command_Allow)
+				gid = GetAdminGroup(adm, i, NULL);
+				/* First get group-level override */
+				override = GetGroupCommandOverride(gid, cmd, Override_CommandGroup, &rule);
+				/* Now get the specific command override */
+				if (GetGroupCommandOverride(gid, cmd, Override_Command, &rule))
```
JoinedSenses, line 2056: Could this be
```cpp
if (!override && GetGroupCommandOverride(gid, cmd, Override_Command, &rule))
```
or is the result needed from this lookup even if the group-level override was true?

JoinedSenses (commented):
> Why would the input be empty?

It normally shouldn't/wouldn't be - I've found myself and others using empty strings to use the default flags in some instances. My original assumption was that it didn't perform a lookup if the string were empty. It's either this or throw an error, since empty strings also shouldn't be in the overrides, although @asherkin mentioned an instance where that has mistakenly happened with someone.

dvander:
What does "default flags" mean here? I might be missing a use case or something which would help me understand this. CheckCommandAccess is supposed to be checking access to a specific command, so it wouldn't normally make sense to have an empty string. There's other functions that can grab effective flags and test them.

JoinedSenses:
> What does "default flags" mean here?

`default flags` being the third parameter of CheckCommandAccess.
`Flag string to use as a default, if the command or override is not found`

> I might be missing a use case or something which would help me understand this. CheckCommandAccess is supposed to be checking access to a specific command, so it wouldn't normally make sense to have an empty string. There's other functions that can grab effective flags and test them.

`CheckCommandAccess(client, "", ADMFLAG_ROOT, true);`
A cleaner/lazy alternative to `GetUserFlagBits(client) & ADMFLAG_ROOT`

There are also no instances where a valid override should be an empty string, which means checking overrides for it would be a waste of resources or could cause issues for others i've seen also using empty strings.

dvander:
I'd just use GetUserFlagBits() which is much clearer. If you want a shorthand stock that returns a bool (called something like "CheckUserFlags") that sounds good too. Both of those options would be better than adding edge cases to CheckCommandAccess.

JoinedSenses:
> I'd just use GetUserFlagBits() which is much clearer. If you want a shorthand stock that returns a bool (called something like "CheckUserFlags") that sounds good too. Both of those options would be better than adding edge cases to CheckCommandAccess.

sure i agree with that, although i would also think it would be sane to only do override lookups on non-empty strings

KyleSanderson:
so, this is working as intended. The command should never be empty as this would defeat the purpose behind allowing users to change flags for a command without having to write code (read: admin_overrides.cfg). I'm closing this one as I think it's invalid (and we should maybe even error out here), but if someone else feels strongly the other way I would appreciate hearing that side of things.

PR #1488, opened 2021-05-06
Start using Github Actions

peace-maker:
Build on windows and linux. Cannot build for macos, since the builders only support xcode 10+ which dropped x86 support. It appears that `-stdlib=libstdc++` isn't supported anymore either, so we can't even build csgo on macx64 due to the pb_proxy hack.

This builds on the latest default compiler of the respective latest image version of the platform. In this case vs2019 and clang 10. I've added another build against clang 3.9, since that is the oldest version packaged in bionic. Could add a build against vs2017 as well if needed, but GHA doesn't support the compiler versions used on buildbot.

Next to the continuous integration workflow there's another one to only build the sourcepawn tooling and package it up. This creates smaller packages containing only spcomp and the include files for faster plugin development setup.

Thanks to @shqke for the inspiration!


KyleSanderson (approved):
We should probably turn down Travis after all the projects have flipped.

tools/buildbot/PackageScript_scripting
```diff
@@ -0,0 +1,138 @@
+# vim: set ts=8 sts=2 sw=2 tw=99 et ft=python:
```
dvander, line 1: nit but can we call this ToolsPackageScript or something?

tools/buildbot/PackageScript_scripting
```diff
@@ -0,0 +1,138 @@
...
+    if 'version_auto.inc' in header.path:
+      builder.AddCopy(header, include_folder)
+
+# Export PDB files. We write to a file in the build folder which is pretty
+# verboten, but it's okay if it's in the root since AMBuild will never try
+# to rmdir the root.
+full_binary_list = SM.spcomp_bins
+with open(os.path.join(builder.buildPath, 'pdblog.txt'), 'w') as fp:
+  for task in full_binary_list:
+    fp.write(task.debug.path + '\n')
```
dvander, line 46: do we need a pdb log for the tools package?

tools/buildbot/PackageScript_scripting
```diff
@@ -0,0 +1,138 @@
...
+# Do all straight-up file copies from the source tree.
+def CopyFiles(src, dest, files):
+  if not dest:
+    dest = src
+  for source_file in files:
+    source_path = os.path.join(builder.sourcePath, src, source_file)
+    builder.AddCopy(source_path, dest)
+
+CopyFiles('plugins/include', include_folder,
+  [ 'admin.inc',
```
dvander, line 58: Oh god, is this something we had to do in PackageScript as well? Looks like it. I'll add a "copy *" command to AMBuild when I get a chance...

In the meantime, it'd be nice to not duplicate all this. You could accomplish that by having a "PackageConstants" or "PackageHelpers" script that returns a dictionary or class with all the functions/lists attached to it. The root build script would run this script and propagate the values down via the SM object or something.

Or, just stick to one thing (PackageScript) and put some "if" conditions in to control what gets built. That might be gross, but either would probably be better than having to update both places when new files are added?

dvander (approved):
LGTM, have comments but approving so we can drop Travis support faster

tools/buildbot/PackageScript_scripting
peace-maker: Probably not, I've removed it.

tools/buildbot/PackageScript_scripting
peace-maker: I've went the `PackageHelpers` route and implemented the "copy *" command as well making this way more readable.

tools/buildbot/PackageHelpers
```diff
@@ -0,0 +1,73 @@
...
+          self.builder.AddCopy(bin_task.binary, os.path.join(target_folder, file))
+        else:
+          self.builder.AddCopy(bin_task.binary, self.folder_map[target_folder])
+    
+    if self.builder.host.platform == 'windows':
+      self.CopyFiles('tools/batchtool', target_folder, '.exe')
+    else:
+      self.CopyFiles('plugins', target_folder, '.sh')
+  
+  def CopyFiles(self, src, dest, filter_ext=None):
```
dvander, line 49: This has to be implemented in AMBuild as a special graph command, which is why we hardcoded the file list. The build scripts are only reparsed if they change, so if you add or remove a file in this folder, the build will either silently omit or will fail not finding it.

This is more of a PSA to avoid it in the future (until native support comes) - the file list here rarely changes, and we can work around it by purging the build cache on buildbot.

dvander (approved):

tools/buildbot/PackageHelpers
dvander: (aka, please go with these changes, they're good)

PR #1494, opened 2021-05-28
Update basetriggers ShowFriendlyFire to use the same chat text visibility system as ShowTimeLeft

FlaminSarge:
The entire setup is sketchy but I am not going to rework it all at the moment to something like ReplyToCommand.

This also allows server to call 'ff' if it wants to.


FlaminSarge:
(This makes it so that one client spamming 'ff' in console doesn't cause chat to be spammed.)

asherkin (approved):

PR #1495, opened 2021-06-03
Update CGameConfig members to use std::string instead of fixed-size char arrays

### LINKED ISSUE #1381: Silent failure when accessing gamedata offset names >= 64 chars
# Help us help you

# Environment
  * ~~Current~~ Running SourceMod version: 1.10.0.6492
  * Current SourceMod snapshot: 1.11.0.6642

# Description

A plugin I was prototyping has a gamedata offset named "ISchemaAttributeTypeBase::BSupportsGameplayModificationAndNetworking()", which is 70 chars long.

When attempting to read the value via `GameData.GetOffset`, it returns -1.  When using `PrepSDKCall_SetFromConf`, it returns `false`.

The name matches what is listed in the gameconf, so this ends up being an unexpected failure case.

Proposed solutions:

- Make the char array longer.
- Refactor to support strings of arbitrary length.
- Throw a proper exception at parse / retrieval time if handling a string that is too long (so the plugin author knows that the string needs to be shorter).

I don't want to go straight to PRing the first solution even if it's the easiest fix (as far as I know), would prefer to run it through the issue tracker first.

1. [See linked sample code and gameconfig.](https://gist.github.com/nosoop/060becd2936881faa6d77827233734ac)
2. Compile plugin and install plugin and gameconfig.  On plugin load, a message will be logged with the returned value of `GameConfGetOffset` on names with 64 and 63 characters.

nosoop:
Fixes #1381.

~~Currently marked as a draft since it's untested and there's still some fixed-size char arrays I need to look into under `AddressConf`.  Will test once those remaining things are looked at.~~ Tested in a TF2 server loaded up with a lot of gamedata.

### DIFF
--- a/core/logic/GameConfigs.cpp
+++ b/core/logic/GameConfigs.cpp
@@ -374,7 +374,7 @@ SMCResult CGameConfig::ReadSMC_NewSection(const SMCStates *states, const char *n
 				if (strcmp(name, "linux") != 0 && strcmp(name, "windows") != 0 && strcmp(name, "mac") != 0 &&
 					strcmp(name, "linux64") != 0 && strcmp(name, "windows64") != 0 && strcmp(name, "mac64") != 0)
 				{
-					logger->LogError("[SM] Error while parsing Address section for \"%s\" (%s):", m_Address, m_CurFile);
+					logger->LogError("[SM] Error while parsing Address section for \"%s\" (%s):", m_Address.c_str(), m_CurFile);
 					logger->LogError("[SM] Unrecognized platform \"%s\"", name);
 				}
 				m_IgnoreLevel = 1;
@@ -476,7 +476,7 @@ SMCResult CGameConfig::ReadSMC_KeyValue(const SMCStates *states, const char *key
 			int limit = sizeof(m_AddressRead)/sizeof(m_AddressRead[0]);
 			if (m_AddressLastIsOffset)
 			{
-				logger->LogError("[SM] Error parsing Address \"%s\", 'offset' entry must be the last entry (gameconf \"%s\")", m_Address, m_CurFile);
+				logger->LogError("[SM] Error parsing Address \"%s\", 'offset' entry must be the last entry (gameconf \"%s\")", m_Address.c_str(), m_CurFile);
 			}
 			else if (m_AddressReadCount < limit)
 			{
@@ -489,7 +489,7 @@ SMCResult CGameConfig::ReadSMC_KeyValue(const SMCStates *states, const char *key
 			}
 			else
 			{
-				logger->LogError("[SM] Error parsing Address \"%s\", does not support more than %d read offsets (gameconf \"%s\")", m_Address, limit, m_CurFile);
+				logger->LogError("[SM] Error parsing Address \"%s\", does not support more than %d read offsets (gameconf \"%s\")", m_Address.c_str(), limit, m_CurFile);
 			}
 		} else if (strcmp(key, "signature") == 0) {
 			m_AddressSignature = value;
--- a/core/logic/GameConfigs.cpp
+++ b/core/logic/GameConfigs.cpp
@@ -1078,7 +1078,7 @@ CGameConfig::AddressConf::AddressConf(std::string&& sigName, unsigned readCount,
 {
 	unsigned readLimit = minOf(readCount, sizeof(this->read) / sizeof(this->read[0]));
 
-	this->signatureName = std::move(sigName);
+	this->signatureName = sigName;
 	this->readCount = readLimit;
 	memcpy(&this->read[0], read, sizeof(this->read[0])*readLimit);
 
--- a/core/logic/GameConfigs.cpp
+++ b/core/logic/GameConfigs.cpp
@@ -548,8 +548,7 @@ SMCResult CGameConfig::ReadSMC_LeavingSection(const SMCStates *states)
 	case PSTATE_GAMEDEFS_OFFSETS_OFFSET:
 		{
 			/* Parse the offset... */
-			if (m_Class[0] != '\0'
-				&& m_Prop[0] != '\0')
+			if (!m_Class.empty() && !m_Prop.empty())
 			{
 				SendProp *pProp = gamehelpers->FindInSendTable(m_Class.c_str(), m_Prop.c_str());
 				if (pProp)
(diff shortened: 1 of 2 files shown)


core/logic/GameConfigs.cpp
```diff
@@ -1074,11 +1074,11 @@ static inline unsigned minOf(unsigned a, unsigned b)
 	return a <= b ? a : b;
 }
 
-CGameConfig::AddressConf::AddressConf(char *sigName, unsigned sigLength, unsigned readCount, int *read, bool lastIsOffset)
+CGameConfig::AddressConf::AddressConf(std::string sigName, unsigned sigLength, unsigned readCount, int *read, bool lastIsOffset)
```
nosoop, line 1077: `sigLength` doesn't seem to be used. Might be safe to remove?

core/logic/GameConfigs.cpp
```diff
@@ -241,7 +241,7 @@ SMCResult CGameConfig::ReadSMC_NewSection(const SMCStates *states, const char *n
 			{
 				m_ParseState = PSTATE_GAMEDEFS_KEYS;
 			}
-			else if ((strcmp(name, "#supported") == 0) && (strcmp(m_Game, "#default") == 0))
+			else if ((strcmp(name, "#supported") == 0) && (strcmp(m_Game.c_str(), "#default") == 0))
```
dvander, line 244: (m_Game == "#default")

if you prefer

core/logic/GameConfigs.cpp
```diff
@@ -551,22 +551,22 @@ SMCResult CGameConfig::ReadSMC_LeavingSection(const SMCStates *states)
...
 					int val = gamehelpers->GetSendPropOffset(pProp);
-					m_Offsets.replace(m_offset, val);
-					m_Props.replace(m_offset, pProp);
+					m_Offsets.replace(m_offset.c_str(), val);
+					m_Props.replace(m_offset.c_str(), pProp);
 				} else {
 					/* Check if it's a non-default game and no offsets exist */
-					if (((strcmp(m_Game, "*") != 0) && strcmp(m_Game, "#default") != 0)
-						&& (!m_Offsets.retrieve(m_offset)))
+					if (((strcmp(m_Game.c_str(), "*") != 0) && strcmp(m_Game.c_str(), "#default") != 0)
```
dvander, line 562: m_Game != "*" && m_Game != "#default"

core/logic/GameConfigs.cpp
```diff
@@ -692,8 +692,8 @@ SMCResult CGameConfig::ReadSMC_LeavingSection(const SMCStates *states)
 
 			if (m_Address[0] != '\0' && m_AddressSignature[0] != '\0')
 			{
-				AddressConf addrConf(m_AddressSignature, sizeof(m_AddressSignature), m_AddressReadCount, m_AddressRead, m_AddressLastIsOffset);
-				m_Addresses.replace(m_Address, addrConf);
+				AddressConf addrConf(m_AddressSignature, m_AddressSignature.length(), m_AddressReadCount, m_AddressRead, m_AddressLastIsOffset);
```
dvander, line 695: nit: prefer the canonical size() here

core/logic/GameConfigs.cpp
dvander: Yup, feel free to remove.

core/logic/GameConfigs.cpp
```diff
@@ -1074,11 +1074,11 @@ static inline unsigned minOf(unsigned a, unsigned b)
 	return a <= b ? a : b;
 }
 
-CGameConfig::AddressConf::AddressConf(char *sigName, unsigned sigLength, unsigned readCount, int *read, bool lastIsOffset)
+CGameConfig::AddressConf::AddressConf(std::string sigName, unsigned sigLength, unsigned readCount, int *read, bool lastIsOffset)
 {
 	unsigned readLimit = minOf(readCount, sizeof(this->read) / sizeof(this->read[0]));
 
-	strncopy(signatureName, sigName, sizeof(signatureName) / sizeof(signatureName[0]));
+	this->signatureName = sigName;
```
dvander, line 1081: std::move(sigName)

dvander (changes requested):
Thanks for doing this, it's great to see more C++-ification. This is fine to go in as-is, but there are a few tiny improvements to make.

core/logic/GameConfigs.cpp
nosoop: Wasn't too sure early on if `m_AddressSignature` was going to be reused later, but yeah, moving it should be fine.

core/logic/GameConfigs.cpp
```diff
@@ -551,22 +551,22 @@ SMCResult CGameConfig::ReadSMC_LeavingSection(const SMCStates *states)
 			if (m_Class[0] != '\0'
```
nosoop, line 551: Any opinion on whether to leave this as-is or switch these and other ones over to `.empty()`?

nosoop:
Tested it as far as putting it on my test TF2 instance with all sorts of gamedata and verifying that it fixes #1381; will go ahead and toggle it off draft.

core/logic/GameConfigs.cpp
```diff
@@ -1074,11 +1074,11 @@ static inline unsigned minOf(unsigned a, unsigned b)
 	return a <= b ? a : b;
 }
 
-CGameConfig::AddressConf::AddressConf(char *sigName, unsigned sigLength, unsigned readCount, int *read, bool lastIsOffset)
+CGameConfig::AddressConf::AddressConf(std::string sigName, unsigned readCount, int *read, bool lastIsOffset)
 {
 	unsigned readLimit = minOf(readCount, sizeof(this->read) / sizeof(this->read[0]));
 
-	strncopy(signatureName, sigName, sizeof(signatureName) / sizeof(signatureName[0]));
+	this->signatureName = std::move(sigName);
```
nosoop, line 1081: Wait a second; this still copies, doesn't it.  Should be a move assignment.

core/logic/GameConfigs.cpp
dvander: empty() would be a nice bonus!

core/logic/GameConfigs.cpp
```diff
@@ -1074,11 +1074,11 @@ static inline unsigned minOf(unsigned a, unsigned b)
 	return a <= b ? a : b;
 }
 
-CGameConfig::AddressConf::AddressConf(char *sigName, unsigned sigLength, unsigned readCount, int *read, bool lastIsOffset)
+CGameConfig::AddressConf::AddressConf(std::string&& sigName, unsigned readCount, int *read, bool lastIsOffset)
 {
 	unsigned readLimit = minOf(readCount, sizeof(this->read) / sizeof(this->read[0]));
 
-	strncopy(signatureName, sigName, sizeof(signatureName) / sizeof(signatureName[0]));
+	this->signatureName = sigName;
```
dvander, line 1081: You still need std::move() here to avoid a copy. "sigName" becomes an l-value, or something (I can't cite the rule book, but an explicit std::move is needed).

nosoop: Gotcha; not at all familiar with how moves work myself.

dvander (approved):

Headline (approved):
Looks great!

Thanks for taking the time to work on this

PR #1506, opened 2021-06-21
core: Fix not working MENU_DEBUG

Wend4r:
Fixed a compilation error
```
error: use of undeclared identifier 'g_Logger'
```

### DIFF
--- a/core/MenuStyle_Base.cpp
+++ b/core/MenuStyle_Base.cpp
@@ -36,7 +36,7 @@
 #include "MenuManager.h"
 #include "CellRecipientFilter.h"
 #if defined MENU_DEBUG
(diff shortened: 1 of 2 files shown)


dvander (approved):

asherkin (changes requested):
This doesn't look right, you can't use an internal header from `logic` inside `core` - the extern won't be populated at runtime as it is part of the logic binary instead.

The code needs updating to use the `logger` global provided by `logic_bridge.h` (already included here), which'll call over into the correct library.

Wend4r:
Works
image

asherkin (approved):
LGTM, thanks!

Issue #1461, opened 2021-04-03
Add SetCollisionGroup to API


xMaZax:
Hello everyone, some people know and have encountered the fact that when installing a collision, an object can fall under the map or have some other side effects.

# Problematic Code (or Steps to Reproduce)
  ```
 SetEntProp(client, Prop_Data, "m_CollisionGroup", flags);
  ```

If we use a function that directly causes a collision to be set then this problem will not occur

void CBaseEntity::SetCollisionGroup( int collisionGroup )

https://github.com/perilouswithadollarsign/cstrike15_src/blob/f82112a2388b841d72cb62ca48ab1846dfcc11c8/game/shared/baseentity_shared.cpp#L2704

I suggest adding this feature to the main sourcemod API.

I think this is because another CollisionRulesChanged() function is not called; when we expose just via SetEntProp.

I think a direct function call will be more correct than we use now " crutches"

xMaZax:
Other info bug:
https://bugs.alliedmods.net/show_bug.cgi?id=6348
https://web.archive.org/web/20190406220848/https://forum.facepunch.com/gmoddev/likl/Crazy-Physics-cause-found/1/

And even there is a plugin to try to fix this problem, I do not see the point in it when you can add it to the main API.
https://github.com/ashort96/SetCollisionGroup

ashort96:
The plugin does fix it, and yes you are spot on about the reasoning. `CollisionRulesChanged()` getting called after changing an entity's collision group does prevent the "Physics Mayhem" bug.

ashort96:
I believe this Issue can be closed as #1507 addresses this.

KyleSanderson:
Indeed, thanks for the PR.

PR #1507, opened 2021-06-21
Add SetCollisionGroup to API (#1461)

ashort96:
Addresses issue #1461 and adds SetCollisionGroup to the API. Only CS:S and TF2 signatures are added - requesting feedback of where the CS:GO signature should go.


ashort96:
Added CS:GO signature (linux only).

plugins/include/sdktools_functions.inc
```diff
@@ -349,3 +349,12 @@ native void SetClientName(int client, const char[] name);
  * @return              Amount of ammo actually given.
  */
 native int GivePlayerAmmo(int client, int amount, int ammotype, bool suppressSound=false);
+
+/**
+ * Changes a client's collision group
+ * 
+ * @param client            The client index
+ * @param collisionGroup    Collision group to use
+ * @error                   Invalid client index, or client not connected.
```
KyleSanderson, line 358: In-game, no?

extensions/sdktools/vnatives.cpp
```diff
@@ -1521,6 +1569,7 @@ sp_nativeinfo_t g_Natives[] =
 	{"SetClientInfo",			SetClientInfo},
 	{"SetClientName",           SetClientName},
 	{"GetPlayerResourceEntity", GetPlayerResourceEntity},
-	{"GivePlayerAmmo",		GivePlayerAmmo},
+	{"GivePlayerAmmo",			GivePlayerAmmo},
```
KyleSanderson, line 1572: Revert this one

extensions/sdktools/vnatives.cpp
```diff
@@ -1491,6 +1501,44 @@ static cell_t GivePlayerAmmo(IPluginContext *pContext, const cell_t *params)
 	return ammoGiven;
 }
 
+// SetCollisionGroup(int client, int collisionGroup)
+static cell_t SetCollisionGroup(IPluginContext *pContext, const cell_t *params)
+{
+	if (!g_pSetCollisionGroup)
```
KyleSanderson, line 1507: Should be a static, no?

KyleSanderson (changes requested):
I'm going to sleep on this one. It's likely the right thing to do (years late, but still the right thing) but there has to be a better way than the signatures.

extensions/sdktools/vnatives.cpp
ashort96: Reverted

plugins/include/sdktools_functions.inc
ashort96: Changed "not connected" -> "client not in game"

plugins/include/sdktools_functions.inc
```diff
@@ -349,3 +349,12 @@ native void SetClientName(int client, const char[] name);
  * @return              Amount of ammo actually given.
  */
 native int GivePlayerAmmo(int client, int amount, int ammotype, bool suppressSound=false);
+
+/**
+ * Changes a client's collision group
+ * 
+ * @param client            The client index.
```
asherkin, line 356: This function should work for any entity, no?

extensions/sdktools/vnatives.cpp
ashort96: What should be static here? I essentially copied from here: [outputnatives.cpp](https://github.com/alliedmodders/sourcemod/blob/267eb90da5027e5e7c378fd6a8494f1bc0b1231b/extensions/sdktools/outputnatives.cpp#L305-L391).

plugins/include/sdktools_functions.inc
ashort96: I believe so - the method is from the CBaseEntity class. I will change any reference of client to entity.

asherkin:
> but there has to be a better way than the signatures.

`CollisionRulesChanged` (which is the important bit of this that `SetCollisionGroup` calls internally) is not completely trivial to re-implement and neither of them are virtual - this seems like the best option. This has been in [the queue](https://bugs.alliedmods.net/show_bug.cgi?id=6348) for almost a decade.

extensions/sdktools/vnatives.cpp
KyleSanderson: The difference is this is localized to this single file, and to this function specifically. If it's used as an interface or something changes then it demands a global scope. For now, you can reduce the scope of this to the function, by using static.

KyleSanderson:
> > but there has to be a better way than the signatures.
> 
> `CollisionRulesChanged` (which is the important bit of this that `SetCollisionGroup` calls internally) is not completely trivial to re-implement and neither of them are virtual - this seems like the best option. This has been in [the queue](https://bugs.alliedmods.net/show_bug.cgi?id=6348) for almost a decade.

There's a better option, which is `PhysEnableEntityCollisions` / `PhysDisableEntityCollisions` - no? Ultimately that's what people actually want out of this, changing collision groups themselves no one particularly cares about. I could be misunderstanding this function though but if we're sigscanning already perhaps it isn't optimized out.

ashort96:
> > > but there has to be a better way than the signatures.
> > 
> > 
> > `CollisionRulesChanged` (which is the important bit of this that `SetCollisionGroup` calls internally) is not completely trivial to re-implement and neither of them are virtual - this seems like the best option. This has been in [the queue](https://bugs.alliedmods.net/show_bug.cgi?id=6348) for almost a decade.
> 
> There's a better option, which is `PhysEnableEntityCollisions` / `PhysDisableEntityCollisions` - no? Ultimately that's what people actually want out of this, changing collision groups themselves no one particularly cares about. I could be misunderstanding this function though but if we're sigscanning already perhaps it isn't optimized out.

The most common use of changing an entity's collision group I've seen is setting the collision group to to either `COLLISION_GROUP_DEBRIS_TRIGGER` (noblock) or `COLLISION_GROUP_PLAYER` (block), although I've seen people in the AM Discord mention using other collision groups, such as `COLLISION_GROUP_WEAPON`. I took a quick peak at the two methods you mentioned, and it doesn't look like it gives the ability to specify the collision group you want - I didn't the `RemoveObjectPair()` method, so maybe there was something more in there I missed. I do think that while the majority will probably only use the first two collision groups for players, there might be some other uses cases that could be supported by adding this.

asherkin (approved):
Nice, this all looks good to me.

KyleSanderson (approved):

Einyux:
Maybe too late, but shouldn't the native named for the sake of clarity something like "Set**Entity**CollisionGroup" (To be consistent with the other entity natives) ?

psychonic:
> Maybe too late, but shouldn't the native named for the sake of clarity something like "SetEntityCollisionGroup" (To be consistent with the other entity natives) ?

I agree.

ashort96:
The native's first param is `entity`, but being consistent with other natives makes sense - I can make a follow on PR. Should this [enum](https://github.com/ashort96/SetCollisionGroup/blob/master/addons/sourcemod/scripting/includes/SetCollisionGroup.inc) also be added? That would make a transition from anyone using my plugin seamless (just delete `#include <SetCollisionGroup>` and remove the running plugin).

Kenzzer:
Another late comment on this PR. But now that `SetCollisionGroup` is added, odds are people will want `SetOwnerEntity` exposed as it also calls `CollisionRulesChanged` to fix some physics issues.

Much agreeing with @KyleSanderson , depending on a signature seems a bit too much, especially if this native is going to be wanted on more games than just TF2/CS.

Therefore I suggest the following alternative:
`CBaseEntity::SetOwnerEntity` is virtual, and its vtable offset is quite low (Linux:18 Win:17 on TF2, L4D2, HL2:DM, Insurgency, Days Of Defeat, and CSS, I'm willing to bet it's also the same on CSGO, unfortunately I lack binaries before the symbol strip to confirm this).

The function is implemented as follow:
```
void CBaseEntity::SetOwnerEntity( CBaseEntity* pOwner )
{
	if ( m_hOwnerEntity.Get() != pOwner )
	{
		m_hOwnerEntity = pOwner;

		CollisionRulesChanged();
	}
}
```
Considering `m_hOwnerEntity` is a netprop, to obtain a dirty access to `CollisionRulesChanged`. One simple way would be to retrieve whatever entity is inside  `m_hOwnerEntity`.
If the entity is null, we force set `m_hOwnerEntity` to some random entity (worldspawn/first entity)
If the retrieved entity is not null, we force set `m_hOwnerEntity` to null.
Then we call the function, with as parameter whatever was previously in `m_hOwnerEntity`.

This effectively only calls `CollisionRulesChanged` which is what we want.

With that dirty hack done, we can then reimplement `SetCollisionGroup` ourselves, by changing the netprop collision group netprop value and then calling `SetOwnerEntity` with the twist I just mentioned.
The added bonus, is we can also offer plugins a new native to call `CollisionRulesChanged` and `SetOwnerEntity`.

If I wasn't clear in my explanations, I could always make a PR.

Extra note: it is possible that this virtual function gets overriden, although I'm not aware of any classes that do this. But should it be overridden, we could parse the vtable on `worldspawn` entity to retrieve `CBaseEntity::SetOwnerEntity` as it isn't overidden there.

Einyux:
> The native's first param is `entity`, but being consistent with other natives makes sense - I can make a follow on PR. Should this [enum](https://github.com/ashort96/SetCollisionGroup/blob/master/addons/sourcemod/scripting/includes/SetCollisionGroup.inc) also be added? That would make a transition from anyone using my plugin seamless (just delete `#include <SetCollisionGroup>` and remove the running plugin).

About the enum `Collision_Group`, I don't know if all games supported by SM have the exact same list (At least PORTAL2 doesn't, but it isn't supported so I guess it's ok). If the enum is added, it would be nice to have the `GetEntityCollisionGroup` native to avoid a view_as<> cast everytime.

ashort96:
It is very well possible that things like `CBaseEntity::SetMoveType()` could break physics as well - maybe it would just be more beneficial to do what @Kenzzer suggested with `SetOwnerEntity()` OR change the signature dependency to rely on `CollisionRulesChanges()` - the method is very easy to find in the binary (easily searchable string), so any updates that might break it should be easily fixable.

Anyone who might call something that would break physics would just have to do something like this
```cpp
SetEntProp(client, Prop_Data, "m_CollisionGroup", COLLISION_GROUP_DEBRIS_TRIGGER, 4);
CollisionRulesChanged();
```

Einyux:
Same issue for `CBaseEntity::SetSolid` & `CBaseEntity::SetSolidFlags`.

Kenzzer:
Not necessarily applying my method. Only the vtable offset (not signature) for `CBaseEntity::SetOwnerEntity` would need to be maintained. And the signature for `CBaseEntity::SetCollisionGroup` no longer required.

Secondly, if we need a better method for SetMoveType than the current SM's implementation, we can fallback on the IServerTools interface. Which is already available through Sourcemod.
`https://github.com/ValveSoftware/source-sdk-2013/blob/0d8dceea4310fde5706b3ce1c70609d72a38efdf/sp/src/game/server/entity_tools_server.cpp#L308`
`CServerTools::SetMoveType` calls `CBaseEntity::SetMoveType`

So we would need to maintain 2 signatures (not 5)
`CBaseEntity::SetSolid`
`CBaseEntity::SetSolidFlags`
And 1 vtable offset
`CBaseEntity::SetOwnerEntity`

`CBaseEntity::SetCollisionGroup` & `CBaseEntity::SetMoveType` can be remade in code. We don't need to maintain signatures for those.

Finally, I don't know if plugins really need access to `CBaseEntity::SetSolidFlags` & `CBaseEntity::SetSolid`, do you have examples where it could come handy?

Einyux:
No, I was just tracking where CollisionRulesChanged() was called, and found SetSolid & SetSolidFlags. Don't know if issues have been detected with these ones, but looking at the last topic comment (https://forums.alliedmods.net/showpost.php?p=2247647&postcount=173), it looks like there's something with SetSolid.

KyleSanderson:
@Kenzzer if you're up for making a PR for the virtual solution I'm up for reviewing and/or co-authoring to help it get in. I agree that's the much better way, and we already do this with other functions by grabbing the function from 0 so there's precedence.

IServerTools is not available on all games, so while it's handy for new stuff, old games will still need a classic fallback.

Kenzzer:
Alright, in that case I'll open a PR in the next couple of days!

PR #1509, opened 2021-06-22
Update Team Fortress 2 gamedata

Mikusch:
TF2 update on 22/06/2021 broke a lot of SM gamedata

Thanks to @Scags for looking up the signature for `MakeBleed`


asherkin (approved):
Thanks!

PR #1511, opened 2021-06-25
Add StringToInt64() and Int64ToString() natives

Wend4r:


Wend4r:
I need this change to convert a big int array to a string value, working with unix bigstamp, SteamID64, EconItem itemid and etc.. 
For compatibility with these natives:
https://sourcemod.dev/#/sourcemod/function.GetTime
https://github.com/alliedmodders/sourcemod/blob/master/plugins/include/clients.inc#L320 by AuthId_SteamID64
https://github.com/komashchenko/PTaH/blob/master/PTaH.inc#L510
https://github.com/alliedmodders/sourcemod/blob/master/plugins/include/protobuf.inc#L56
https://github.com/alliedmodders/sourcemod/blob/master/plugins/include/protobuf.inc#L144
https://github.com/alliedmodders/sourcemod/blob/master/plugins/include/protobuf.inc#L214
https://github.com/ErikMinekus/sm-ripext/blob/main/pawn/scripting/include/ripext/json.inc#L116
https://github.com/ErikMinekus/sm-ripext/blob/main/pawn/scripting/include/ripext/json.inc#L172
...

asherkin:
Nice! This is a good interim solution and something I've been intending to add a format specifier for doing for ages.

Could you add some test cases to the test suite? (Or share a test plugin and test results here.)

peace-maker:
`BigInt` usually stands for some way to store any large number not restricted to 64 bits. Using Int64 might be more descriptive.

plugins/include/string.inc
```diff
@@ -204,6 +214,16 @@ native int StringToIntEx(const char[] str, int &result, int nBase=10);
...
 
+/**
+ * Converts a big integer to a string.
+ *
+ * @param num           Big integer to convert.
+ * @param str           Buffer to store string in.
+ * @param maxlength     Maximum length of string buffer.
+ * @return              Number of cells written to buffer.
+ */
+native int BigIntToString(const int num[2], char[] str, int maxlength);
```
dvander, line 225: "Int64" instead of "BigInt"

plugins/include/string.inc
```diff
@@ -204,6 +214,16 @@ native int StringToIntEx(const char[] str, int &result, int nBase=10);
  */
 native int IntToString(int num, char[] str, int maxlength);
 
+/**
+ * Converts a big integer to a string.
+ *
+ * @param num           Big integer to convert.
+ * @param str           Buffer to store string in.
+ * @param maxlength     Maximum length of string buffer.
+ * @return              Number of cells written to buffer.
```
dvander, line 223: The return comment should be:

"Number of characters written to the buffer, not including the null terminator"

plugins/include/string.inc
```diff
@@ -204,6 +214,16 @@ native int StringToIntEx(const char[] str, int &result, int nBase=10);
  */
 native int IntToString(int num, char[] str, int maxlength);
 
+/**
+ * Converts a big integer to a string.
+ *
+ * @param num           Big integer to convert.
```
dvander, line 220: "Array containing the upper and lower 32-bits of a 64-bit integer"

plugins/include/string.inc
```diff
@@ -194,6 +194,16 @@ native int StringToInt(const char[] str, int nBase=10);
  */
 native int StringToIntEx(const char[] str, int &result, int nBase=10);
 
+/**
+ * Converts a string to a big integer.
+ *
+ * @param str           String to convert.
+ * @param result        Variable to store the result in.
```
dvander, line 201: "Array to store the upper and lower 32-bits of the 64-bit integer."

dvander (changes requested):

dvander (approved):

asherkin:
I wonder if `Int64ToString` should be a format specifier <instead / as well>? I imagine most of the use cases are going to be embedding one in a larger string such as a URL or SQL query.

dvander:
I'm ambivalent. Using int64 ids as anything other than strings is a bad idea unless there's some mathematical calculation going on. As numbers they are not portable to other languages (eg JavaScript/JSON).

asherkin (approved):

PR #1512, opened 2021-06-26
Increase buffer size for map name & DisplayName

xerox8521:
This increases the buffer to store the map name to be consistent with the other calls to GetCurrentMap. Also `currentmap` now uses the map's display name similar how `nextmap` uses it.

### DIFF
--- a/plugins/basetriggers.sp
+++ b/plugins/basetriggers.sp
@@ -272,9 +272,9 @@ public void OnClientSayCommand_Post(int client, const char[] command, const char
 	}
 	else if (strcmp(sArgs, "currentmap", false) == 0)
 	{
-		char map[64];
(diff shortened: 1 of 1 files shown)


asherkin (approved):
Thanks!

dragokas:
What now?
After this commit I'll receive an empty output instead of map name for some maps.

asherkin:
> After this commit I'll receive an empty output instead of map name for some maps.

This PR is just aligning one chat trigger with the behaviour of everything else displaying maps, if you're having an issue generally you should open an issue with some details, especially an example of a map with that problem. `GetMapDisplayName` explicitly uses the input string on failure.

dragokas:
Ok, thanks.
And strange, the issue is no more reproduced.

Issue #1520, opened 2021-07-02
admin-sql-threaded doesn't support Steam64 unlike admin-sql-prefetch


tooolz:
Thanks to asherkin for solving this problem when it occurred.

It turns out admin-sql-threaded.smx doesn't handle SteamIDs similarly to admin-sql-prefetch.smx which breaks compatibility with Steam64 ID's (i.e. 76561197974481263) and so only works with Steam2 ID's (i.e. STEAM_0:1:7107767).

This makes the core difference whether a players' authority is loaded or not.

PR #1521, opened 2021-07-03
AuthID unified converter as native

CrazyHackGUT:
#1520 reminds us of the internal function of SourceMod to convert SteamID in any format - to unified one (part of the account ID from v2):
```
STEAM_0:0:55665612 -> 0:55665612
STEAM_1:0:55665612 -> 0:55665612
[U:1:111331224] -> 0:55665612
76561198071596952 -> 0:55665612
```
This PR just moves this method to `PlayerManager` class, exposes to `IPlayerManager` interface (with interface version bump), implements native for scripters and adopts AdminCache and SQL admin loaders with manager to use this method.

### DIFF
--- a/core/PlayerManager.cpp
+++ b/core/PlayerManager.cpp
@@ -1521,6 +1521,61 @@ void PlayerManager::RecheckAnyAdmins()
 	}
 }
 
+bool PlayerManager::GetUnifiedAuthId(const char *ident, char *out, size_t maxlen)
+{
+	int len = strlen(ident);
+	if (!strcmp(ident, "BOT"))
+	{
+		// Bots
+		strncpy(out, ident, maxlen);
+		return true;
+	}
+	else if (len >= 11 && !strncmp(ident, "STEAM_", 6) && ident[8] != '_')
+	{
+		// non-bot/lan Steam2 Id, strip off the STEAM_* part
+		ke::SafeStrcpy(out, maxlen, &ident[8]);
+		return true;
+	}
+	else if (len >= 7 && !strncmp(ident, "[U:", 3) && ident[len-1] == ']')
+	{
+		// Steam3 Id, replicate the Steam2 Post-"STEAM_" part
+		uint32_t accountId = strtoul(&ident[5], nullptr, 10);
+		ke::SafeSprintf(out, maxlen, "%u:%u", accountId & 1, accountId >> 1);
+		return true;
+	}
+	else
+	{
+		// 64-bit CSteamID, replicate the Steam2 Post-"STEAM_" part
+
+		// some constants from steamclientpublic.h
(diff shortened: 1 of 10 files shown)


Wend4r:
I would like to have a mechanism for casting to some unified integer type (AccountID or SteamID64) for working with databases, searching by integers will be much faster.

asherkin:
I don't think this helps to solve #1520 as the problem is with the query, `admin-sql-prefetch` already works correctly because it just loads everything and goes though the internal SM conversion logic.

CrazyHackGUT:
This can **help solve** #1520 with simple conversion all Auth IDs to unified format. And also provide to custom plugins API for converting SteamID in any format - to "unified".
If you disagree with me - i can just drop branch and close PR.

I agree with @Wend4r and use SteamID integer representation (what can be retrieved with `GetSteamAccountID`), but this is not usable for all cases.

TheByKotik:
Suggested format (`0:55665612`) contain absolute identical information to Account ID. But in string format.
We really want to use char[] instead int32?
I think this native will be very messy.

CrazyHackGUT:
SourcePawn don't supports unsigned integers. In one day, we can get more error reports because this function can return negative value.

Wend4r:
> SourcePawn don't supports unsigned integers. In one day, we can get more error reports because this function can return negative value.

https://github.com/alliedmodders/sourcemod/blob/18f9d65938c6eda7e4c51df11c017ee11ffdbe16/core/logic/sprintf.cpp#L747-L762
Supports for output via formatting

TheByKotik:
> SourcePawn don't supports unsigned integers. In one day, we can get more error reports because this function can return negative value.

Have no difference signed or unsigned we use for storing.
Account ID no requiring `<` or `>` operators.
But, if you're really requiring these operators, you can use my defines.
```sourcepawn
#define uint_operator_greater(%1,%2) ((%1 & (1 << 31)) ? (%2 & (1 << 31)) ? (%1 < %2) : true  : (%2 & (1 << 31)) ? false : (%1 > %2))
#define uint_operator_less(%1,%2)    ((%1 & (1 << 31)) ? (%2 & (1 << 31)) ? (%1 < %2) : false : (%2 & (1 << 31)) ? true :  (%1 < %2))
```

Also your are absolutely right, SourcePawn really needs support for unsigned integers and 64-bit variables.
However, this does not change the fact that the proposed native is not good idea.
It will create even more confusion and SteamID variations.

peace-maker:
Thank you for the PR, but as can be seen from the discussion in this thread, adding another standard to unify all standards isn't the right solution. Using the accountid directly and storing it as an INT in the databases would be the sane way to handle this if you're writing a new plugin now instead of using a string representation. admin-sql-threaded was fixed by updating the query, so I'm closing this.

PR #1523, opened 2021-07-04
Add parameter to configure updating page permissions in StoreToAddress

### LINKED ISSUE #1375: Add StoreToAddress() alternative for data manipulation.
# Description
Feature request: StoreToAddress() for data manipulation as a raw data in memory. Currently StoreToAddress() is ``mainly designed for code patching`` as I was told by @asherkin in discord, and it has poor performance when used extensively (pretty slow and unusable in this cases). The main use case of the extensive usage is when you need to work with data structures in memory directly via raw data pointers received from dhook detour or some other way and when the code is called very frequently (i.e. once per tick). 

Can't really provide small code snippet as an example, but that was a problem in [here](https://github.com/GAMMACASE/MomSurfFix/blob/master/addons/sourcemod/scripting/momsurffix/gametrace.sp#L248-L271), there's methodmaps that are wrappers for real c structures, and StoreToAddressCustom() there is used as an alternative to default StoreToAddress() that in fact just an [asm function](https://github.com/GAMMACASE/MomSurfFix/blob/master/addons/sourcemod/scripting/momsurffix2.sp#L195-L235). That resulted in speed increase up to ~45 times compared to default StoreToAddress(). The comparison was done using default sm profiler.

nosoop:
Closes #1375.

This adds a new optional parameter to `StoreToAddress`, allowing plugins to opt out of having SourceMod update page permissions every time it writes to memory.

Tested in TF2 with the following micro benchmark:

<details>
<summary>Micro benchmark source code:</summary>

```
#include <sourcemod>

#pragma semicolon 1
#pragma newdecls required

#include <profiler>

public void OnPluginStart() {
	RegAdminCmd("test_memwrite", MemWrite, ADMFLAG_ROOT);
}

#define NUM_ITERATIONS (1 << 19)

public Action MemWrite(int client, int argc) {
	if (!client) {
		return Plugin_Handled;
	}
	
	int health = GetEntProp(client, Prop_Send, "m_iHealth");
	int healthOffs = FindSendPropInfo("CTFPlayer", "m_iHealth");
	
	Address pHealthOffs = view_as<Address>(view_as<int>(GetEntityAddress(client)) + healthOffs);
	
	Profiler prof = new Profiler();
	
	prof.Start();
	for (int i; i < NUM_ITERATIONS; i++) {
		SetEntData(client, healthOffs, health);
	}
	prof.Stop();
	ReplyToCommand(client, "SetEntData: %f", prof.Time);
	
	prof.Start();
	for (int i; i < NUM_ITERATIONS; i++) {
		StoreToAddress(pHealthOffs, health, NumberType_Int32);
	}
	prof.Stop();
	ReplyToCommand(client, "StoreToAddress(): %f", prof.Time);
	
	prof.Start();
	for (int i; i < NUM_ITERATIONS; i++) {
		StoreToAddress(pHealthOffs, health, NumberType_Int32, false);
	}
	prof.Stop();
	ReplyToCommand(client, "StoreToAddress(NoPageMod): %f", prof.Time);
	
	delete prof;
	
	return Plugin_Handled;
}
```
</details>

Micro benchmark results:

```
SetEntData: 0.044856
StoreToAddress(): 0.372449
StoreToAddress(NoPageMod): 0.009297
```

### DIFF
--- a/core/logic/smn_core.cpp
+++ b/core/logic/smn_core.cpp
@@ -747,7 +747,7 @@ static cell_t StoreToAddress(IPluginContext *pContext, const cell_t *params)
 
 	NumberType size = static_cast<NumberType>(params[3]);
 
-	// new parameter added after SM 1.10
+	// new parameter added after SM 1.10; defaults to true for backwards compatibility
 	bool updateMemAccess = true;
 	if (params[0] >= 4)
 	{
--- a/core/logic/smn_core.cpp
+++ b/core/logic/smn_core.cpp
@@ -747,18 +747,34 @@ static cell_t StoreToAddress(IPluginContext *pContext, const cell_t *params)
 
 	NumberType size = static_cast<NumberType>(params[3]);
 
+	// new parameter added after SM 1.10
+	bool updateMemAccess = true;
+	if (params[0] >= 4)
+	{
+		updateMemAccess = params[4];
+	}
+
 	switch(size)
 	{
 	case NumberType_Int8:
-		SourceHook::SetMemAccess(addr, sizeof(uint8_t), SH_MEM_READ|SH_MEM_WRITE|SH_MEM_EXEC);
+		if (updateMemAccess)
+		{
+			SourceHook::SetMemAccess(addr, sizeof(uint8_t), SH_MEM_READ|SH_MEM_WRITE|SH_MEM_EXEC);
+		}
 		*reinterpret_cast<uint8_t*>(addr) = data;
 		break;
 	case NumberType_Int16:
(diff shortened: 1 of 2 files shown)


Wend4r:
If speed is so important to you, then I suggest that you the write of a cell in memory without access changes in a separate native.

Also, I had the idea to make a additional `methodmap Address`, where there would be a `Address.SetAccess()` and etc.

KyleSanderson (approved):

KyleSanderson:
The use-case is pretty terrible and is prone to corrupting state without extreme safety taken. However, writing memory should not be crazy more expensive than EntData which is ever so slightly more safe, so I've taken this in.

Please do not use this, or EntData, and use EntProp whenever possible when manipulating networked entity data.

PR #1527, opened 2021-07-12
Add Event.GetDataTypes() method

Wend4r:
This pull request is necessary for the possibility of iterating the parameters of the event, if the task is to log events in the SourcePawn environment and not update the valid parameters from the Wiki every time.

### DIFF
--- a/core/smn_events.cpp
+++ b/core/smn_events.cpp
@@ -35,7 +35,7 @@
 #include "PlayerManager.h"
 #include "logic_bridge.h"
 
-#if SOURCE_ENGINE >= SE_ORANGEBOX
+#if SOURCE_ENGINE == SE_BLADE || SOURCE_ENGINE == SE_CSGO
 #include "smn_keyvalues.h"
 #endif
 
--- a/core/smn_events.cpp
+++ b/core/smn_events.cpp
@@ -367,6 +367,8 @@ static cell_t sm_GetEventDataTypes(IPluginContext *pContext, const cell_t *param
 	KeyValueStack *pStk = new KeyValueStack;
 	pStk->pBase = new KeyValues(pEvDataTypes->GetName());
 	*(pStk->pBase) = *pEvDataTypes;
+	pStk->pCurRoot.push(pStk->pBase);
+
 	return handlesys->CreateHandle(g_KeyValueType, pStk, pContext->GetIdentity(), g_pCoreIdent, NULL);
 #else
 	return pContext->ThrowNativeError("Event.GetDataTypes is not supported on this game.");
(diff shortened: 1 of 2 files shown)


Wend4r:
Example code
[test_event_data_types.zip](https://github.com/alliedmodders/sourcemod/files/6803108/test_event_data_types.zip)

Output
```
L 07/12/2021 - 17:22:58: [test_event_data_types.smx] Event "round_start"
Data types KeyValues:
"descriptor"
{
        "timelimit"             "3"
        "fraglimit"             "3"
        "objective"             "1"
}

L 07/12/2021 - 17:22:58: [test_event_data_types.smx] Paramenter "timelimit" is equal "999" (long)
L 07/12/2021 - 17:22:58: [test_event_data_types.smx] Paramenter "fraglimit" is equal "0" (long)
L 07/12/2021 - 17:22:58: [test_event_data_types.smx] Paramenter "objective" is equal "BOMB TARGET" (string)
```

eyal282:
I want to add that if you must replicate an event with different parameters, this beats the updates of CSGO

asherkin:
Have you considered parsing `modevents.res` / `gameevents.res` / `serverevents.res` directly? It seems that is all this API does under the hood - Source has always had these event descriptor files (and they're what is used to update the wiki), it looks like CS:GO has just now made them authoritative rather than advisory.

The KV structure is very simple and almost exactly mirrors the API this PR is exposing (just with readable string values rather than opaque integer identifiers), and a plugin providing a wrapping API around the res files would work for almost all Source games (just with not-guaranteed accuracy for non-CS:GO/Blade), so I think that is probably a much better path forward rather than adding this new API to SM.

https://github.com/SteamDatabase/GameTracking-CSGO/blob/ac9963b29e7d33adc190b36648f208a0b33cf20a/csgo/pak01_dir/resource/modevents.res#L25-L46

Wend4r:
> Have you considered parsing `modevents.res` / `gameevents.res` / `serverevents.res` directly?

Yes, but I thought that it would be more convenient to use `methodmap Event`. When reading `.res` files with events, will have to search for the desired event in several KeyValues at once, and it works slowly in contrast to that game already knows in the `CGameEventDescriptor` which structures with types belong to a certain event.

asherkin:
Thanks for the PR but I'm going to close it based on the reasoning in my previous comment. You should be able to avoid any perf impact by pre-parsing all the relevant res files (which is just what the game does) in a central plugin and offering an API (which I look forward to seeing released on the forum!)

Wend4r:
image
I will restore the branch, because in the end it turned out that the events in these files may matches. The order of loading event files is called via `CGameEventManager`.

PR #1528, opened 2021-07-16
Set macro on __pl_mapchooser_SetNTVOptional

Kailo97:
Macro `#if !defined REQUIRE_PLUGIN` on __pl_mapchooser_SetNTVOptional was missied in mapchooser.inc

### DIFF
--- a/plugins/include/mapchooser.inc
+++ b/plugins/include/mapchooser.inc
@@ -135,6 +135,8 @@ forward void OnNominationRemoved(const char[] map, int owner);
  */
 forward void OnMapVoteStarted();
 
+/* DO NOT EDIT BELOW THIS LINE */
+
 public SharedPlugin __pl_mapchooser =
 {
 	name = "mapchooser",
@@ -146,6 +148,7 @@ public SharedPlugin __pl_mapchooser =
 #endif
 };
 
+#if !defined REQUIRE_PLUGIN
 public void __pl_mapchooser_SetNTVOptional()
 {
 	MarkNativeAsOptional("NominateMap");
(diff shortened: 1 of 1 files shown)


asherkin:
What's the issue this is solving? Looking at the code it appears that SM will only call the `__pl_*_SetNTVOptional` function if `required = 0` in the linkage info struct.

Kailo97:
1) Code rules, identity.
Manual recommend such and all other plugins and extensions follow it.
https://wiki.alliedmods.net/Optional_Requirements_(SourceMod_Scripting)#Creating_a_Dependency
https://github.com/alliedmodders/sourcemod/blob/master/plugins/include/adminmenu.inc#L145-L152
https://github.com/alliedmodders/sourcemod/blob/master/plugins/include/basecomm.inc#L101-L109
https://github.com/alliedmodders/sourcemod/blob/master/plugins/include/clientprefs.inc#L342-L368
https://github.com/alliedmodders/sourcemod/blob/master/plugins/include/cstrike.inc#L460-L486
https://github.com/alliedmodders/sourcemod/blob/master/plugins/include/geoip.inc#L224-L241
https://github.com/alliedmodders/sourcemod/blob/master/plugins/include/regex.inc#L277-L291
https://github.com/alliedmodders/sourcemod/blob/master/plugins/include/tf2.inc#L481-L500
https://github.com/alliedmodders/sourcemod/blob/master/plugins/include/topmenus.inc#L422-L446
2) Unused code.
So, as REQUIRE_PLUGIN force `required = 0`, `__pl_*_SetNTVOptional` will never called. There is no point in adding.

peace-maker (approved):

peace-maker:
Having all stock includes follow the same coding guidelines is good, good catch.

PR #1531, opened 2021-07-16
Fix broken libraries with the SteamWorks SDK (CS:GO)

Wend4r:
These changes fix the malfunction of libraries working with the SteamWorks SDK. For example, https://github.com/KyleSanderson/SteamWorks

After changes #614

Co-Authored-By: @komashchenko

### DIFF
--- a/extensions/cstrike/rulesfix.cpp
+++ b/extensions/cstrike/rulesfix.cpp
@@ -102,12 +102,12 @@ void RulesFix::OnLoad()
 
 		if(SteamAPI_SteamGameServer == nullptr)
 		{
-			g_pSM->LogError(myself, "[CStrike] Failed to get \"%s\" function", pSteamGameServerFuncName);
+			g_pSM->LogError(myself, "[CStrike] Failed to get %s function", pSteamGameServerFuncName);
 		}
 
 		if(SteamAPI_ISteamGameServer_SetKeyValue == nullptr)
 		{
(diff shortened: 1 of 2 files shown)


DaLLuZZ:
Yes, it fixes some issues with steamworks ext

Wend4r:
For Windows, `hl2sdk-csgo\lib\public\steam_api.lib` does not contain links on
```
__imp__SteamAPI_SteamGameServer_v013
__imp__SteamAPI_ISteamGameServer_SetKeyValue
```

I'll go take a look at `hl2sdk-csgo`

asherkin:
What is the issue with SteamWorks.ext and the existing code?

DaLLuZZ:
> What is the issue with SteamWorks.ext and the existing code?

`FixSteam()` doesn't return correct `g_pSteamClientGameServer`

asherkin:
So the code in SM doesn't work when SteamWorks.ext is loaded? Or does the code in SM break SteamWorks.ext?

The bug report template is super useful for this reason.

DaLLuZZ:
The code in SM break https://github.com/KyleSanderson/SteamWorks/blob/e51d2be1ffb0a0064afa6182b0bfea34c94cdc0e/Extension/swgameserver.cpp#L62

```cpp
if (g_pSteamClientGameServer != NULL)
		return g_pSteamClientGameServer;
```
`g_pSteamClientGameServer` is not null cause of SM code in FixSteam(), but it is also incorrect

asherkin (approved):
Thanks!

PR #1534, opened 2021-07-17
Remove entity lump manipulation from OnLevelInit

### LINKED ISSUE #1470: OnLevelInit mapEntities / g_szMapEntities buffer too small and crashes for with some maps
# Help us help you

# Environment
  * Operating System version: 
  * Game/AppID (with version if applicable):
  * Current SourceMod version:
  * Current SourceMod snapshot: 
  * Current Metamod: Source snapshot: 
  - [ ] I have updated SourceMod to the [latest version](https://www.sourcemod.net/downloads.php) and it still happens.
  - [ ] I have updated SourceMod to the [latest snapshot](https://www.sourcemod.net/downloads.php?branch=dev) and it still happens.
  - [ ] I have updated SourceMM to the [latest snapshot](https://sourcemm.net/downloads.php?branch=dev) and it still happens.

# Description
[`char g_szMapEntities[2097152];`](https://github.com/alliedmodders/sourcemod/blob/b14c18ee64fc822dd6b0f5baea87226d59707d5a/extensions/sdkhooks/extension.cpp#L129) (2 * 1024 * 1024) is too small for some maps and crashes (from [`strcpy(g_szMapEntities, pMapEntities);`](https://github.com/alliedmodders/sourcemod/blob/2b1cc433558896a88c0461d0123f18c6ad7acf16/extensions/sdkhooks/extension.cpp#L945) in `SDKHooks::Hook_LevelInit` probably)

Try out [bhop_miku_v2](https://gamebanana.com/maps/109298) or [bhop_miku_csgo](https://steamcommunity.com/sharedfiles/filedetails/?id=1713703801). These maps have about ~16 MiB of mapEntities
```
public Action OnLevelInit(const char[] mapName, char mapEntities[2097152])
{
	return Plugin_Continue;
}
```

# Logs
* Please attach in separate files: game output, library logs, kernel logs, and any other supporting information.
* In case of a crash, please attach minidump or dump analyze output.

asherkin:
Newer Source engine versions now use a dynamically allocated buffer for
the map entity lump, and some maps have over 16MB of entity data - far
larger than our 2MB limit.

There is no sane way we can currently handle this, so just remove the
functionality from the forward until a more comprehensive API can be
designed.

The change in behaviour to the OnLevelInit forward params isn't obvious
when compiling a plugin, deprecate it to make it a lot more obvious that
something has changed.

Some plugins rely just on the timing of OnLevelInit rather than doing
anything with the entity lump, for these plugins offer a new OnMapInit
forward that is implemented in core rather than sdkhooks. If / when we
offer a new entity lump manipulation API in the future this'll be the
forward where it can be used to make changes.

Fixes #1470

### DIFF
--- a/core/sourcemod.cpp
+++ b/core/sourcemod.cpp
@@ -63,7 +63,8 @@ ISourcePawnEngine *g_pSourcePawn = NULL;
 ISourcePawnEngine2 *g_pSourcePawn2 = NULL;
 ISourcePawnEnvironment *g_pPawnEnv = NULL;
 IdentityToken_t *g_pCoreIdent = NULL;
-IForward *g_pOnMapEnd = NULL;
+IForward *g_pOnMapInit = nullptr;
+IForward *g_pOnMapEnd = nullptr;
 IGameConfig *g_pGameConf = NULL;
 bool g_Loaded = false;
 bool sm_show_debug_spew = false;
@@ -403,12 +404,20 @@ bool SourceModBase::LevelInit(char const *pMapName, char const *pMapEntities, ch
 		pBase = pBase->m_pGlobalClassNext;
 	}
 
+	if (!g_pOnMapInit)
+	{
+		g_pOnMapInit = forwardsys->CreateForward("OnMapInit", ET_Ignore, 1, NULL, Param_String);
+	}
+
 	if (!g_pOnMapEnd)
 	{
 		g_pOnMapEnd = forwardsys->CreateForward("OnMapEnd", ET_Ignore, 0, NULL);
 	}
 
 	g_LevelEndBarrier = true;
+
+	g_pOnMapInit->PushString(pMapName);
+	g_pOnMapInit->Execute();
 
 	RETURN_META_VALUE(MRES_IGNORED, true);
 }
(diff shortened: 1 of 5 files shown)


Headline (approved):
Definitely the right direction regarding this. 

Would the future new API be an extension of this new forward or would be be through a different avenue? Any fcompat we can do now to anticipate that coming new API?

dvander (approved):
This does seem the right way to go, unfortunately.

asherkin:
> Would the future new API be an extension of this new forward or would be be through a different avenue? Any fcompat we can do now to anticipate that coming new API?

I'm imagining some new natives where the read-only ones can be used inside and after this forward, and the write ones can be used only inside this forward - no changes to the actual forward itself.

lazyuselessman:
Hi.
What do you guys think about creation another function OnLevelInit2 with next set of parameters
```
public Action OnLevelInit2(const char[] mapName, ArrayList mapEntities)
```
each element of mapEntities will have type char, so you can set up such ArrayList on higher level and pass it like this. And `Plugin_Changed` will left same.
Also it will be nice if each individual element wasn't splitted in the middle, i mean it should be that last element of individual will be `}` and the first one `{`.


So idea is something like pack big data in ArrayList and on return if data was changed unpack it back.

dvander:
Are entities still passed in a key/value map? I feel like a specialized set of natives for interacting with the entity lump would be ideal. Such a big list/string is unworkable in Pawn.

lazyuselessman:
> Are entities still passed in a key/value map? I feel like a specialized set of natives for interacting with the entity lump would be ideal. Such a big list/string is unworkable in Pawn.

If it will be splitted as describe then yes it won't damage key/value map. It will be just few key/value map / strings, whatever u called it.

PR #1535, opened 2021-07-17
Handle detour patches across page boundaries

### LINKED ISSUE #984: ProtectMemory called with wrong address in public/CDetour/detourhelpers.h
# Help us help you

# Environment
(not applicable)

# Description
This line in detourhelpers.h:
[`ProtectMemory(address, 20, PAGE_EXECUTE_READWRITE);`](https://github.com/alliedmodders/sourcemod/blob/ae00144e84263c934d4df9b3cf0c10204e1a18a3/public/CDetour/detourhelpers.h#L121)
should be changed to this:
`ProtectMemory((void *)addr, 20, PAGE_EXECUTE_READWRITE);`
but this may be even more correct:
`ProtectMemory((void *)addr, patch->bytes, PAGE_EXECUTE_READWRITE);`

_UPD: ProtectMemory also may have bug, see comment._

# Explanation
A friend is coding own mods for his server, and found what appears to be a bug. The first proposed change above has stopped certain segfaults from occurring.

The second proposed change appears more correct to me from simple inspection of the code.

His explanation:

> on mine I've changed it to this:
> 	ProtectMemory((void *)addr, 20, PAGE_EXECUTE_READWRITE);
> and the segfaults went away
> I mean
> it looks like Sourcemod is using it not for generic patches
> but for detouring
> so it hasn't been an issue for them
> detours are installed at the start of functions
> within the first 20 bytes
> the problem only comes when you try to use that ApplyPatch function for general purpose patching in random locations
> and it only protects the first 20 bytes of the function

Note that "on mine" refers to his private copy of `detourhelpers.h`. He's using ApplyPatch to patch functions, not just create detours, and finding that some calls to ApplyPatch cause segfaults. The proposed fix eliminates these segfaults.

Note that neither of us are familiar with how ProtectMemory is meant to be used, but this appears to be correct from inspection and the elimination of crashes.

Use ApplyPatch() to change bytes "far" from the start of the function. I assume "far" means "any subsequent memory pages", but again I don't know how it works.

asherkin:
On Linux if a detour crossed a page boundary we would only change the
memory protection of the first page (as we were aligning the address as
required, but not taking into account the length).

I don't have an easy way to test this but it looks correct. `addr + len`
doesn't appear to need to be aligned though, so another option could be
to use `(addr - startPage) + length` as len.

Also fixed a non-zero offset being passed into CDetour's ApplyPatch
function - this is never done internally anywhere, but it doesn't hurt
to fix it.

Fixes #984

### DIFF
--- a/public/CDetour/detourhelpers.h
+++ b/public/CDetour/detourhelpers.h
@@ -56,7 +56,7 @@ inline void ProtectMemory(void *addr, int length, int prot)
 #if defined PLATFORM_POSIX
 	long pageSize = sysconf(_SC_PAGESIZE);
 	void *startPage = ke::AlignedBase(addr, pageSize);
-	void *endPage = ke::AlignedBase((intptr_t)addr + length, pageSize);
+	void *endPage = ke::AlignedBase((void *)((intptr_t)addr + length), pageSize);
 	mprotect(startPage, ((intptr_t)endPage - (intptr_t)startPage) + pageSize, prot);
 #elif defined PLATFORM_WINDOWS
 	DWORD old_prot;
--- a/public/CDetour/detourhelpers.h
+++ b/public/CDetour/detourhelpers.h
@@ -56,7 +56,7 @@ inline void ProtectMemory(void *addr, int length, int prot)
 #if defined PLATFORM_POSIX
 	long pageSize = sysconf(_SC_PAGESIZE);
 	void *startPage = ke::AlignedBase(addr, pageSize);
-	void *endPage = ke::AlignedBase(addr + length, pageSize);
+	void *endPage = ke::AlignedBase((intptr_t)addr + length, pageSize);
 	mprotect(startPage, ((intptr_t)endPage - (intptr_t)startPage) + pageSize, prot);
 #elif defined PLATFORM_WINDOWS
 	DWORD old_prot;
--- a/public/CDetour/detourhelpers.h
+++ b/public/CDetour/detourhelpers.h
@@ -34,13 +34,10 @@
 
 #if defined PLATFORM_POSIX
 #include <sys/mman.h>
-#ifndef PAGE_SIZE
-#define	PAGE_SIZE	4096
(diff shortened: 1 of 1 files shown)


public/CDetour/detourhelpers.h
```diff
@@ -57,8 +54,10 @@ struct patch_t
 inline void ProtectMemory(void *addr, int length, int prot)
 {
 #if defined PLATFORM_POSIX
-	void *addr2 = (void *)ALIGN(addr);
-	mprotect(addr2, sysconf(_SC_PAGESIZE), prot);
+	long pageSize = sysconf(_SC_PAGESIZE);
+	void *startPage = ke::AlignedBase(addr, pageSize);
+	void *endPage = ke::AlignedBase((intptr_t)addr + length, pageSize);
```
dvander, line 59: Should this be addr + length - 1 if the sequence runs right up to the page boundary?

asherkin: I did thing that initially but I *think* this is correct, if you imagine a zero-length sequence doing -1 would take us one page back, and a sequence equal to the length of a page seems like it'll do the right thing here as well (endPage would be the same as startPage, but one more byte would take it to the next page).

dvander: startPage = AlignedBase(0, 4096) = 0
endPage = AlignedBase(0 + 4096, 4096) = 4096
len = endPage - startPage + pageSize = 4096 - 0 + 4096 = 8192

So I think you'd wind up with two pages? That said, it's pretty edge casey and doesn't really matter.

dvander (approved):

public/CDetour/detourhelpers.h
asherkin: Ah, yes, that does make sense! Yeah, I think I'd prefer going over in that edge case than trying to special-case zero length.

The code here can only use a maximum length of 20 currently :D

PR #1537, opened 2021-07-17
Track the creating plugin for convars

### LINKED ISSUE #1492: Add ConVar.PluginHandle property
Could you, please, add the new field in ConVar struct to know which plugin is created the specific ConVar?

There was many use cases for this time.
 - The last one was an ability to find what "... _version" Cvar name has the specific plugin and vice versa (used in planned "Plugin updater" to search updated version by cvar name in a database).
 - Also, e.g. for [[ANY] ConVars Anomaly Fixer](https://forums.alliedmods.net/showthread.php?t=307804) to indicate in log which plugin the concrete ConVar "belongs" to.

I know that ConVar objects are global, but the ability to know its creator would be very useful.

Thank you.

asherkin:
Similar to the recent work for commands, track and expose the creating
plugin for convars. The first plugin to register a given cvar becomes
the owner until that plugin is unloaded. If a plugin attempts to
register a convar that was already registered and the originally
registering plugin has been unloaded, that plugin becomes the owner.
This isn't quite as nice as the way commands shift "ownership" as
plugins are unloaded, but we don't have a sane data structure currently
to implement that, and it seemed like a lot of unnecessary work as there
shouldn't really be multiple plugins with conflicting cvars.

Closes #1492

### DIFF
--- a/core/ConVarManager.cpp
+++ b/core/ConVarManager.cpp
@@ -210,18 +210,27 @@ void ConVarManager::OnUnlinkConCommandBase(ConCommandBase *pBase, const char *na
 void ConVarManager::OnPluginUnloaded(IPlugin *plugin)
 {
 	ConVarList *pConVarList;
-	List<ConVarQuery>::iterator iter;
-
 	/* If plugin has a convar list, free its memory */
 	if (plugin->GetProperty("ConVarList", (void **)&pConVarList, true))
 	{
 		delete pConVarList;
 	}
 
+	/* Clear any references to this plugin as the convar creator */
+	for (List<ConVarInfo *>::iterator iter = m_ConVars.begin(); iter != m_ConVars.end(); ++iter)
+	{
+		ConVarInfo *pInfo = (*iter);
+
+		if (pInfo->pPlugin == plugin)
+		{
+			pInfo->pPlugin = nullptr;
+		}
+	}
+
 	const IPluginRuntime * pRuntime = plugin->GetRuntime();
 
 	/* Remove convar queries for this plugin that haven't returned results yet */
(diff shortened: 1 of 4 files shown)


asherkin:
Tested with:

```sourcepawn
public void OnPluginStart()
{
    PrintConVar("sm_chat_mode");
    PrintConVar("sv_alltalk");
}

void PrintConVar(const char[] name)
{
    ConVar cvar = FindConVar(name);
    Handle plugin = cvar.Plugin;

    char filename[PLATFORM_MAX_PATH];
    if (plugin)
        GetPluginFilename(plugin, filename, sizeof(filename));

    PrintToServer(">>> %s: %x %x - \"%s\"", name, cvar, plugin, filename);
}
```

Unloaded, loaded, reloaded basechat (provider of `sm_chat_mode`) and the test plugin several times in various orders to ensure there were no dangling references and everything was updated correctly.

core/ConVarManager.cpp
```diff
@@ -210,18 +210,27 @@ void ConVarManager::OnUnlinkConCommandBase(ConCommandBase *pBase, const char *na
...
-	List<ConVarQuery>::iterator iter;
-
 	/* If plugin has a convar list, free its memory */
 	if (plugin->GetProperty("ConVarList", (void **)&pConVarList, true))
 	{
 		delete pConVarList;
 	}
 
+	/* Clear any references to this plugin as the convar creator */
+	for (List<ConVarInfo *>::iterator iter = m_ConVars.begin(); iter != m_ConVars.end(); ++iter)
```
Headline, line 220: We should probably get used to using ranged-for, but it's just a nit

Headline (approved):
All this looks good - thanks!

PR #1540, opened 2021-07-18
Switch internal SM concept of frames to use Think

asherkin:
This has been asked for and debated in some form since Valve introduced
hibernation into the Source engine. The changes here are based on quite
a deep dive into the engine's frame/think logic (mainly in CS:GO which
has "legacy" hibernation and TF2 which has modern "frameless" ticking)
and all seem to be sane.

I think I've managed to maintain all the oddities around time keeping,
and the simulated bool (even though we don't really use it for anything)
should have a sane value. There is a slight behaviour change for
anything needing exact timings as we're now run earlier in the frame
before gpGlobals are updated, this should generally be fine but it might
affect some plugins such as bhop timers that are trying to be extremely
precise (often more precise than the underlying data they're using).

We'll probably want to add a native for plugins to detect if the server
is not completely simulating so they can opt out of work, but I think
defaulting to having things work like this makes more sense than adding
a 2nd set of per-frame forwards and natives (#540), and this makes
timers and any extension callbacks work automatically.

### DIFF
--- a/core/TimerSys.cpp
+++ b/core/TimerSys.cpp
@@ -35,6 +35,7 @@
 #include "frame_hooks.h"
 #include "ConVarManager.h"
 #include "logic_bridge.h"
+#include <bridge/include/IProviderCallbacks.h>
 
 #define TIMER_MIN_ACCURACY		0.1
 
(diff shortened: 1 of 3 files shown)


KyleSanderson (approved):
Timers have an inherient resolution of 0.1s + ~15ms on the next frame depending on where they are in the cycle - so the bhop comment I don't necessarily agree with that we've impacted their ability to maintain accuracy.

Thanks for taking this one on, it's a bit late in the cycle but this will likely help a lot of people with their expectations of what SM will allow them to do during a hibernation.

Issue #1529, opened 2021-07-16
NameHashSet


Kailo97:
https://crash.limetech.org/7txceegvqyfk
I was traking the crash and find that NameHashSet is weird.
For example:
`key` declarated but not used.
https://github.com/alliedmodders/sourcemod/blob/38eecd5ece26b2469560db1822511a7e2685286e/public/sm_namehashset.h#L128-L136
https://github.com/alliedmodders/sourcemod/blob/38eecd5ece26b2469560db1822511a7e2685286e/public/sm_namehashset.h#L148-L153
I'm not sure that find all problems in class.
Crash may not to be connected with bad code.

Crash results:
Coruppted entry in HashTable, whitch may cause by:
1) Whole entry is incorrect (coruppted table)
2) Wrong entry's payload (pointer)
3) Pointer correct but virtual table pointer was coruppted
For cases 1 and 2 need to looking in SM code for reason, in case 3 it's external problem.
How to figure it out that exactly, no idea yet.
SM version 1.10-6502, but upstream 1.10-dev (0b468f2) and master (38eecd5) same in this part.
Crash tracing:
```
sourcemod.2.csgo.so!CommandFlagsHelper::ConCommandPolicy::matches												INLINED	https://github.com/alliedmodders/sourcemod/blob/617fb77ea02076cc70be84c6092ebcfc81d54cc2/core/smn_console.cpp#L185
sourcemod.2.csgo.so!ke::HashTable<SourceMod::NameHashSet::Policy<ConCommandBase *, CommandFlagsHelper::ConCommandPolicy>, ke::SystemAllocatorPolicy>::lookup	INLINED	https://github.com/alliedmodders/amtl/blob/e00a845c6bc415995ddc4b7ec538d1704fdd0122/amtl/am-hashtable.h#L304
sourcemod.2.csgo.so!ke::HashTable<SourceMod::NameHashSet::Policy<ConCommandBase *, CommandFlagsHelper::ConCommandPolicy>, ke::SystemAllocatorPolicy>::find	INLINED	https://github.com/alliedmodders/amtl/blob/e00a845c6bc415995ddc4b7ec538d1704fdd0122/amtl/am-hashtable.h#L430
[...]

PR #1542, opened 2021-07-18
Make all command lookup case-insensitive

### LINKED ISSUE #1480: CommandExists() case sensitive alternative
At the moment, [CommandExists()](https://sm.alliedmods.net/new-api/console/CommandExists) is case insensitive, so it is useless for cases when I need to check for:

```
CommandExists("sm_ADMIN")
```
returns true, however, ~~such command "sm_ADMIN"  cannot be handled by console.~~ such chat command "!ADMIN" cannot be handled.

Is it possible to add a new command like CommandExistsEx() or an optional argument, allowing to check "Is Command Exists" considering case sensitivity?

Thank you.

asherkin:
SM internally maintained both a case-sensitive and a case-insensitive
lookup method for commands, where the case-sensitive hashmap was used as
a fast path, and case-insensitive iteration over a list used as the slow
path if a command was not found in the hashmap. But only command
dispatch handling used this dual path approach, chat triggers for
example only did a loopup in the hashmap.

Over the years Valve has made more and more of the command dispatch
logic case-insensitive to the point where all console commands are now
case-insensitive, so maintaining case sensitivity when using chat
triggers does not make a lot of sense. There are somewhat popular
plugins that attempt to "correct" this behaviour - but at least one is
having issues after the previous case-sensitivity fixes for commands -
see #1480.

We still have to keep the list around for the sorted help use case and
command iteration, but this PR changes the hashmap to use a
case-insensitive hashing policy (as previously done for convars, and
more recently for game command lookup) and changes all by-name lookup to
exclusively use the hashmap (as there is no need to fall back to the
list any more).

Tested a bunch in TF2, I don't know of any games that still have a
case-sensitive command dispatch pipeline to test. I think the worst case
would be that we'd accept a chat command in the "wrong" case then fail
to execute the underlying command. If that turns out to be an issue in
practice, we should be able to fix it easily enough by replacing the
command name in the buffer with the correct casing of the command we
looked up.

Also fixed a couple of very minor Lookup vs. Key issues in NameHashSet
(noted in #1529) that were being masked due to CharsAndLength's
converting constructor. I tried to make the constructor explicit to
avoid this happening in the future but HashTable's add function relies
on being able to do an implicit conversion so that wasn't possible. We
might want to just rely on the implicit conversion up here as well, but
it doesn't really matter either way.

Fixes #1480, #1529

### DIFF
--- a/core/ChatTriggers.cpp
+++ b/core/ChatTriggers.cpp
@@ -363,7 +363,7 @@ bool ChatTriggers::PreProcessTrigger(edict_t *pEdict, const char *args)
 	if (!g_ConCmds.LookForSourceModCommand(cmd_buf))
 	{
 		/* Check if we had an "sm_" prefix */
-		if (strncmp(cmd_buf, "sm_", 3) == 0)
+		if (strncasecmp(cmd_buf, "sm_", 3) == 0)
 		{
 			return false;
 		}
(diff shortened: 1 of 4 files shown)


dvander (approved):
Weird, why is HashTable trying to convert to a CharsAndLength?

asherkin:
> Weird, why is HashTable trying to convert to a CharsAndLength?

It's something to do with the key type being std::string and the lookup type being CharsAndLength, the add function that takes a key needs to be able to convert it to the lookup type to call findForAdd, but also needs it in key form later to do the actual insert.

dvander:
Weird, sounds like a design flaw or bug in how it uses the policy. The intent was to have lookup types separate from the key, to avoid unnecessary allocations. But it's probably just over-complicated in any situation SM uses hashtables.

PR #1544, opened 2021-07-18
Always use our cached name value

### LINKED ISSUE #1315: IGamePlayer::GetName with multibyte characters
Following https://forums.alliedmods.net/showthread.php?t=305561 
Game: Left 4 Dead 2
```Dump of assembler code for function _ZN7CPlayer7GetNameEv:
   0xe9519b50 <+0>:     push   %ebp
   0xe9519b51 <+1>:     mov    %esp,%ebp
   0xe9519b53 <+3>:     sub    $0x8,%esp
   0xe9519b56 <+6>:     mov    0x8(%ebp),%ecx
   0xe9519b59 <+9>:     mov    0x38(%ecx),%eax
   0xe9519b5c <+12>:    test   %eax,%eax
   0xe9519b5e <+14>:    je     0xe9519b72 <_ZN7CPlayer7GetNameEv+34>	// if (m_Info && ...
   0xe9519b60 <+16>:    mov    0x34(%ecx),%edx
   0xe9519b63 <+19>:    cmpl   $0x0,0xc(%edx)
   0xe9519b67 <+23>:    je     0xe9519b72 <_ZN7CPlayer7GetNameEv+34>
   0xe9519b69 <+25>:    mov    (%eax),%ecx				// ... m_pEdict->GetUnknown())
   0xe9519b6b <+27>:    mov    %eax,(%esp)
   0xe9519b6e <+30>:    call   *(%ecx)
   0xe9519b70 <+32>:    jmp    0xe9519b7f <_ZN7CPlayer7GetNameEv+47>	// return m_Info->GetName();
   0xe9519b72 <+34>:    mov    0x8(%ecx),%ecx
   0xe9519b75 <+37>:    test   %ecx,%ecx
   0xe9519b77 <+39>:    mov    $0xe95af817,%eax
=> 0xe9519b7c <+44>:    cmovne %ecx,%eax				// return m_Name.c_str();
   0xe9519b7f <+47>:    add    $0x8,%esp
   0xe9519b82 <+50>:    pop    %ebp
   0xe9519b83 <+51>:    ret
End of assembler dump.
```
^ Marked instruction only gets executed during initial connect, then #545 appears to be useless.

Could you tell why not return m_Name [here](https://github.com/alliedmodders/sourcemod/blob/master/core/PlayerManager.cpp#L2281)?
Could I make a PR to return m_Name or should I change IGamePlayer interface's method into `const char *GetName(fixed =  true)` (following change for sp include) with corresponding return values?

asherkin:
In #545 we started automatically fixing up invalid UTF8 characters
caused by truncated names from Steam, but since the dawn of time CPlayer
has preferred directly returning the engine's name pointer if we have
once available, so our corrected name is almost never used.

Lightly tested in CS:GO and TF2 with no ill effects. Fixes #1315

### DIFF
--- a/core/PlayerManager.cpp
+++ b/core/PlayerManager.cpp
@@ -2282,11 +2282,6 @@ void CPlayer::SetName(const char *name)
 
 const char *CPlayer::GetName()
 {
-	if (m_Info && m_pEdict->GetUnknown())
-	{
-		return m_Info->GetName();
-	}
-	
 	return m_Name.c_str();
 }
 


KyleSanderson:
Can we return the engine value if our cache is invalid (and then store the result in our cache?).

asherkin:
I don't think there can ever be a point that the name isn't set - the previous code here was explicitly using it early when the engine player info wasn't available.

KyleSanderson (approved):

NotPowered (commented):
This modification has integrated a regression

Names set with SetInfo are now not returned properly.

So we have a display name = B (in scoreboard, chat etc ... with SetInfo and when we do GetClientName we have a name A

PR #1545, opened 2021-07-18
Clear sm_nextmap so we don't get stuck in a loop

### LINKED ISSUE #1429: Populated sm_nextmap causes looping on changelevel failures
# Help us help you

# Environment
  * Operating System version: Linux
  * Game/AppID (with version if applicable): Team Fortress 2
  - [x] I have updated SourceMod to the [latest snapshot](https://www.sourcemod.net/downloads.php?branch=dev) (1.11.0.6655) and it still happens.

# Description
Team Fortress 2 (and possibly other games, since this is in the base teamplay system) has an event `server_changelevel_failed` that the game uses during the gameover state to move to the next map in the cycle.  SourceMod's `sm_nextmap` isn't cleared at this stage; if the game was in that state and the `sm_nextmap` ~~was invalid~~ couldn't be validated (e.g. `workshop/1`), SourceMod will intercept the game's attempts to change level and switch back to the invalid map, getting into a loop.

1. While in TF2, use `sm_setnextmap workshop/1` in the game / server console to set the next map, or change `sm_nextmap` directly to `workshop/1`.
2. Get to the post-game screen, then observe the server console as it changes maps.

# Logs
```
In gameover, but failed to load the next map. Trying next map in cycle.
CHANGE LEVEL: cp_snakewater_final1
L 02/18/2021 - 06:46:07: [SM] Changed map to "workshop/1"
[TF Workshop] Preparing map ID 1
[TF Workshop] Error fetching updated information for map id 1
[TF Workshop] Info lookup failed for workshop file 1 ( EResult 1 )
[TF Workshop] Map failed to sync, load will not go well :(
CModelLoader::Map_IsValid:  No such map 'maps/workshop/1.bsp'
Unable to change level!
In gameover, but failed to load the next map. Trying next map in cycle.
CHANGE LEVEL: cp_sunshine
L 02/18/2021 - 06:46:07: [SM] Changed map to "workshop/1"
[TF Workshop] Preparing map ID 1
[TF Workshop] Error fetching updated information for map id 1
[TF Workshop] Info lookup failed for workshop file 1 ( EResult 1 )
[TF Workshop] Map failed to sync, load will not go well :(
CModelLoader::Map_IsValid:  No such map 'maps/workshop/1.bsp'
Unable to change level!
In gameover, but failed to load the next map. Trying next map in cycle.
CHANGE LEVEL: cp_metalworks
L 02/18/2021 - 06:46:07: [SM] Changed map to "workshop/1"
[TF Workshop] Preparing map ID 1
[TF Workshop] Error fetching updated information for map id 1
[TF Workshop] Info lookup failed for workshop file 1 ( EResult 1 )
[TF Workshop] Map failed to sync, load will not go well :(
CModelLoader::Map_IsValid:  No such map 'maps/workshop/1.bsp'
Unable to change level!
In gameover, but failed to load the next map. Trying next map in cycle.
[...]
```

asherkin:
This seems to work well, but it requires a change to the nextmap plugin as well as the implementation in core - which will be a pain for users as they almost never update the base plugins on upgrade. If the old plugin is used against a newer core version the `sm_nextmap` cvar will never get re-populated after map change, so SM will stop controlling the map cycle.

Fixes #1429

### DIFF
--- a/plugins/nextmap.sp
+++ b/plugins/nextmap.sp
@@ -88,7 +88,9 @@ public void OnPluginStart()
 	RegAdminCmd("sm_maphistory", Command_MapHistory, ADMFLAG_CHANGEMAP, "Shows the most recent maps played");
 	RegConsoleCmd("listmaps", Command_List);
 
-	// Set to the current map so OnMapStart() will know what to do
+	HookEventEx("server_changelevel_failed", OnChangelevelFailed, EventHookMode_Pre);
+
+	// Set to the current map so OnConfigsExecuted() will know what to do
 	char currentMap[PLATFORM_MAX_PATH];
 	GetCurrentMap(currentMap, sizeof(currentMap));
 	SetNextMap(currentMap);
@@ -110,8 +112,16 @@ public void OnConfigsExecuted()
 	// not in mapcyclefile. So we keep it set to the last expected nextmap. - ferret
 	if (strcmp(lastMap, currentMap) == 0)
 	{
-		FindAndSetNextMap();
+		FindAndSetNextMap(currentMap);
 	}
+}
+
+public void OnChangelevelFailed(Event event, const char[] name, bool dontBroadcast)
+{
+	char failedMap[PLATFORM_MAX_PATH];
+	event.GetString("levelname", failedMap, sizeof(failedMap));
+
+	FindAndSetNextMap(failedMap);
 }
 
 public Action Command_List(int client, int args) 
@@ -129,7 +139,7 @@ public Action Command_List(int client, int args)
 	return Plugin_Handled;
 }
   
-void FindAndSetNextMap()
+void FindAndSetNextMap(char[] currentMap)
 {
 	if (ReadMapList(g_MapList, 
 			g_MapListSerial, 
@@ -149,14 +159,11 @@ void FindAndSetNextMap()
 	
 	if (g_MapPos == -1)
 	{
-		char current[PLATFORM_MAX_PATH];
-		GetCurrentMap(current, sizeof(current));
-
 		for (int i = 0; i < mapCount; i++)
 		{
 			g_MapList.GetString(i, mapName, sizeof(mapName));
 			if (FindMap(mapName, mapName, sizeof(mapName)) != FindMap_NotFound && 
-				strcmp(current, mapName, false) == 0)
+				strcmp(currentMap, mapName, false) == 0)
 			{
 				g_MapPos = i;
 				break;
--- a/core/NextMap.cpp
+++ b/core/NextMap.cpp
@@ -109,6 +109,8 @@ bool NextMapManager::SetNextMap(const char *map)
 	return true;
 }
 
+static char g_nextMap[PLATFORM_MAX_PATH];
+
 #if SOURCE_ENGINE != SE_DARKMESSIAH
 void NextMapManager::HookChangeLevel(const char *map, const char *unknown)
 #else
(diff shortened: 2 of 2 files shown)


asherkin:
Got two potential solutions for the bug here (in each commit), one on the C++ side that is more generic but a slightly hairy implementation (I think only TF2 can take this path though, other games don't recover from bad map changes), or one purely in the nextmap plugin that listens to the event added for TF2 specifically. The plugin code has a nice advantage of preserving the SM-enforced map cycle, whereas the C++ impl just gives up for a map. I'm really not sure which I prefer.

nosoop:
I think having a hybrid of both would be best &mdash; the changes on the C++ side are necessary since core manages `sm_nextmap` and there's no guarantee that the plugin will be up to date, let alone actually be installed.

plugins/nextmap.sp
```diff
@@ -110,10 +112,18 @@ public void OnConfigsExecuted()
 	// not in mapcyclefile. So we keep it set to the last expected nextmap. - ferret
 	if (strcmp(lastMap, currentMap) == 0)
 	{
-		FindAndSetNextMap();
+		FindAndSetNextMap(currentMap);
 	}
 }
 
+public void OnChangelevelFailed(Event event, const char[] name, bool dontBroadcast)
```
nosoop, line 119: Note that this event also fires for failed changelevels mid-round; will ideally want to introduce a dependency in SDKTools and check against `GameRules_GetRoundState() == RoundState_GameOver`.

asherkin:
Ah, I hadn't spotted that any other plugins were using SetNextMap - yeah, probably best to take both of these then (they do work well together). The core changes will ensure nothing gets stuck, and the plugin changes will keep it on the right mapcycle if it would have gotten stuck.

plugins/nextmap.sp
asherkin: I think this triggering even on a failed mid-round change is probably sane, can you see any problem with that? I guess we'd end up picking the "wrong" next map perhaps (it'd be the one after the one they tried to change to).

nosoop: The server (well, TF2) forces a map change to advance to the next map in its cycle only during the game over state, so without checking for that, a failed mid-round change would overwrite any existing nextmap even if it would've been okay to use.

Round state may not be the most exact one though; [the SDK actually uses `g_fGameOver`](https://github.com/ValveSoftware/source-sdk-2013/blob/0d8dceea4310fde5706b3ce1c70609d72a38efdf/mp/src/game/shared/teamplayroundbased_gamerules.cpp#L588-L599).

KyleSanderson:
> If the old plugin is used against a newer core version the sm_nextmap cvar will never get re-populated after map change, so SM will stop controlling the map cycle.

This sounds rather undesirable - no?

nosoop:
> > If the old plugin is used against a newer core version the sm_nextmap cvar will never get re-populated after map change, so SM will stop controlling the map cycle.
> 
> This sounds rather undesirable - no?

I believe that comment was for the original version of the PR, which (if I remember correctly) blanked out nextmap at the start of every map change.  The latest revision for core should be fine with an older version of the plugin; the effect of a failed map change will be one off-cycle map before getting back on track (with an updated plugin, a failed map change will advance to the next available map in the cycle).

nosoop:
I've tested the latest commit against TF2 - including running a copy of `nextmap.smx` from before this change - and it should be fine, aside from the current behavior the plugin where a failed changelevel mid-game unconditionally modifies `sm_nextmap` (which wouldn't cause a softlock as the game only continues on the current map).

KyleSanderson:
I don't like this - I think there's a better approach.

This has messed up my bench for ages (readas: 15 years) - the reason why is we don't do any validation of the level before setting the convar.  Things are OK as the engine takes over and ignores case sensitivity which is excellent as levels are delivered using netchannel at a speedy `rate` speed (usually 30 to 100KB/s). However, the requests that are serviced against sv_downloadurl are still in the wrong case which frequently results in a 404 situation as the requested file does not actually exist on the webserver.

What I think we should do: I think we should resolve the relative path, then verify the file exists, and then set the convar. It's going to result in a "no workshop2" support situation, but we can add support for that situation when it happens (like when workshop shipped previously).

nosoop:
Could you clarify?  There is no change in that the engine still handles checking the validity of the map and preparing resources here; I don't quite understand where `sv_downloadurl` ties into this.  `newmap` will be whatever was stored in `sm_nextmap`; there's no change in that behavior either.

psychonic (approved):

KyleSanderson:
Yeah, boiling the ocean lead to nothing happening... this is still an improvement, feel free to merge.

peace-maker:
Sounds like this should be merged a while ago and nobody pressed da button. Handling failed mid-round mapchanges slightly wrong is better than deadlocking in a map loop and can be addressed later.

ambaca:
I actually liked that forced change map behaviour...

Anyway, this change affected DOD:S game, because it have a bug inside game.
https://forums.alliedmods.net/showpost.php?p=2831497&postcount=4

PR #1550, opened 2021-07-20
Support reading legacy sendprop arrays

### LINKED ISSUE #1386: Implement array element parameter on GameRules_SetPropString
# Environment
  * Operating System version: Windows OS
  * Game/AppID (with version if applicable): L4D2
  * Current SourceMod version: 1.10.0.6499
  * Current MetaMod version: Source version 1.10.7-dev
  
# Description
  In L4D2, there is a feature called EMS (Extended Mutation System) that allows us to use some available huds (15 in total) to draw things on the screen.
  The netprops related to this are on CTerrorGameRules, but one of them ("m_szScriptedHUDStringSet") seems to be a string array.
  
  Currently the "GameRules_SetPropString" function doesn't support passing an element (from array) in its parameters, unlike the others (e.g: GameRules_SetProp, GameRules_SetPropFloat)
  
  https://sourcemod.dev/#/sdktools_gamerules/function.GameRules_SetPropString
  https://sm.alliedmods.net/new-api/sdktools_gamerules/GameRules_SetPropString

  I'm not sure but maybe having this implementation also on GameRules_SetPropString, I could use the element parameter to set the text from the HUD [1-15].
  
  Currently, the only way that I achieved that was by setting
  
  ```
  GameRules_SetPropString("m_szScriptedHUDStringSet", "some string here");
  ```
  
  Seems that at every strlen of 128 the text goes to the next hud, but the engine has a limit of networking a strlen of 512. If I pass a higher value, the game crashes. (also, probably is the reason that in the core, the buffer is limited to strlen of 512)
  So, I can only set the text of 4 huds in the best scenario.
  
  This is all the info that I got:
  
  It all started here https://forums.alliedmods.net/showpost.php?p=2725233&postcount=14,
  where is mentioned that the EMS system has some netprops related to the GameRules, then I did several tests.
  
  I was able to set most of the netprops to their specific huds but the text.
  
  From a discord talk, according to asherkin:
  > looks like array support isn't there just because there were no string arrays when it was implemented
  > the netprop dump output is odd, they don't appear to be any different than the other fixed arrays"
  > it looks like this is just the pattern for this kind of array prop, there just really aren't that many of them (18 in l4d2)
  > most of the array props we deal with, use something called SendPropArray3, but these use SendPropArray which looks like an older implementation that is possibly more flexible but is encoded as two props
  > it is of course Valve, so I can't find anything explaining it
  > https://developer.valvesoftware.com/wiki/SendPropArray3()
  > SendPropArray3()
  > looks like there are probably several bits of work here:
  > 1) SDKTools' netprop dumper has no idea what to do with these props and outputs the internal impl detail
  > 2) SM probably can't read these in any form currently (interestingly one of them is m_hViewModel on players - I wonder what that 2nd index is used for...)
  > 3) The gamerules string prop will need array support
  > on the upside it doesn't look like there is any sdk hackery to do, it's all covered in the public SDK
  > heh, and pose params are one of the other ones
  > MVM has a couple of string_t ones that will probably be hairy (m_iszMannVsMachineWaveClassNames)

  Some info from netprop dump:
  ```
  Member: m_szScriptedHUDStringSet (offset 1656) (type string) (bits 0) (InsideArray)
  Member: m_szScriptedHUDStringSet (offset 0) (type array) (bits 0) ()
  ```
  
  Which probably on SM would look like that:
  ```
  char m_szScriptedHUDStringSet[15][128];
  ```
  
  Some info found by asherkin:
  > CTerrorGameRules::ResetScriptedHUD has useful info about the data layout
  > InternalSendPropArray((int)&unk_F66154, 15, (char *)0x80, "m_szScriptedHUDStringSet", 0);
  
  Some valve wiki related to the EMS:
  https://developer.valvesoftware.com/wiki/L4D2_EMS/Appendix:_HUD

asherkin:
This rounds out the work started in #1548 to complete support for
reading the older SendPropArray type array netprops, along with bringing
SDKTools' GameRule netprop code in sync with core to add string array
support.

There aren't many SendPropArray type props around but this opens up a
few interesting opportunities for plugin developers, particularly in
L4D2 with manipulation of the EMS HUD.

Tested reading the `m_vCPPositions` array in TF2, and reading/writing
the `m_szScriptedHUDStringSet` EMS HUD netprop in L4D2. Closes #1386.

### DIFF
--- a/extensions/sdktools/gamerulesnatives.cpp
+++ b/extensions/sdktools/gamerulesnatives.cpp
@@ -108,12 +108,43 @@ enum PropFieldType
 	{ \
 	case type: \
 		{ \
-			if (element > 0) \
+			if (element != 0) \
 			{ \
 				return pContext->ThrowNativeError("SendProp %s is not an array. Element %d is invalid.", \
 					prop, \
 					element); \
 			} \
+			break; \
+		} \
+	case DPT_Array: \
+		{ \
+			int elementCount = pProp->GetNumElements(); \
+			int elementStride = pProp->GetElementStride(); \
+			if (element < 0 || element >= elementCount) \
+			{ \
+				return pContext->ThrowNativeError("Element %d is out of bounds (Prop %s has %d elements).", \
+					element, \
+					prop, \
+					elementCount); \
+			} \
+			\
+			pProp = pProp->GetArrayProp(); \
+			if (!pProp) { \
+				return pContext->ThrowNativeError("Error looking up ArrayProp for prop %s", \
+					prop); \
+			} \
+			\
+			if (pProp->GetType() != type) \
+			{ \
+				return pContext->ThrowNativeError("SendProp %s type is not " type_name " ([%d,%d] != %d)", \
+					prop, \
+					pProp->GetType(), \
+					pProp->m_nBits, \
+					type); \
+			} \
+			\
+			offset += pProp->GetOffset() + (elementStride * element); \
+			bit_count = pProp->m_nBits; \
 			break; \
 		} \
 	case DPT_DataTable: \
@@ -142,7 +173,7 @@ enum PropFieldType
 	} \
 	\
 	int elementCount = pTable->GetNumProps(); \
-	if (element >= elementCount) \
+	if (element < 0 || element >= elementCount) \
 	{ \
 		return pContext->ThrowNativeError("Element %d is out of bounds (Prop %s has %d elements).", \
 			element, \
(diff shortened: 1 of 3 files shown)


extensions/sdktools/gamerulesnatives.cpp
```diff
@@ -577,29 +602,46 @@ static cell_t GameRules_SetPropString(IPluginContext *pContext, const cell_t *pa
...
-
+	size_t len;
 	pContext->LocalToString(params[2], &src);
-	size_t len = strncopy(dest, src, maxlen);
+
+	if (bIsStringIndex)
+	{
+		return pContext->ThrowNativeError("Setting string_t gamerules prop %s not supported yet.", prop);
+
+		// *(string_t *)((intptr_t)pGameRules + offset) = g_HL2.AllocPooledString(src);
```
dvander, line 637: intentional?

dvander (approved):

extensions/sdktools/gamerulesnatives.cpp
asherkin: Yeah, the extension API needs some work to expose `AllocPooledString` from core.

PR #1554, opened 2021-07-28
Fix out of bounds write in CDataPack::Write*Array

asherkin:
WriteCellArray and WriteFloatArray were allocating N+1 slots, but due to
a copy-paste error were writing N+2 slots. Much later in the process the
CRT would catch this and cause a crash - this was pretty painful to
debug but thankfully running SRCDS in CRT debug mode caught it much
sooner in CDataPack::RemoveItem.

Thanks to @dysphie for the initial report.


dvander:
Nice catch!

PR #1556, opened 2021-07-29
Fix clientprefs prefab menu crash

Natanel-Shitrit:
While working with `clientprefs` and Cookies i stubbled upon a [bug](https://discord.com/channels/335290997317697536/335290997317697536/870399001806123019):

1. you register a cookie (and add it to the prefab menu).
2. you open the prefab menu.
3. you unload the plugin (resulting the cookie to be removed).
4. you click on the removed cookie in the menu.
5. server crashed.

From what i found in the crash dumps i got: [1](https://crash.limetech.org/4nmm6lkdvmox), [2](https://crash.limetech.org/t5pjlbafhaey).

I saw that the variable `info` is NULL and it's sent to the `strtoul()` function which causes the crash to happen. (`SIGSEGV /SEGV_MAPERR accessing 0x0`)

while discussing this in the discord asherkin [pointed out](https://discord.com/channels/335290997317697536/335290997317697536/870403278263976048) to me that he doesn't think the bug is in the clientprefs extension, hence it probably won't be fixed there, but after applying this check everything seem to work just fine and the server doesn't crash anymore.

### DIFF
--- a/extensions/clientprefs/menus.cpp
+++ b/extensions/clientprefs/menus.cpp
@@ -40,7 +40,7 @@ void ClientMenuHandler::OnMenuSelect(IBaseMenu *menu, int client, unsigned int i
 
 	const char *info = menu->GetItemInfo(item, &draw);
 
-	if (!info)
+	if (info == NULL)
 	{
 		return;
 	}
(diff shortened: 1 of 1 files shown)


Natanel-Shitrit:
Just noticed you check the same check in a different place in the extension:
https://github.com/alliedmodders/sourcemod/blob/fc905641bf6f68f4d27dc5aa8a2c80b18ab945f2/extensions/clientprefs/cookie.cpp#L340-L343
So i guess you just forgot to check it here.

asherkin:
Thanks for the PR - we're not generally a fan of papering over issues so I'm going to close this out in favour of #1559 (which documents the root cause), as I don't think fixing the underlying issue will be too much more work and it will have much wider impact.

PR #1557, opened 2021-07-31
Catch exceptions from TraceRay filters/enumerators

asherkin:
When a custom TraceRay filter / EnumerateEntities enumerator callback
throws an exception we currently continue execution and then return
execution to the calling code as if there were no problems. This
currently causes a heap tracking issue in SourcePawn, but even ignoring
that it is likely the wrong behaviour and differs from our other
synchronous callbacks.

This change causes the exception to be caught, immediately terminates
the trace / enumeration, and propagates the exception state back to the
calling plugin correctly. The implementation here is based on how
SortCustom1D handles exceptions.

See also alliedmodders/sourcepawn#607. Thanks again to @dysphie for the report.

### DIFF
--- a/extensions/sdktools/trnatives.cpp
+++ b/extensions/sdktools/trnatives.cpp
@@ -56,20 +56,26 @@ class CSMTraceFilter : public CTraceFilter
 public:
 	bool ShouldHitEntity(IHandleEntity *pEntity, int contentsMask)
 	{
+		if (m_pEh->HasException()) {
+			return true;
+		}
+
 		cell_t res = 1;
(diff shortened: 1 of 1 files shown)


KyleSanderson:
I'm not sure we want this. Each plugin callback is fired individually, so it enters the function, has an error, it should continue firing until done, no? I could go either way, though.

asherkin:
Unfortunately in both cases if the callback has errored at all the trace/enumeration probably hasn't done what was intended - a big part of the change here is that now the error propagates back up so that execution doesn't continue as the state is bad, so it doesn't make sense to keep calling the callback.

KyleSanderson (approved):

PR #1561, opened 2021-08-05
Add missing return values to plugins

Wend4r:
For optional build support with the new SourcePawn 1.12 branch


dvander (approved):
Thanks for doing this!

PR #1569, opened 2021-08-22
Clarify SQL_EscapeString usage

sirdigbot:
It's hard to see the difference between 2 single-quotes and 1 double-quote unless you highlight it.

### DIFF
--- a/plugins/include/dbi.inc
+++ b/plugins/include/dbi.inc
@@ -635,8 +635,8 @@ native bool SQL_GetError(Handle hndl, char[] error, int maxlength);
  * the database's character set.
  *
  * NOTE: SourceMod only guarantees properly escaped strings when the query
- * encloses the string in ''. While drivers tend to allow " instead, the string
- * may be not be escaped (for example, on SQLite)!
+ * encloses the string in ''. While drivers tend to allow "" instead,
+ * the string may be not be escaped (for example, on SQLite)!
  *
  * @param database      A database Handle.
  * @param string        String to quote.


sneak-it:
What's there isn't technically wrong, however could be "worded" differently for clarity based upon the way it is displayed on different websites/applications, which appears to be the purpose of this PR.

Taking note what is in the previous paragraph:
https://github.com/alliedmodders/sourcemod/blob/f4ff2ad45a0a33e6a85845e675e586e3701bb7c4/plugins/include/dbi.inc#L632

It is likely better to institute the usage of `single quotes (')` and `double quotes (")`, rather than replacing with `""`, as it is currently indicating a single double quote, which is grammatically correct.

plugins/include/dbi.inc
```diff
@@ -374,8 +374,8 @@ methodmap Database < Handle
 	// the database's character set.
 	//
 	// NOTE: SourceMod only guarantees properly escaped strings when the query
-	// encloses the string in ''. While drivers tend to allow " instead, the string
-	// may be not be escaped (for example, on SQLite)!
+	// encloses the string in single quotes. While drivers tend to allow double
```
sneak-it, line 377: Almost!
`single quotes (')`

plugins/include/dbi.inc
```diff
@@ -635,8 +635,8 @@ native bool SQL_GetError(Handle hndl, char[] error, int maxlength);
  * the database's character set.
  *
  * NOTE: SourceMod only guarantees properly escaped strings when the query
- * encloses the string in ''. While drivers tend to allow " instead, the string
- * may be not be escaped (for example, on SQLite)!
+ * encloses the string in single quotes. While drivers tend to allow double
```
sneak-it, line 638: ditto

sneak-it (changes requested):
You displayed the double quotes in parentheses, but not the single quotes.

PR #1572, opened 2021-08-25
Revert "Switch internal SM concept of frames to use Think"

asherkin:
I think all the problems here are around `realtime` being used to update `g_fUniversalTime` and that feeding into the timer run calculations, but it takes several days for things to go wonky enough here that it is just too time consuming to investigate right now.

Notes for the next person to try this:
* The feedback on the concept here was generally very positive and people *want this to work*.
* Some people are adamant about about timers being aligned to simulation time, despite them being measured in seconds.
* For the love of god do not imply changes have been tested on anything less than every game and obscure Source mod ever released for weeks of time each.
* The old Source engine time keeping is really bad, (for L4D2) frame execution runs in a 1ms sleep loop where it accumulates time until it hits `fps_max` (which a number of servers set to 0, so it triggers every time), then it runs an engine frame (`::Think`) which then decides how many game frames to run based on the target tickrate (anything from 30 to 128). All in, this means it almost never hits exact game frame start timing and variance is all over the place.
* `realtime` accumulates in a float, which is probably the cause of the most of the issues seen - we tried to account for that in the same way we do `curtime`, but I suspect the engine was just failing to accumulate properly due to the tight frame loop (whereas `curtime` accumulates much later in the process where more time has elapsed).
* Hibernation in some games (*cough* L4D2, CS:GO) completely destroys all the engine's time keeping against reality as it inserts a 20ms sleep very late in the process (right before it would have run a game frame) which isn't accounted for anywhere else - this is why the console is sluggish in these games.
* `Plat_FloatTime` seems sane, and returns a double.

I think the next attempt should look like this:
* Use the Think hook changes from this PR to implement a callback that is run continuously even when hibernating.
* In there, cache `Plat_FloatTime` into a global, call it the reference time for that frame. Update pretty much everything other than timers using "universal time" to use that.
* Dispatch all the *internal* SM time-based logic off that callback, but nothing exposed (this'll require decoupling the existing frame hook list). This'll make DBI work correctly during hibernation, which is a top priority.
* Add a new extension API to register a think hook so that extensions can do background work, which is the other top priority and really the most important bit of these changes.
* Add a new timer flag `TIMER_FLAG_REAL_TIME` that works based on the frame reference time instead of "universal time" (both one-shot and repeating), for the few plugins that want true always-running execution and the ability to dispatch new work during hibernation.

Requiring an API version bump and not being able to make existing extensions magically just work is unfortunate, but the nuance of the engine being able to run multiple game frames per think in the older games probably means `OnGameFrame` and friends should stay as they currently are.

Reverts alliedmodders/sourcemod#1540

### DIFF
--- a/core/TimerSys.cpp
+++ b/core/TimerSys.cpp
@@ -35,7 +35,6 @@
 #include "frame_hooks.h"
 #include "ConVarManager.h"
 #include "logic_bridge.h"
-#include <bridge/include/IProviderCallbacks.h>
 
 #define TIMER_MIN_ACCURACY		0.1
 
@@ -171,8 +170,9 @@ void ITimer::Initialize(ITimedEvent *pCallbacks, float fInterval, float fToExec,
 TimerSystem::TimerSystem()
 {
 	m_pMapTimer = NULL;
+	m_bHasMapTickedYet = false;
+	m_bHasMapSimulatedYet = false;
 	m_fLastTickedTime = 0.0f;
-	OnSourceModLevelEnd();
 }
 
 TimerSystem::~TimerSystem()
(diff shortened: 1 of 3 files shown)


sapphonie:
I know this is a low priority, but as someone who has a use for a "hibertimer" (tracking server memory usage Pawn side, even on a hibernating empty server), this would be neat to see in 1.11 🙏

PR #1575, opened 2021-09-06
Fix crash in `FindSendPropInfo()`

Wend4r:
After #1548 changes

### DIFF
--- a/core/smn_entities.cpp
+++ b/core/smn_entities.cpp
@@ -825,7 +825,7 @@ static cell_t FindSendPropInfo(IPluginContext *pContext, const cell_t *params)
 {
 	char *cls, *prop;
 	sm_sendprop_info_t info;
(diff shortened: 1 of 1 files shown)


Wend4r:
Writing `pTable->GetNumProps()` to the address of garbage from stack led to a crash
https://github.com/alliedmodders/sourcemod/blob/3cc5e463df9b095533700afdb40f3b39f4aa3bf8/core/smn_entities.cpp#L872-L874
if there is no new 6th argument
https://github.com/alliedmodders/sourcemod/blob/3cc5e463df9b095533700afdb40f3b39f4aa3bf8/core/smn_entities.cpp#L828
https://github.com/alliedmodders/sourcemod/blob/3cc5e463df9b095533700afdb40f3b39f4aa3bf8/core/smn_entities.cpp#L842-L845

dvander (approved):

PR #1583, opened 2021-09-22
Update gamedata after latest CSGO update (22/09/21)

GAMMACASE:
Unsure why diff change acts weird for the sdkhooks gamedata, encoding and line endings should be the same. Tested both on linux and windows according to SMs discord.

### DIFF
--- a/gamedata/sdktools.games/engine.csgo.txt
+++ b/gamedata/sdktools.games/engine.csgo.txt
@@ -292,7 +292,7 @@
 			}
 			"CommitSuicide"
 			{
-				"windows"	"507"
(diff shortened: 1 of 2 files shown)


gamedata/sdktools.games/engine.csgo.txt
```diff
@@ -293,9 +293,9 @@
 			"CommitSuicide"
 			{
 				"windows"	"507"
```
wanted241, line 295: You forgot to update "windows" offset, it should be "508", too.

asherkin (approved):
Thanks!

PR #1586, opened 2021-09-24
Updates for Blade Symphony and x64

psychonic:
Multiple changes here to fix support for Blade Symphony and some things on x64 in general. BS is largely on the CS:GO engine now, and for Linux, only supports x64.

- Enable more CS:GO-specific logic on Blade Symphony (fixes and workarounds for engine quirks)
- Massive gamedata updates
- Remove some gamedata in favor of using the IServerTools helper functions that will soon be available (like in TF2, etc.)
- Fix a crash bug in Bintools for x64 funcs with void return
- Fix assumption of 4-byte pointer sizes in multiple SDKTools calls

This won't build without errors until the blade-2021 branch of the hl2sdk is merged into the blade branch, after the next game update.


AMBuildScript
```diff
@@ -31,6 +31,11 @@ class SDK(object):
 WinOnly = ['windows']
 WinLinux = ['windows', 'linux']
 WinLinuxMac = ['windows', 'linux', 'mac']
+Blade = {
+  'windows': ['x86', 'x86_64'],
+  'linux': ['x86_64'],
+  'max': ['x86_64']
```
bakugo, line 37: Shouldn't this be `mac`?

peace-maker:
We'll have to build against mmsource-1.11 for x64 support in sourcehook. I think the buildbots still use mms-1.10.

There are lots of interfaces using `int` as an equivalent of `size_t` which throws a bunch of compile warnings, but those can be addressed later after we ship some usable x64 builds.

AMBuildScript
psychonic: Thanks! Fixed.

psychonic:
> We'll have to build against mmsource-1.11 for x64 support in sourcehook. I think the buildbots still use mms-1.10.

Thanks. I'll get that updated as well.

PR #1590, opened 2021-09-29
Fix 'GiveNamedItem' in the game left4dead2.

A1mDev:
The method 'CCSPlayer::GiveNamedItem(char const*, int, CBaseEntity*)' does not work in game left4dead2, any given weapon or item immediately falls to the ground, besides, the code is missing a parameter to call this method (even with the added parameter does not give a weapon or item to hands). Another method 'CTerrorPlayer::GiveNamedItem(char const*, int, bool, CBaseEntity*)' works great, besides it makes it possible to give out all the items that exist in the game.

### DIFF
--- a/extensions/sdktools/vnatives.cpp
+++ b/extensions/sdktools/vnatives.cpp
@@ -229,6 +229,39 @@ static cell_t GiveNamedItem(IPluginContext *pContext, const cell_t *params)
 
 	return gamehelpers->EntityToBCompatRef(pEntity);
 }
+#elif SOURCE_ENGINE == SE_LEFT4DEAD2
+// CBaseEntity *CTerrorPlayer::GiveNamedItem( const char *pchName, int iSubType, bool bForce, CBaseEntity *pUnk)
+static cell_t GiveNamedItem(IPluginContext *pContext, const cell_t *params)
+{
+	static ValveCall *pCall = NULL;
+	if (!pCall)
+	{
+		ValvePassInfo pass[5];
+		InitPass(pass[0], Valve_String, PassType_Basic, PASSFLAG_BYVAL);
+		InitPass(pass[1], Valve_POD, PassType_Basic, PASSFLAG_BYVAL);
+		InitPass(pass[2], Valve_Bool, PassType_Basic, PASSFLAG_BYVAL);
+		InitPass(pass[3], Valve_CBaseEntity, PassType_Basic, PASSFLAG_BYVAL);
+		InitPass(pass[4], Valve_CBaseEntity, PassType_Basic, PASSFLAG_BYVAL);
+		
+		if (!CreateBaseCall("GiveNamedItem", ValveCall_Player, &pass[4], pass, 4, &pCall)) {
+			return pContext->ThrowNativeError("\"GiveNamedItem\" not supported by this mod");
+		} else if (!pCall) {
+			return pContext->ThrowNativeError("\"GiveNamedItem\" wrapper failed to initialize");
+		}
+	}
+
+	CBaseEntity *pEntity = NULL;
+	START_CALL();
+	DECODE_VALVE_PARAM(1, thisinfo, 0);
+	DECODE_VALVE_PARAM(2, vparams, 0);
+	DECODE_VALVE_PARAM(3, vparams, 1);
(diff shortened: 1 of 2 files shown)


extensions/sdktools/vnatives.cpp
```diff
@@ -229,6 +229,39 @@ static cell_t GiveNamedItem(IPluginContext *pContext, const cell_t *params)
...
+		}
+	}
+
+	CBaseEntity *pEntity = NULL;
+	START_CALL();
+	DECODE_VALVE_PARAM(1, thisinfo, 0);
+	DECODE_VALVE_PARAM(2, vparams, 0);
+	DECODE_VALVE_PARAM(3, vparams, 1);
+	*(bool *)(vptr + 12) = false;
+	*(void **)(vptr + 16) = NULL;
```
psychonic, line 259: Instead of hardcoding sizes, we should be making them dependent on the actual types and the platform. This has bit us when games gain support for x64. I've cleaned up most of them in https://github.com/alliedmodders/sourcemod/pull/1586. If you want to update here, you can, else I can do it as an update to that PR since I'll likely have to account for merge conflicts from this change anyway.

psychonic (approved):
Good find!

extensions/sdktools/vnatives.cpp
A1mDev: I can add of course, but left 4 dead2 does not support x64 and is unlikely to receive support). I have a question about gamedata. Servers may encounter problems when auto-updating gamedata, do I need to create a unique key in the gamedata?

psychonic: I don't think it's out of the realm of possibility that they at least add x64 support on Mac so that the game can run on modern macOS versions, whether or not those changes propagate to other platforms.

That's a good point about the gamedata key. I think that it's OK enough as-is. For users to be adversely affected, they would have to both actually using the semi-broken function, and be on a dev build older than when this lands. (The gamedata updater differentiates between the major versions).

I think some rockiness in the dev builds is expected, and is worth being able to keep the consistency in naming.

A1mDev: I don't quite understand how can I make changes to your PR?

A1mDev: You can somehow make changes to the old versions of the sourcemod, as it will not be very good to force everyone to use the new version. Because people need time to fix plugins. It would be nice if the functionality was also available in older versions of sourcemod.

extensions/sdktools/vnatives.cpp
```diff
@@ -229,6 +229,38 @@ static cell_t GiveNamedItem(IPluginContext *pContext, const cell_t *params)
...
+	}
+
+	CBaseEntity *pEntity = NULL;
+	START_CALL();
+	DECODE_VALVE_PARAM(1, thisinfo, 0);
+	DECODE_VALVE_PARAM(2, vparams, 0);
+	DECODE_VALVE_PARAM(3, vparams, 1);
+	*(bool *)(vptr + sizeof(void *) + sizeof(void *) + sizeof(int)) = false;
+	*(void **)(vptr + sizeof(void *) + sizeof(void *) + sizeof(int) + sizeof(int)) = NULL;
+	FINISH_CALL_SIMPLE(&pEntity);
```
psychonic, line 260: The last sizeof should be sizeof(bool) rather than int

extensions/sdktools/vnatives.cpp
psychonic: We plan for 1.11 to be the new stable version much sooner than later, so this fix wouldn't be exclusive to the dev build for long.

extensions/sdktools/vnatives.cpp
A1mDev: I am using the Int type because the boolean is 4 bytes on the stack, not 1 byte, based on what I see in ida, of course i could be wrong here.

A1mDev: There is a stack alignment and sizeof bool returns 1, it will be a mistake.

A1mDev: @psychonic, i fixed the type!

PR #1620, opened 2021-10-27
Add EntityCollisionRulesChanged & SetEntityOwner natives

Kenzzer:
## Introduction

#1507 
> Alright, in that case I'll open a PR in the next couple of days!

_Famous last words... or not!_

First I'd like to apologise for not having made this PR earlier.

## What changed ?

`SetCollisionGroup` signature has been removed, this signature was used by sdktools to call `CBaseEntity::SetCollisionGroup`.
```
void CBaseEntity::SetCollisionGroup( int collisionGroup )
{
	if ( (int)m_CollisionGroup != collisionGroup )
	{
		m_CollisionGroup = collisionGroup;
		CollisionRulesChanged();
	}
}
```
The main point, was to call `CBaseEntity::CollisionRulesChanged` after setting the collision group.

So instead now we go through `CBaseEntity::SetEntityOwner`, a virtual function which should ideally provide easier maintenance of gamedata (if ever needed) and easier support accross all source games.
```
void CBaseEntity::SetOwnerEntity( CBaseEntity* pOwner )
{
	if ( m_hOwnerEntity.Get() != pOwner )
	{
		m_hOwnerEntity = pOwner;

		CollisionRulesChanged();
	}
}
```
Bypassing the if statement is trivial, as `m_hOwnerEntity` is both a datamap and netprop. With this little hack, we get an easy access to `CBaseEntity::CollisionRulesChanged`.

The native `SetEntityCollisionGroup` has been rewritten to mimic what `CBaseEntity::SetCollisionGroup` does.
Added `EntityCollisionRulesChanged` as native, since we have an access to it.
And finally added `SetEntityOwner` as native, because we have to maintain that vtable offset in order to keep the other two natives working.

### Final note
As said in introduction, I have not tested anything yet and I'd be very grateful if anyone could test this on csgo, css and l4d.
I also want to mention, that the vtable offsets for CSGO and Left 4 Dead **1** were guessed, there's a chance they're invalid. If someone can provide or confirm  them that'd be helpful too!

I'm planning on testing this for CSGO & TF2 this weekend.


extensions/sdktools/vhelpers.cpp
```diff
@@ -310,6 +310,98 @@ bool FindNestedDataTable(SendTable *pTable, const char *name)
...
+	{
+		int offset = -1;
+		if (!g_pGameConf->GetOffset("SetOwnerEntity", &offset))
+		{
+			return false;
+		}
+
+#if SOURCE_ENGINE < SE_ORANGEBOX
+		CBaseEntity *pWorldEntity = nullptr;
+		for (int i = 0; i < gpGlobals->maxEntities && pWorldEntity == nullptr; ++i)
```
KyleSanderson, line 336: It's "dumb" but this can blow the array past maxEntities if it can't find anything. It's better as a branch in-loop because of the following check after kills the helper if it can't find anything.

extensions/sdktools/vnatives.cpp
```diff
@@ -1564,75 +1564,113 @@ static cell_t GivePlayerAmmo(IPluginContext *pContext, const cell_t *params)
...
+
+	// Reimplementation of CBaseEntity::SetCollisionGroup
+	int *collisionGroup = (int *)((uint8_t *)pEntity + offsetCollisionGroup);
+	if ((*collisionGroup) != params[2])
+	{
+		*collisionGroup = params[2];
+		// Returns false if CollisionRulesChanged hack isn't supported for this game
+		if (!CollisionRulesChanged(pEntity))
+		{
+			return pContext->ThrowNativeError("\"SetEntityCollisionGroup\" unsupported entity");
```
KyleSanderson, line 1591: "is not supported on this engine." ? Something like that - it shouldn't be the same message if it fails the datamap lookup.

extensions/sdktools/vnatives.cpp
```diff
@@ -1564,75 +1564,113 @@ static cell_t GivePlayerAmmo(IPluginContext *pContext, const cell_t *params)
...
+static cell_t EntityCollisionRulesChanged(IPluginContext *pContext, const cell_t *params)
+{
+	CBaseEntity *pEntity;
+	ENTINDEX_TO_CBASEENTITY(params[1], pEntity);
+	// Returns false if CollisionRulesChanged hack isn't supported for this game
+	if (!CollisionRulesChanged(pEntity))
 	{
-		void *addr;
-		if (!g_pGameConf->GetMemSig("SetCollisionGroup", &addr) || !addr)
+		return pContext->ThrowNativeError("\"EntityCollisionRulesChanged\" unsupported entity");
```
KyleSanderson, line 1604: ditto on messaging.

extensions/sdktools/vnatives.cpp
```diff
@@ -1564,75 +1564,113 @@ static cell_t GivePlayerAmmo(IPluginContext *pContext, const cell_t *params)
...
-	pSetCollisionGroup->Execute(vstk, nullptr);
+	CBaseEntity *pNewOwner = gamehelpers->ReferenceToEntity(params[2]);
+	ArgBuffer<CBaseEntity *, CBaseEntity *> vstk(pEntity, pNewOwner);
+	pSetOwnerEntity->Execute(vstk, nullptr);
 
 	return 1;
-
 }
 
 sp_nativeinfo_t g_Natives[] =
```
KyleSanderson, line 1641: can you reset this mess please.

KyleSanderson (changes requested):
were you able to test over the weekend? I'm really optimistic about this PR as it's going to help a lot of people.

extensions/sdktools/vnatives.cpp
Kenzzer: I believe that's web github showing you a weird indentation, it's already "weird" currently on the repo.
https://github.com/alliedmodders/sourcemod/blob/4bebdc18a2e4820b234a85c3a1e38eb868a84bf2/extensions/sdktools/vnatives.cpp#L1494L1526

However when edited with a code editor, say VSCode or just Visual. It showed all the native functions aligned.
And when I added `EntityCollisionRulesChanged` I had to put an extra tab on every other lines to keep everything aligned, like it was before.

Do you still want me to revert this ? Alternatively, should I indent this manually with spaces ?

Kenzzer:
> were you able to test over the weekend?
Not yet, unfortunately time flew by without me noticing, I also need to come up with a good test plugin to ensure everything's fine. But I've built the ext for tf2 & csgo (linux) and setup my local server already, hopefully this shouldn't take much longer.

extensions/sdktools/vhelpers.cpp
Kenzzer: Sorry I'm little confused by what you mean, the condition to ensure we're under maxEntities is here `i < gpGlobals->maxEntities` and we an extra null check to abort the loop early when something is found.

KyleSanderson: Yeah sorry I've been working on other weird platforms. You're right, thanks for thinking through it.

extensions/sdktools/vnatives.cpp
KyleSanderson: Don't do spaces, just modify your single line and reset the others. Functionality commits shouldn't change style.

KyleSanderson:
> > were you able to test over the weekend?
> 
> Not yet, unfortunately time flew by without me noticing, I also need to come up with a good test plugin to ensure everything's fine. But I've built the ext for tf2 & csgo (linux) and setup my local server already, hopefully this shouldn't take much longer.

Damn. Let us know when this is closer (I think we're close, too).

Kenzzer:
The requested changes have been done, and the PR has been tested on TF2 both linux & windows.

Here's a test plugin for those interested to test this out on their own
(With the merge of dhooks, this works on stock SM)
[collisionrules.txt](https://github.com/alliedmodders/sourcemod/files/7558340/collisionrules.txt)
[collisionrules.sp.txt](https://github.com/alliedmodders/sourcemod/files/7558352/collisionrules.sp.txt)

Here's the console output
unknown

KyleSanderson (approved):

einyux:
Like often a little too late, but I think it would have been better to keep the real function name "SetOwnerEntity()" instead of "SetEntityOwner()" :).

psychonic:
> Like often a little too late, but I think it would have been better to keep the real function name "SetOwnerEntity()" instead of "SetEntityOwner()" :).

SetEntityOwner is more in line with the rest of the SourceMod API. The full original function name is CBaseEntity::SetOwnerEntity, implying that you're operating on an entity to set it's owner entity. SetEntityOwner also implies that you're operating on an entity to set it's owner. The only part lost is that the owner is an entity, but I think that's better left implied rather than SetEntityOwnerEntity.

einyux:
Indeed, I understand the issue for SetOwnerEntity().

However, I don't really see a problem with SetEntityOwnerEntity(), which is in line with CBaseEntity::**SetOwnerEntity**, knowing that CXX::SetOwner() exists too.

Here we're loosing the info that we're setting "m_hOwnerEntity" and not "m_hOwner". 
Its not indicated in the sdktools_functions.inc either.

PR #1621, opened 2021-10-30
Fix up gamedata CRC functionality

psychonic:
- Changes the CRC implementation to match what appears to be most other popular implementations.
- Fixes server bin name lookup, broken for most engines.

### DIFF
--- a/bridge/include/CoreProvider.h
+++ b/bridge/include/CoreProvider.h
@@ -115,6 +115,8 @@ public:
 	virtual void ConsolePrint(const char *fmt, ...) = 0;
 	virtual void ConsolePrintVa(const char *fmt, va_list ap) = 0;
 
+	virtual void FormatSourceBinaryName(const char *basename, char *buffer, size_t maxlength) = 0;
+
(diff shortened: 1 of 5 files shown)


peace-maker (approved):

asherkin:
With the CRC impl change, do we want a different gamedata key so we don't break old versions? Or just avoid using it for a few years?

peace-maker:
Since this was undocumented ever since, we could just say it's a "new" feature. You're required to update to a newer SourceMod build to use it. It's partially broken due to the hardcoded server.so path on older versions anyway and nobody complained, which seems to be a good indicator, that this isn't used anywhere.

asherkin:
I guess as long as we're always restricting the "old" gamedata rather than the "new" gamedata that does work!

psychonic:
> I guess as long as we're always restricting the "old" gamedata rather than the "new" gamedata that does work!

I think that's the only way we'd plan to use it. That way it's less critical to go back and edit later.

asherkin (approved):

PR #1625, opened 2021-11-01
Bring languages into the tree

KyleSanderson:
The translation tool has been an excellent resource since its inception and has added atleast 30 languages to the Source ecosystem. However, since moving to Git from Mercurial (and other source versioning tools) it's received less and less love and has been a cause of concern to merging PRs due to i18n issues as these files do not live in the tree.

This PR brings the latest languages into the tree in hopes of avoiding situations where merge requests get stalled out. There's a buildbot step that needs to be removed before this is merged to no longer pull from the other database, centralizing the language database to the repository.


asherkin (commented):
Is this a direct export from the translator tool or pulled from a SM release?

Unmodified?

Any idea if partial translations are all included here?

KyleSanderson:
> Is this a direct export from the translator tool or pulled from a SM release?
> 
> Unmodified?
> 
> Any idea if partial translations are all included here?

This is from `1.11.0-git6804-linux` as-is. I haven't watched the tree for a minute but the last time the directory was changed (in git) was 2 years ago so I'm feeling confident it's OK.

peace-maker:
This probably needs an update and the mentioned change to the buildbot packaging script.

When switching to all languages being in the repo, we lose the easy way to see which phrases are missing in which language. We could try to solve that using a CI job which never fails but just prints the missing phrases for each language whenever the translations folder is changed.

Or even let it update a TODO.md and push it into the repo? Can you post in github "projects" from a CI job?

dvander (approved):

Wend4r:
I future propose to completely synchronize files from sourcemod.net to AMBuild package after binaries built (`addons/sourcemod/*.txt` licenses, actual GeoIP2 database and maybe more)

peace-maker:
Sorry, I didn't get this. What are you suggesting?

Wend4r:
> Sorry, I didn't get this. What are you suggesting?

In simple terms,
so that after the manual `ambuild`, the `package/**` is same as on https://www.sourcemod.net/downloads.php?branch=dev (Of course, excluding the other current compiler and distribution ABI compatibility on which `ambuild` was executed)

Wend4r:
Or I suggest making a proxy mirror from GitHub to AlliedModders backend (auto-updater and buildbot) so that changes in files do not conflict

Issue #1414, opened 2020-12-28
Extend sig scanning for other libraries


dragokas:
Hi,
Is there a reason why sig scanning is only limited to "server", "engine" and "matchmaking_ds" libraries?
Can you extend it to be able to use other loaded libs, like "tier0" etc. in gamedata?
Thank you.

PR #1626, opened 2021-11-02
Add support for other engine binaries in game configs (#1414).

psychonic:


core/logic/GameConfigs.cpp
```diff
@@ -1231,3 +1211,54 @@ void GameConfigManager::RemoveCachedConfig(CGameConfig *config)
...
+
+	// Don't bother trying to get CRC if we couldn't find the bin loaded
+	if (info.m_pAddr)
+	{
+		FILE* fp;
+		char path[PLATFORM_MAX_PATH];
+		char name[64];
+		bridge->FormatSourceBinaryName(pszName, name, sizeof(name));
+
+		// This pattern doesn't work for every game. Instead, we need to use the GAMEBIN search path properly
```
peace-maker, line 1228: Is this a TODO? Or does this magically open the `$BASE/bin/engine.so` as well as the `$BASE/csgo/bin/server.so`?

peace-maker (commented):
This is cool, it's easier to hook stuff in e.g. tier0 now. I'm unsure about the CRC refactoring though.

core/logic/GameConfigs.cpp
psychonic: Yes and no... and good catch.

The TODO was in regard to games that don't use standard paths (/bin, /<gamedir>/bin). However, you're absolutely right that it's not accounting for the far more common case of all of the files normally just in /bin. I'll have to revisit that. The easiest solution to both problems may be to just open/read the files via IFIleSystemBridge.

psychonic:
Some other issues with this current method of using Valve's filesystem layer is that if MM:S is loaded via the gameinfo method, it's loader will be picked up instead of the game's server.dll.

psychonic:
I updated with a different approach. The aforementioned issues appear to be solved now on Windows. It's currently completely broken on Linux, but it shouldn't be too bad to fix.

peace-maker:
What's missing for linux support here? We could have used this for the optional game update #1766

psychonic:
> What's missing for linux support here? We could have used this for the optional game update #1766

I've now found and fixed the remaining issue on Linux with https://github.com/alliedmodders/sourcemod/pull/1626/commits/67aea1796615b91f2b2f681b58a9419c65c36042

peace-maker (approved):
Could use the same `link` cast as in #576, but limiting the allowed binaries to engine ones seems reasonable given those are the libraries targeted.

psychonic:
> Could use the same `link` cast as in #576, but limiting the allowed binaries to engine ones seems reasonable given those are the libraries targeted.

Cool. That's good info.

I had a small back-and-forth about this in my head. I'm not a fan of just casting the handle since that does not seem to be an API guarantee. However, it looks like `dlinfo` can be safely used to the same end. I then wondered if we shouldn't be relying on the Windows HMODULE to be a pointer in the library either. Rather than then looking up an equivalent method there, I landed on this being fine as-is for now. We can revisit if/when there is a use case.

Alienmario:
Is this supposed to automatically recognize the _srv filename suffix in some linux binaries?

psychonic:
Yes. To clarify, any platform-specific prefixes and/or suffixes are handled automatically. File system case sensitivity can still apply.

PR #981, opened 2019-04-21
Incorporate DHooks into base SourceMod

Headline:
Through discussion in `#sourcemod` we are planning on bringing DHooks in-tree, so I guess I volunteered for the dirty work.

There's still concerns about this and I also have yet to scour DHooks source for possible improvements so this is currently WIP.

PR #1629, opened 2021-11-03
Import DHooks as first-party extension

peace-maker:
Since DHooks with dynamic detouring is widely used by plugin developers nowadays, we decided to adopt the extension as a base feature of SourceMod. #981 made a first proposal of importing the original DHooks extension, but since most users switched to the fork which adds detouring support in addition to dynamic vtable hooking, it seemed better to import the forked version.

This is the latest DHooks version from https://github.com/peace-maker/DHooks2/tree/1314f2d1b4d870677a1b9f628c770f2799878f9b

@Drifter321 and I both agreed to transfer the copyright of the code to Alliedmodders LLC. To implement dynamic detouring for plugins the [DynamicHooks](https://github.com/Ayuto/DynamicHooks) library by @Ayuto was adopted. Thank you again!

This is a straight import of the latest version including the fixes and enhancements that didn't make it into a new release yet. There was no effort made to streamline the implementation with other code inside SourceMod.

Supercedes #981


Headline (approved):

Drifter321:
I have a few questions (I probably looked at but forgot now…)

What happens if two plugins detour the same function? Does one clobber the other?

I think it’s still better to add a detour library to MM:S (along the lines of sourcehook) so we don’t have plugins/extensions clobbering each other.

Also, I can’t remember this. Is it possible to access object param address to get data? I know it’s possible to get object values.

Example object->object->thisval ?

If not, we should consider adding this probably?

on another note, thanks for all your hard work on dhooks over the years!

peace-maker:
> What happens if two plugins detour the same function? Does one clobber the other?

They are added to the plugin callback list of the same detour and get their callbacks called in sequence. The highest `MRES` return value is used. This only works for detours done through DHooks though. Detours through an extension or other server plugin could still cause problems during unloading/reloading similar to the problems with `CDetour`.

> I think it’s still better to add a detour library to MM:S (along the lines of sourcehook) so we don’t have plugins/extensions clobbering each other.

Yes, very true. It just needs someone to do it and DHooks can be moved to that other library when it's available.

> Also, I can’t remember this. Is it possible to access object param address to get data? I know it’s possible to get object values.
> 
> Example object->object->thisval ?

[`DHookParam.GetAddress`](https://github.com/peace-maker/DHooks2/pull/24) was added recently if that's what you're after. I don't think it's possible to nest object parameters though?

🎉

asherkin (approved):
LGTM - would definitely like the sigscan cache in before 1.11 is branched though.

PR #1630, opened 2021-11-04
Add new natives for request/send file from client

Accelerator74:
I thought these natives would be useful in sourcemod. They can help validate client files (`RequestFile`) or additionally identify them by unique files (`SendFile`) for each client. In addition, sdktools includes forwards `OnFileSend` and` OnFileReceive`.


KyleSanderson:
I'm under the impression that this doesn't work on newer games like csgo. Is that wrong?

Accelerator74:
This works for csgo. But in order for the server to download files, the `sv_allowupload` value must be equal to 1. Also, in order for the client to receive files from the server, the client needs to bring the `sv_allowupload` cvar to the value 1. In general, the functions work fine :+1:
test

extensions/sdktools/vnatives.cpp
```diff
@@ -1638,6 +1638,60 @@ static cell_t SetEntityOwner(IPluginContext *pContext, const cell_t *params)
...
+	}
+
+	char path[PLATFORM_MAX_PATH];
+	g_pSM->FormatString(path, sizeof(path), pContext, params, 2);
+
+#if SOURCE_ENGINE < SE_CSGO
+	return pNetChan->RequestFile(path);
+#else
+	return pNetChan->RequestFile(path, false);
+#endif
```
asherkin, line 1692: Returning the transfer ID doesn't seem valuable unless we expose it elsewhere - is there value in adding it to the forwards for cross-referencing with requests here, or just ignore it?

plugins/include/sdktools_functions.inc
```diff
@@ -377,3 +377,23 @@ native void EntityCollisionRulesChanged(int entity);
  * @error                   Invalid entity or lack of mod support.
  */
 native void SetEntityOwner(int entity, int owner=INVALID_ENT_REFERENCE);
+
+/**
+ * Sends a file to a client
+ *
+ * @param client          Client to send file.
+ * @param file            File to send.
+ * @param ...             Variable number of format parameters.
```
asherkin, line 386: It's not obvious what these are format parameters for - these natives being format-class seems wrong though, `BuildPath` should be used for paths in many cases and it would be very easy to break the format rules here and pass a non-literal as the format string.

asherkin (changes requested):
There are at least two extensions that expose these functions for plugins already (LateDL and FileNetMessages), do we have any numbers on how widely those are used?

How do these natives interact with the existing forwards in SM? Particularly when the forward denies the transfer.

A user on Discord had a bunch of performance issues with the existing extensions that expose these functions due to not having a way to detect if a file transfer was already in progress, do we need to expose that here to make these usable? https://discord.com/channels/335290997317697536/335290997317697536/904900799778410576

Accelerator74:
I think you are right that adding `RequestFile` and `SendFile` natives is not practical. These are very narrowly focused functions, which, moreover, do not work quite correctly everywhere, besides, we can still limit yourself to the extensions you mentioned. 👍

PR #1637, opened 2021-11-12
Lift raw pointers out of DatabaseConfBuilder

Headline:
This patch `ke::RefPtr`'s the database configs in storage instead of storing raw `ConfDbInfo*`s. I'm not certain if the original form caused any problems, but in my recent review of this area I was very dissatisfied.

Also, the config lists have been converted to be value types instead of pointers. I'm not certain why I made these lists pointers, but this patch overall makes this section much safer.

untested

### DIFF
--- a/core/logic/DatabaseConfBuilder.h
+++ b/core/logic/DatabaseConfBuilder.h
@@ -38,6 +38,8 @@
 
 #include <am-vector.h>
 #include <am-string.h>
+#include <am-maybe.h>
+#include <am-refcounting.h>
 #include <am-refcounting-threadsafe.h>
 
 class ConfDbInfo : public ke::RefcountedThreadsafe<ConfDbInfo>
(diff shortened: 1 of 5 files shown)


core/logic/DatabaseConfBuilder.h
```diff
@@ -66,37 +68,34 @@ class ConfDbInfoList : public std::vector<ConfDbInfo *>
 		return m_DefDriver;
 	}
 	
-	ConfDbInfo *GetDatabaseConf(const char *name) {
+	ke::Maybe<ke::RefPtr<ConfDbInfo>> GetDatabaseConf(const char *name) {
```
dvander, line 71: Is Maybe<> needed here? null refptr would imply the same thing.

peace-maker:
Were you able to test your changes?

Headline:
> Were you able to test your changes?

@peace-maker Just did - looks okay to me!

Headline:
@peace-maker I apologize for the poke, but does this look good to you?

peace-maker (approved):

peace-maker:
Sorry for the delay. LGTM

PR #1638, opened 2021-11-13
No reason docs for TF2

Naleksuh:


JoinedSenses:
The kick reason is still part of the event, so it could be useful to keep the reason as an arg - it just wont be displayed to the client.

Naleksuh:
> The disconnect reason is still part of the event, so it could be useful to keep the reason as an arg for logging - it just wont be displayed to the client.

The reason hasn't gone away, I simply altered the help text to be more accurate. People can still log if they want to (although I'd consider that not a good enough reason either, especially when SM has its own logs)

> Also, your formatting isn't consistent with the rest of SM plugins

Looks pretty much identical to me. Unless you mean how I indents/line endings, in which case maybe less so, I don't know much about it.

JoinedSenses:
> The reason hasn't gone away, I simply altered the help text to be more accurate. People can still log if they want to (although I'd consider that not a good enough reason either, especially when SM has its own logs)

It would be misleading to remove `[reason]` from the response, even though that argument is still valid.

asherkin:
The reason still gets logged for other admins even though it's not included in the client disconnect message currently, there is actually nothing changed here indicating that it would be displayed to clients.

Valve have said they're looking to restore the kick message functionality, so let's wait for them either way before making changes.

PR #1639, opened 2021-11-15
Add Support for Direct Arrays into OnMenuVoteResults handler

PerfectLaugh:
Support 1.11 compiled plugins for VoteMenu handlers

### DIFF
--- a/core/logic/smn_menus.cpp
+++ b/core/logic/smn_menus.cpp
@@ -501,68 +501,121 @@ void CMenuHandler::OnMenuVoteResults(IBaseMenu *menu, const menu_vote_result_t *
 		bool no_call = false;
 		int err;
 
-		/* First array */
 		cell_t client_array_address = -1;
 		cell_t *client_array_base = NULL;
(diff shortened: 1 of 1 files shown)


dvander:
Thanks for catching this. I apparently failed to sweep the include files properly for uses of double-arrays.

There is so much complexity here, I think it would be good to move the responsibility over to the VM. It would simplify this a lot. I'll try to do that sometime this week.

PerfectLaugh:
I think it is good enough to keep this situation on.
The core_logic and extension part are still required to using underlying array anyway.
Even we created the API for that operation, extensions that use 2D-Arrays are still requires to modify.

dvander:
They need modification, but this is extremely tricky code that is easy to get wrong. And it doesn't make sense to embed it right into the menu code (which is already very complicated). So I'm happy to do some refactoring to clean this up.

dvander:
I just posted https://github.com/alliedmodders/sourcemod/pull/1643 to fix this, but I'm unable to test at the moment. Any help would be appreciated.

PerfectLaugh:
Closed in favor of #1643

PR #1640, opened 2021-11-15
Fix translation buffers not terminating at last cell

### LINKED ISSUE #1594: Translated buffers are early null terminated
I've just noticed what seems to be an important issue.

When holding a string buffer of size `N`, formatting it without involving translations allow for the null terminator to be placed at the last cell as much (`N - 1`), correctly. However, when a translation is used while formatting, the null terminator is written as much late at the `N - 2` cell. This causes to always render one cell of the max. length useless.

See attached picture for an illustrative sample.

Captura

Thank you.

Adrianilloo:
Fixes #1594 per @dvander instructions and self tests

### DIFF
--- a/core/logic/sprintf.cpp
+++ b/core/logic/sprintf.cpp
@@ -1218,7 +1218,7 @@ reswitch:
 				cell_t *target;
 				pCtx->LocalToString(params[arg++], &key);
 				pCtx->LocalToPhysAddr(params[arg++], &target);
-				res = Translate(buf_p, llen, pCtx, key, *target, params, &arg, &error);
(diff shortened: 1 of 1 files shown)


asherkin:
Please could you rebase and retarget this PR against the master branch.

Adrianilloo:
Ok, followed your steps now

asherkin (approved):
Thanks - LGTM

Adrianilloo:
Will this be made into 1.10 as well?

asherkin:
> Will this be made into 1.10 as well?

We're not currently backporting non-critical things to 1.10 as we're in the final push for 1.11 - please do upgrade to 1.11 and test it out!

PR #1642, opened 2021-11-16
Cache copy of library on first symbol/signature lookup

peace-maker:
Always perform signature searches on an unaltered copy of the binary. This avoids signature mismatches if the same function is detoured twice and thus the first bytes of the function were replaced by the detour. It is implemented by creating a copy of the current (hopefully unaltered) state of the library on first symbol or signature lookup through `IMemoryUtils` and using that copy to find matching signatures.

Now you're able to find signatures even if something else already messed with the bytes. Before this patch, code like this would fail to locate the signature again after the memory was altered (assuming both gamedata files are identical).

```sourcepawn
public void OnPluginStart() {
    GameData gd = new GameData("test1.games");
    Address testfunc = gd.GetMemSig("testfunc");
    PrintToServer("Found sig: %x", testfunc);
    delete gd;

    StoreToAddress(testfunc, 0xff, NumberType_Int32);

    gd = new GameData("test2.games");
    testfunc = gd.GetMemSig("testfunc");
    PrintToServer("Found sig again: %x", testfunc);
    delete gd;
}
```

So it's not necessary to wildcard the first 6 bytes of a signature to still find it even if the function was detoured by something else. This is only true for SourceMod and its extensions or rather anything using the `IMemoryUtils` to find a signature in memory. It'd be desired to have such an interface exposed in MetaMod:Source for all other server plugins to use in the future.

### DIFF
--- a/core/logic/MemoryUtils.cpp
+++ b/core/logic/MemoryUtils.cpp
@@ -49,6 +49,7 @@ MemoryUtils g_MemUtils;
 
 MemoryUtils::MemoryUtils()
 {
+	m_InfoMap.init();
 #ifdef PLATFORM_APPLE
 
 	task_dyld_info_data_t dyld_info;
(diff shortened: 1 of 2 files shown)


asherkin (approved):

dragokas:
This is the genial and demanded thing.

> Cache copy of library on first symbol/signature lookup
When does the first symbol/signature lookup happens by default (except GetMemSig e.t.c.)?

I mean, if e.g. I manually detour some function (not via dhooks) from the plugin context, does the most common libs like server/engine had already been precached?

dragokas:
Ok, I see. It only preserves a single symbol by FindPattern request.

peace-maker:
The current state of the whole library in memory will be preserved whenever the first call to `IMemoryUtils::FindPattern` or `IMemoryUtils::ResolveSymbol` happens on it. If there were other modifications on the library beforehand where the address wasn't looked up through one of the two functions that modification is preserved as well, but that only applies to other metamod plugins or Valve Server Plugins not using `IMemoryUtils`.

So this isn't linked to dhooks at all, but only to the address lookup process. You can assume that SourceMod itself already performed some lookup and triggered the caching before the first sourcepawn plugin code runs, yes.

PR #1643, opened 2021-11-18
Fix menu vote handler to work with direct arrays.

dvander:


core/logic/smn_menus.cpp
```diff
@@ -496,99 +492,61 @@ void CMenuHandler::OnMenuVoteResults(IBaseMenu *menu, const menu_vote_result_t *
...
+		if (!pContext->HeapAlloc2dArray(results->num_items, 2, &item_array_address, init.get())) {
+			g_DbgReporter.GenerateError(pContext, m_fnVoteResult, -1,
+										"Menu callback could not allocate %d cells for item list.",
+										item_array_size);
+			ok = false;
 		}
 	}
+
+	/* Finally, push everything */
+	if (!ok) {
```
PerfectLaugh, line 534: This is not okay

core/logic/smn_menus.cpp
```diff
@@ -496,99 +492,61 @@ void CMenuHandler::OnMenuVoteResults(IBaseMenu *menu, const menu_vote_result_t *
...
-			m_pVoteResults->PushCell(menu->GetHandle());
-			m_pVoteResults->PushCell(results->num_votes);
-			m_pVoteResults->PushCell(results->num_clients);
-			m_pVoteResults->PushCell(client_array_address);
-			m_pVoteResults->PushCell(results->num_items);
-			m_pVoteResults->PushCell(item_array_address);
-			m_pVoteResults->Execute(NULL);
+		if (!pContext->HeapAlloc2dArray(results->num_clients, 2, &client_array_address, init.get())) {
+			g_DbgReporter.GenerateError(pContext, m_fnVoteResult, -1,
+										"Menu callback could not allocate cells for client list.");
```
PerfectLaugh, line 512: better to add %d into the message

PerfectLaugh (changes requested):
Have to fix the sourcepawn with this PR https://github.com/alliedmodders/sourcepawn/pull/716

core/logic/smn_menus.cpp
dvander: The number is kind of meaningless. It's not accurate given that we allocate metadata behind the scenes, and at most it'll be a pretty small number.

dvander:
Latest patch uses AutoEnterHeapScope instead of HeapPop, which is now basically deprecated.

asherkin (approved):
I assume this doesn't need the SP pin update any more?

Headline (approved):

PR #1648, opened 2021-11-22
Exposed SDKCall_Engine call type to use CVEngineServer methods

accelerator74:
Similar to https://github.com/alliedmodders/sourcemod/pull/1612 :)

I think this will be useful when used in plugins, where a pointer engine is needed. It will be convenient to use for vtable offsets to make it easier to call the necessary functions.

### DIFF
--- a/extensions/sdktools/vcaller.cpp
+++ b/extensions/sdktools/vcaller.cpp
@@ -336,15 +336,25 @@ static cell_t SDKCall(IPluginContext *pContext, const cell_t *params)
 			}
 			break;
 		case ValveCall_Server:
-            {
-                if (iserver == NULL)
-                {
-                    vc->stk_put(ptr);
-                    return pContext->ThrowNativeError("Server unsupported or not available; file a bug report");
-                }
-                *(void **)ptr = iserver;
-            }
-            break;
+			{
+				if (iserver == NULL)
+				{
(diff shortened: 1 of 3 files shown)


Accelerator74:
My mistake :( Didn't know it was `engine = iserver`

psychonic:
> My mistake :( Didn't know it was `engine = iserver`

?

They are not the same. This was valid.

Accelerator74:
I checked some functions and they work with `SDKCall_Server` too. For example, `GetPlayerNetInfo`.

Accelerator74:
Looked at this function again. It turns out that pointer is simply not very important to her, so in general the function is called as it should. Well, I'll reopen this PR if you think it should be :)

psychonic:
> Looked at this function again. It turns out that pointer is simply not very important to her, so in general the function is called as it should. Well, I'll reopen this PR if you think it should be :)

Yes, they are separate interfaces with separate interface pointers. The PR is valid.

Accelerator74:
In general, there is little usefulness in this PR. There are very few functions using this pointer, and in most of them it itself does not play any role. It will be convenient to use it only for vtable offsets to make it easier to call the necessary functions.

peace-maker (approved):
Less gamedata for plugins through providing readily available pointers is great.

Issue #1555, opened 2021-07-29
Expose 'CBaseAnimating::GetAttachment' function


KoNLiG:
# Help us help you
  - [x] I have checked that my issue [doesn't exist yet](https://github.com/alliedmodders/sourcemod/issues).
  
# Description

As discussed [here](https://discord.com/channels/335290997317697536/335290997317697536/870079382079799317) there is a function to retrieve an attachment origin & angles called `CBaseAnimating::GetAttachment`.

It would be nice if a native will be added to access this function.

asherkin:
This is a very common SDKCall used in plugins and I think it'd be a great idea to have a native for it.

Marking as good-first-issue as the implementation should be relatively straightforward for anyone familiar with SDKCall looking to get a start in C++ land - we have a lot of very similar natives already that can be cribbed from.

PR #1653, opened 2021-11-26
Add LookupEntityAttachment & GetEntityAttachment natives

Mikusch:
#1555

> This is a very common SDKCall used in plugins and I think it'd be a great idea to have a native for it.

I agree.

## Overview

This PR adds two new natives:

```cpp
// Same as SDKCall to:  int CBaseAnimating::LookupAttachment(const char *)
native int LookupEntityAttachment(int entity, const char[] name);

// Same as SDKCall to:  bool CBaseAnimating::GetAttachment(int, Vector &, QAngle &)
native bool GetEntityAttachment(int entity, int attachment, float origin[3], float angles[3]);
```

Example usage:

```cpp
int attachment = LookupEntityAttachment(client, "head");
if (attachment == 0)
	return;

float origin[3], angles[3];
if (!GetEntityAttachment(client, attachment, origin, angles))
	return;

PrintToServer("Attachment %d at: (%f %f %f)", attachment, origin[0], origin[1], origin[2]);
```

Gamedata for CS:GO, CS:S, L4D, L4D2 and TF2 is provided, but has not been tested by me.

## Implementation

Using the virtual `CBaseAnimating::GetAttachment(int, matrix3x4_t &)` was a deliberate choice because virtual offsets are generally easier to maintain than signatures. The `matrix3x4_t` is converted to world position and world angles internally.
Some of the other overloads are also inlined on a few games, making this the best choice.

Since this call can only be used on classes inheriting `CBaseAnimating`, we check if the `DT_BaseAnimating` SendTable exists on the entity, throwing a native error if it doesn't.
This safeguard could be greatly improved with a call to `CBaseEntity::GetBaseAnimating`, but would require more gamedata (maybe something to consider for the future?)

## Footnote

I have not been able to test this for any game other than TF2. I'd be grateful if other people could assist with testing
this on different games and engine versions.

This is my first time actively working with C++, so excuse some rookie mistakes.


Headline (approved):
Tested for CS:GO Windows, functions properly. 

Looks like a great patch, thank you for seeing this through!

extensions/sdktools/vnatives.cpp
```diff
@@ -1638,6 +1643,113 @@ static cell_t SetEntityOwner(IPluginContext *pContext, const cell_t *params)
 	return 1;
 }
 
+static cell_t LookupEntityAttachment(IPluginContext* pContext, const cell_t* params)
+{
+	CBaseEntity* pEntity;
+	ENTINDEX_TO_CBASEENTITY(params[1], pEntity);
+
+	ServerClass* pClass = ((IServerUnknown*)pEntity)->GetNetworkable()->GetServerClass();
```
KyleSanderson, line 1651: This needs a null check on networkable.

extensions/sdktools/vnatives.cpp
```diff
@@ -1638,6 +1643,113 @@ static cell_t SetEntityOwner(IPluginContext *pContext, const cell_t *params)
...
+
+	return ret;
+}
+
+static cell_t GetEntityAttachment(IPluginContext* pContext, const cell_t* params)
+{
+	CBaseEntity* pEntity;
+	ENTINDEX_TO_CBASEENTITY(params[1], pEntity);
+
+	ServerClass* pClass = ((IServerUnknown*)pEntity)->GetNetworkable()->GetServerClass();
```
KyleSanderson, line 1697: Ditto

extensions/sdktools/vnatives.cpp
```diff
@@ -1638,6 +1643,113 @@ static cell_t SetEntityOwner(IPluginContext *pContext, const cell_t *params)
...
+	if (!FindNestedDataTable(pClass->m_pTable, "DT_BaseAnimating"))
+	{
+		return pContext->ThrowNativeError("Entity %d (%d) is not a CBaseAnimating", gamehelpers->ReferenceToIndex(params[1]), params[1]);
```
KyleSanderson, line 1700: In?

KyleSanderson (changes requested):
Nice one - thank you for the contribution. Can you add sdk2013 Gamedata outside of the comments? The networkable stuff was a big issue with ents a decade ago.

extensions/sdktools/vnatives.cpp
```diff
@@ -1638,6 +1643,113 @@ static cell_t SetEntityOwner(IPluginContext *pContext, const cell_t *params)
 	return 1;
 }
 
+static cell_t LookupEntityAttachment(IPluginContext* pContext, const cell_t* params)
+{
+	CBaseEntity* pEntity;
+	ENTINDEX_TO_CBASEENTITY(params[1], pEntity);
+
+	ServerClass* pClass = ((IServerUnknown*)pEntity)->GetNetworkable()->GetServerClass();
+	if (!FindNestedDataTable(pClass->m_pTable, "DT_BaseAnimating"))
```
KyleSanderson, line 1652: @asherkin am I losing it or is this not calling back to the safe nesting datatable code from core?

extensions/sdktools/vnatives.cpp
Mikusch: Could you elaborate what you would like changed here?

Mikusch:
> Can you add sdk2013 Gamedata outside of the comments?

What exactly do you mean by this? There isn't any gamedata that would make sense to add for sdk2013.

extensions/sdktools/vnatives.cpp
Kenzzer: The networkable null check should be unnecessary on this.
IServerUnknown is inherited by IServerEntity which itself is inherited by CBaseEntity, and if we look at the function implementation by valve : 
```cpp
inline IServerNetworkable *CBaseEntity::GetNetworkable()
{
	return &m_Network;
}
```
And CBaseEntity definition
```cpp
private:
	// NOTE: Keep this near vtable so it's in cache with vtable.
	CServerNetworkProperty m_Network;
```
The networkable interface should never be null. And valve also doesn't seem to ever null check this in their code base
https://github.com/ValveSoftware/source-sdk-2013/search?q=GetNetworkable

It's also used under CBaseEntity constructor regardless of whether it's server side only or not
```cpp
CBaseEntity::CBaseEntity( bool bServerOnly )
{
	// code
	NetworkProp()->Init( this );
	// code
}
inline CServerNetworkProperty *CBaseEntity::NetworkProp()
{
	return &m_Network;
}
```
https://github.com/ValveSoftware/source-sdk-2013/blob/0d8dceea4310fde5706b3ce1c70609d72a38efdf/mp/src/game/server/baseentity.cpp#L367
https://github.com/ValveSoftware/source-sdk-2013/blob/0d8dceea4310fde5706b3ce1c70609d72a38efdf/mp/src/game/server/baseentity.h#L2378

plugins/include/sdktools_functions.inc
```diff
@@ -377,3 +379,25 @@ native void EntityCollisionRulesChanged(int entity);
  * @error                   Invalid entity or lack of mod support.
  */
 native void SetEntityOwner(int entity, int owner=INVALID_ENT_REFERENCE);
+
+/**
+ * Returns the index number of a given named attachment.
+ * 
+ * @param entity        The entity index.
+ * @param name          The attachment name.
+ * @return              An attachment index on success, INVALID_ATTACHMENT otherwise.
```
asherkin, line 388: I feel like we should just specify this returns 0 on failure rather than adding a define for it, we can't type it correctly and code is simpler if it can just check truthiness (which people are going to do anyway with `INVALID_ATTACHMENT` being 0).

Probably "The attachment index or 0 if the attachment name is invalid or unused"?

asherkin (changes requested):
This all looks good to me, the error handling and datatable bits all seem fine - one tiny comment on how the API is presented but otherwise I'd say the meat of this is all good to go.

plugins/include/sdktools_functions.inc
Mikusch: Thanks for the feedback, this has been changed in 0bd4b04

asherkin (approved):
Thanks!

asherkin:
@KyleSanderson please could you re-review this or dismiss your stale review.

mixern6:
I might have an idea how to get attachment points without using signatures, though can't check if that will work:
Get IMDLCache* from MDLCACHE_INTERFACE_VERSION
That will provide methods to get studiohdr_t*, studiohwdata_t*, vcollide_t*, virtualmodel_t *, vertexFileHeader_t* by model name.
Providing the structures and the methods are correct, it should be possible to get attachment names this way: studiohdr_t::pAttachment(i).pszName( )

PR #1657, opened 2021-11-28
Change ShouldCollide to a posthook

### LINKED ISSUE #1650: ShouldCollide originalResult different with SM 1.11
Per updating my SM install from 1.10 to 1.11 recently, the value of `originalResult` changed from `true` to `false`, as of my plugin usage:

```
SDKHook(client, SDKHook_ShouldCollide, Hook_ShouldCollide);

// .......

public bool:Hook_ShouldCollide(entity, collisiongroup, contentsmask, bool:originalResult)
{
	return (isKillProtected[entity] && bNoBlock) ? false : originalResult;
}
```

What the value of `originalResult` depends on? Is it possible that something is wrong from SM?

I had noted improper collisions were being filtered out instead of succeeding after upgrading, breaking my server (players passing through bullets and other players). And for now I replaced `originalResult` with true. I was assumming the value of it was the one from the game function, but I see that doesn't make sense on a non-post hook, right?

This is the only plugin that hooks `ShouldCollide` in my server.

Kenzzer:
Fixes #1650

As pointed by @asherkin over discord the use of `META_RESULT_OVERRIDE_RET` or `META_RESULT_ORIG_RET` within a non post hook is wrong. This PR fixes that.

I'm not super familiar with sourcehook, so my fix could be horribly wrong. If I'm wrong, at the very least this PR will kick off a discussion? I was also tempted to replace the `META_RESULT_ORIG_RET` line with an `SH_CALL` instead, but this would end to essentially turning the function into a post hook.

_Smallest PR in the small west ?_

### DIFF
--- a/extensions/sdkhooks/extension.cpp
+++ b/extensions/sdkhooks/extension.cpp
@@ -729,7 +729,7 @@ HookReturn SDKHooks::Hook(int entity, SDKHookType type, IPluginFunction *callbac
 				hookid = SH_ADD_MANUALVPHOOK(Weapon_Switch, pEnt, SH_MEMBER(&g_Interface, &SDKHooks::Hook_WeaponSwitchPost), true);
 				break;
 			case SDKHook_ShouldCollide:
-				hookid = SH_ADD_MANUALVPHOOK(ShouldCollide, pEnt, SH_MEMBER(&g_Interface, &SDKHooks::Hook_ShouldCollide), false);
+				hookid = SH_ADD_MANUALVPHOOK(ShouldCollide, pEnt, SH_MEMBER(&g_Interface, &SDKHooks::Hook_ShouldCollide), true);
 				break;
 			case SDKHook_Blocked:
 				hookid = SH_ADD_MANUALVPHOOK(Blocked, pEnt, SH_MEMBER(&g_Interface, &SDKHooks::Hook_Blocked), false);


asherkin:
This fix looks correct to me, but I'd like it to get some testing around the behaviour of ShouldCollide before merging.

Kenzzer:
Will try to test this weekend then. Though by behavior what exactly are you looking into being tested ? 

All I can think right now, is retrieving the address of `CBaseEntity::ShouldCollide` make an SDKCall of it. (no virtual call)
Then write a plugin that makes use of SDKHook_ShouldCollide, and verify that the boolean parameter in the callback, matches what I get from the SDKCall. Then testing this plugin on current version of SM, then again but with this PR's change.
And we should get at least one random mismatch without the fix, and the other no mismatch ever.

I believe this would satisfy testing the behaviour ?

Mikusch:
I've tested this on TF2 Windows.

**Without** the changes in this PR, a couple entities report a mismatch between SDKHook's `originalResult` and the return value of the SDKCall.
The only thing all of these entities have in common is that their collision group is set to `COLLISION_GROUP_DEBRIS`, but none of them actually override `CBaseEntity::ShouldCollide`.

**With** the changes in this PR, the only entities that report a mismatch are the ones that override `CBaseEntity::ShouldCollide`.

TF2 Windows sig in case someone else would like to test: `\x55\x8B\xEC\x83\xB9\xF0\x01\x00\x00\x01`

asherkin (approved):

KyleSanderson:
> I've tested this on TF2 Windows.
> 
> **Without** the changes in this PR, a couple entities report a mismatch between SDKHook's `originalResult` and the return value of the SDKCall. The only thing all of these entities have in common is that their collision group is set to `COLLISION_GROUP_DEBRIS`, but none of them actually override `CBaseEntity::ShouldCollide`.
> 
> **With** the changes in this PR, the only entities that report a mismatch are the ones that override `CBaseEntity::ShouldCollide`.
> 
> TF2 Windows sig in case someone else would like to test: `\x55\x8B\xEC\x83\xB9\xF0\x01\x00\x00\x01`

Thank you very much for your feedback

PR #1662, opened 2021-12-11
Update rulesfix.cp

xMaZax:


psychonic:
What is the purpose of this pull request?

The Steam API has backward compatibility so the old interface versions are always available. And the SOURCE_ENGINE define will always be equal to SE_CSGO here because this file only gets compiled into the CS:GO build.

xMaZax:
L 12/12/2021 - 02:00:39: [CSTRIKE] [CStrike] Failed to get SteamAPI_SteamGameServer_v013 function

PR #1665, opened 2021-12-12
[Buildbot] Delete package after successful upload.

psychonic:
Delete package file after successful upload rather than eventually running out of space.

### DIFF
--- a/tools/buildbot/package.pl
+++ b/tools/buildbot/package.pl
@@ -147,7 +147,13 @@ $ftp->put($tmpfile, $latest)
 
 $ftp->close();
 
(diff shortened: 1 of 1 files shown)


asherkin (approved):
How hard is adding error handling for the `system` calls?

psychonic:
> How hard is adding error handling for the `system` calls?

I think that the best way to handle this is with the [autodie module](https://perldoc.perl.org/autodie) rather than continuing to add the `or die` on every call. However, that's going to requiring some maintenance to the Perl installs on our builders. I've started on that now and I think it's mostly ready, but I don't have time to finish up it yet.

PR #1667, opened 2021-12-13
Fix getting SteamAPI_SteamGameServer

komashchenko:
There no backward compatibility in ``flat api`` and only latest version of interface will always be available.

> Notice: together with ``1.38.1.2`` update, ``Steamworks SDK`` was also updated, fix is needed in this or any other form.


psychonic:
> There no backward compatibility in `flat api`

I think it's safe for us to switch to the C++ API now, which does have backward compatibility built-in. I believe the usage of the flat API was a temporary measure as the SteamWorks SDK that CS:GO used at one time was a never-released version during a time when the interface accessors and compatibility were in a state of change (around the 1.38-1.40 time).

PR #1668, opened 2021-12-13
Use SteamWorks C++ API in CStrike RulesFix, for better interface compatibility

psychonic:
See https://github.com/alliedmodders/sourcemod/pull/1662 and https://github.com/alliedmodders/sourcemod/pull/1667

At the time of the implementation of the "RulesFix" module in the CStrike extension, there was very poor forward and backward compatibility with the SteamWorks interfaces in the version of the SW SDK used by CS:GO (1.39a). Because of that, the "flat" C API for it was used. However, it turns out that this also has poor backward compatibility.

Since then, the SW C++ API has gained much better backward compatibility and pins the interface version at compile time, per library. And all interface versions do continue to be available internally in the Steam client. Now we can simplify this code while also fixing the current problem with it not finding the function.

This requires version 1.40 or later of the SteamWorks SDK for full compatibility. Our CS:GO SDK has been [synced to v1.53a](https://github.com/alliedmodders/hl2sdk/commit/3466d3b425f26f2e3071da71713d5f7c37dc4817).

### DIFF
--- a/extensions/cstrike/rulesfix.cpp
+++ b/extensions/cstrike/rulesfix.cpp
@@ -39,9 +39,6 @@ bool bPatched = false;
 
 RulesFix rulesfix;
 
-ISteamGameServer *(*SteamAPI_SteamGameServer)();
-void (*SteamAPI_ISteamGameServer_SetKeyValue)(ISteamGameServer *self, const char *pKey, const char *pValue);
-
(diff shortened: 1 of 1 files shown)


asherkin (approved):

komashchenko (changes requested):
https://github.com/alliedmodders/sourcemod/blob/7febbe0da9c4364dcfede9eee7d5af42881265ea/extensions/cstrike/rulesfix.cpp#L80-L95

psychonic:
> I think this can be removed

Yes, thanks.

Issue #1670, opened 2021-12-14
error?


UnrealKaraulov:
CRITICAL ERROR 🐕‍🦺 https://github.com/alliedmodders/sourcemod/blob/547ac5b02621a6fdd6dc789a7079ab16a4a82dcf/extensions/sdkhooks/natives.cpp#L299
???

PR #1671, opened 2021-12-14
Fix issue #1670

naydef:
Out-of-bounds write
    Not tested If it actually fixes the issue the [issue](https://github.com/alliedmodders/sourcemod/issues/1670) reporter actually
    experiences...
    Not sure why there are other changes written. Probably line endings


dvander:
Did line endings change or something? There's a lot of noise in this patch.

naydef:
Yea, probably line endings. I'm not sure how change the file without my editor doing all of the other changes
Probably will try with hex editor :)

naydef:
Will try again

UnrealKaraulov:
@naydef I tried to change file via GitHub web editor, and got same problem))))

PR #1673, opened 2021-12-18
Add functions for working with entity lumps

### LINKED ISSUE #1547: Add functions for working with entity lumps
With the merge of #1534 there is currently no way to access the level entity string.  [LevelKeyValues](https://github.com/nosoop/SM-LevelKeyValues) among others was built around having that ability.

Here's my proposal for a new API to deal with entity lumps. @asherkin thoughts?

(I'll be writing a draft implementation of it as a SourceMod extension, so input on the API design is greatly appreciated.)
```sourcepawn
/**
 * A handle to a weakref of a list of key / value pairs; any write operations on this handle will affect the resulting map entity string.
 * If the entry in the EntityLump is removed, the handle will error on all operations.  (The handle will still need to be deleted.)
 */
methodmap EntityLumpEntry < Handle {
	/**
	 * Copies the key / value at the given index into buffers.
	 *
	 * @error Index is out of bounds.
	 */
	native void Get(int index, char[] keybuf = "", int keylen = 0, char[] valbuf = "", int vallen = 0);
	
	/**
	 * Updates the key / value pair at the given index.
	 * NULL_STRING may be used to preserve the existing value.
	 *
	 * @error Index is out of bounds or entity lump is read-only.
	 */
	native void Update(int index, const char[] key = NULL_STRING, const char[] value = NULL_STRING);
	
	/**
	 * Inserts a new key / value pair at the given index, shifting the existing entry and ones past it up.
	 * If EntityLumpEntry.Length is passed in, this is an append operation.
	 *
	 * @error Index is out of bounds or entity lump is read-only.
	 */
	native void Insert(int index, const char[] key, const char[] value);
	
	/**
	 * Removes the key / value pair at the given index, shifting all entries past it down.
	 *
	 * @error Index is out of bounds or entity lump is read-only.
	 */
	native void Erase(int index);
	
	/**
	 * Inserts a new key / value pair at the end.
	 *
	 * @error Index is out of bounds or entity lump is read-only.
	 */
	native void Append(const char[] key, const char[] value);
	
	/**
	 * Returns the next index with the matching key, or -1 if not found.
	 *
	 * NOTE: Just a convenience function; probably fine to omit in favor of .Get()
	 */
	native int FindKey(const char[] key, int start = -1);
	
	/**
	 * Returns the number of key / value pairs in the list.
	 */
	property Length {
		native int get();
	};
};

/**
 * A group of natives for a singleton entity lump.
 * EntityLumpEntry instances are only available for writing during OnMapInit().
 */
methodmap EntityLump {
	/**
	 * Returns the EntityLumpEntry at the given index.
	 * This handle should be freed by the calling plugin.]
	 *
	 * @error Index is out of bounds.
	 */
	static native EntityLumpEntry Get(int index);
	
	/**
	 * Erases an EntityLumpEntry at the given index.
	 * Any handles referencing that EntityLumpEntry are invalidated.
	 *
	 * @error Index is out of bounds or entity lump is read-only.
	 */
	static native void Erase(int index);
	
	/**
	 * Inserts an empty EntityLumpEntry at the given index.
	 *
	 * @error Index is out of bounds or entity lump is read-only.
	 */
	static native void Insert(int index);
	
	/**
	 * Creates an empty EntityLumpEntry, returning its index.
	 *
	 * @error Entity lump is read-only.
	 */
	static native int Append();
	
	/**
	 * Returns the number of entities currently in the lump.
	 */
	static native int Length();
};
```

nosoop:
Closes #1547.

This PR merges my [Entity Lump Manager extension][] into SourceMod's core mostly as-is, with the primary change being that the lump is unlocked for writing during `OnMapInit` instead of during its own forward.

Making use of `OnMapInit` just about the only reason for merging into core &mdash; I'm open to alternative options if the team would like to push forward with its inclusion in some form.

Parsing logic tested against ~700 maps for Team Fortress 2, and a bug has been reported and fixed for CS:GO.

Also open to other changes like bringing the native names more in line with the existing `ArrayList` methods and such.

~~(PR is a WIP and not in working order &mdash; have to do some additional work for creating the handle type and to clean up the includes / license headers.  Just putting the PR in now in case anyone has any input regarding its inclusion.)~~

[Entity Lump Manager extension]: https://github.com/nosoop/SMExt-EntityLump


nosoop:
Should be confirmed working at this stage.  [This plugin is available as a sample for entity lump access](https://gist.github.com/nosoop/66b818407a1e1231808ede949e668ba6); the main difference is replacing `OnMapEntitiesParsed` with SourceMod's `OnMapInit` and a few extra diagnostic prints.

asherkin (commented):
This looks solid, I like the API a lot - thanks for doing the work here @nosoop! The usage does seem niche in total plugin count, but I think it's powerful enough (and stripper popular enough) that it does make sense to ship with SM for sure.

The implementation looks good, I haven't super dug into it yet but I know it's been well tested. The interaction between `weakref` and `Handle` could be a bit interesting, but I think it's definitely an easier way to go than dealing with handle cloning (and handles aren't the sanest in native land currently). @Headline I'd appreciate your eyes on the C++ stdlib usage here as an extra sanity check.

I think the only change to the implementation that I think it'd be good to attempt is moving the bulk of the code into the logic binary (LumpManager and the natives) - the per-engine builds really are slow and we're trying to share as much as possible, and it looks like we won't need much added to the bridge to accommodate this.

EDIT: Please could you include that test plugin in `plugins/testsuite/` too - it looks like a really great reference.

nosoop:
As long as there isn't any need for SDK-specific flags, moving it into the logic binary would be fine.

I'm not familiar with how the bridge itself would work.  From skimming the source, it sounds like I should move the files into `core/logic/`, add either an interface or some functions to `bridge/include/LogicProvider.h` / `core/logic/common_logic.cpp`, then access them from `core/sourcemod.cpp` via `logicore` &mdash; is that right?

Aside, I'm also doing some minor refactors to move the SourceMod dependencies out of `LumpManager.{cpp,h}` so it can be compiled into a [standalone binary like before](https://github.com/nosoop/SMExt-EntityLump#standalone-binary).  

>Please could you include that test plugin in `plugins/testsuite/` too

Will do; I'll push a commit with that.

Before I forget to mention it, there are currently no checks in place to ensure users don't insert double quotes.  I've considered it before, but I'm not sure which of the options would be preferable:

- Throw a native error if the string contains double quotes
- Implicitly convert double quotes to two single quotes
- Process the string as-is and have the server blow up in the case of malformed input (not ideal, but there weren't any safeguards in place in `OnLevelInit`)

nosoop:
I've moved the entity lump code into the logic binary and confirmed that it builds and runs as expected on Windows / Linux TF2, and have also added the standalone entity lump parsing utility under `tools/entlumpparser/`; forgot to fix up the pathing so it reads from `core/logic/` but it's otherwise buildable.  Open to removing / fixing it up.

I'd also request to please add `Co-authored-by: Mat <matan@mansheroff.net>` to the squash / merge commit message if this PR gets accepted; they contributed the CS:GO-specific fix.

Ilusion9:
Any news regarding this pull request?
Isn't a waste of memory to have lumps stored after the call of OnMapInit?

nosoop:
> Any news regarding this pull request?

Nothing on my end; just waiting on the merge or further feedback.  I might retract the PR since it missed 1.11 to fill in for the now-removed `OnLevelInit` and there's no schedule for 1.12.

> Isn't a waste of memory to have lumps stored after the call of OnMapInit?

The lumps remain available in case plugins have any reason to read the original data in the future.  Should be able to pool the strings, though.

nosoop:
Retracting this PR; extension builds will be made available within the next few days.

I'm open to resubmitting this in a form that is backward-compatible with users of the extension in the future.

nosoop:
Reopening this and moving the informal discussion with @Headline from Discord into the PR conversation here.  Hopefully we can get this sorted out and into stable soon.

> nosoop: [I assume 1.11] gets no feature updates.
> asherkin: that's not true / the plan, and merging [this] PR into [1.11] is 100% the goal (and was a condition discussed when 1.11 was branched)

> headline: What was the decision for weak_ptrs made from? Does it make sense for these references to expire?
> headline: My guess is to handle erasures?

If I remember correctly, the idea was that `EntityLumpEntry` should only be retrieved on an as-needed basis and has no guarantee of existing once control is handed off to another plugin.  If something else removes the entry from the global `EntityLump`, expiry is a way to indicate that attempted changes to the entry won't take effect on the map.

Still, a plugin shouldn't remove an entry it's working with, and it shouldn't expect anything to be preserved if it calls into another plugin, so I consider it developer error.  I can expose a property to check for expiry, but in my mind, plugins should not have to worry about it; granted I'm sure people will find creative ways to break my assumptions.

Headline (approved):

nosoop:
Fixed up a bug introduced in the core-merged version that would cause issues if parsing failed partway into a file.

----

For those interested in testing, attached are builds of Windows and Linux versions compiled for supported SDK versions.  The builds were compiled against [entlump-1.11](https://github.com/nosoop/sourcemod/commits/entlump-1.11), which is this PR rebased against a somewhat recent 1.11 commit (5d9ddeee2f1a6deba590640b254310b4a21eb93d causes rebase conflicts, presumably due to a file list being reverted to a previous ordering).

⚠️ The build is provided as-is and may not reflect the final state of the PR.  If you are a server operator and don't want to worry about changes that may break plugins, wait until the PR lands.

[package.zip](https://github.com/alliedmodders/sourcemod/files/9449836/package.zip) (Windows) / [package.tar.gz](https://github.com/alliedmodders/sourcemod/files/9449835/package.tar.gz) (Linux)

I've tested the build on TF2, so I assume there won't be any regressions for the few others that have run the extension; the lumpmanager component is practically unchanged.

dysphie:
Would it make more sense to turn `EntityLump.Length()` into a property? Especially since `EntityLumpEntry.Length` also exists

nosoop:
Thanks for testing!

> Would it make more sense to turn `EntityLump.Length()` into a property?

`EntityLump` is a methodmap with only static methods; properties can't be declared as static as far as I'm aware.

PR #1676, opened 2021-12-25
Add ArrayStack.Clear native

sirdigbot:
Provide an ArrayStack.Clear method since this really has no reason not to exist.
Left out the non-methodmap native since I was told it wasn't needed.

I didn't write a test plugin since because ArrayStack just uses a CellArray under the hood, this _completely_ re-uses the code from ArrayList.Clear. I assume that's ok?


peace-maker (approved):
This looks good, thank you.
Usually I'd ask for a use case where this feature was needed, but in this case mirroring `ArrayList.Clear` doesn't add any maintenance burden.

PR #1656, opened 2021-11-28
Add block parameter to various ArrayList functions

Mikusch:
This PR adds the `block` parameter to the following functions:
* `ArrayList.FindString`
* `ArrayList.GetArray`
* `ArrayList.SetArray`
* `ArrayList.GetString`
* `ArrayList.SetString`

Default value is 0.

This can be useful especially when operating on enum structs:
```sourcepawn
enum struct MyStruct
{
	int intValue;
	int arrValue[3];
	char strValue[64];
}

public void OnPluginStart()
{
	MyStruct struct1;
	struct1.intValue = 42;
	struct1.arrValue[0] = 1;
	struct1.arrValue[1] = 2;
	struct1.arrValue[2] = 3;
	strcopy(struct1.strValue, sizeof(struct1.strValue), "foo");
	
	MyStruct struct2;
	struct2.intValue = 1337;
	struct2.arrValue[0] = 4;
	struct2.arrValue[1] = 5;
	struct2.arrValue[2] = 6;
	strcopy(struct2.strValue, sizeof(struct2.strValue), "bar");
	
	int arrBuffer[3];
	char strBuffer[64];
	
	ArrayList list = new ArrayList(sizeof(MyStruct));
	list.PushArray(struct1);
	list.PushArray(struct2);
	
	// ArrayList.FindString
	int index = list.FindString("bar", MyStruct::strValue);
	PrintToServer("bar is at index: %d", index); // "1"
	
	// ArrayList.GetArray
	list.GetArray(1, arrBuffer, sizeof(arrBuffer), MyStruct::arrValue);
	PrintToServer("arrValue at index 1 contains: (%d, %d, %d)", arrBuffer[0], arrBuffer[1], arrBuffer[2]);	// "(4, 5, 6)"
	
	// ArrayList.SetArray
	arrBuffer[2] = 322;
	list.SetArray(1, arrBuffer, sizeof(arrBuffer), MyStruct::arrValue);
	list.GetArray(1, arrBuffer, sizeof(arrBuffer), MyStruct::arrValue);
	PrintToServer("arrValue at index 1 contains: (%d, %d, %d)", arrBuffer[0], arrBuffer[1], arrBuffer[2]);	// "(4, 5, 322)"
	
	// ArrayList.GetString
	list.GetString(0, strBuffer, sizeof(strBuffer), MyStruct::strValue);
	PrintToServer("strValue at index 0: %s", strBuffer);	// "foo"
	
	// ArrayList.SetString
[...]

PR #1679, opened 2021-12-26
Fix ArrayList return types

sirdigbot:
[ArrayList.SetString](https://github.com/alliedmodders/sourcemod/blob/86363dd3ecf27d57ad1bfe01d63fee68b435291a/core/logic/smn_adt_array.cpp#L383) and [ArrayList.SetArray](https://github.com/alliedmodders/sourcemod/blob/86363dd3ecf27d57ad1bfe01d63fee68b435291a/core/logic/smn_adt_array.cpp#L416) both have int return types but their methodmap natives use void.

There might be others but I only found these 2.

### DIFF
--- a/plugins/include/adt_array.inc
+++ b/plugins/include/adt_array.inc
@@ -159,7 +159,7 @@ methodmap ArrayList < Handle {
 	// @param value         String value to set.
 	// @return              Number of characters copied.
 	// @error               Invalid index.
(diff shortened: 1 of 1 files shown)


JoinedSenses:
This PR exists to accomplish the same thing and more https://github.com/alliedmodders/sourcemod/pull/1656

peace-maker (approved):
Thanks!

PR #1165, opened 2020-01-28
Fix heap corruption in CUtlVector destructor

peace-maker:
`CHalfLife2::FindMap` works around the missing `IVEngineServer::FindMap` function by using the command autocompletion feature of the `changelevel` command.

The function populates a `CUtlVector<CUtlString>` object with the auto completion results. The game allocates memory for the vector and strings and we try to free it. This crashes when the C Run-time library version differs, which can happen when compiling using a newer toolchain.

Fixes #910 like @psychonic suggested.

I've only tested this on CS:GO Windows. This allowed me to run my server with a self-compiled build of SourceMod.

PR #1689, opened 2022-01-07
SourceMod SDK: introduction into Valve memory system 

Wend4r:
These changes to the overload of memory operators are necessary for components that include HL2SDK and subsequently static linking with the `tier1`! It was originally written to use the `Valve memory system` from `tier0`, so that through it there was an abstract opportunity to profile working with memory (`g_pMemAlloc`) to find leaks.

AMTL was not overloaded - it has a separate working environment with memory.

### DIFF
--- a/core/AMBuilder
+++ b/core/AMBuilder
@@ -138,6 +138,10 @@ for sdk_name in SM.sdks:
       binary.sources += pb_sources
       binary.compiler.cxxdefines += ['PROTOBUF_ENABLE']
 
+    # Instead of using the external SourceMod SDK.
+    if cxx.target.platform in ['linux', 'mac']:
+      binary.sources += [os.path.join(sdk.path, 'public', 'tier0', 'memoverride.cpp')]
+
     if cxx.target.platform == 'mac' and sdk.name in ['csgo']:
       # We need a proxy library since the game uses libstdc++.
       pb_binary = SM.HL2Library(builder, cxx, 'pbproxy.' + sdk.ext, sdk)
--- a/public/smsdk_ext.cpp
+++ b/public/smsdk_ext.cpp
@@ -474,8 +474,9 @@ bool SDKExtension::SDK_OnMetamodPauseChange(bool paused, char *error, size_t max
 
 #endif
 
+#if defined(__linux__) || defined(__APPLE__)
 /* Overload a few things to prevent libstdc++ linking */
-#if defined __linux__ || defined __APPLE__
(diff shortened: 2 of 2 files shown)


asherkin:
Thank you for the PR, but this is going to need a lot more description and rationale to consider merging.

Wend4r:
I really don't like that in my extension with the `tier1` used, memory can first be allocated from `operator new` -> `malloc`, and then closed in `IMemAlloc::Free`

image

Wend4r:
If I connect separately `tier0/memoverride.cpp` , then there will be a conflict of operator overloads at the linking stage with
https://github.com/alliedmodders/sourcemod/blob/5611ec54a21c3045cc1680b954631c6ca049c768/public/smsdk_ext.cpp#L483-L501

so I decided to embed `g_pMemAlloc` overload in the SourceMod SDK for components using the HL2SDK with including `tier1`

asherkin:
I do think this might end up a little painful across games, there was a bunch of Source games that had memoveride disabled on Linux builds, so a lot of testing will be required (and CI isn't going to validate anything here).

If your root issue is just cross-crt allocation around CUtlVector, you might be interested in a solution similar to #1165.

Other than seeming like a large-ish linkage change here (which we've done worse of recently, and paid the support price), I'm not really for or against it, but would like the thoughts of @psychonic who knows the engine much better than I.

Wend4r:
> I do think this might end up a little painful across games, there was a bunch of Source games that had memoveride disabled on Linux builds, so a lot of testing will be required (and CI isn't going to validate anything here).

Games that do not support MemAlloc overloading/override will be ignored
https://github.com/alliedmodders/sourcemod/blob/4e3df763582e06e6308c1c88305764f14f8311cf/AMBuildScript#L623-L625
https://github.com/alliedmodders/sourcemod/blob/62cb6a0458184c2c03500cc934e5eb125413f2f9/public/sample_ext/AMBuildScript#L348-L350
https://github.com/alliedmodders/sourcemod/blob/62cb6a0458184c2c03500cc934e5eb125413f2f9/public/sample_ext_nosdk/AMBuildScript#L244

I built and looked, they use malloc/free

Wend4r:
Another motivation to do this PR was what I saw in Valve binaries that they use MemAlloc overloading/override for binaries using `tier1` (in `server.so`, `engine.so` and etc.)

psychonic:
> but would like the thoughts of @psychonic who knows the engine much better than I.

I think this would probably be a net positive change, but the blast radius is huge. We need better ways to test these kinds of changes across all or most supported engines (or someone needs to test this one with brute force) before we can take this change.

PR #1690, opened 2022-01-07
Add VProf to frame functions

Wend4r:
I was wondering how much SM slows down the server.
The added VProf will only be used during profiling.
Maybe someone will come in handy.

Сan run profiling via 
```
sm prof start
```
and stop with the output of statistics in `vprof.log`
```
sm prof stop

con_logfile "vprof.log"
sm prof dump vprof
con_logfile ""
```

### DIFF
--- a/core/TimerSys.cpp
+++ b/core/TimerSys.cpp
@@ -220,7 +220,7 @@ void TimerSystem::OnSourceModLevelEnd()
 
 void TimerSystem::GameFrame(bool simulating)
 {
-	VPROF("SourceMod::TimerSystem::GameFrame");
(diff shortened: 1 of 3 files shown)


peace-maker:
Did you find out anything interesting regarding SM's performance using these scopes? I don't like adding a call to those hot paths which are noops 99% of the time. It might be fine to add the `ENTER_SCOPE` and `EXIT_SCOPE` macros, but don't set `VPROF_ENABLED` by default, but add a new flag for profiling sourcemod to the configure.py script. So you can enable vprof support in your local build easily if you need it, but don't have it in the production builds.

I might overestimate the impact of this though.

psychonic:
I agree with Peace-Maker, but regardless, am closing this due to lack of response.

PR #1694, opened 2022-01-18
Fix crash when calling TF2_IsHolidayActive while no map is running

Mikusch:
Currently, the server crashes when attempting to call `TF2_IsHolidayActive` while no map is running. Throw a native error instead.

Can be tested with this snippet:
```sourcepawn
#include <tf2>

public void OnEntityCreated(int entity, const char[] classname)
{
	PrintToServer("Is it Halloween? %d", TF2_IsHolidayActive(TFHoliday_HalloweenOrFullMoon));
}
```

### DIFF
--- a/extensions/tf2/natives.cpp
+++ b/extensions/tf2/natives.cpp
@@ -475,6 +475,11 @@ cell_t TF2_IsPlayerInDuel(IPluginContext *pContext, const cell_t *params)
 // native bool:TF2_IsHolidayActive(TFHoliday:holiday);
 cell_t TF2_IsHolidayActive(IPluginContext *pContext, const cell_t *params)
 {
+	if (!g_pSM->IsMapRunning())
+	{
+		return pContext->ThrowNativeError("Cannot check active holiday when no map is running");
+	}
+
 	void *pGameRules;
 	if (!g_pSDKTools || !(pGameRules = g_pSDKTools->GetGameRules()))
 	{


asherkin:
Is there a more concrete thing we can check here that is actually related to the crash? E.g. is it actually that the gamerules ptr is invalid between maps and sdktools should be clearing it out?

KyleSanderson:
Is there a stack from a Linux box or similar for this one? Asher is right if the ptr is invalid at a global level we should correct that instead.

sapphonie:
> Is there a stack from a Linux box or similar for this one? Asher is right if the ptr is invalid at a global level we should correct that instead.

https://crash.limetech.org/ib3armhzana5

I let the server start up (on itemtest), 

then did

`killserver`
`map itemtest`

and it crashed.

peace-maker:
TF2 uses the `GameRulesProxy` way to grab the `g_pGameRules` pointer instead of using the other way of a pointer to that pointer which probably gets cleared by the game on mapchange.

https://github.com/alliedmodders/sourcemod/blob/master/extensions/sdktools/vglobals.cpp#L131-L152

That `s_pGameRules` isn't set to nullptr on LevelShutdown, so `ISDKTools::GetGamerules()` returns the old value in between.
I guess the correct fix here is to set `s_pGameRules` to `nullptr` in `LevelShutdown`.

Mikusch:
Thanks for the input. I'll create a new PR that fixes this issue in a more concrete way.

PR #1698, opened 2022-01-23
Update Makefile

luckyweNda:
fix these bugs below:
fatal error: am-string.h: No such file or directory
fatal error: am-bits.h: No such file or directory

In file included from ../IExtensionSys.h:37,
                 from ../smsdk_ext.h:41,
                 from smsdk_ext.cpp:34:
../../public/amtl/amtl/am-string.h: In function ‘std::unique_ptr<char []> ke::detail::SprintfArgsImpl(size_t*, const char*, va_list)’:
../../public/amtl/amtl/am-string.h:110:28: error: ‘make_unique’ is not a member of ‘std’
  110 |         auto buffer = std::make_unique<char[]>(1);
      |                            ^~~~~~~~~~~
../../public/amtl/amtl/am-string.h:110:28: note: ‘std::make_unique’ is only available from C++14 onwards
../../public/amtl/amtl/am-string.h:110:40: error: expected primary-expression before ‘char’
  110 |         auto buffer = std::make_unique<char[]>(1);
      |                                        ^~~~
../../public/amtl/amtl/am-string.h:116:24: error: ‘make_unique’ is not a member of ‘std’
  116 |     auto buffer = std::make_unique<char[]>(len + 1);
      |                        ^~~~~~~~~~~
../../public/amtl/amtl/am-string.h:116:24: note: ‘std::make_unique’ is only available from C++14 onwards
../../public/amtl/amtl/am-string.h:116:36: error: expected primary-expression before ‘char’
  116 |     auto buffer = std::make_unique<char[]>(len + 1);

### DIFF
--- a/public/sample_ext/Makefile
+++ b/public/sample_ext/Makefile
@@ -15,6 +15,8 @@ HL2SDK_L4D2 = ../../../hl2sdk-l4d2
 HL2SDK_CSGO = ../../../hl2sdk-csgo
 MMSOURCE19 = ../../../mmsource-1.9
 
+ENGINE = 
+
 #####################################
 ### EDIT BELOW FOR OTHER PROJECTS ###
 #####################################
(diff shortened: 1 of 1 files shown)


peace-maker:
The Makefile is unmaintained and scheduled for removal for a long time. SourceMod uses [AMBuild](https://github.com/alliedmodders/ambuild) as build tool instead of make. Thank you for your contribution, but please consider [using ambuild](https://wiki.alliedmods.net/Building_SourceMod#Configuring) as well since we don't want to maintain two build systems at once.

```
pip install git+https://github.com/alliedmodders/ambuild
cd sourcemod/extensions/sample_ext
mkdir build && cd build
python ../configure.py
ambuild
```

PR #1699, opened 2022-01-24
Remove hardcoded a question mark

TheByKotik:
My little arguments:
1. Hardcoded strings is not a good idea.
2. In some cases it is really looks amiss (in my language).


peace-maker (approved):
This seems fine. I've seen many `How much is the fish??` votes, where the question already included the questionmark.

PR #1704, opened 2022-01-30
better CBaseEntity::m_fFlags definition

GamerXCZ:
Previously FL_ANIMDUCKING was missing and the entire definition list was shifted by 1 after FL_DUCKING.

### DIFF
--- a/plugins/include/entity_prop_stocks.inc
+++ b/plugins/include/entity_prop_stocks.inc
@@ -127,42 +127,46 @@ enum RenderFx
 // Note: these are only for use with GetEntityFlags and SetEntityFlags
 //       and may not match the game's actual, internal m_fFlags values.
 // PLAYER SPECIFIC FLAGS FIRST BECAUSE WE USE ONLY A FEW BITS OF NETWORK PRECISION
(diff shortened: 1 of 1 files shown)


JoinedSenses:
Hello.

Two things for ya.
Could you explain then purpose of this?
Could you fix the alignment of the defines to be consistent using spaces as they previously were?
markup_176

sneak-it:
Along With JoinedSenses' statements, you have removed and changed some defines which will break existing plugins, how do you plan on accounting for backwards compatibility?

GamerXCZ:
better?

asherkin:
These are SM-internal defines, they're not directly passed to the game - this change doesn't seem correct.

GamerXCZ:
> Toto jsou interní definice SM, nejsou přímo předávány hře – tato změna se nezdá správná.

I tested it and these definitions don't match.

If player in water player have FL_FLY flag.
When player start crouching have FL_WATERJUMP flag.
If player is BOT have flag FL_INWATER.
and more...

So mine definition is correct and SM definition is bad.

PR #1718, opened 2022-02-19
Optional autoload extensions

Kxnrl:
sometimes, we don't need auto load that extension.

### DIFF
--- a/plugins/include/clientprefs.inc
+++ b/plugins/include/clientprefs.inc
@@ -331,7 +331,11 @@ public Extension __ext_cprefs =
 {
 	name = "Client Preferences",
 	file = "clientprefs.ext",
+#if defined AUTOLOAD_EXTENSIONS
 	autoload = 1,
+#else
+	autoload = 0,
+#endif
 #if defined REQUIRE_EXTENSIONS
 	required = 1,
 #else
--- a/plugins/include/cstrike.inc
+++ b/plugins/include/cstrike.inc
@@ -451,11 +451,7 @@ public Extension __ext_cstrike =
 {
 	name = "cstrike",
 	file = "games/game.cstrike.ext",
-#if defined AUTOLOAD_EXTENSIONS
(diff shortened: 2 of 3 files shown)


peace-maker:
What is your use case where those extensions cause you trouble? Some extensions provide passive game fixes like fixing precaching issues in `OnMapStart` in CS:GO, so not loading them could do more harm than good.

asherkin:
> Some extensions provide passive game fixes like fixing precaching issues in `OnMapStart` in CS:GO, so not loading them could do more harm than good.

The game support extensions are actually autoloaded by core explicitly: https://github.com/alliedmodders/sourcemod/blob/c3e6428ff7c605d85bb082dfb00fe4cdf7eb84a5/core/sourcemod.cpp#L477-L484

Kxnrl:
> What is your use case where those extensions cause you trouble? Some extensions provide passive game fixes like fixing precaching issues in `OnMapStart` in CS:GO, so not loading them could do more harm than good.

Just make them as optinal.
I reach extension load limit.
I don't know why I can not load more then 40 extensions (windows 2016/2019).
for multi mod server, the extension list is changeable.
examples:
https://github.com/Kxnrl/Store/blob/master/store.sp#L35
https://github.com/Kxnrl/MiniGames/blob/master/MiniGames.sp#L47

peace-maker (changes requested):
Not being able to load more than X extensions seems to be a bug, could you file a bug with as much info about that as you can please?

Adding the `AUTOLOAD_EXTENSIONS` guard for clientprefs seems fine, eventhough the clientprefs plugin strongly depends on the extension, so if you have that plugin loaded the extension will be loaded anyways. Since the game extensions are autoloaded automatically as asherkin pointed out, adding the guard to the cstrike and tf2 includes does nothing. Could you switch this up to only change the clientprefs include file, please?

peace-maker (approved):

PR #1719, opened 2022-02-20
ZPS: Add missing offsets

xerox8521:
Adds support for the SDKHook Types: OnTakeDamage_Alive, GroundEntChanged, GetMaxHealth, Blocked and Reload.
Adds support for GivePlayerAmmo native.

I would have added them earlier but wasn't aware they were missing entirely.

### DIFF
--- a/extensions/sdktools/vnatives.cpp
+++ b/extensions/sdktools/vnatives.cpp
@@ -516,25 +516,19 @@ static cell_t ForcePlayerSuicide(IPluginContext *pContext, const cell_t *params)
 			return pContext->ThrowNativeError("\"CommitSuicide\" wrapper failed to initialize");
 		}
 	}
-#if SOURCE_ENGINE == SE_SDK2013
+	bool bForce = false;
 	if (!strcmp(g_pSM->GetGameFolderName(), "zps"))
 	{
-		START_CALL();
-		DECODE_VALVE_PARAM(1, thisinfo, 0);
-		*(bool *)(vptr + pCall->vparams[0].offset) = false;
-		*(bool *)(vptr + pCall->vparams[1].offset) = true;
-		FINISH_CALL_SIMPLE(NULL);
+		// ZPS requires force to be set as true otherwise the action itself is delayed.
+		// Which affects Slay and Timebomb.
+		bForce = true;
 	}
-	else
-	{
-		START_CALL();
-		DECODE_VALVE_PARAM(1, thisinfo, 0);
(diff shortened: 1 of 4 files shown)


xerox8521:
The last 2 commits weren't supposed to be in this PR. I wanted them in a seperate one and I'm not entirely sure how to split these.

extensions/sdktools/vnatives.cpp
```diff
@@ -516,12 +516,24 @@ static cell_t ForcePlayerSuicide(IPluginContext *pContext, const cell_t *params)
 			return pContext->ThrowNativeError("\"CommitSuicide\" wrapper failed to initialize");
 		}
 	}
-
-	START_CALL();
-	DECODE_VALVE_PARAM(1, thisinfo, 0);
-	*(bool *)(vptr + pCall->vparams[0].offset) = false;
-	*(bool *)(vptr + pCall->vparams[1].offset) = false;
-	FINISH_CALL_SIMPLE(NULL);
+#if SOURCE_ENGINE == SE_SDK2013
```
KyleSanderson, line 519: Something is wrong here.

KyleSanderson (changes requested):

extensions/sdktools/vnatives.cpp
xerox8521: Can you clarify? Apart that the change was supposed to be in a different PR.
The check is basically like this one here: https://github.com/alliedmodders/sourcemod/blob/master/extensions/sdktools/vnatives.cpp#L405

psychonic: What is the issue with leaving bForce as false? Does it just not do anything?

peace-maker: The preprocessor directive you added makes the native do nothing on any other engine than sdk2013.
You could refactor it to use a `bool bForce` variable which defaults to `false` and is set to true for zps.

xerox8521: > What is the issue with leaving bForce as false? Does it just not do anything?

If set to false the action is delayed. So if you slay someone or timebomb explode the explosion happens and then nothing until the delay is finished.

> The preprocessor directive you added makes the native do nothing on any other engine than sdk2013. You could refactor it to use a `bool bForce` variable which defaults to `false` and is set to true for zps.

Oh yea I see will fix that up

Headline:
@KyleSanderson updates?

psychonic (approved):

PR #1727, opened 2022-03-11
Clientprefs helpers for integers and strings

eyal282:
Title

### DIFF
--- a/plugins/include/clientprefs.inc
+++ b/plugins/include/clientprefs.inc
@@ -126,12 +126,37 @@ methodmap Cookie < Handle {
 
 	public static native Cookie Find(const char[] name);
 
-	// Set the value of a Client preference cookie.
+	// Set the string value of a Client preference cookie.
 	//
 	// @param client        Client index.
 	// @param value         String value to set.
 	// @error               Invalid cookie handle or invalid client index.
 	public native void Set(int client, const char[] value);
+	
+	// Set the integer value of a Client preference cookie.
+	//
+	// @param client        Client index.
+	// @param value         Integer value to set.
+	// @error               Invalid cookie handle or invalid client index.
+	public void SetInt(int client, int value) {
+		char sValue[11];
+		IntToString(value, sValue, sizeof(sValue));
+		
+		this.Set(client, sValue);
+	}
+	
+	// Set the float value of a Client preference cookie.
+	//
+	// @param client        Client index.
+	// @param value         Float value to set.
+	// @error               Invalid cookie handle or invalid client index.
+	public void SetFloat(int client, float value) {
+		char sValue[11];
+		FloatToString(value, sValue, sizeof(sValue));
+		
+		this.Set(client, sValue);
+	}
+
 
 	// Retrieve the value of a Client preference cookie.
 	//
@@ -141,6 +166,30 @@ methodmap Cookie < Handle {
 	// @error               Invalid cookie handle or invalid client index.
 	public native void Get(int client, char[] buffer, int maxlen);
 
+	// Retrieve the integer value of a Client preference cookie.
+	//
+	// @param client        Client index.
+	// @return              Integer value of cookie
+	// @error               Invalid cookie handle or invalid client index.
+	public float GetInt(int client) {
+		char sValue[11];
+		FloatToString(value, sValue, sizeof(sValue));
+		
+		return StringToInt(sValue)
+	}
+	
+	// Retrieve the float value of a Client preference cookie.
+	//
+	// @param client        Client index.
+	// @return              Float value of cookie
+	// @error               Invalid cookie handle or invalid client index.
+	public float GetFloat(int client) {
+		char sValue[11];
+		FloatToString(value, sValue, sizeof(sValue));
+		
+		return StringToFloat(sValue)
+	}
+	
 	// Sets the value of a Client preference cookie based on an authID string.
 	//
 	// @param authID        String Auth/STEAM ID of player to set.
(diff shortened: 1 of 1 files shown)


MAGNAT2645:
1) You shouldn't use `MarkNativeAsOptional` because these methods are not natives.
2) `Cookie.Get` returns empty string if there's no cookie stored for specific client. `StringToInt` in this case will return `0` and you can't really tell if that's the real value stored or that's the default value returned by `StringToInt` due to invalid (non-numeric or empty) string. I think it should be useful to provide default value if there's no cookie stored. For example:
```sourcepawn
	public int GetInt(int client, int defaultValue = 0) {
		char sValue[11];
		this.Get(client, sValue, sizeof(sValue));

                int value;
                if (!StringToIntEx(sValue, value))
                       value = defaultValue;
		return value;
	}

	public float GetFloat(int client, float defaultValue = 0.0) {
		char sValue[11];
		this.Get(client, sValue, sizeof(sValue));

                float value;
                if (!StringToFloatEx(sValue, value))
                       value = defaultValue;
		return value;
	}
```

eyal282:
> 1. You shouldn't use `MarkNativeAsOptional` because these methods are not natives.
> 2. `Cookie.Get` returns empty string if there's no cookie stored for specific client. `StringToInt` in this case will return `0` and you can't really tell if that's the real value stored or that's the default value returned by `StringToInt` due to invalid (non-numeric or empty) string. I think it should be useful to provide default value if there's no cookie stored. For example:
> 
> ```sourcepawn
> 	public int GetInt(int client, int defaultValue = 0)
> 	{
> 		char buffer[11];
> 		this.Get(client, buffer, sizeof(buffer));
> 
>                 int value;
>                 if (!StringToIntEx(buffer, value))
>                 {
>                        value = defaultValue;
>                 }
> 		return value;
> 	}
> 
> 	public float GetFloat(int client, float defaultValue = 0.0)
> 	{
> 		char buffer[11];
> 		this.Get(client, buffer, sizeof(buffer));
> 
>                 float value;
>                 if (!StringToFloatEx(buffer, value))
>                 {
>                        value = defaultValue;
>                 }
> 		return value;
> 	}
> ```

I made the suggested changes you gave.

MartLegion:
In case this is approved, would be nice having something like that for DispatchKeyValue as well

plugins/include/clientprefs.inc
```diff
@@ -141,6 +166,42 @@ methodmap Cookie < Handle {
...
+	//
+	// @param client        Client index.
+	// @return              Integer value of cookie
+	// @error               Invalid cookie handle or invalid client index.
+	public int GetInt(int client, int defaultValue = 0)
+	{
+		char buffer[11];
+		this.Get(client, buffer, sizeof(buffer));
+
+                int value;
```
peace-maker, line 179: Mix of tabs and spaces.

plugins/include/clientprefs.inc
```diff
@@ -141,6 +166,42 @@ methodmap Cookie < Handle {
...
+	}
+	
+	// Retrieve the float value of a Client preference cookie.
+	//
+	// @param client        Client index.
+	// @return              Float value of cookie
+	// @error               Invalid cookie handle or invalid client index.
+	public float GetFloat(int client, float defaultValue = 0.0)
+	{
+		char buffer[11];
```
peace-maker, line 194: Is the largest float really only 10 characters long? I wouldn't prematurely optimize buffer sizes and just use some large-enough power of two.

peace-maker (changes requested):
Those seem nice to have. Some comments inline.

plugins/include/clientprefs.inc
```diff
@@ -126,12 +126,37 @@ methodmap Cookie < Handle {
...
+	// @error               Invalid cookie handle or invalid client index.
+	public void SetFloat(int client, float value) {
+		char sValue[11];
+		FloatToString(value, sValue, sizeof(sValue));
+		
+		this.Set(client, sValue);
+	}
+
 
 	// Retrieve the value of a Client preference cookie.
```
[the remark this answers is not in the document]
MartLegion: SetFloat still small [11]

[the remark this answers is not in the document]
peace-maker: I'm talking about the comment I've commented on on line 136/161. You changed the docs for `.Set` to say "string value", so the docs for `.Get` could say that as well. It's not that important, sorry for not being clear.

plugins/include/clientprefs.inc
```diff
@@ -132,6 +132,31 @@ methodmap Cookie < Handle {
...
+		this.Set(client, sValue);
+	}
+	
+	// Set the float value of a Client preference cookie.
+	//
+	// @param client        Client index.
+	// @param value         Float value to set.
+	// @error               Invalid cookie handle or invalid client index.
+	public void SetFloat(int client, float value) {
+		char sValue[11];
```
peace-maker, line 154: Same comment about buffer size for large float values applies here too I think.

plugins/include/clientprefs.inc
```diff
@@ -141,7 +166,43 @@ methodmap Cookie < Handle {
...
+	public int GetInt(int client, int defaultValue = 0)
+	{
+		char buffer[11];
+		this.Get(client, buffer, sizeof(buffer));
+
+		int value;
+                if (!StringToIntEx(buffer, value))
+                {
+                       value = defaultValue;
+                }
```
peace-maker, line 183: Still lots of mixed tabs and spaces. I didn't comment on every line, but only wanted to point out the general problem with your changes.

peace-maker (changes requested):

peace-maker (approved):
Thank you for this!

PR #1729, opened 2022-03-21
Add native IsClientAdminChecked to expose m_bAdminCheckSignalled

JoinedSenses:
As title mentions, this adds a native to return the value of m_bAdminCheckSignalled for a player.

Although unlikely, it can be incorrect to assume that an authorized player has had their admin stuff processed.
```sourcepawn
native bool IsClientAdminChecked(int client);
```

```sourcepawn
bool g_Late;

public APLRes AskPluginLoad2(Handle myself, bool late, char[] error, int err_max) {
    g_Late = late;
}

public void OnPluginStart() {
    if (g_Late) {
        for (int i = 1; i <= MaxClients; ++i) {
            if (!IsClientConnected(i)) { continue; }

            OnClientConnected(i);

            if (IsClientInGame(i)) { OnClientPutInServer(i); }
           
            if (IsClientAuthorized(i)) {
                char auth[32];
                GetClientAuthId(i, AuthId_Steam2, auth, sizeof(auth), false);
                OnClientAuthorized(i, auth);

                if (IsClientAdminChecked(i)) { OnClientPostAdminCheck(i); }
            }
        }
    }
}

public void OnClientConnected(int client) { /* do stuff */ }

public void OnClientPutInServer(int client) { /* do stuff */ }

public void OnClientAuthorized(int client, const char[] auth) { /* do stuff */ }

public void OnClientPostAdminCheck(int client) { /* do stuff */ }
```

### DIFF
--- a/core/logic/smn_players.cpp
+++ b/core/logic/smn_players.cpp
@@ -486,6 +486,17 @@ static cell_t sm_IsClientAuthorized(IPluginContext *pCtx, const cell_t *params)
 	return (playerhelpers->GetGamePlayer(index)->IsAuthorized()) ? 1 : 0;
 }
 
+static cell_t sm_IsClientAdminChecked(IPluginContext *pCtx, const cell_t *params)
+{
+	int index = params[1];
+	if ((index < 1) || (index > playerhelpers->GetMaxClients()))
+	{
+		return pCtx->ThrowNativeError("Client index %d is invalid", index);
+	}
+
+	return (playerhelpers->GetGamePlayer(index)->IsAdminChecked()) ? 1 : 0;
+}
+
 static cell_t sm_IsClientFakeClient(IPluginContext *pCtx, const cell_t *params)
 {
 	int index = params[1];
@@ -1630,6 +1641,7 @@ REGISTER_NATIVES(playernatives)
 	{ "GetUserAdmin", GetUserAdmin },
 	{ "GetUserFlagBits", GetUserFlagBits },
 	{ "IsClientAuthorized", sm_IsClientAuthorized },
+	{ "IsClientAdminChecked", sm_IsClientAdminChecked },
 	{ "IsClientConnected", sm_IsClientConnected },
 	{ "IsFakeClient", sm_IsClientFakeClient },
 	{ "IsClientSourceTV", sm_IsClientSourceTV },
--- a/core/PlayerManager.cpp
+++ b/core/PlayerManager.cpp
@@ -2393,6 +2393,11 @@ bool CPlayer::IsAuthorized()
 	return m_IsAuthorized;
 }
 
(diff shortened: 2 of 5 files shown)


Headline:
https://github.com/alliedmodders/sourcemod/blob/0e716833d9a774e90729bd6a050eaf7311b11dce/public/IPlayerHelpers.h#L44

pretty sure we're gonna need to bump this

core/logic/smn_players.cpp
```diff
@@ -486,6 +486,17 @@ static cell_t sm_IsClientAuthorized(IPluginContext *pCtx, const cell_t *params)
...
 
+static cell_t sm_IsClientAdminChecked(IPluginContext *pCtx, const cell_t *params)
+{
+	int index = params[1];
+	if ((index < 1) || (index > playerhelpers->GetMaxClients()))
+	{
+		return pCtx->ThrowNativeError("Client index %d is invalid", index);
+	}
+
+	return (playerhelpers->GetGamePlayer(index)->IsAdminChecked()) ? 1 : 0;
```
Headline, line 497: Is `playerhelpers->GetGamePlayer()` returning nullptr on an invalid client something we should consider here?

JoinedSenses: It was a thought and I'm uncertain, this pattern was borrowed from IsClientAuthorized https://cs.alliedmods.net/sourcemod/source/core/logic/smn_players.cpp#478

Headline: Got it - being proven in such a hot path should be enough justification. Thanks!

JoinedSenses: To verify, I tested and it appears to be safe:
```sourcepawn
public void OnPluginStart() {
    RegAdminCmd("sm_testauth", cmdTest, ADMFLAG_ROOT);
}

public Action cmdTest(int client, int args) {
    char result[256];

    for (int i = 1; i <= MaxClients; ++i) {
        Format(result, sizeof(result), "%s%i %s\n", result, i, IsClientAuthorized(i) ? "Yes" : "No");
    }

    ReplyToCommand(client, "%s", result);

    return Plugin_Handled;
}
```
>1 No
2 No
3 No
4 No
5 No
6 No
7 No
8 No
9 No
10 No

asherkin:
I would prefer not to add this unless there is a clear need where the existing natives are inadequate and it has actually caused a problem for plugins in practice.

It binds the public API onto a particularly quirky and difficult behaviour of the admin system, which will make it more difficult to refactor in the future. While SM does have this bool, with multiple admin source plugins and deferred (async) admin loading I believe the current implementation is known to fall apart, and none of it accounts for plugins making changes later.

JoinedSenses:
Since it appears to not be desired and I have not received a reply from asher in discord, im gonna go ahead and close this.

PR #1730, opened 2022-03-22
Allow ROOT flag to call public votes

AGabenMan:
This patch allow ROOT flag to call votes even if a delay is set. Its nice to have the possibility to set a delay betweens votes to limit public votes. Generally, we do this when we have lots of admins. But as a ROOT flag this limit shall no be set.

### DIFF
--- a/plugins/basevotes.sp
+++ b/plugins/basevotes.sp
@@ -406,14 +406,12 @@ void VoteSelect(Menu menu, int param1, int param2 = 0)
 
 bool TestVoteDelay(int client)
 {
- 	int delay = CheckVoteDelay();
- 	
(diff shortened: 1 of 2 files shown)


asherkin (changes requested):
This seems like a good feature to have, but please change the implementation to use CheckCommandAccess with a suitable override string.

AGabenMan:
I propose an override sm_vote_access

AGabenMan:
@asherkin what do you think of the new update ?

AGabenMan (commented):
Fixed

GabenManPowered:
Hello, I was the originator of #1730 but I have lost my git account, so I have created a new one.

Can you reopen PR #1733 and close this PR instead please ?

Then I can make the new update as proposed by  @Kenzzer 

Thanks

PR #1733, opened 2022-04-02
Allow ROOT flag to call public votes

GabenManPowered:
This patch allow ROOT flag to call votes even if a delay is set. Its nice to have the possibility to set a delay betweens votes to limit public votes. Generally, we do this when we have lots of admins. But as a ROOT flag this limit shall no be set.

### DIFF
--- a/plugins/basevotes.sp
+++ b/plugins/basevotes.sp
@@ -406,8 +406,13 @@ void VoteSelect(Menu menu, int param1, int param2 = 0)
 
 bool TestVoteDelay(int client)
 {
+	if (CheckCommandAccess(client, "sm_vote_access", ADMFLAG_ROOT, true))
+	{
+		return true;
+	}
+	
  	int delay = CheckVoteDelay();
- 	
+	
  	if (delay > 0)
  	{
  		if (delay > 60)
--- a/plugins/basevotes.sp
+++ b/plugins/basevotes.sp
@@ -406,7 +406,7 @@ void VoteSelect(Menu menu, int param1, int param2 = 0)
 
 bool TestVoteDelay(int client)
 {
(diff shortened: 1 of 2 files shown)


plugins/basevotes.sp
```diff
@@ -406,8 +406,13 @@ void VoteSelect(Menu menu, int param1, int param2 = 0)
 
 bool TestVoteDelay(int client)
 {
+	if (CheckCommandAccess(client, "sm_vote_access", ADMFLAG_ROOT, true))
```
Kenzzer, line 409: `sm_vote_access` seems a bit off as a name, and is probably going to clash with any future changes to vote cmds.
May I recommend `sm_vote_delay_bypass`, also the default flag for this shouldn't be the admin flag `root` but `cvar`, in order to line up with sm's `sm_vote_delay` convar as any admin with such permission could disable delays anyways. This will still allow root admins to bypass delays regardless.

Kenzzer: Just noticed this is a duplicate of #1730, making a reference to it here so my comment can be relayed on both without having to send it twice.

peace-maker:
Duplicate of #1730
Please keep the conversation in one PR.

GabenManPowered:
Hello, I was the originator of #1730 but I have lost my git account, so I have create a new one.

Can you reopen this PR and close the #1730 please ?

Then I can make the new update as proposed by  @Kenzzer 

Thanks

GabenManPowered:
Thanks for reopening this PR @asherkin

@Kenzzer, you can have a look, i have repalced flag ROOT with flag CVAR and set the command acces to sm_vote_delay_bypass

asherkin (approved):
Thanks for working on this!

PR #1735, opened 2022-04-08
Timer semantics & remove unnecessary typeset entry

JoinedSenses:
- The typeset entry for `(Handle, Handle)` is already covered by `(Handle, any)`, so has no reason to be here.

- Modified default parameter of CreateTimer for consistency with other includes and because it isnt exclusively a Handle type.

### DIFF
--- a/plugins/include/timers.inc
+++ b/plugins/include/timers.inc
@@ -51,16 +51,6 @@ typeset Timer
 	 * Called when the timer interval has elapsed.
 	 *
 	 * @param timer         Handle to the timer object.
-	 * @param hndl          Handle passed to CreateTimer() when timer was created.
-	 * @return              Plugin_Stop to stop a repeating timer, any other value for
-	 *                      default behavior.
-	 */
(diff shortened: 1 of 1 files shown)


peace-maker:
Hm, I agree that the Handle data param is covered by the other signature, but I've always looked at it as being connected to the `TIMER_DATA_HNDL_CLOSE` flag, making it explicit that there's potential special handling of handles passed through. But I'm not married to the idea.

The `INVALID_HANDLE` change is good.

dvander:
Agreed it's somewhat serving as documentation which isn't a bad thing.

JoinedSenses:
> Agreed it's somewhat serving as documentation which isn't a bad thing.

what if instead of using a typedef to serve as documentation, we use documentation to serve as documentation and tweak the comment for the `(Handle,any)` entry?

peace-maker:
Yes, that's good as well.

Headline (approved):

PR #1736, opened 2022-04-08
Always add `FCVAR_RELEASE` to `CreateConVar()`

Wend4r:
In my advanced HL2SDK (not AlliedModders), that is assembled from pieces of Valve games code, there was a problem in
https://github.com/Wend4r/hl2sdk/blob/29858017a1e7cd2216af4a31ef6f81d42f55e59a/tier1/convar.cpp#L1127-L1134 

If a convar is declared without flags, then it will be marked as `FCVAR_DEVELOPMENTONLY` by analogy with the games code.
`FCVAR_RELEASE` flag is added to `ConVarManager::CreateConVar()` to access from the console/usual command

### DIFF
--- a/core/ConVarManager.cpp
+++ b/core/ConVarManager.cpp
@@ -429,7 +429,7 @@ Handle_t ConVarManager::CreateConVar(IPluginContext *pContext, const char *name,
 
 	/* Since an existing convar (or concmd with the same name) was not found , now we can finally create it */
 	pConVar = new ConVar(sm_strdup(name), sm_strdup(defaultVal), flags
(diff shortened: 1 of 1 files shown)


Wend4r:
I just building SourceMod with my HL2SDK, that can't executing convars from cfg/sourcemod by `AutoExecConfig()`.
They output `Unknown command` because of the added `FCVAR_DEVELOPMENTONLY` due to the `FCVAR_RELEASE` lack

psychonic:
We're not going to make a change to SourceMod just because you choose to build against an unsupported SDK where you intentionally made the change necessary.

PR #1737, opened 2022-04-09
Fix formatting nomination items

Sikarii:
This should fix nomination items appearing with workshop prefix in the results menu and when items are disabled.

Other solution to fix this is to **always** redraw the nomination items, lemme know if that's preferred.
This issue was found by @busheezy, thanks to him.


Headline (approved):
Looks good, thanks for submitting the patch!

PR #1741, opened 2022-04-10
Change sort optional parameters from Handle to any

JoinedSenses:
Not seeing any particular reason why the optional parameter needs to exclusively be a handle.

This change shouldn't break any API.

### DIFF
--- a/plugins/include/sorting.inc
+++ b/plugins/include/sorting.inc
@@ -90,12 +90,12 @@ native void SortStrings(char[][] array, int array_size, SortOrder order = Sort_A
  * @param elem1         First element to compare.
  * @param elem2         Second element to compare.
  * @param array         Array that is being sorted (order is undefined).
- * @param hndl          Handle optionally passed in while sorting.
+ * @param data          Handle or value optionally passed in while sorting.
  * @return              -1 if first should go before second
  *                      0 if first is equal to second
  *                      1 if first should go after second
  */
-typedef SortFunc1D = function int (int elem1, int elem2, const int[] array, Handle hndl);
+typedef SortFunc1D = function int (int elem1, int elem2, const int[] array, any data);
 
 /**
  * Sorts a custom 1D array.  You must pass in a comparison function.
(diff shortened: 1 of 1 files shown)


JoinedSenses:
This looks like it would require some change to https://github.com/alliedmodders/sourcemod/blob/2a06680252324fc8f81fc717027de7cc10f320d1/core/logic/smn_sorting.cpp#L342

plugins/include/sorting.inc
```diff
@@ -133,7 +129,7 @@ typeset SortFunc2D
  * @param sortfunc      Sort comparison function to use.
  * @param hndl          Optional Handle to pass through the comparison calls.
```
Headline, line 130: Missed one :)

Headline (commented):
Other than that one doc comment, this seems fine to me!

Go ahead and switch that out and I'll pull this in

peace-maker:
any and floats aren't friends, since the float operators aren't used if a float is tagged as `any`. This could cause confusion when sorting float arrays.

Can you add a warning for float arrays?

peace-maker:
any and floats aren't friends, since the float operators aren't used if a float is tagged as `any`. This could cause confusion when sorting float arrays.

Can you add a warning for float arrays?

JoinedSenses:
> any and floats aren't friends, since the float operators aren't used if a float is tagged as `any`. This could cause confusion when sorting float arrays.
> 
> Can you add a warning for float arrays?

This seems like a general sp nuisance not specific to sort stuff that folks should be aware of in general, so I feel like this isn't the best place to put that info

Headline:
> any and floats aren't friends, since the float operators aren't used if a float is tagged as `any`. This could cause confusion when sorting float arrays.
> 
> 
> 
> Can you add a warning for float arrays?

If they're detached in the function signature the correct operator will be used, no?

JoinedSenses:
> > any and floats aren't friends, since the float operators aren't used if a float is tagged as `any`. This could cause confusion when sorting float arrays.
> > Can you add a warning for float arrays?
> 
> If they're retagged in the function prototype for comparison the correct operator will be used, no?
> 
> iirc you can pass in a float[] as any[], and in the comparison accept float type and not any

yep, that is also true

asherkin (approved):
LGTM

The `SortCustom2D` / `SortFunc2D` change is interesting because I'm pretty sure this used to work correctly (as long as the callback matched) before passing `char[]` to `any[]` was prevented. With that change and now this actually removing the unusable prototype I think we've lost the ability to sort arrays of strings.

FortyTwoFortyTwo:
My pull request at #1124 is very identical to this apart from mine having optional data param in callbacks.
No idea why my PR gets ignored, but this PR could be updated to have SortFunc(1D/2D/ADTArray) without having data param as optional.

KyleSanderson:
@asherkin are you still OK with this?

PR #1742, opened 2022-04-11
Add GetCmdArgFloat(Ex) stocks

JoinedSenses:
Adding since there are stocks to handle int values but nothing for float.

### DIFF
--- a/plugins/include/console.inc
+++ b/plugins/include/console.inc
@@ -466,7 +466,7 @@ stock bool GetCmdArgIntEx(int argnum, int &value)
  */
 stock float GetCmdArgFloat(int argnum)
 {
-    char str[12];
+    char str[18];
     GetCmdArg(argnum, str, sizeof(str));
 
     return StringToFloat(str);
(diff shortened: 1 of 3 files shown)


JoinedSenses:
image

Headline (approved):
Yeah I think this is great and makes sense, thank you!

MartLegion:
Yeah is useful for me as well, sometimes, as JS mentioned, I create commands in plugins to set some netprops. 
As a use case, e.g. in L4D2 you can set a burning percentage from 0.0 to 1.0 to change the Special Infected appearence)
Where 0.5 means 50%, etc.

plugins/include/console.inc
```diff
@@ -443,17 +444,51 @@ stock int GetCmdArgInt(int argnum) {
...
+ * Retrieves a float command argument given its index, from the current
+ * console or server command. Will return 0.0 if the argument can not be
+ * parsed as a number. Use GetCmdArgFloatEx to handle that explicitly.
+ *
+ * @param argnum        Argument number to retrieve.
+ * @return              Value of the command argument.
+ */
+stock float GetCmdArgFloat(int argnum)
+{
+    char str[12];
```
asherkin, line 469: Some quick research indicates this buffer is likely too small for reasonable floats, it's hard to find a good reference but 17 might be a better value.

asherkin (changes requested):

plugins/include/console.inc
JoinedSenses: Made it 18 because why not

asherkin (approved):

PR #1743, opened 2022-04-12
Added PVKII game with related team colors (beacon)

MartLegion:
Spectators = Yellow
Pirates = Red
Vikings = Green
Knights = Blue

### DIFF
--- a/gamedata/funcommands.games.txt
+++ b/gamedata/funcommands.games.txt
@@ -84,4 +84,15 @@
 			"Team3BeaconColor"		"255,75,75,255"
 		}
 	}
+	
+	"pvkii"
+	{
+		"Keys"
+		{
+			"Team1BeaconColor"		"255,255,75,255"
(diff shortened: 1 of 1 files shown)


psychonic (changes requested):
It looks like these were nested inside the L4D section instead of put into their own section.

psychonic:
It works, but it's inconsistent with how the rest of the file is laid out, which can easily lead to confusion. There's currently a section for each game (or each family of game in the case of l4d/2). The pvkii can be above or below the l4d/2 block, just not inside of it.

MartLegion:
Ok I will move it below, but is NOT inside the section, just check:

https://github.com/alliedmodders/sourcemod/blob/1d71b399409e00e8cdc778e13371f95a3b930674/gamedata/funcommands.games.txt

psychonic (approved):

PR #1749, opened 2022-04-18
Add OnKickClient() forward

GabenManPowered:
This forward allow kick to be detected (ie catch kick reason) or block kicking process.

### DIFF
--- a/core/PlayerManager.cpp
+++ b/core/PlayerManager.cpp
@@ -59,6 +59,7 @@ bool g_OnMapStarted = false;
 IForward *PreAdminCheck = NULL;
 IForward *PostAdminCheck = NULL;
 IForward *PostAdminFilter = NULL;
+IForward *OnKickClient = NULL;
 
 const unsigned int *g_NumPlayersToAuth = NULL;
 int lifestate_offset = -1;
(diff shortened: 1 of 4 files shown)


Wend4r:
I think it is necessary to intercept the common `PerformKick` (`kick` and `kicked` commands) (in `engine/host_cmd.cpp`) to call not only from the SourceMod abstraction trigger

Wend4r:
Or even more common, can take `CBaseClient::Disconnect()`

GabenManPowered:
> I think it is possible to hook the common PerformKick() (kick and kicked commands) (from engine/host_cmd.cpp) to call not only from the SourceMod abstraction trigger

@Wend4r I have used the same design as BanClient (ie banid with forwards OnBanClient, OnBanIdentity, OnRemoveBan)

> Or even more common, can use `CBaseClient::Disconnect()` with a reason

@Wend4r For CBaseClient::Disconnect(), we dont want to have the disconnect reason! Here we want to detect a kick and if needed catch the reason when a player/bot/tv etc is kicked

Regards

Wend4r:
> For CBaseClient::Disconnect(), we dont want to have the disconnect reason! Here we want to detect a kick and if needed catch the reason when a player/bot/tv etc is kicked

@GabenManPowered In this regard, I suggest adding an argument to `OnClientDisconnect()`/`OnClientDisconnected()` by `CBaseClient::Disconnect()`

Wend4r:
From the kick it follows disconnecting with the kick reason

GabenManPowered:
I think its fine for the moment to used the same design as BanClient.

xerom999 (approved):
Sounds good to me, can this request be approved please? Will be very usefull...

KyleSanderson:
I apologize for the delay on this.

Functionally this looks ok, but what's the use-case? In what situation would you need to deny a kick? We have immunity rules that should take care of this so I'm coming to this from a place of ignorance.

sapphonie:
I dont have a reason to block a kick, but it would be wonderful to have at least a fwd for when it gets called instead of rolling my own

Kenzzer:
I'm also against the idea to block kicks on the fly, as stated by Kyle the permission system more than cover this aspect.
If one truly wants control over the game kick system, then they should do what one would when it comes to anything game related, and dhook the engine kick function.

But I agree with sappho, having a notification forwards for kicks could have its benefits, but I can't think of any right now.

KyleSanderson:
Can you make this a Post notification for now?

Kenzzer:
This PR has gone stale, and contains conflict. Re-open if this feature is still desired & conflicts are resolved.

PR #1752, opened 2022-04-20
Fix TF2_OnIsHolidayActive forward not getting called after map change

### LINKED ISSUE #1747: TF2_OnIsHolidayActive forward does not get called anymore after a map change
# Help us help you

# Environment
  * Game/AppID (with version if applicable): Team Fortress 2 (App 440)
  * Current SourceMod version: 1.11.0.6867
  * Current Metamod: Source version: 1.11.0-dev+1145

# Description
After changing the map, the `TF2_OnIsHolidayActive` forward does not get called anymore. Reloading a plugin that makes use of this forward fixes the issue.

# Steps to Reproduce

1. Load the plugin below. The console should get spammed.
2. Change the map.
3. `TF2_OnIsHolidayActive` does not get called anymore. Reloading the plugin causes the forward to get called again.

  ```cpp
#include <sourcemod>
#include <tf2>

public Action TF2_OnIsHolidayActive(TFHoliday holiday, bool &result)
{
	PrintToServer("Called TF2_OnIsHolidayActive");
	return Plugin_Continue;
}
  ```

Mikusch:
Fixes #1747

After a map change, the virtual hook on `IsHolidayActive` will be invalid due to the gamerules object being destroyed, but current code does not attempt to hook it again if a plugin using the `TF2_OnIsHolidayActive` forward is still loaded.

Always unhook on level change, and check for plugins using the forward only in `HolidayManager::OnPluginUnloaded`.

### DIFF
--- a/extensions/tf2/holiday.cpp
+++ b/extensions/tf2/holiday.cpp
@@ -118,10 +118,6 @@ void HolidayManager::UnhookIfNecessary()
 	if (!m_iHookID)
 		return;
 
-	// We're still wanted
-	if (m_isHolidayForward->GetFunctionCount() > 0)
-		return;
-
 	SH_REMOVE_HOOK_ID(m_iHookID);
 	m_iHookID = 0;
 }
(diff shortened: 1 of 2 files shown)


peace-maker:
Maybe it's better to just call the function `Unhook` now that it always does that.

peace-maker (approved):

PR #1758, opened 2022-04-23
Add weapon_bayonet to the list of blocked knives

TrueProfesional:


asherkin:
This needs a description with the reason for the change.

TrueProfesional:
> This needs a description with the reason for the change.

Using GiveNamedItem (GivePlayerItem) for giving players items that are not in their inventory is not allowed in CS:GO.

It's said in the [Game Server Operation Guidelines](https://blog.counter-strike.net/index.php/server_guidelines/).
> – Allowing players to claim temporary ownership of CS:GO items that are not in their inventory (Weapon skins, knives, etc.).

The code in which I made a change is used to block giving those items if you don't set "FollowCSGOServerGuidelines" to "no" in the core.cfg SourceMod config.

@psychonic made that check in #486, at https://github.com/alliedmodders/sourcemod/pull/486/commits/b65de29c920701b48310bd7bff579a0c3689cbe3 ([here](https://github.com/alliedmodders/sourcemod/pull/486/commits/b65de29c920701b48310bd7bff579a0c3689cbe3#diff-839ccf13838a98272d4e89028f66b030ae541f6206118f1d59be7cee5557f9d8)) so people wouldn't get banned.

His comments say that the knives that aren't allowed follow the pattern weapon_knife_*, however that's not true for one knife - weapon_bayonet, so that's why I added it.

Vauff:
You might want to add `weapon_karambit` too.

TrueProfesional:
> You might want to add `weapon_karambit` too.

But that.. doesn't exist (?) I don't see it defined anywhere and it doesn't work.

> Attempted to create unknown entity type weapon_karambit!
NULL Ent in GiveNamedItem!

Vauff:
> But that.. doesn't exist (?) I don't see it defined anywhere and it doesn't work.

Yeah I searched a bit and it seems I mistaken. I could've sworn it was similar to the bayonet as I had similar trouble doing weapon_knife_* matching on a karambit before.

extensions/sdktools/vnatives.cpp
```diff
@@ -163,10 +163,10 @@ static cell_t GiveNamedItem(IPluginContext *pContext, const cell_t *params)
 		char *pWeaponName;
 		pContext->LocalToString(params[2], &pWeaponName);
 
-		// Don't allow knives other than weapon_knife,  weapon_knifegg, and weapon_knife_t.
+		// Don't allow knives other than weapon_knife, weapon_knifegg, and weapon_knife_t.
 		// Others follow pattern weapon_knife_*
 		size_t len = strlen(pWeaponName);
-		if (len >= 14 && strnicmp(pWeaponName, "weapon_knife_", 13) == 0 && !(pWeaponName[13] == 't' && pWeaponName[14] == '\0'))
+		if ((len >= 14 && strnicmp(pWeaponName, "weapon_knife_", 13) == 0 && !(pWeaponName[13] == 't' && pWeaponName[14] == '\0')) || (len == 14 && stricmp(pWeaponName, "weapon_bayonet") == 0))
```
Headline, line 169: should be a call to `strnicmp` instead

Headline (changes requested):

extensions/sdktools/vnatives.cpp
peace-maker: Why? `weapon_bayonet` isn't a prefix but the whole classname, so matching the whole string is correct? `stricmp(pWeaponName, "weapon_bayonet", 14)` would match `weapon_bayonet_headline` as well. Do we want that?

TrueProfesional: > Why? `weapon_bayonet` isn't a prefix but the whole classname, so matching the whole string is correct? `strnicmp(pWeaponName, "weapon_bayonet", 14)` would match `weapon_bayonet_headline` as well. Do we want that?

I agree, I don't think the change Headline requested is right, I was just too scared to ask about this.

psychonic (approved):

extensions/sdktools/vnatives.cpp
psychonic: What you have is fine. Thanks.

ambaca:
Hello @asherkin @psychonic @peace-maker etc. etc.

I want inform that between these build:
```
DEV SM 1.12 - 6958
DEV SM 1.12 - 6954
```

...server crash when use SDKTools ```DispatchSpawn("weapon_hegrenade")``` on version 6958.
Again, it works fine on version 6954 and on other previous build, included  STABLE SM 1.11

Plugin [Weapon Restrict](https://github.com/Drifter321/csgo-css-weapon-restrict) would cause that crash.
And I managed to follow code to  ```DispatchSpawn("weapon_hegrenade")```

Here is simple problematic code.
```c++
#include <sdktools>

public void OnConfigsExecuted()
{
	CreateTimer(10.0, delay)
}

public Action delay(Handle timer)
{
	int iEnt = CreateEntityByName("weapon_hegrenade");

	if(iEnt)
	{
		DispatchSpawn(iEnt);
		//iGrenadeAmmoIndex[i] = GetEntProp(iEnt, Prop_Send, "m_iPrimaryAmmoType");
		//AcceptEntityInput(iEnt, "Kill");
	}

	return Plugin_Continue;
}
```

- I tested on Windows. 
- This person mentioned Linux #[2247](https://forums.alliedmods.net/showpost.php?p=2796680&postcount=2247)


Hope you find this bug. Good day!

peace-maker:
From a quick look, we're crashing in `CBaseCombatWeapon::GetWorldDroppedModel` because `CCSWeaponSystem::WeaponData(defidx)` returns null and the game tries to dereference it. Wasn't there some problem with creating weapon entities using `CreateEntityByName` vs. `GivePlayerItem` ever since csgo introduced the ECON system? @Drifter321 

I don't see how that would behave differently between build 6954 and 6958. The hl2sdk-csgo wasn't changed in between, but I think the buildbot was updated to build against metamod 1.12. Only the MCV changes could be a problem, but I don't see how? We'd have to bisect a bit to be sure.

psychonic:
https://github.com/alliedmodders/sourcemod/pull/1887/files#diff-839ccf13838a98272d4e89028f66b030ae541f6206118f1d59be7cee5557f9d8

Line 1023 in vnatives.cpp should have been an `&&` instead of `||`

psychonic:
Fixed in d5a26adc.

PR #1760, opened 2022-04-24
Add a OnPlayerRunCmdPre forward

### LINKED ISSUE #1759: Feature Request: OnPlayerRunCmdPre
# Description
Having a version of OnPlayerRunCmd with a guarantee that none of the values have been touched by other plugins or extensions would be absolutely *wonderful*. I know I need it for [StAC](https://github.com/sapphonie/StAC-tf2/issues/88), I know it'd be helpful for things like [Shavit's OryxAC fork](https://github.com/shavitush/Oryx-AC), and it'd probably also be nice for any other server side anticheats.

I *really* don't want to do something [evil like dhooking sdktools](https://github.com/sapphonie/sm-onplayerruncmd-pre-evil-dontuse/blob/main/scripting/onplayerruncmd_pre.sp) (because it hooks PlayerRunCommand with a VMT hook, meaning detouring it normally would be useless) in prod to get this functionality. Please sourcemod developers, you're my only hope.

neko-pm:
As requested in this issue #1759, adds a OnPlayerRunCmdPre forward for use with certain plugins which want access to unmodified usercmds.

The name isn't ideal, but after a little discussion in the Discord, we've not been able to come up with anything better.

I'm also a little unsure if `hasUsercmdsPreFwds` and `hasUsercmdsFwds` are the correct formatting for the new varaibles.

I don't have an easy means to test this right now, so the changes are currently untested.

Closes #1759

### DIFF
--- a/plugins/include/sdktools_hooks.inc
+++ b/plugins/include/sdktools_hooks.inc
@@ -51,8 +51,6 @@
  * @param tickcount     Tick count. A client's prediction based on the server's GetGameTickCount value.
  * @param seed          Random seed. Used to determine weapon recoil, spread, and other predicted elements.
  * @param mouse         Mouse direction (x, y).
- *
- * @note To see if all 11 params are available, use FeatureType_Capability and FEATURECAP_PLAYERRUNCMD_11PARAMS.
  */
 forward Action OnPlayerRunCmdPre(int client, int buttons, int impulse, float vel[3], float angles[3], int weapon, int subtype, int cmdnum, int tickcount, int seed, int mouse[2]);
 
--- a/plugins/include/sdktools_hooks.inc
+++ b/plugins/include/sdktools_hooks.inc
@@ -52,7 +52,7 @@
  * @param seed          Random seed. Used to determine weapon recoil, spread, and other predicted elements.
  * @param mouse         Mouse direction (x, y).
  */
-forward Action OnPlayerRunCmdPre(int client, int buttons, int impulse, float vel[3], float angles[3], int weapon, int subtype, int cmdnum, int tickcount, int seed, int mouse[2]);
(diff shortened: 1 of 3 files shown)


plugins/include/sdktools_hooks.inc
```diff
@@ -37,6 +37,25 @@
...
+ * @param vel           Players desired velocity.
+ * @param angles        Players desired view angles.
+ * @param weapon        Entity index of the new weapon if player switches weapon, 0 otherwise.
+ * @param subtype       Weapon subtype when selected from a menu.
+ * @param cmdnum        Command number. Increments from the first command sent.
+ * @param tickcount     Tick count. A client's prediction based on the server's GetGameTickCount value.
+ * @param seed          Random seed. Used to determine weapon recoil, spread, and other predicted elements.
+ * @param mouse         Mouse direction (x, y).
+ *
+ * @note To see if all 11 params are available, use FeatureType_Capability and FEATURECAP_PLAYERRUNCMD_11PARAMS.
```
peace-maker, line 55: This note isn't necessary for this forward since it doesn't exist in previous versions.

neko-pm: Removed it.

Headline (approved):

asherkin:
I was about to merge this, but I wonder if we need a feature capability for it existing? There doesn't appear to be a way to detect if a forward has been registered (which is probably sane - there is no guarantee they'd be called), but it seems like it would be important for plugins that want to support both 1.10 and 1.11.

Headline:
> There doesn't appear to be a way to detect if a forward has been registered 

IMO this is merge-able with the understanding that we know a forward isn't registered if it's never called, by the time OPRC is called a plugin would be able to tell if the `Pre` variant was fired, with authors able to maintain bcompat if they wish through a simple boolean flag

PR #1763, opened 2022-05-03
completing operator% forward

assyrianic:
lack of return type to the forward is breaking sptools.

### DIFF
--- a/plugins/include/float.inc
+++ b/plugins/include/float.inc
@@ -390,9 +390,9 @@ stock bool operator<=(int oper1, float oper2)
 /**
  * Forbidden operators.
  */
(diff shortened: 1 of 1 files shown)


psychonic:
Wouldn't that be a bug in sptools then if SM itself is fine?

assyrianic:
@psychonic sptools only parses new syntax. It's giving a parse error because the forwards do not have an explicit return type.

peace-maker (approved):
Thanks, this seems to be an oversight while updating the float include.

Adjusting the include files because a third party parser is having issues isn't a good reason to change things. But since this was supposed to change initially, it's good.

PR #1764, opened 2022-05-06
Add "DispatchKeyValueInt" stock

### LINKED ISSUE #1761: Feature Request: DispatchKeyValueInt
SM has DispatchKeyValue/DispatchKeyValueFloat/DispatchKeyValueVector available.

Having a wrapper to int would be useful for scenarios where we use int and int cvars as the value.

To achieve that with int values, currently is necessary to create a buffer string, format it, then uses it as DispatchKeyValue;

e.g: DispatchKeyValue(entity, "key", buffer);

From what I researched the current methods read some symbols and seem that "Int" is not available as one of them.

But I believe this could be implemented by changing "extensions/sdktools/vnatives.cpp"

IMO would help in scripting, some similar stuff was done like "GetCmdArgInt", etc.

Anyway, is just a request, if it doesn't make sense, just be polite and close.

Psykotikism:
Closes #1761

### DIFF
--- a/plugins/include/sdktools_functions.inc
+++ b/plugins/include/sdktools_functions.inc
@@ -180,7 +180,7 @@ native bool DispatchKeyValue(int entity, const char[] keyName, const char[] valu
  */
 stock bool DispatchKeyValueInt(int entity, const char[] keyName, int value)
 {
-    char str[256];
+    char str[12];
     FormatEx(str, sizeof str, "%d", value);
 
     return DispatchKeyValue(entity, keyName, str);
--- a/plugins/include/sdktools_functions.inc
+++ b/plugins/include/sdktools_functions.inc
@@ -181,7 +181,7 @@ native bool DispatchKeyValue(int entity, const char[] keyName, const char[] valu
 stock bool DispatchKeyValueInt(int entity, const char[] keyName, int value)
 {
     char str[12];
-    FormatEx(str, sizeof str, "%d", value);
+    FormatEx(str, sizeof(str), "%d", value);
 
     return DispatchKeyValue(entity, keyName, str);
 }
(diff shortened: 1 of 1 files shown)


Wend4r:
I suggest adding 2 more stock functions by `Vector` and `float` as in the games code
```c


bool CBaseEntity::KeyValue( const char *szKeyName, float flValue ) 
{
	char	string[256];

	Q_snprintf(string,sizeof(string), "%f", flValue );

	return KeyValue( szKeyName, string );
}

bool CBaseEntity::KeyValue( const char *szKeyName, int nValue ) 
{
	char	string[256];

	Q_snprintf(string,sizeof(string), "%d", nValue );

	return KeyValue( szKeyName, string );
}

bool CBaseEntity::KeyValue( const char *szKeyName, const Vector &vecValue ) 
{
	char	string[256];

	Q_snprintf(string,sizeof(string), "%f %f %f", vecValue.x, vecValue.y, vecValue.z );

	return KeyValue( szKeyName, string );
}
```

plugins/include/sdktools_functions.inc
```diff
@@ -169,6 +169,23 @@ native bool DispatchSpawn(int entity);
...
+ * @param entity        Destination entity index.
+ * @param keyName       Name of the key.
+ * @param value         Integer value.
+ * @return              True on success, false otherwise.
+ * @error               Invalid entity index or lack of mod support.
+ */
+stock bool DispatchKeyValueInt(int entity, const char[] keyName, int value)
+{
+    char str[12];
+    FormatEx(str, sizeof str, "%d", value);
```
asherkin, line 184:
```suggestion
    FormatEx(str, sizeof(str), "%d", value);
```

Prevailing style uses parentheses with `sizeof`.

asherkin (approved):
Thanks, LGTM.

PR #1765, opened 2022-05-06
fix WeaponPrice offset on windows

BoneTM:
seems like https://github.com/alliedmodders/sourcemod/commit/f7cd28cfd36e73652172fe81c93892dba6d87a94 pr a wrong gamedata for win

### DIFF
--- a/gamedata/sm-cstrike.games/game.csgo.txt
+++ b/gamedata/sm-cstrike.games/game.csgo.txt
@@ -31,7 +31,7 @@
 			// -*(_DWORD *)(v34 + 8) in ida 7 -v34[52]
 			"WeaponPrice"
 			{
-				"windows"		"200"
+				"windows"		"208"
 				"linux"			"208"
 				"linux64"		"328"
 				"mac64"			"328"


Vauff:
Went and double checked those changes, that gamedata was correct at the time, but things shifted again sometime after that and before December 27th (don't have all Windows binaries on hand to fully nail it down). Surprised it went unnoticed for this long haha.

Did you test if this worked though? From what I can see there isn't any register at offset 208 in the latest binary.

BoneTM:
> ~Went and double checked those changes, that gamedata was correct at the time, but things shifted again sometime between then and December 27th (don't have all Windows binaries on hand to fully nail the day down). Surprised it went unnoticed for this long haha.~
> 
> ~Did you test if this worked though? From what I can see there isn't any register at offset 208 in the latest binary, just `-1`. image~
> 
> Ok nevermind all that, I completely misunderstood what this was offsetting on and thought it was the same as CCSPlayerInventoryOffset. Can confirm that 208 has been the correct offset and it works as it should 👍

I'm not sure if it's broken because of the game update. In my experience, windows and linux offsets should change at the same time, but I've only seen windows offsets change.
Last night when testing a plugin I found that CS_OnGetWeaponPrice doesn't work on windows, but it does on linux.

Drifter321 (approved):

PR #1766, opened 2022-05-10
Update WriteBaselines signature for CS:GO update

Vauff:
Note this came as an optional update on 2022/05/09, and may cause issues for people with outdated servers.


Headline (approved):
Thanks!

Very quick

PR #1775, opened 2022-06-04
[Feature] Allow sending panels to SourceTV client

### LINKED ISSUE #1563: Sourcemod menus do not show for Source TV (L4D2)
Not compatible apparently.

https://github.com/shqke/sourcetvsupport/issues/24#issuecomment-894728629

A1mDev:
Code tested in game Left4Dead2, menu successfully sent to SourceTV client.

Close #1563


peace-maker:
Why do you want menus to show up? SourceTV clients won't be able to interact with them.

A1mDev:
Yes, I understand that SourceTV clients cannot interact with the menu. The menu is needed to send information to SourceTV clients.

asherkin:
I think we could enable panels for bots, but menus seem like they could be problematic as they'd remain stuck active forever until a new menu was displayed, which could have knock on effects for plugins.

peace-maker:
SourceMod keeps track of menus and could keep some resources allocated for the sourcetv client indefinetely. Panels are one-shot and won't cause trouble after being displayed, so it's safe to send them to the sourcetv bot.

A1mDev:
Ok, I think it's enough for me that I can only send the menu panel to the SourceTV bot. I don't know if anyone needs the same ability to send a menu. If necessary, I can make another separate PR to add this feature with problem fixes.

A1mDev:
Hello, what about my PR?

A1mDev:
Hello, I changed my PR a little, added the ability to enable menu sending for SourceTV, this was done in order not to change the default behavior of the menu.

PR #1777, opened 2022-06-05
Expand ShowHudText message buffer for CS:GO

Vauff:
This allows for ShowHudText to use the full message size available in CS:GO, rather than getting cut off compared to an ingame game_text as seen in https://cdn.discordapp.com/attachments/335290997317697536/983095891802066984/2022-06-05_15-50-19.mp4


psychonic:
You might as well make the same change for Blade Symphony as well. It also uses Protobuf for UserMessages.

psychonic (approved):

psychonic:
Thanks.

PR #1780, opened 2022-06-12
Fix typo in sdkhooks/natives.cpp

naydef:


naydef:
Something went wrong with all these changes...

PR #1782, opened 2022-06-16
SDKTools: Add parameters to ForcePlayerSuicide native

Mikusch:
The function `CBasePlayer::CommitSuicide` has two additional parameters which SourceMod silently passes `false` for when calling the `ForcePlayerSuicide` native. There are situations where you want to force the player to gib (explode) or want to bypass the 5 second suicide cooldown.
This PR allows users to pass the parameters to the underlying call.

```sourcepawn
native void ForcePlayerSuicide(int client, bool explode = false, bool force = false);
```

Tested on TF2 Windows.

### DIFF
--- a/extensions/sdktools/vnatives.cpp
+++ b/extensions/sdktools/vnatives.cpp
@@ -516,19 +516,22 @@ static cell_t ForcePlayerSuicide(IPluginContext *pContext, const cell_t *params)
 			return pContext->ThrowNativeError("\"CommitSuicide\" wrapper failed to initialize");
 		}
 	}
-	bool bForce = false;
+
+	START_CALL();
+	DECODE_VALVE_PARAM(1, thisinfo, 0);
+	DECODE_VALVE_PARAM(2, vparams, 0);
 	if (!strcmp(g_pSM->GetGameFolderName(), "zps"))
 	{
 		// ZPS requires force to be set as true otherwise the action itself is delayed.
 		// Which affects Slay and Timebomb.
-		bForce = true;
+		*(bool*)(vptr + pCall->vparams[1].offset) = true;
 	}
-
-	START_CALL();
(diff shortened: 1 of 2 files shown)


asherkin:
> or want to bypass the 5 second suicide cooldown.

Seems that should just be the default native behaviour here - it likely wasn't known what the boolean was for when this was added.

What's game support like for explode?

peace-maker (changes requested):

xerox8521:
ZPS Supports exploding of models with their own gibs even and also requires the force parameter to be set to true to avoid the delay.

peace-maker (approved):

peace-maker:
> Seems that should just be the default native behaviour here - it likely wasn't known what the boolean was for when this was added.

It is weird that a native called "ForceSomething" isn't forcing something by default indeed. I don't know if changing the `force` default value to `true` would cause problems in some games now that it's been off for so long, but I'm inclined to try.

asherkin:
I don't think we should even expose `force`, just use `true` internally.

KyleSanderson:
KickClient was expanded because it (well documented) crashed, RemoveEdict is still a nightmare in-frame where it's simply not safe. The Force* on these is a wart, I think defaulting these to safe is the appropriate way forward.

Mikusch:
@peace-maker @asherkin I've set the `force` param to true by default. Is this good to go?

peace-maker:
> KickClient was expanded because it (well documented) crashed, RemoveEdict is still a nightmare in-frame where it's simply not safe. The Force* on these is a wart, I think defaulting these to safe is the appropriate way forward.

The `force` parameter isn't removing entities or edicts when they shouldn't be. The game just prevents spamming CommitSuicide for some reason and ignores future calls for 5 seconds after the last call. Plugins should be able to override this.

peace-maker (approved):

PR #1785, opened 2022-06-20
Allow existing datapack Handles in CreateDataTimer

dysphie:
Currently, `CreateDataTimer` will always initialize a new datapack and assign it to the by-reference param.
This is a common pitfall for new developers who may pass an existing datapack Handle to it, effectively leaking it.
I propose it skips creating a new datapack if one is provided

### DIFF
--- a/plugins/include/timers.inc
+++ b/plugins/include/timers.inc
@@ -187,14 +187,16 @@ native bool IsServerProcessing();
  *
  * @param interval      Interval from the current game time to execute the given function.
  * @param func          Function to execute once the given interval has elapsed.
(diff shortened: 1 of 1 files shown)


Wend4r:
I use `decl` knowing how to use it, but it's not just about him. Scripter can use a variable outside of the loop with `CreateDataTimer()` is located and applied the value from the previous iteration. Same can be said it about `static` variable attribute.

dvander:
decl already has no effect on scalars. I'd be happy to neuter it for arrays as well. But probably a new helper is better here (CreateTimerWithDataPack?)

PR #1797, opened 2022-07-08
Add support for SDKCall returning non-networked entity

FortyTwoFortyTwo:
Caught this from TF2 update where some serverside ents are no longer networked. This is supposedly missed when support for non-networked ents was added.
Quick-tested this in TF2 windows.


psychonic (approved):
Looks good. Thanks!

PR #1810, opened 2022-07-26
Add GetClientOriginalLanguage

sirdigbot:
It seems reasonable that while there are functions to change a client's language, there should also be function to restore it back--both with the natives and also internally.

My use case for this is I was making a language selection menu and had to implement this logic manually.
In fact, every plugin that modifies the client's language would have to do so (particularly during OnPluginEnd) to be correctly written.

It makes sense then that a correct implementation is built in natively instead.

I'm not entirely sure that I'm supposed to set it inside PlayerManager::HandleConVarQuery(), but it looks like that's just an async way of getting the client's language for certain games, so I did.

Tested with the following code
```sourcepawn
public void OnPluginStart()
{
    LoadTranslations("common.phrases");

    RegConsoleCmd("lang_set", Lang_Set);
    RegConsoleCmd("lang_reset", Lang_Reset);
}

public Action Lang_Set(int client, int args)
{
    if (args < 1)
    {
        ReplyToCommand(client, "lang_set <code>");
        return Plugin_Handled;
    }
    char arg1[32];
    GetCmdArg(1, arg1, sizeof(arg1));
    
    int lang = GetLanguageByCode(arg1);
    if (lang == -1)
    {
        ReplyToCommand(client, "bad code");
        return Plugin_Handled;
    }

    SetClientLanguage(client, lang);
    ReplyToCommand(client, "Set lang to %s", arg1);
    ReplyToCommand(client, "%t", "No matching client");
    return Plugin_Handled;
}

public Action Lang_Reset(int client, int args)
{
    ResetClientLanguage(client);
    ReplyToCommand(client, "Reset lang to default");
    ReplyToCommand(client, "%t", "No matching client");
    return Plugin_Handled;
}
```

### DIFF
--- a/core/PlayerManager.cpp
+++ b/core/PlayerManager.cpp
@@ -2646,15 +2646,6 @@ void CPlayer::SetLanguageId(unsigned int id)
 	}
 }
 
-void CPlayer::ResetLanguageId()
-{
-	if(m_LangId != m_OriginalLangId)
-	{
-		m_LangId = m_OriginalLangId;
-		g_Players.OnClientLanguageChanged(m_iIndex, m_OriginalLangId);
-	}
-}
-
 int CPlayer::GetUserId()
 {
 	if (m_UserId == -1)
--- a/core/PlayerManager.cpp
+++ b/core/PlayerManager.cpp
@@ -528,6 +528,7 @@ bool PlayerManager::OnClientConnect(edict_t *pEntity, const char *pszName, const
 			pPlayer->m_LangId = translator->GetServerLanguage();
 		}
 #endif
+		pPlayer->m_OriginalLangId = pPlayer->m_LangId;
 	}
 	
 	List<IClientListener *>::iterator iter;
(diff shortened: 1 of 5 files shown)


asherkin (dismissed):
The `IGamePlayer` change here will break extensions, you need to add the functions to the end of the interface and adjust the compatibility check function so that old exts can load on the new version (and new ones only on the new one).

I'm not sure it really makes sense to add `ResetLanguageId` / `ResetClientLanguage` - it seems fairly straightforward that it's just `SetLanguage(GetOriginalLanguage())`. I'm also not sure about the term "reset" - as the "initial" language is the server language.

Storing and exposing the original language ID definitely seems useful though.

asherkin:
> > adjust the compatibility check function
> 
> Where would I find that? My guess is I would increment `SMINTERFACE_PLAYERMANAGER_VERSION`?

Yeah - look at `IsVersionCompatible`.

asherkin:
Yep the logic looks fine now and we're safe with the iface version bump.

I think it'd just be good to validate that it's non-breaking and my brain isn't up to checking the `IsVersionCompatible` logic - so if you could just grab the extensions from the latest 1.11 and 1.12 snapshot builds and ensure you can load them against your recompiled core that'd be great.

peace-maker:
I'd prefer following SMs other API naming scheme by mentioning the target of the function `GetOriginalLanguage` -> `GetClientOriginalLanguage`.

dvander:
Don't worry, that name isn't even in the same league as Java names for length.

peace-maker (approved):

Headline:
This looks good.

Thank you!

I did catch that the line below is unnecessary and can be pulled (see `pPlayer->Initialize`) but as far as the scope of this patch, all is well.
https://github.com/alliedmodders/sourcemod/blob/b2265dad31210f8ceb636a503644f6b25116d551/core/PlayerManager.cpp#L518

PR #1811, opened 2022-07-30
Correction to documentation, `ThrowNativeError` returns nothing.

IL0co:


Wend4r:
I think it is necessary to return the throw value to prevent the execution of the plugin that calls the native (at least this is how it should work in theory, but when building SourceMod on Ubuntu 20, execution continues)

peace-maker:
This is intentional, so you can
```sourcepawn
return ThrowNativeError(SP_ERROR_NATIVE, "Broken.");
```

instead of

```sourcepawn
ThrowNativeError(SP_ERROR_NATIVE, "Broken.");
return;
```

Thanks for the contribution, but since you didn't provide any explanation in the description I'll just close this.

IL0co:
> This is intentional, so you can
> 
> ```sourcepawn
> return ThrowNativeError(SP_ERROR_NATIVE, "Broken.");
> ```

There is no point in doing `return` after `ThrowNativeError()`, since after this function further code execution stops.
I checked it with `PrintToServer("%i", ThrowNativeError(4, "test"))`, except for the error itself, there was nothing else in the console.

And the fact that this function has `int`, but it does not return anything `@noreturn` is misleading, it writes in the doc-genes, but instead of `return` - `No description`.

I don't think it's correct to tell a function that it returns something, but in fact the code after it is not executed.
We can equate it as `return` so that the compiler understands that if this function is present, then `return` was not required?

peace-maker:
Yes, I agree some noreturn attribute to signal the compiler calls to that function never return would be nice. But until we have something like that, it's required to have a `return` afterwards to please the compiler and having the function return something saves us an extra line.

We should add documentation for the return value to explain that it's never used and just for convenience. The C++ side has such a comment iirc.

dvander:
Or better yet, a "throw" keyword.

PR #1813, opened 2022-08-01
Update CreateDirectory

Impact123:
`mode` now has a default argument and there's a note about creating nested directories.

### DIFF
--- a/plugins/include/files.inc
+++ b/plugins/include/files.inc
@@ -575,6 +575,8 @@ native bool RemoveDir(const char[] path);
 /**
  * Creates a directory.
  *
+ * Note: Does not work recursively. Parent directory has to exist.
+ *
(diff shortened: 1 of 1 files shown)


peace-maker:
https://github.com/alliedmodders/sourcemod/blob/da3e868df5da5c56c7b760584047cb9cf46ce05c/core/logic/smn_filesystem.cpp#L738

Judging by the function name it appears directories are created recursively when using the valve_fs layer.
The doc comment `create a local directory structure` isn't clear about it. Has anyone used it before and can tell?

https://github.com/alliedmodders/hl2sdk/blob/96df7cdd539cee59e10afe76f0334236e769b6c4/public/filesystem.h#L613-L614

peace-maker (approved):

PR #1815, opened 2022-08-10
raise a2s rules patch to 65000

### LINKED ISSUE #1814: CS:GO A2S_Rules fix does not patch MTU high enough to accommodate all use cases
# Help us help you

# Environment
  * Game/AppID (with version if applicable): CS:GO/730

# Description
https://github.com/alliedmodders/sourcemod/pull/614 brought the functionality of the [A2S_Rules hax plugin](https://forums.alliedmods.net/showthread.php?t=236521) directly into SM to fix CS:GO A2S_Rules responses. However while that plugin increased the MTU to 65k bytes, SM's implementation went with a [value of 5k instead](https://github.com/alliedmodders/sourcemod/blob/2f0ba0f34516157c7d5a873dbd16c9e5c4ed2a5d/extensions/cstrike/rulesfix.cpp#L83). This has proven to not be enough for some people that have plugins adding many `FCVAR_NOTIFY` flagged convars, and causes the A2S_Rules response to fail again after going past 5k bytes.

This should be be a fairly simple PR, but I was unsure of the reasoning for going with a lower value in the first place, and want to make sure there aren't any concerns with 65k or whatever other value may be preferred.

Enough convars created like this will eventually trigger the issue
```
ConVar cvar;

public void OnPluginStart()
{
	cvar = CreateConVar("sm_cvar_with_notify_flag", "0", "", FCVAR_NOTIFY);
}
```

# Logs
Some real world examples posted by several users in the SM Discord
```
[NET] Cannot send 6031-byte packet to ***:7130.  MTU is 1200.  ff ff ff ff 45
[NET] Cannot send 5171-byte packet to 195.201.249.154:47566.  MTU is 1200.  ff ff ff ff 45
[NET] Cannot send 5602-byte packet to [IP]. MTU is 1200. ff ff ff ff 45 
```

sapphonie:
fixes #1814

### DIFF
--- a/extensions/cstrike/rulesfix.cpp
+++ b/extensions/cstrike/rulesfix.cpp
@@ -80,7 +80,7 @@ void RulesFix::OnLoad()
 	host_rules_show = g_pCVar->FindVar("host_rules_show");
 	if (host_rules_show)
 	{
-		if (SetMTUMax(5000))
+		if (SetMTUMax(65000))
 		{
 			// Default to enabled. Explicit disable via cfg will still be obeyed.
 			host_rules_show->SetValue(true);
--- a/extensions/cstrike/rulesfix.cpp
+++ b/extensions/cstrike/rulesfix.cpp
@@ -124,7 +124,7 @@ static void OnConVarChanged(IConVar *var, const char *pOldValue, float flOldValu
 		{
 			if (!bPatched)
 			{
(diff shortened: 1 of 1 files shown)


sapphonie:
Cvar would be ideal here, instead of just magic hardcoding, but this should work fine for now

sapphonie:
#1829 !

PR #1816, opened 2022-08-10
Add natives to get chat trigger strings

sirdigbot:
The main utility of being able to know what the chat triggers are is if you want to display to clients how to type commands.
e.g. `Welcome to the server! Type !rules for the server rules.`

Or for chat processing, I guess.

Although I've never even encountered a server that has changed the chat triggers--so it's usually fine for plugins to just assume that `!` and `/` are correct--it's still allowed to be changed, and currently the only way to get the actual chat triggers is processing the `core.cfg` file manually.

Tested with the following code:
```sourcepawn
#include <sourcemod>

public void OnPluginStart()
{
    char buffer[32];
    int written;
    
    written = GetPublicChatTriggers(buffer, sizeof(buffer));
    PrintToServer("Public Triggers: '%s', written=%i", buffer, written);
    
    written = GetSilentChatTriggers(buffer, sizeof(buffer));
    PrintToServer("Silent Triggers: '%s', written=%i", buffer, written);
    
    // Outputs
    // Public Triggers: '!', written=1
    // Silent Triggers: '/', written=1

}
```

### DIFF
--- a/core/ChatTriggers.cpp
+++ b/core/ChatTriggers.cpp
@@ -461,3 +461,23 @@ bool ChatTriggers::WasFloodedMessage()
 {
 	return m_bWasFloodedMessage;
 }
+
+const char *ChatTriggers::GetPublicChatTrigger()
+{
+	if (!m_PubTrigger.length())
+	{
+		return NULL;
+	}
+
+	return m_PubTrigger.c_str();
+}
+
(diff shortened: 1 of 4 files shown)


peace-maker (approved):
I could have used that as well for some help messages. Thanks for working on this and sorry for the delay!

Issue #1363, opened 2020-10-10
CommandListener documentation LIES about Plugin_Handled


rtldg:
# Help us help you
  - [x] I have checked that my issue [doesn't exist yet](https://github.com/alliedmodders/sourcemod/issues).
  - [x] I have tried my absolute best to reduce the problem-space and have provided the absolute smallest test-case possible.
  - [x] I can always reproduce the issue with the provided description below.

# Environment
  * Operating System version: Windows
  * Game/AppID (with version if applicable): 232330 (Counter-Strike: Source Dedicated Server)
  * Current SourceMod version: 1.10.0.6492
  * Current SourceMod snapshot: 1.11.0.6633
  * Current Metamod: Source snapshot: 971:6c8495f
  - [x] I have updated SourceMod to the [latest version](https://www.sourcemod.net/downloads.php) and it still happens.
  - [x] I have updated SourceMod to the [latest snapshot](https://www.sourcemod.net/downloads.php?branch=dev) and it still happens.
  - [x] I have updated SourceMM to the [latest snapshot](https://sourcemm.net/downloads.php?branch=dev) and it still happens.

# Description
The [documentation for the CommandListener callback](https://sm.alliedmods.net/new-api/console/CommandListener) says
`Returning Plugin_Handled or Plugin_Stop will prevent the original, baseline code from running.`
In my test code it seems to only block when returning Plugin_Stop.
I think the [check is here](https://github.com/alliedmodders/sourcemod/blob/67f0e4be60fc698264c7a2b93772198a9ad5c1c2/core/PlayerManager.cpp#L1205-L1209)
Is this a documentation issue or an implementation issue?

# Problematic Code (or Steps to Reproduce)
  ```C
#include <sourcemod>

public void OnPluginStart()
{
	AddCommandListener(CommandListener_Noclip, "sm_noclip");
}

Action CommandListener_Noclip(int client, const char[] command, int argc)
{
	if (IsPlayerAlive(client) && argc < 1) {
		ReplyToCommand(client, "[SM] aaaa");
		return Plugin_Handled;
	}
	return Plugin_Continue;
}
  ```
With the funcommands plugin loaded, run `sm_noclip`

# Logs
Console output is:
```
] sm_noclip
[SM] aaaa
[SM] Usage: sm_noclip <#userid|name>
```

Headline:
[...]

PR #1819, opened 2022-08-11
Fix CommandListener ignoring Plugin_Handled

sirdigbot:
For issue #1363

I'm not entirely sure if this is the correct way to fix this.
But it works, and I couldn't find any other instances where `RETURN_META(MRES_SUPERCEDE);` or `RETURN_META_VALUE(MRES_SUPERCEDE, ...` were used that were using Pl_Stop instead of Pl_Handled

Tested with the same code as the issue:
```sourcepawn
#include <sourcemod>

public void OnPluginStart()
{
  AddCommandListener(CommandListener_Noclip, "sm_noclip");
}

Action CommandListener_Noclip(int client, const char[] command, int argc)
{
  if (IsPlayerAlive(client) && argc < 1) {
  	ReplyToCommand(client, "[SM] aaaa");
  	return Plugin_Handled;
  }
  return Plugin_Continue;
}
```
fix

### DIFF
--- a/core/PlayerManager.cpp
+++ b/core/PlayerManager.cpp
@@ -1214,7 +1214,7 @@ void PlayerManager::OnClientCommand(edict_t *pEntity)
 	if (g_ConsoleDetours.IsEnabled())
 	{
 		cell_t res2 = g_ConsoleDetours.InternalDispatch(client, &cargs);
-		if (res2 >= Pl_Stop)
+		if (res2 >= Pl_Handled)
 		{
 			RETURN_META(MRES_SUPERCEDE);
 		}


asherkin:
Sorry for the delay, not sure how this one snuck by! As a code fix, this is definitely correct, and is what the behaviour should be.

I'm wary of the bcompat change, it'll need a quick validation on the corpus and I'm not sure how difficult that'll be to do. That said, I could easily be persuaded to just land it for the next release from master, the doc behaviour is sane and any incorrect plugins will fail-closed, so there should be no security impact.

asherkin (approved):

peace-maker:
I'd expect plugin authors to trust the documentation and not verifying if `Plugin_Handled` actually prevented it. If they did check and switched to `Plugin_Stop`, this change won't hurt them. If you rely on `Plugin_Handled` still calling the original command dispite the docs telling otherwise, it's not really something we should support. The docs should be the accountable source here.

🚢

PR #1787, opened 2022-06-23
Add support for gamedata lookups from `soundemittersystem` library

Wend4r:

PR #1826, opened 2022-08-19
Allow gamedata to use vscript binary

Vauff:
Similar to https://github.com/alliedmodders/sourcemod/pull/1787, as https://github.com/alliedmodders/sourcemod/pull/1626 seems to be frozen at the moment.

This is necessary for a plugin modifying some vscript VM behaviour.


peace-maker (approved):
This looks fine, but we should really focus attention on a generic solution instead of these little patches.

PR #1829, opened 2022-08-28
A2S_Rules fix: Move MTU patch value to gamedata

### LINKED ISSUE #1814: CS:GO A2S_Rules fix does not patch MTU high enough to accommodate all use cases

Vauff:
Fixes #1814, supersedes #1815.

Implements the suggestion from https://github.com/alliedmodders/sourcemod/pull/1459#pullrequestreview-638382759, and increases the MTU to 65000 (what was used in the old plugin).

### DIFF
--- a/extensions/cstrike/rulesfix.cpp
+++ b/extensions/cstrike/rulesfix.cpp
@@ -36,6 +36,7 @@
 // Grab the convar ref
 ConVar *host_rules_show = nullptr;
 bool bPatched = false;
+int iPatchSize = -1;
 
 RulesFix rulesfix;
 
@@ -53,7 +54,7 @@ bool SetMTUMax(int iValue)
 
 	//If we never changed skip resetting
 	if (iOriginalValue == -1 && iValue == -1)
-		return true;
+		return false;
 
 	if (m_pMaxMTU == nullptr)
 	{
@@ -77,10 +78,16 @@ bool SetMTUMax(int iValue)
 
 void RulesFix::OnLoad()
 {
+	const char *patchSize = g_pGameConf->GetKeyValue("MTUPatchSize");
+	if (patchSize != NULL)
+	{
+		iPatchSize = atoi(patchSize);
+	}
+
 	host_rules_show = g_pCVar->FindVar("host_rules_show");
 	if (host_rules_show)
 	{
-		if (SetMTUMax(5000))
+		if (SetMTUMax(iPatchSize))
 		{
 			// Default to enabled. Explicit disable via cfg will still be obeyed.
 			host_rules_show->SetValue(true);
@@ -124,7 +131,7 @@ static void OnConVarChanged(IConVar *var, const char *pOldValue, float flOldValu
 		{
 			if (!bPatched)
 			{
-				if (SetMTUMax(5000))
+				if (SetMTUMax(iPatchSize))
 				{
 					bPatched = true;
 					NotifyAllCVars();
--- a/gamedata/sm-cstrike.games/game.csgo.txt
+++ b/gamedata/sm-cstrike.games/game.csgo.txt
@@ -269,7 +269,7 @@
 	{
 		"Keys"
 		{
-			"MTUPatchSize"		"65000"
+			"MTUPatchSize"		"5000"
 		}
 		"Addresses"
 		{
--- a/gamedata/sm-cstrike.games/game.csgo.txt
+++ b/gamedata/sm-cstrike.games/game.csgo.txt
@@ -267,6 +267,10 @@
 	
 	"#default"
 	{
+		"Keys"
+		{
+			"MTUPatchSize"		"65000"
+		}
 		"Addresses"
 		{
 			// Offset from NET_SendPacket sig to MTU


WhiteAngel4 (approved):

psychonic:
The MTU is already being patched to more than twice a sane MTU value. Making it configurable is fine, but we don't want it getting too huge and people potentially opening themselves up to DOSing if they actually do have that many convars with the notify flag or large values due to what would surely be hugely fragmented oob packets.

Vauff:
> The MTU is already being patched to more than twice a sane MTU value. Making it configurable is fine, but we don't want it getting too huge and people potentially opening themselves up to DOSing if they actually do have that many convars with the notify flag or large values due to what would surely be hugely fragmented oob packets.

What value would be recommended here then? Or no increase at all?

extensions/cstrike/rulesfix.cpp
```diff
@@ -53,7 +54,7 @@ bool SetMTUMax(int iValue)
 
 	//If we never changed skip resetting
 	if (iOriginalValue == -1 && iValue == -1)
-		return true;
+		return false;
```
asherkin, line 57: What's this change for?

asherkin (commented):
Definitely 👍 to making this configurable, not sure about changing the default value though.

> people potentially opening themselves up to DOSing if they actually do have that many convars with the notify flag or large values due to what would surely be hugely fragmented oob packets.

Is fragmented OOB reply packets much of a concern? I've never really understood Valve introducing a single-packet limit here as the OOB protocol is explicitly designed to be fragmented. Biggest concern I can see is SRCDS as a source of reflection attacks, but that's better solved through the challenge mechanism and ISP-level controls.

extensions/cstrike/rulesfix.cpp
Vauff: This is to prevent the lines 91-95 and 135-138 code blocks from running when the patch does not happen, since -1 can now be passed in those spots through iPatchSize's default value if the MTUPatchSize key is missing from gamedata.

Vauff:
10k could possibly be a better middle ground? It would still cover all the reported "cannot send packet" instances so far without going too far overboard.

KyleSanderson:
> 10k could possibly be a better middle ground? It would still cover all the reported "cannot send packet" instances so far without going too far overboard.

I do think 10k might be ok, but there's too many unknowns right now with us not understanding the full implications of what this does (such as the vulnerability potential mentioned). 10k is much larger than the internal we're already patching.

If you're open to it, if you change the gamedata value back to 5k we can land this, and then you can report back how 10k has worked for you (and others) after a period of time.

peace-maker (approved):

crashzk:
@Vauff using the latest version **SourceMod 1.11.0.6934** I keep getting a lot of these errors.

```
[NET] Cannot send 5561-byte packet to XX.XXX.XX.XX:XXXXX.  MTU is 1200.  ff ff ff ff 45

[NET] Cannot send 5561-byte packet to XX.XXX.XX.XX:XXXXX.  MTU is 1200.  ff ff ff ff 45

[NET] Cannot send 5561-byte packet to XX.XXX.XX.XX:XXXXX.  MTU is 1200.  ff ff ff ff 45

[NET] Cannot send 5561-byte packet to XX.XXX.XX.XX:XXXXX.  MTU is 1200.  ff ff ff ff 45
```

peace-maker:
Try increasing the `MTUPatchSize` in a custom gamedata override. It defaults to 5000 and you're having loads of convars with 5561 bytes.

crashzk:
> Try increasing the `MTUPatchSize` in a custom gamedata override. It defaults to 5000 and you're having loads of convars with 5561 bytes.

Thanks for the reply. In any case, I noticed that SM 1.11 does not have this feature implemented yet, only in version 1.12+.

I even manually looked to change it in gamedata but it doesn't have this `MTUPatchSize`. For now, leave this spam on the console until I update to SM version 1.12.

PR #1831, opened 2022-08-31
Block Hidden Commands from Help Command

Batfoxkid:
Allows using `FCVAR_HIDDEN` to block commands from appearing in `sm_help` and `sm_searchcmd` which helps with creating alias to commands without filling up the list with those aliases.

For example:
```sourcepawn
RegAdminCmd("sm_setmyperk", MyCommand, ADMFLAG_CHEATS);
RegAdminCmd("sm_setmyperks", MyCommand, ADMFLAG_CHEATS, _, FCVAR_HIDDEN);

RegConsoleCmd("ff2_boss", MyCommand);
RegConsoleCmd("ff2boss", MyCommand, _, FCVAR_HIDDEN);
```

### DIFF
--- a/plugins/adminhelp.sp
+++ b/plugins/adminhelp.sp
@@ -98,7 +98,7 @@ public Action HelpCmd(int client, int args)
 			cmdIter.GetName(name, sizeof(name));
 			cmdIter.GetDescription(desc, sizeof(desc));
 
-			if ((StrContains(name, arg, false) != -1) && CheckCommandAccess(client, name, cmdIter.Flags))
+			if ((StrContains(name, arg, false) != -1) && ((FindConVar(name).Flags & FCVAR_HIDDEN) == 0) && CheckCommandAccess(client, name, cmdIter.Flags))
 			{
 				PrintToConsole(client, "[%03d] %s - %s", i++, name, (desc[0] == '\0') ? noDesc : desc);
 			}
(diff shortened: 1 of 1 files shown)


plugins/adminhelp.sp
```diff
@@ -98,7 +98,7 @@ public Action HelpCmd(int client, int args)
 			cmdIter.GetName(name, sizeof(name));
 			cmdIter.GetDescription(desc, sizeof(desc));
 
-			if ((StrContains(name, arg, false) != -1) && CheckCommandAccess(client, name, cmdIter.Flags))
+			if ((StrContains(name, arg, false) != -1) && ((FindConVar(name).Flags & FCVAR_HIDDEN) == 0) && CheckCommandAccess(client, name, cmdIter.Flags))
```
Sikarii, line 101: Doesn't `cmdIter.Flags` already give you the flags, why the additional lookup?

Batfoxkid: > Doesn't `cmdIter.Flags` already give you the flags, why the additional lookup?

That is for admin flags and not command flags.

peace-maker (approved):
This seems like a nice non-intrusive addition to the help menu. All commands should be visible by default and allowing you to hide aliases this way does help the readability of the help menu.

plugins/adminhelp.sp
asherkin: Seems like it would be a good idea to add it to the iterator before making this change - we don't need workarounds here.

PR #1846, opened 2022-10-22
Fix `SetEntityHealth` to support 'any' entities

Kenzzer:
This PR closes #1770 

### What changed ?

Instead of assuming the prop name given in core gamedata is a networked property. We will instead assume it's the datamap equivalent, and is on the same offset, plus visible on every single entity as it is a CBaseEntity datamap, unlike the netprop which is only visible if the deriving class decided to have it networked.
To maintain the behaviour the old native had (mainly firing a network update, since it used `Prop_Send`) we notify the game the entity changed in order to fire a network update. And if the entity isn't networked we do nothing.

### Benefits

This will allow for a wider range of entity support. Granted plugins could just do `SetEntProp(entity, Prop_Data, "m_iHealth", amount)`. But Sourcemod aims at being strongly backwards compatible, therefore it is safe to say this native isn't going to be removed anytime soon. So since we already keep track of the property name in the core gamedata might as well just make a native that works on any source games (not that I'm aware of any that changed this property name but still).

### DIFF
--- a/plugins/include/entity_prop_stocks.inc
+++ b/plugins/include/entity_prop_stocks.inc
@@ -534,19 +534,10 @@ stock void SetEntityHealth(int entity, int amount)
 		gotconfig = true;
 	}
 
-	char cls[64];
 	PropFieldType type;
 	int offset;
 
-	if (!GetEntityNetClass(entity, cls, sizeof(cls)))
-	{
-		ThrowError("SetEntityHealth not supported by this mod: Could not get serverclass name");
-		return;
-	}
-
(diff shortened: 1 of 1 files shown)


psychonic (approved):
I went back and forth on this in my head a bit, but I think it's fine enough. I am slightly concerned that there could also be reverse cases where the necessary var only exists as a sendprop and not at all in the datadesc. That is probably rarer yet, but we are trading breakage type either way.

Let's toss it in master and see if anyone complains.

PR #1847, opened 2022-10-23
Split PVKII into its own engine branch

Spirrwell:
Our most recent update saw some changes to tier0 that broke binary compatibility causing MM/SM to fail to load on Linux. This PR splits PVKII into its own engine branch for SourceMod, and I will be doing a corresponding PR for Metamod as well.

This will also require a new hl2sdk branch for 'pvkii' which I have done here: https://github.com/Octoshark/hl2sdk-pvkii/tree/pvkii ~however I see no way to do a pull request that would create this branch on the original repo. The `pvkii` branch is based off of `sdk2013`.~ Looks like we've got that sorted.

As a side note since I've been talking about it in the SourceMod Discord, I did go ahead and backport the threadtools changes in that hl2sdk branch from C++20 to C++14 which as I understand is the default SM/MM builds with. Dealing with compiler differences here seemed like too much of a pain, so I just backported to C++14.

EDIT: CI will fail until the hl2sdk and MM changes are merged in.


extensions/sdktools/vglobals.cpp
```diff
@@ -235,11 +235,12 @@ void GetIServer()
 	|| SOURCE_ENGINE == SE_BMS     \
 	|| SOURCE_ENGINE == SE_DOI     \
 	|| SOURCE_ENGINE == SE_BLADE   \
-	|| SOURCE_ENGINE == SE_INSURGENCY
+	|| SOURCE_ENGINE == SE_INSURGENCY \
+	|| SOURCE_ENGINE == SE_PVKII
 
-#if SOURCE_ENGINE == SE_SDK2013
+#if SOURCE_ENGINE == SE_SDK2013 || SOURCE_ENGINE == SE_PVKII
```
psychonic, line 241: Ditto here. Since the interface version is predictable on PVKII, we can skip this addition, only keeping the change right above it.

extensions/sdktools/vglobals.cpp
```diff
@@ -235,11 +235,12 @@ void GetIServer()
...
 	|| SOURCE_ENGINE == SE_BLADE   \
-	|| SOURCE_ENGINE == SE_INSURGENCY
+	|| SOURCE_ENGINE == SE_INSURGENCY \
+	|| SOURCE_ENGINE == SE_PVKII
 
-#if SOURCE_ENGINE == SE_SDK2013
+#if SOURCE_ENGINE == SE_SDK2013 || SOURCE_ENGINE == SE_PVKII
 	if (g_SMAPI->GetEngineFactory(false)("VEngineServer022", nullptr))
-#endif // SE_SDK2013
+#endif // SE_SDK2013 || SE_PVKII
```
psychonic, line 243: And here.

psychonic (changes requested):

psychonic (approved):

PR #1849, opened 2022-10-26
Fix DHooks jit code stack memory alignment

Kailo97:
This fixes wrong stack memory alignment, because stack already have 4 bytes of return address at this moment in both cases: pre and post calls of hook handler.

Cherry pick in 1.11 if it needed.

Found with crash after local recompiling of SM 1.11. After dhooks jit code stack was wrongly aligned and crashed on `movdqa` instruction. 
https://crash.limetech.org/inzoqhjqfcrl

### DIFF
--- a/extensions/dhooks/DynamicHooks/hook.cpp
+++ b/extensions/dhooks/DynamicHooks/hook.cpp
@@ -355,13 +355,13 @@ void CHook::Write_CallHandler(sp::MacroAssembler& masm, HookType_t type)
 	Write_SaveRegisters(masm, type);
 
 	// Align the stack to 16 bytes.
-	masm.subl(esp, 8);
+	masm.subl(esp, 4);
 
 	// Call the global hook handler
 	masm.push(type);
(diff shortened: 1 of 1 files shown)


Ciallo-Ani:
In SM 1.11, i was able to reproduce crashes on a simple DHooks detour in Debian OS, but it seems fixed after updated to SM 1.12.
Is this pr related to it?

peace-maker (approved):
Thank you! Looks like that stack alignment fix was wrong from start. My test case seems to have been bad. Glad @Ayuto didn't just copy&paste my fix 😄 
https://github.com/Ayuto/DynamicHooks/commit/7b141acb815614ee30bc968a63080733dfd7b870

peace-maker:
> In SM 1.11 (snapshot 6914), i was able to reproduce crashes on a simple DHooks detour in **Debian OS**, but it seems fixed after updated to SM 1.12. Is this pr related to it?

Can't tell without a crash trace. But 1.12 dhooks is fine as well, if it works for you.

PR #1854, opened 2022-10-28
Enable CI on release branches

asherkin:
While we don't commonly have PRs against release branches, it seems like we should have CI runs when we do. It also seems useful to have the scripting archive for them.

I'm not 100% this syntax is correct but there isn't a super easy way to test it without just doing it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,9 +3,11 @@ on:
   push:
     branches:
      - master
+     - '[0-9]+.[0-9]+-dev'
   pull_request:
     branches:
      - master
+     - '[0-9]+.[0-9]+-dev'
 jobs:
   test:
     strategy:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -3,9 +3,11 @@ on:
   push:
     branches:
      - master
+     - '[0-9]+.[0-9]+-dev'
```
psychonic, line 6: Assuming that this is regex. Should we escape the period to treat as literal?

asherkin: Weirdly it doesn't appear to be, it's some regex / glob hybrid syntax.

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

psychonic (approved):

PR #1856, opened 2022-11-04
Add Support TLS in Webternet.

### LINKED ISSUE #1497: Support TLS in Webternet
I suggest support TLS in Webternet. HTTP is not secure way to transfer any data in today web. [Accelerator](https://github.com/asherkin/accelerator) sends all crash dumps over HTTP protocol for this reason, i guess. Dumps can contain sensitive data (such as passwords to DB or RCON).

idk1703:
Resolves #1497.


sapphonie:
image
I'm no alliedmodders developer, but I would consider paring down your merge request a bit...

fuckOff1703:
> изображение Я не разработчик alliedmodders, но я бы подумал о том, чтобы немного сократить ваш запрос на слияние...

I think so too, but I see the full version of cURL in repo.
Need delete all, except include, lib/library and license.

rumblefrog:
I think a git submodule would be appropriate.

asherkin:
> I'm no alliedmodders developer, but I would consider paring down your merge request a bit...

This is fine, the individual commits are suitably sized.

tools/buildbot/PackageScript
```diff
@@ -106,6 +106,7 @@ helpers.CopyFiles('configs/metamod', 'addons/metamod')
 helpers.CopyFiles('configs/sql-init-scripts/mysql', 'addons/sourcemod/configs/sql-init-scripts/mysql')
 helpers.CopyFiles('configs/sql-init-scripts/pgsql', 'addons/sourcemod/configs/sql-init-scripts/pgsql')
 helpers.CopyFiles('configs/sql-init-scripts/sqlite', 'addons/sourcemod/configs/sql-init-scripts/sqlite')
+helpers.CopyFiles('configs/crts', 'addons/sourcemod/configs/crts')
```
asherkin, line 109: The build is failing as the new dir isn't in `folder_list` - but I think it'd make sense for `ca-bundle.crt` to just be in `configs/` directly anyway, as we're not going to read others from that dir without it being rebuilt anyway.

fuckOff1703:
I think this better, for keep clear repo.

extensions/curl/AMBuilder
```diff
@@ -1,18 +1,30 @@
...
 import os
 
+curl_ext_path = os.path.join(builder.sourcePath, 'extensions', 'curl')
+
+os.rename(os.path.join(curl_ext_path, 'AMBuild', 'AMBuilder.curl'), os.path.join(curl_ext_path, 'curl-src' , 'lib', 'AMBuilder'))
+os.rename(os.path.join(curl_ext_path, 'AMBuild', 'AMBuilder.mbedtls'), os.path.join(curl_ext_path, 'mbedtls-src' , 'library', 'AMBuilder'))
+os.rename(os.path.join(curl_ext_path, 'AMBuild', 'config-sm-linux.h'), os.path.join(curl_ext_path, 'curl-src' , 'lib', 'config-sm-linux.h'))
+os.rename(os.path.join(curl_ext_path, 'AMBuild', 'config-sm-mac.h'), os.path.join(curl_ext_path, 'curl-src' , 'lib', 'config-sm-mac.h'))
+os.rename(os.path.join(curl_ext_path, 'AMBuild', 'config-sm-win.h'), os.path.join(curl_ext_path, 'curl-src' , 'lib', 'config-sm-win.h'))
+os.replace(os.path.join(curl_ext_path, 'AMBuild', 'curl_setup.h'), os.path.join(curl_ext_path, 'curl-src' , 'lib', 'curl_setup.h'))
```
asherkin, line 11: I'm not sure this is safe with the build graph (@dvander?) - I think these will need to be able to be out-of-tree and found where they are.

asherkin (commented):
This looks really good - great work, and nice to get a cURL update out of it too.

I think the deps are probably a question that'll take some debate - @alliedmodders/sourcemod how do we feel about introducing submodules for external dependencies? I was initially a little against it and preferred having things vendored in-tree for reliability, but especially given the provenance of these projects (and our other deps that could benefit from this treatment - sqlite and libpcre) and the fact we have SP submoduled so a recursive checkout is already required, keeping the repo cleaner is appealing.

psychonic:
> how do we feel about introducing submodules for external dependencies?

I am in favour of this.

asherkin (changes requested):
One additional request to go with the AMBuild changes above, please could you add a feature provider for this so that exts can check if it is available? Take a look at the search results for `AddCapabilityProvider` for inspiration.

extensions/curl/AMBuilder
fuckOff1703: I thinked of this, but don't know how better.
Now build failed after first compilation, files not found in dir `AMBuild`.
On copy: files changed on dir `AMBuild` not tracked, and after change any file in this dir - show message `Build succeeded, no changes.` This working on first build but next i don't known.
Need help.

fuckOff1703: Maybe symlink, this working on windows ?

Frenzoid (approved):

extensions/curl/extension.cpp
```diff
@@ -70,6 +70,8 @@ bool CurlExt::SDK_OnLoad(char *error, size_t maxlength, bool late)
 		return false;
 	}
 
+	sharesys->AddCapabilityProvider(myself, this, "Webternet");
```
peace-maker, line 73: A more descriptive name like "Webternet TLS" would be better here. Consider placing a FEATURECAP_* #define in `public/extensions/IWebternet.h` to streamline the usage.
We could bump the interface version too, but the interface itself didn't change, so a capability seems good.

peace-maker:
Symlinks on windows require special capabilitities or SYSTEM privileges. So this breaks casual builds on windows :(
@dvander knows AMBuild the best and probably knows how to integrate this best. Maybe we could get curl to look for the config files in other directories as well?

dvander:
symlink isn't the only problem. (1) third-party links in submodules are risky. They can and do disappear. They should be cloned into the alliedmodders organization. (2) modifying the source tree during the build is asking for trouble. instead, once the clone is available, these can be checked in as prebuilts, and updated on downstream merges.

fuckOff1703:
Need import or fork repos in [AlliedModders](https://github.com/alliedmodders).

[curl](https://github.com/fuckOff1703/curl/tree/sm-ext), [mbedtls](https://github.com/fuckOff1703/mbedtls/tree/sm-ext)

idk1703:
How create pr to new branch ? on curl and mbedtls repos.
`sm-ext` <- `sm-ext`
`sm-ext` - new branch from tag `curl-7_86_0`

KyleSanderson:
@idk1703 are you still blocked on this?

idk1703:
I think after https://github.com/alliedmodders/mbedtls/pull/1 and https://github.com/alliedmodders/curl/pull/1, this PR is done.

idk1703:
So i think this done

`curl` - `7.86.0`
`mbedtls` - `3.3.0`

EricZhang456:
If this is still being worked on, wouldn't it be better to use [`CURLSSLOPT_NATIVE_CA`](https://curl.se/libcurl/c/CURLOPT_SSL_OPTIONS.html#CURLSSLOPTNATIVECA) instead of shipping a CA bundle inside SM?

Edit: On Windows, I think curl should be built with Schannel and for Linux I don't know if it's okay to dynamically link against OpenSSL

EricZhang456:
Ok, this is gonna require more effort than what's available, since no one is reviewing this patch, but hear me out

I'd imagine dynamically linking OpenSSL will get messy pretty quickly, which is why I wasn't sure if it's viable. Plus AMBuild doesn't have any functionality for finding system packages like reading pkg-config and such

The native CA option also supports WolfSSL and [it has instructions on how to build it from scratch](https://www.wolfssl.com/documentation/manuals/wolfssl/chapter02.html#building-with-various-vendor-ides). But that would require a new repository under the AM org

dvander:
AMBuild has pkgconfig support. Not super useful on windows though.

You don't need a new repo if you use a git submodule and put the AMbuild script outside of it. This is what sourcepawn does, see third_party.

EricZhang456:
Should this PR be redone? It's been 4 years

PR #1857, opened 2022-11-05
Add support for TF2's `vscript_test` branch

nosoop:
> psychonic: If someone submits gamedata that still uses current build as default with overrides by crc for the branch, we could take it

This PR intends to add support for TF2's `vscript_test` staging branch.  This should be current for game version 7682888 (latest as of 2022-11-23).

Since VScripts are likely to be integrated into the game in the future, this also serves as preparation for that change regardless of whether or not the maintainers decide to add support for the branch.

~~Currently untested and unsure if the syntax is correct.  Will review when I'm more awake if nobody confirms this is working.~~

~~Do note that Linux servers are not usable at this time as the game itself is missing dependencies.  Even stock installations currently fail.~~

Summary of changes:

- Account for additions to entity vtables that have shifted entries by up to 4.
- `CTFPlayerShared::m_pOuter` was shifted down by 4 bytes, breaking Windows signatures in `sm-tf2.games` that relied on member offsets.  At the moment this is resolved by changing the bytes to match, but we could also wildcard them instead.
- Account for new location of `EntityListeners` for SDKHooks.

### DIFF
--- a/gamedata/sdkhooks.games/common.games.txt
+++ b/gamedata/sdkhooks.games/common.games.txt
@@ -23,31 +23,4 @@
 			}
 		}
 	}
-	
-	/**
-	 * Special handling for TF2's vscript_test branch.
-	 * There appears to be an issue where the #default value in this file
-	 * takes priority over the CRC-specific match in the engine-specific
-	 * one.
-	 */
-	"#default"
-	{
-		"CRC"
-		{
-			"server"
-			{
-				"windows"	"A801CADC"
-				"linux"		"DCB7584C"
-			}
-		}
-		"Offsets"
-		{
-			"EntityListeners"
-			{
-				"windows"	"131108"
-				"linux"		"131108"
-				"mac"		"131108"
-			}
-		}
-	}
 }
--- a/gamedata/sdkhooks.games/common.games.txt
+++ b/gamedata/sdkhooks.games/common.games.txt
@@ -23,4 +23,31 @@
 			}
 		}
 	}
+	
+	/**
+	 * Special handling for TF2's vscript_test branch.
+	 * There appears to be an issue where the #default value in this file
+	 * takes priority over the CRC-specific match in the engine-specific
+	 * one.
+	 */
+	"#default"
+	{
+		"CRC"
+		{
+			"server"
+			{
+				"windows"	"A801CADC"
+				"linux"		"DCB7584C"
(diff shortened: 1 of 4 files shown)


nosoop:
Most of the signature breakages are due to shifted member offsets.  Could probably mask them out no problem so they work on both platforms.

nosoop:
Looks like SDKHooks is hanging on Windows; will investigate.

nosoop:
`EntityListeners` has also changed to 131108.  It doesn't look like changing that makes it hang any less, though.

It's still hanging on this operations:

https://github.com/alliedmodders/sourcemod/blob/58f0ca475887dc697890431ec08a5e09ad1d7314/extensions/sdkhooks/extension.cpp#L236

nosoop:
For some reason SDKHooks isn't reading my changes for `EntityListeners`.  Hardcoding the offset makes it load fine.
Should be good to go once I figure out why that isn't taking.

nosoop:
I don't really have the capacity to look into the parsing routine right now to avoid touching `common.games.txt`, but these should be all the changes required for TF2.

Kenzzer:
Considering the rapid succession of CRC value changes, this is starting to feel like that PR shouldn't land and be marked as a draft at least until tf2's vscript branch becomes 'stable' and/or is merged into main, otherwise this is gonna be a stream of new PRs.

Also this is more of a question towards the repo maintainers but, wouldn't this PR existence be a good argument for a new gamedata config feature that looks up a custom file existence and its optionally its crc as well (I believe the vscript branch has a few files the main one doesn't) ? I'm willing to make a PR for that.

nosoop:
Yeah; I don't expect this to get merged upstream since CRCs by nature change after every build, and `vscript_test` is moving fast.  Other more stable heuristics would be nice though.

Regardless, I'll continue maintaining it for community server operators / developers that are planning ahead.  Less so with the former because third-party gamedata is unsupported.

nosoop:
VScript has been merged upstream, so my work here's done.

Now I have to process the third-party stuff&hellip;

psychonic:
Thank you for doing this. While we ended up not merging it while the update was in flux due to the constant CRC updates it was needing, I'm glad that we finally (almost) had a real potential use for this feature. Your gamedata also made it super easy to fix upstream when the update went live.

PR #1862, opened 2022-11-20
Return the full list of exts/plugins with "sm exts/plugins"

### LINKED ISSUE #1611: Return the full list of exts/plugins with "sm exts/plugins"
# Help us help you

# Description
Right now `sm exts/plugins` only returns 10/11 extensions/plugins at once which was a limitation in old SM versions until [this PR](https://github.com/alliedmodders/sourcemod/pull/1071), allegedly.

Right now if a client want to fetch the full list of plugins, they would have to repeat `sm plugins ##` many times, making the process quite tedious. Having `sm plugins` displaying all plugins at once similar to `sm plugins list` in console would be a nice QoL change.

zer0k-z:
Resolves #1611. Allegedly #1071 allows this to work.

### DIFF
--- a/core/PlayerManager.cpp
+++ b/core/PlayerManager.cpp
@@ -1006,11 +1006,6 @@ void ListExtensionsToClient(CPlayer *player, const CCommand &args)
 			continue;
 		}
 
-		if (id - start > 10)
-		{
-			break;
-		}
-
 		IExtensionInterface *api = ext->GetAPI();
 
 		const char *name = api->GetExtensionName();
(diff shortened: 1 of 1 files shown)


peace-maker:
Allegedly? Have you tested this change?

zer0k-z:
I tested on CS:GO with around 90 plugins printed out 10 times which makes is effectively 900 plugins. One possible downside is that if the client has packet loss, they risk not receiving every plugin, but this also happens on current version as well when you do `sm plugins; sm plugins 11;...` in the console anyway.

peace-maker:
Sounds good. We can drop the `start` parameters of the `sm exts` and `sm plugins` commands now, since they're obsolete after this change. I'm fine with the current state as well though if you don't want to keep working on this PR. :)

zer0k-z:
The `start` param should be gone as well, it should be good now.

peace-maker (approved):

asherkin (approved):
Wild to see something like this change after all these years.

PR #1869, opened 2022-12-01
Add CommandIterator.ConVarFlags property

peace-maker:
Allow to access the convar flags `FCVAR_*` value from the iterator using a `ConVarFlags` property.
The `CommandIterator.Flags` property should have been called `AdminFlags`, but it's too late to change that now.

This was brought up in #1831.


asherkin (approved):
Is it worth aliasing `Flags` as `AdminFlags` and deprecating the former now?

peace-maker:
Yes, but deprecation of methodmap properties isn't implemented. https://github.com/alliedmodders/sourcepawn/issues/676

PR #1872, opened 2022-12-02
Fix SDKHook_[Use|Spawn|GetMaxHealth] callback result value handling

### LINKED ISSUE #1870: SDKHook_Use doesn't respect return Plugin_Handled.
# Help us help you
  - [ ] I have checked that my issue [doesn't exist yet](https://github.com/alliedmodders/sourcemod/issues). (not sure, i tried to find it but couldn't find it)

# Environment
  * Operating System version: Ubuntu 22.04.1
  * Game/AppID (with version if applicable): Left 4 Dead 2 v2.2.2.6 / 550
  * Current SourceMod version: 1.11.0.6919
  * Current SourceMod snapshot: 1.11.0.6919
  * Current Metamod: Source snapshot: 1.11.0-dev+1148

# Description
When trying to prevent players to use entities with sdkhooks, SDKHook_Use doesn't prevent it when function return Plugin_Handled. But if we unload plugin and then load it again, plugin works fine (late load?).

  ```
public void OnEntityCreated(int iEnt, const char[] szClsName) {
    if (strcmp(szClsName, "prop_minigun") == 0 || strcmp(szClsName, "prop_minigun_l4d1") == 0 || strcmp(szClsName, "prop_mounted_machine_gun") == 0)
        SDKHook(iEnt, SDKHook_Use, OnUseMinigun);
}

Action OnUseMinigun(int iEnt, int iActivator, int iCaller, UseType uType, float fValue) {
    if (iCaller <= 0)                return Plugin_Continue;
    if (iCaller > MaxClients)        return Plugin_Continue;
    if (!IsClientInGame(iCaller))    return Plugin_Continue;
    if (GetClientTeam(iCaller) != 2) return Plugin_Continue;
    if (uType != Use_Toggle)         return Plugin_Continue;
    ...
    return Plugin_Handled; << it works but anyway it allows players to use entity
}
  ```

peace-maker:
The returned result of the last callback in the list was used instead of the highest value. This differs from the behavior of the other hooks.

The `GetMaxHealth` hook has an outparam. I've changed the return value handling to only return the changed value if `Pl_Changed` was the highest value a plugin wanted. If some plugin returned `Pl_Handled` it returns the original value. This is similar to the `TraceAttack` hook handling.

Fixes #1870

### DIFF
--- a/extensions/sdkhooks/extension.cpp
+++ b/extensions/sdkhooks/extension.cpp
@@ -1046,7 +1046,7 @@ int SDKHooks::Hook_GetMaxHealth()
 
 		int new_max = original_max;
 
-		cell_t res = Pl_Continue;
+		cell_t ret = Pl_Continue;
 
 		std::vector<IPluginFunction *> callbackList;
 		PopulateCallbackList(vtablehooklist[entry]->hooks, callbackList, entity);
@@ -1055,10 +1055,20 @@ int SDKHooks::Hook_GetMaxHealth()
 			IPluginFunction *callback = callbackList[entry];
 			callback->PushCell(entity);
 			callback->PushCellByRef(&new_max);
+
+			cell_t res;
 			callback->Execute(&res);
+
+			if (res > ret)
+			{
+				ret = res;
+			}
 		}
 
-		if (res >= Pl_Changed)
+		if (ret >= Pl_Handled)
+			RETURN_META_VALUE(MRES_SUPERCEDE, original_max);
+
+		if (ret >= Pl_Changed)
 			RETURN_META_VALUE(MRES_SUPERCEDE, new_max);
 
 		break;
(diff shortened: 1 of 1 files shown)


KyleSanderson (approved):

KyleSanderson:
Hopefully wasn't when I did the vectorization. Are you sure this goes far enough? Plugin_Stop doesn't appear to be implemented.

There was a patch for this to use fwdsys, but preserving the behaviour in one of the callbacks wasn't consistent so I think it died in a branch.

peace-maker:
You're right, I didn't touch breaking the loop on Plugin_Stop yet. I wanted to check how fwdsys does it first.

Do you remember which hook was a special snowflake when switching to private forwards?

KyleSanderson:
> You're right, I didn't touch breaking the loop on Plugin_Stop yet. I wanted to check how fwdsys does it first.
> 
> Do you remember which hook was a special snowflake when switching to private forwards?

I think it's all of them with an action that doesn't break in a loop (I just looked, it's a fairly sizeable chunk). Right now you can return Handled or Stop and it will still spin through all the callbacks. The concession in 2014~ was that there's already too much code out there to tolerate the breakage. There may have been something else as well where Handled was False and Stop was true, but it's cold and I didn't see it.

PR #1873, opened 2022-12-05
entitylump: Output separator as spaces instead of tabs

### LINKED ISSUE #1833: Crash from 6922 version
Crash on Nmrih server with Sourcemod 6922.
Report:"classname missing from entity!"

nosoop:
On NMRiH and possibly other games that use the Maphack system, entries that are modified using that system are rendered with tab characters stripped out - see `CNMRiHMapHackManager::GetEntDataString`.

That results in there being no separators at all between keys and values, as Maphack receives the serialized string from Entity Lump Manager, which uses tabs as its key / value separator.

This crash only occurs if `GetMapHackManager()->HasEntData()` is set to true, so to my understanding it would not crash on removals of entire entities, and it would not crash on maps that do not use Maphack at all.

This commit changes the key / value separator character to use spaces instead.

Fixes #1833.

### DIFF
--- a/core/logic/LumpManager.cpp
+++ b/core/logic/LumpManager.cpp
@@ -102,7 +102,7 @@ std::string EntityLumpManager::Dump() {
 		}
 		stream << "{\n";
 		for (const auto& pair : *entry) {
-			stream << '"' << pair.first << "\"	\"" << pair.second << '"' << '\n';
+			stream << '"' << pair.first << "\" \"" << pair.second << '"' << '\n';
 		}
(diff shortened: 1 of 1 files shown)


peace-maker (approved):

peace-maker:
Good work!

KyleSanderson:
> @peace-maker can this also get backported into 1.11? The entity lump code was recently merged there.
> 
> I can open a new PR for it if that's necessary.

Yes

PR #1883, opened 2022-12-16
Add missing null pointer check to protobuf messages

zer0k-z:
`StartProtobufMessage` can sometimes return a null pointer but the check for null pointer is only implemented in 7 out of 9 occurences in the repository, causing the crashes below.

https://crash.limetech.org/lf4n2nhxysdr
https://crash.limetech.org/k5iw7z2mfxe7

I am not sure how the function can return null pointer in these crashes though, I am fairly certain there was no hook that happened there.


psychonic:
Thanks.

It will return null if passed an invalid message id, or probably more commonly if there's already another usermessage in progress, or if from inside of a usermessage hook callback.

psychonic (approved):

PR #1885, opened 2022-12-16
Update MM:S version in CI, dependency checkout scripts and sample extensions

zer0k-z:
The CI and dependency checkout scripts name the folder as 1.10 even though it gets the latest version (1.12).

The sample extensions want the 1.10 metamod but sourcemod itself wants 1.12, so that's changed as well.

The tutorial for building sourcemod/sm extension is broken because of this.

Should I make a PR for 1.11 as well?


peace-maker:
While it's strange for the `checkout-deps` scripts to clone `master` but call the folder something different. This is good maintenance though, thank you.

peace-maker (approved):

PR #1886, opened 2022-12-18
Enable math functions in sqlite

rtldg:
I'd like to use `POW` in sqlite please... (and also have this in 1.11) (also I didn't test this but this this should:tm: work)

### DIFF
--- a/extensions/sqlite/AMBuilder
+++ b/extensions/sqlite/AMBuilder
@@ -16,6 +16,7 @@ for cxx in builder.targets:
     'SQLITE_THREADSAFE',
     'SQLITE_USE_URI',
     'SQLITE_ALLOW_URI_AUTHORITY',
+    'SQLITE_ENABLE_MATH_FUNCTIONS',
   ]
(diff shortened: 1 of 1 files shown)


peace-maker:
Please report back once you had time to test this change!

rtldg:
Just now tested on Windows and it worked fine. The github actions should upload assets imo

rtldg:
```
#include <sourcemod>

public void OnPluginStart()
{
	char error[256];
	Database db = SQL_DefConnect(error, sizeof(error));
	PrintToServer("%s", error);
	DBResultSet query = SQL_Query(db, "SELECT POW(5, 2);");
	query.FetchRow();
	PrintToServer("%f", query.FetchFloat(0));
}
```

peace-maker (approved):
The math functions were added in sqlite 3.35.0 and appear useful. I'm not sure moving business logic into sqlite is a good idea and postgres seems like a better choice if you really want to. But why not have sin/tan in a query?

PR #1887, opened 2022-12-19
Add initial support for Military Conflict: Vietnam

psychonic:
~This is still missing gamedata for non-networked entity support, and for SDKTools / SDKHooks. All other core functionality is available.~

~CI is going to fail until https://github.com/alliedmodders/metamod-source/pull/100 lands~


dvander (approved):

Prefix:
Please add Linux support, now it runs on Linux too.

psychonic:
> Please add Linux support, now it runs on Linux too.

At least as of yesterday, their Linux server build still only has been able to runs on zero or one distro/version out of many that people have tried. Because of that, they've been iterating near daily on it and will likely have a build environment change (or several).

It doesn't make sense to look up all of the signatures now, knowing that they'll likely be broken within days, when most people can't seem to successfully run it yet anyway.

PR #1889, opened 2022-12-19
Add "sm_dsay" command to basechat for sending HUD messages

psychonic:
This saves me from having to find/create a test plugin when checking for what games support this, has very low maintenance burden, and just may actually be useful for someone other than me.

"d" was chosen for the prefix over the more intuitive "h" since hint messages use that.

### DIFF
--- a/plugins/basechat.sp
+++ b/plugins/basechat.sp
@@ -72,6 +72,7 @@ public void OnPluginStart()
 		RegAdminCmd("sm_hsay", Command_SmHsay, ADMFLAG_CHAT, "sm_hsay <message> - sends hint message to all players");	
 	}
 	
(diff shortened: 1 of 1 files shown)


dvander (approved):

Rushaway:
Can't make it work on CS:S using SM 1.11

psychonic:
> Can't make it work on CS:S using SM 1.11

Thanks for letting me know. In the specific case that I was using it for, I was already expecting it to not display (and it still doesn't, after having fixed the plugin). What I should have done, was also tested the positive case. Since no error was thrown, I had assumed the code was fine.

I was missing the SetHudTextParams call. Fixed now. I'm also going to add a check that could have caught this case and thrown an error.

Rushaway:
With [914e253](https://github.com/alliedmodders/sourcemod/pull/1889/commits/914e253556de7b3f70447c91c41d0a79a9f0477c), it's now working properly on CS:S. 👍

PR #1891, opened 2022-12-20
Fix ReadMapList ignoring file's last modified time

clague:
```
if (!libsys->FileTime(pMapList->path, FileTime_LastChange, &last_time)
	|| last_time > pMapList->last_modified_time)
{
	/* Reparse */
	FileHandle_t fp;
```
Here the `pMapList->path` is not a real path, so `libsys->FileTime` will always return false so that the map file will be always reparsed.

### DIFF
--- a/core/logic/smn_maplists.cpp
+++ b/core/logic/smn_maplists.cpp
@@ -592,7 +592,10 @@ private:
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


psychonic (approved):

psychonic:
Thanks. Nice catch.

PR #1893, opened 2022-12-25
Stop EntIndexToEntRef throwing

sirdigbot:
Some forwards such as `OnEntityCreated` provide either an entref or an index, which is a bit of gotcha.
A simple fix, in most cases, could be to always force the value to either an index or an entref with the `EntIndexToEntRef` or `EntRefToEntIndex` natives depending on what you're doing.

However, mixing entrefs and indexes with `EntIndexToEntRef` will currently throw an error because entrefs are always an out-of-bounds index.

To me it *seems like* the throw was added because `CHalfLife2::IndexToReference` would call `CHalfLife2::LookupEntity`, which at the time of writing wouldn't validate index bounds.
However a later commit added this bounds checking to `CHalfLife2::LookupEntity`, making the check in the `EntIndexToEntRef` native redundant.

Additionally, `CHalfLife2::IndexToReference` seems to accept both indexes and references so no other checks should be needed.

Unless I'm missing something important, there is no benefit to having `EntIndexToEntRef` throw an error, and no breakage to old plugins by removing it.

### DIFF
--- a/core/smn_halflife.cpp
+++ b/core/smn_halflife.cpp
@@ -546,11 +546,6 @@ static cell_t GetEngineVersion(IPluginContext *pContext, const cell_t *params)
 
 static cell_t IndexToReference(IPluginContext *pContext, const cell_t *params)
 {
-	if (params[1] >= NUM_ENT_ENTRIES || params[1] < 0)
-	{
-		return pContext->ThrowNativeError("Invalid entity index %i", params[1]);
-	}
-
 	return g_HL2.IndexToReference(params[1]);
 }
 
--- a/plugins/include/halflife.inc
+++ b/plugins/include/halflife.inc
@@ -662,7 +662,6 @@ stock void DisplayAskConnectBox(int client, float time, const char[] ip, const c
  *
  * @param entity        Entity index.
  * @return              Entity reference or -1 on invalid entity.
- * @error               Entity index >= GetMaxEntities() or < 0
  */
 native int EntIndexToEntRef(int entity);
 


KyleSanderson:
gut says this is fine, but the current call pattern to handle this today is as simple as EntIndexToEntRef(EntRefToEntIndex(iEntity)));

KyleSanderson:
> `EntIndexToEntRef(EntRefToEntIndex(iEntity)))` still throws when passed an invalid entity; you'd be passing -1 into EntIndexToEntRef which is out of bounds.
> 
> Though I guess you could use something like this if you really wanted:
> 
> ```sourcepawn
> int entRef;
> if (IsValidEntity(iEntity) && (entRef = EntIndexToEntRef(EntRefToEntIndex(iEntity))))
> {
>     // do thing to entRef
> }
> ```

That's a completely different issue then if the reference isn't valid (and is an excellent example of why references are the way to go).

psychonic:
It is intended that passing a non-index to a `EntIndexToEntRef` or the internal `IndexToReference` is an error. As implied by the function names and noted in the `EntIndexToEntRef`, these functions expect and rely on an entity index to be passed in.

If you have an entity reference, whether it's an ehandle-based one (looks negative if treated as signed) or a backcompat ref (an index), there are a couple of options to convert it into an explicit index.

```sp
int myEnt = GetSomeEnt();

// No validity checks. Just strips ent ref/handle data if present
int index = MakeCompatEntRef(myEnt);

// If passed a backcompat ref, same as above
// If passed an ehandle-backed ref, checks that entity is still valid before stripping off the ent ref/handle data.
int index2 = EntIndexToEntRef(EntRefToEntIndex(myEnt));
```

PR #1900, opened 2023-01-04
DHooks: Remove function entry on a new matching section

### LINKED ISSUE #1879: DHooks: Function name collisions across gamedata files
# Help us help you

# Environment
  * Current SourceMod version: 1.11.0.6924
  * Current SourceMod snapshot: 1.12.0.6949

# Description

When two separate gamedata files are loaded in and one wants to declare a function name as a detour and another wants to declare it as a virtual hook, any `*FromConf` operations on the name after the second file will fail with the following message as the `SignatureWrapper` struct contains references to both offsets / signatures:

> Cannot have "signature", "address" or "offset" keys at the same time in one function

Expected behavior is that names don't collide across different gamedata files.

Attached is gamedata, compiled plugins, and corresponding sources for two plugins that will attempt to load detours / hooks for `CBaseCombatWeapon::PrimaryAttack()`.  Gamedata is only provided for Team Fortress 2.

[dhook_name_collide_test.zip](https://github.com/alliedmodders/sourcemod/files/10200001/dhook_name_collide_test.zip)

This includes a special case to test to ensure that any fix doesn't cause reading a game-specific section to clobber items in a `#default` one.

nosoop:
Closes #1879.

This PR changes the behavior in DHooks so new matching functions replaces any previous declarations instead of attempting to apply changes to an existing one.  This does mean `DHookSetup.SetFromConf` needs to be called before other gamedata files are parsed to ensure the hook is initialized as expected.

Note that this does prevent merges of partially-defined functions via `#default` (failing the edge-case test in the linked issue).  I'd like to preserve that behavior, but I don't think there's any workable solution around that at this time without a change in the ~~`IGameConfigManager`~~ parsing interface, because as far as I'm aware there is no way for a custom gamedata section hook to detect when it's in a new file (default and game-specific sections both call into `ReadSMC_ParseStart`).

I considered making offset / signature / address a "variant" of sorts so only one can be provided at a time, but I believe the "arguments" section also has issues with merging.

### DIFF
--- a/extensions/dhooks/signatures.cpp
+++ b/extensions/dhooks/signatures.cpp
@@ -124,9 +124,11 @@ SMCResult SignatureGameConfig::ReadSMC_NewSection(const SMCStates *states, const
 	{
 		auto sig = signatures_.find(name);
 		if (sig.found())
-			g_CurrentSignature = sig->value;
-		else
-			g_CurrentSignature = new SignatureWrapper();
+		{
+			delete sig->value;
+			signatures_.remove(sig);
+		}
+		g_CurrentSignature = new SignatureWrapper();
 		g_CurrentFunctionName = name;
 		g_ParseState = PState_Function;
 		break;


nosoop:
Closing per #1879.  Probably won't be fixed in 1.11 due to likely interface breakages.

PR #1903, opened 2023-01-09
Reconcile the concept of Edict & Networkable across the codebase

### LINKED ISSUE #1902: SDKTools GameRules native crash
# Help us help you
  - [ ] I can always reproduce the issue with the provided description below.

# Environment
  * Operating System version: Debian Bullseye (Docker Debian Buster)
  * Game/AppID (with version if applicable): 440
  * Current SourceMod version: 1.12.0.6945
  * Current Metamod: Source snapshot: 1.11.0-dev+1145

# Description
Through unknown conditions, the native `GameRules_SetProp` and more specifically, the extension sdktools's `GetGameRulesProxyEnt` will cause a segmentation fault.

Call `GameRules_SetProp` repeatedly during any map. This might or might not trigger the crash.

# Logs
https://crash.limetech.org/d42cffz6sta4
But this seems much more widespread than the crash signature I've, if we filter stats
https://crash.limetech.org/stats/sdktools.ext.2.tf2.so/GetGameRulesProxyEnt
Those servers seem to crash in the same location as I, and they're most certainly not running the same plugin setup as I.

Kenzzer:
This fixes #1902 and potential crash in other natives.

# What is this all about ?

Well, the scope of this PR mainly boils down to
`CBaseEdict::GetNetworkable` & `IServerUnknown::GetNetworkable` (CBaseEntity::GetNetworkable)

The former, is the edict that gives a pointer (that might or might not be valid) of the attached entity's networkable
The latter, is the entity/unknown that gives a pointer (that is always valid) of the attached entity's networkable

# What are the benefits ?

The PR is mostly a speculative a fix for PR #1902 but everything points towards an unproperly setup Networkable's vtable. Which is strange, because Sourcemod retrives the networkable interface of any entity in a lot of places.
https://github.com/alliedmodders/sourcemod/search?q=GetNetworkable
But the key difference with all of them, and the gamerules native (as well tf2's objective ressource native, and another function). Is that they use the edict_t ptr, to retrieve the networkable interface. While everywhere else in the codebase, we go from edict ptr to unknown ptr (the cbaseentity) to the networkable interface.

This leads me to believe, the engine doesn't properly clean up the networkable ptr on its edict_t(s) which is assigned by the srcds.

# Why is a core file modified ?

The goal is to keep the retrieval of the networkable interface consistent across all the codebase. That includes core files.
Furthermore, this also fixes a non logical message of the Set/GetEntProp natives
`Edict %d (%d) is not networkable`
Edicts are networkable by definition. We should have never gotten that far to begin with, if we started from an edict. But we don't, we start from a `IServerUnknown`. So we remove this useless check, whose error message never ever showed in sourcemode entire lifetime, I browsed the forums & google never found a post where a plugin maker asks about the error message.

~~# Why is there some null check removed on networkable ?~~

> `IServerUnknown` is only ever implemented by `CBaseEntity`
> `IServerUnknown::GetNetworkable` is implemented like so by `CBaseEntity`
> 
> ```
> inline IServerNetworkable *CBaseEntity::GetNetworkable()
> {
> 	return &m_Network;
> }
> ```
> The ptr can never ever be invalid. Its memory is allocated along with `CBaseEntity (IServerUnknown)`
> 
> And it is always initialised.
> ```cpp
> CBaseEntity::CBaseEntity( bool bServerOnly )
> {
> 	NetworkProp()->Init( this );
> ```
> This statement is true, all the way back to the 2007 release of the source engine. I haven't/couldn't check older versions. But I'm confident.
> 
> Sourcemod itself is "guilty" of not null checking the networkable interface, when retrieving it from `IServerUnknown`
> https://github.com/alliedmodders/sourcemod/blob/57a38636fcb1bccfed0f69705b237e8f028c0d1a/extensions/tf2/criticals.cpp#L169-L170
> https://github.com/alliedmodders/sourcemod/blob/51bba0b8942faef8366b426057f64d10c798a62b/extensions/sdkhooks/natives.cpp#L234-L237
> 
> https://github.com/alliedmodders/sourcemod/blob/775d4f04f25edc10531d615ccd9e977cd82d467a/extensions/sdktools/vnatives.cpp#L1675-L1676
> https://github.com/alliedmodders/sourcemod/blob/775d4f04f25edc10531d615ccd9e977cd82d467a/extensions/sdktools/vnatives.cpp#L1721-L1722
> A no check, that I myself justified using the same arguments used in this PR
> https://github.com/alliedmodders/sourcemod/pull/1653#discussion_r760973885
> 
> https://github.com/alliedmodders/sourcemod/blob/536750b428ebf0ae82fb2b274765b990d0f9c742/extensions/cstrike/natives.cpp#L255-L260

Edited : After discussion, it has been decided null checks should not be removed. Instead null checks have been added where they had been missing.

# This should be field tested

I haven't had the chance to try the changes myself. This is all speculative, but I'm confident in the research I've performed.
Attached to this post is a zip file containing the sourcemod installation compiled with the changes of this PR (linux), minus the pgsql extension + spcomp, which gave me troubles to be compiled because of zlib.
~~[sourcemod.zip](https://github.com/alliedmodders/sourcemod/files/10375229/sourcemod.zip)~~ OUTDATED
~~[sourcemod.zip](https://github.com/alliedmodders/sourcemod/files/10375813/sourcemod.zip)~~ OUTDATED
[sourcemod.zip](https://github.com/alliedmodders/sourcemod/files/10377543/sourcemod.zip) - Latest

### DIFF
--- a/public/IGameHelpers.h
+++ b/public/IGameHelpers.h
@@ -357,7 +357,7 @@ namespace SourceMod
 		 *
 		 * @return				ServerClass pointer on success, nullptr on failure.
 		 */
-		virtual ServerClass *FindServerClass(CBaseEntity *pEntity) = 0;
+		virtual ServerClass *FindEntityServerClass(CBaseEntity *pEntity) = 0;
 	};
 }
 
--- a/public/IGameHelpers.h
+++ b/public/IGameHelpers.h
@@ -40,7 +40,7 @@
  */
 
 #define SMINTERFACE_GAMEHELPERS_NAME		"IGameHelpers"
-#define SMINTERFACE_GAMEHELPERS_VERSION		11
+#define SMINTERFACE_GAMEHELPERS_VERSION		12
 
 class CBaseEntity;
 class CBaseHandle;
@@ -351,6 +351,13 @@ namespace SourceMod
 		 * @return				64-bit server Steam id.
 		 */
 		virtual uint64_t GetServerSteamId64() const =0;
+
+		/**
+		 * @brief Finds a given entity's server class.
+		 *
+		 * @return				ServerClass pointer on success, nullptr on failure.
+		 */
+		virtual ServerClass *FindServerClass(CBaseEntity *pEntity) = 0;
 	};
 }
 
--- a/core/HalfLife2.cpp
+++ b/core/HalfLife2.cpp
@@ -413,6 +413,17 @@ ServerClass *CHalfLife2::FindServerClass(const char *classname)
 	return pInfo->sc;
 }
 
+ServerClass *CHalfLife2::FindServerClass(CBaseEntity *pEntity)
+{
+	IServerNetworkable* pNetwork = ((IServerUnknown *)pEntity)->GetNetworkable();
+	if (pNetwork == nullptr)
+	{
+		return nullptr;
+	}
+
+	return pNetwork->GetServerClass();
+}
+
 DataTableInfo *CHalfLife2::_FindServerClass(const char *classname)
 {
 	DataTableInfo *pInfo = NULL;
--- a/core/HalfLife2.cpp
+++ b/core/HalfLife2.cpp
@@ -413,7 +413,7 @@ ServerClass *CHalfLife2::FindServerClass(const char *classname)
 	return pInfo->sc;
 }
 
-ServerClass *CHalfLife2::FindServerClass(CBaseEntity *pEntity)
+ServerClass *CHalfLife2::FindEntityServerClass(CBaseEntity *pEntity)
 {
 	IServerNetworkable* pNetwork = ((IServerUnknown *)pEntity)->GetNetworkable();
 	if (pNetwork == nullptr)
--- a/core/HalfLife2.h
+++ b/core/HalfLife2.h
@@ -205,6 +205,7 @@ public: //IGameHelpers
 	bool FindSendPropInfo(const char *classname, const char *offset, sm_sendprop_info_t *info);
 	datamap_t *GetDataMap(CBaseEntity *pEntity);
 	ServerClass *FindServerClass(const char *classname);
+	ServerClass *FindServerClass(CBaseEntity *pEntity);
 	typedescription_t *FindInDataMap(datamap_t *pMap, const char *offset);
 	bool FindDataMapInfo(datamap_t *pMap, const char *offset, sm_datatable_info_t *pDataTable);
 	void SetEdictStateChanged(edict_t *pEdict, unsigned short offset);
--- a/core/HalfLife2.h
+++ b/core/HalfLife2.h
@@ -205,7 +205,7 @@ public: //IGameHelpers
 	bool FindSendPropInfo(const char *classname, const char *offset, sm_sendprop_info_t *info);
 	datamap_t *GetDataMap(CBaseEntity *pEntity);
 	ServerClass *FindServerClass(const char *classname);
-	ServerClass *FindServerClass(CBaseEntity *pEntity);
+	ServerClass *FindEntityServerClass(CBaseEntity *pEntity);
 	typedescription_t *FindInDataMap(datamap_t *pMap, const char *offset);
 	bool FindDataMapInfo(datamap_t *pMap, const char *offset, sm_datatable_info_t *pDataTable);
 	void SetEdictStateChanged(edict_t *pEdict, unsigned short offset);
--- a/core/smn_entities.cpp
+++ b/core/smn_entities.cpp
@@ -1252,10 +1252,6 @@ static cell_t SetEntDataString(IPluginContext *pContext, const cell_t *params)
 	SendProp *pProp; \
 	IServerUnknown *pUnk = (IServerUnknown *)pEntity; \
 	IServerNetworkable *pNet = pUnk->GetNetworkable(); \
-	if (!pNet->GetEdict() || pNet->GetEdict()->IsFree()) \
-	{ \
-		return pContext->ThrowNativeError("Entity %d (%d) is not networkable", g_HL2.ReferenceToIndex(params[1]), params[1]); \
-	} \
 	if (!g_HL2.FindSendPropInfo(pNet->GetServerClass()->GetName(), prop, &info)) \
 	{ \
 		const char *class_name = g_HL2.GetEntityClassname(pEntity); \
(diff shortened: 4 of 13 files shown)


Kenzzer:
So compiling Sourcemod with clang-3.8 didn't please a few extensions, that hack themselves into sourcemod. So I moved up to clang-11, still debian buster
[sourcemod.zip](https://github.com/alliedmodders/sourcemod/files/10377543/sourcemod.zip)
Also fully removed the "networkable" check inside the Set/GetEntProp natives. It never made any sense to begin with.

KyleSanderson:
> So compiling Sourcemod with clang-3.8 didn't please a few extensions, that hack themselves into sourcemod. So I moved up to clang-11, still debian buster [sourcemod.zip](https://github.com/alliedmodders/sourcemod/files/10377543/sourcemod.zip) Also fully removed the "networkable" check inside the Set/GetEntProp natives. It never made any sense to begin with.

Have you reviewed https://bugs.alliedmods.net/show_bug.cgi?id=5297 (and the child bug?).

Kenzzer:
> > So compiling Sourcemod with clang-3.8 didn't please a few extensions, that hack themselves into sourcemod. So I moved up to clang-11, still debian buster [sourcemod.zip](https://github.com/alliedmodders/sourcemod/files/10377543/sourcemod.zip) Also fully removed the "networkable" check inside the Set/GetEntProp natives. It never made any sense to begin with.
> 
> Have you reviewed https://bugs.alliedmods.net/show_bug.cgi?id=5297 (and the child bug?).

Ty for sharing this. No I hadn't thought of checking this bug tracker.
But this bug ticket seems to corroborate the issue this PR raise.
https://hg.alliedmods.net/sourcemod-experimental/rev/13ea64fe5237

Instead of trusting the edict's networkable ptr, you instead directly go for the CBaseEntity
```cpp
inline const char *	CBaseEdict::GetClassName() const
{
	if ( !m_pUnk )
		return "";
	return m_pNetworkable->GetClassName();
}
```
We address the same problem once again. Gamerules natives, retrieve the edict's networkable, but we should be retrieving the serverunknow's(CBaseEntity) networkable.

So I addressed that, and I also took the liberty of eliminating all the null checks against networkable interface. As they're artifacts of a bygone time when the interface was instead found through the edict.

Kenzzer:
## New field test build ?
Sorry to the people that wish to field test this. I have no SM builds to provide atm, I'll upload one later today or tomorrow.

## Null checks update
I've re-added the removed null checks, and I added new ones where they had been previously missing.
I've also added null checks around `IServerNetworkable::GetServerClass` because a few parts of SM already null checked it. I believe this is in scope of the PR still, as the ServerClass pointer is currently only retrieved through IServerNetworkable which this PR brings consistency for.

Examples of where ServerClass is currently null checked on `master`
https://github.com/alliedmodders/sourcemod/blob/250dc1b206998daad3ca1e688409c17794a2913f/extensions/tf2/criticals.cpp#L170-L175

## Streamline code further ?

I also thought it might be a good idea to overload IGameHelpers::FindServerClass
https://github.com/alliedmodders/sourcemod/blob/250dc1b206998daad3ca1e688409c17794a2913f/public/IGameHelpers.h#L102
And introduce `SeverClass* IGameHelpers::FindServerClass(CBaseEntity*)`, so that in the event a mod doesn't have a networkable interface for its entities, we can just create a `ifdef SE_GAME` in this function to add support. Without having to do that for all the natives in SourceMod that currently do
```cpp
IServerNetworkable *pNet = pUnk->GetNetworkable();
if (!pNet)
{
	return;
}

ServerClass *pClass = pNet->GetServerClass();
if (!pClass)
{
	return;
}
```
to
```cpp
ServerClass *pClass = gamehelpers/halflife2->GetServerClass(pUnk/pEntity);
if (!pClass)
{
	return;
}
```

## The null status of IServerNetworkable, open discussion

I also wish to bring something new to this discussion. While rebrowsing source engine code, I remembered that `IServerNetworkable` was a return value of `IEntityFactory`, I therefore want to ask again. Do we want to continue staying on the side of safety and null check IServerUnknown::GetNetworkable still ?

Kenzzer:
Decided to roll with the idea of overloading `IGameHelpers::FindServerClass` (let me know if I bumped the interface version correctly, first time I'm doing this).

In my opinion this does streamline the code much further, and overall improves understandability of the code in certain places. The retrieval of the `IServerNetworkable` was only ever used to ultimately call `GetServerClass`, so let's hide that away in the event a game breaks away from this "convention", it will be less code to maintain overall.

Kenzzer:
The gamerules fix has been uploaded to my server for a week now, and the crash hasn't re-surfaced. Every other changes are either a merely extra null check, or the use of the new vtable entry of IGameHelpers.

I haven't noticed any regression or new bugs following those changes. Attached to this post is a build of the latest commit for those that wish to try it privately (tf2 only).
[sourcemod_1.12_3e2520418a4c375c2a9040df4d63aa4d2bbfe152.tar.gz](https://github.com/alliedmodders/sourcemod/files/10440226/sourcemod_1.12_3e2520418a4c375c2a9040df4d63aa4d2bbfe152.tar.gz)

public/IGameHelpers.h
```diff
@@ -351,6 +351,13 @@ namespace SourceMod
 		 * @return				64-bit server Steam id.
 		 */
 		virtual uint64_t GetServerSteamId64() const =0;
+
+		/**
+		 * @brief Finds a given entity's server class.
+		 *
+		 * @return				ServerClass pointer on success, nullptr on failure.
+		 */
+		virtual ServerClass *FindServerClass(CBaseEntity *pEntity) = 0;
```
psychonic, line 360: This isn't safe to add like this since it won't actually get added to the end of the vtable. (It will be next to the other overload). An extension compiled against v11 is going to have a bad time if they access this version. At minimum, IsVersionCompatible would need to be overridden to return `false` if a lower version is requested. If not wanting to break compatibility, that could be taken one step further to add an additional interface registration with a shim providing the previous version. Another alternative is just using a different name.

I'm not sure how much any of that is worth the addition.

asherkin: Just giving it a different name so it's simple compat seems like the way to go.

psychonic: That'd be fine, too.

psychonic (approved):

PR #1907, opened 2023-01-15
fix: Print full map name in Log instead of args

Rushaway:
Print the full map name in LogAction instead of the args.
Exemple : !map de_
**Before:** 
`.Rushaway<150><[U:1:44340774]><> changed map to de_`

**After:**
`.Rushaway<150><[U:1:44340774]><> changed map to de_dust2`


psychonic:
There may be value in logging the original input string as well.

Something like: `"\"%L\" changed map to \"%s\" (input \"%s\")"`. Thoughts?

Rushaway:
@psychonic Here you go.

psychonic (approved):

psychonic:
Thanks!

PR #1915, opened 2023-01-30
Update Military Conflict: Vietnam gamedata

Poggicek:
CommitSuicide and Ignite seem to have shifted at some point and would crash the server when called

### DIFF
--- a/gamedata/sdktools.games/engine.mcv.txt
+++ b/gamedata/sdktools.games/engine.mcv.txt
@@ -229,7 +229,7 @@
 			}
 			"CommitSuicide"
 			{
-				"windows"	"510"
+				"windows"	"511"
 			}
 			"GetVelocity"
 			{


Poggicek:
I didn't check every offset, however for example GiveNamedItem offset still matched even though it's between two shifted offsets. Perhaps other methods were removed from the vtable making other offsets untouched. The removed whitespaces were caused by prettifier in my editor that went unnoticed until I made the commit. Looking at other gamedata there's quite some chaos around whitespaces, some new lines have such, others don't

Poggicek:
Hm, the ignite method seems sketchy, the offset would indeed match the correct one. So perhaps something was added before CommitSuicide and shifted the entire vtable. That however means that Ignite likely changed code-wise which crashes when called.

Poggicek:
Reverted Ignite offset, ValveCall parameters have to be modified instead

PR #1925, opened 2023-02-04
Fix EntityFactoryCaller signature

Poggicek:
Previous signature pointed to an incorrect function resulting into crashes when `GetEntityFactoryDictionary` got called. `sm_dump_datamaps(_xml)` and `sm_dump_classes` use such.
`EntityFactoryOffset` is still correct after the update


psychonic (approved):

Natanel-Shitrit:
Epic reversing 😅

Issue #1936, opened 2023-02-18
Calling SQL_LockDatabase twice is crashing server


Bara:
# Environment
  * SourceMod version: 1.11.0.6916
  * Metamod: version: 1.11.0-dev+1148

# Description
I randomly found the issue and is reproducible with the following code, sqlite and mysql doesn't matter.

# Problematic Code (or Steps to Reproduce)
  ```
public void OnPluginStart()
{
    char sError[512];
    Database db = SQL_Connect("clientprefs", true, sError, sizeof(sError));

    if (db == null)
    {
        SetFailState("Error: %s", sError);
        return;
    }

    SQL_LockDatabase(db);
    SQL_LockDatabase(db);
    SQL_UnlockDatabase(db);
}
  ```

Last message in console was the following:
```
sm plugins load crashtest
**** WARNING: Watchdog timer exceeded, aborting!
```

I'm not really familiar how SQL_LockDatabase works, but this should be checked to avoid a crash like this.

sirdigbot:
Seems the std::mutex used by LockDatabase has undefined behaviour if the currently-owning thread locks again.
Maybe that's it?

Wend4r:
There is a similar problem with server crash when you reload plugin that is waiting for a SQL query response.
But I think it needs to be solved globally, rewriting the basis for working with different databases in other threads (one DB/TCP connection = one thread) to don't clog the queue flow (this problem has old server holders with a long response from DB).
And from the point of view of SourceMod, which is basically designed to work with the Source Engine, also use the Valve Thread Pool (aka IThreadPool) for this (it looks like #1689).

Bara:
#1937 was merged, so I think this can be closed too

peace-maker:
Your server would still deadlock, no? Unbalanced Lock/Unlock calls.

PR #1937, opened 2023-02-19
Fix crash from DB locking

sirdigbot:
Fix for #1936

Not sure if I've gone about this the right way, but it seems to make sense and fixes the crashing with the test code.
```sourcepawn
public void OnPluginStart()
{
  char sError[512];
  Database db = SQL_Connect("clientprefs", true, sError, sizeof(sError));

  if (db == null)
  {
      SetFailState("Error: %s", sError);
      return;
  }

    
  PrintToServer("ABOUT TO LOCK");

  SQL_LockDatabase(db);
  SQL_LockDatabase(db);
  SQL_UnlockDatabase(db);
  
  PrintToServer("REACHED UNLOCK");
}
```
image

### DIFF
--- a/extensions/mysql/mysql/MyDatabase.cpp
+++ b/extensions/mysql/mysql/MyDatabase.cpp
@@ -277,27 +277,13 @@ IPreparedQuery *MyDatabase::PrepareQuery(const char *query, char *error, size_t 
 
 bool MyDatabase::LockForFullAtomicOperation()
 {
-	std::thread::id myId = std::this_thread::get_id();
-	if (m_LockHolder == myId)
-	{
-		return true;
-	}
-
 	m_FullLock.lock();
-	m_LockHolder = myId;
 	return true;
 }
 
 void MyDatabase::UnlockFromFullAtomicOperation()
 {
-	std::thread::id defaultId;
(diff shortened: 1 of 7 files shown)


dvander:
The API should not be used, at all, ever. The best fix is to remove it.

dvander:
And, why: it's fundamentally broken. One, it blocks the main thread, which is a no-no. Two, a native error, spurious or otherwise, could cause the database to become permanently locked. Neither of these are fixable.

So, IMO, deciding on the "right" behavior is moot, when there is a trivial workaround: having two database connections.

Wend4r:
I think a similar thing needs to be done by separate PR/changes for User Messages (I was dealing with Protobuf).
https://github.com/alliedmodders/sourcemod/blob/1fbe5e1daaee9ba44164078fe7f59d862786e612/core/smn_usermsgs.cpp#L52 .
When UM is called inside high UM, the permanently locking to call more because return throwing (breaks the SourcePawn backtrace, jumps to the founder invoke (aka forward)).
I have to restart the server (Issue: SM does not safely unload from MM:Source).

KyleSanderson (approved):

PR #1939, opened 2023-02-28
Test Improvements: Gamedata checker + CI scripts

Mooshua:
Hi folks.

This is a collection of powershell scripts and a workflow file that automate the testing of gamedata files against server binaries with minimal dependencies.

 - Only dependencies on host machine are [Docker](https://docs.docker.com/engine/install/), [PowerShell](https://learn.microsoft.com/en-us/powershell/scripting/install/installing-powershell?view=powershell-7.3), and [Lua](https://ulua.io/).
 - Will produce non-zero exit code if any of the *signatures* are invalid, and will place human-readable text in stdout describing any other warnings GDC produces. This output is very concise ([see for yourself](https://github.com/Mooshua/sourcemod/actions/runs/4319383995/jobs/7538507572))
 - Adds a GitHub workflow file which checks the gamedata for all supported games every 24 hours, after every push, and before every pull request. 
 - It currently natively supports CS:GO, CS:S, DOD:S, L4D2, L4D, TF2, IOSoccer, and NMRIH--but more games can be added fairly easily.

**Of Note--** the scripts are super janky. Like, if the CEO of jank came down from jankville in their private ~~jet~~ jank then they'd say "dang that's jank" and purchase a 20% stake in the thing. So I might be iteratively improving and de-janking these. However, I believe at their current state the scripts are simple, clear, and *very* easy to maintain, so I'm going to ignore the jankening for now.

Thanks! 😄 

### [Sample Run With Bad Signature Here](https://github.com/Mooshua/sourcemod/actions/runs/4319383995)
(The bad sig is from NMRIH, which was recently fixed by dysphie in #1940.)


Mooshua:
The scripts have the functionality to export game binaries as artifacts which can be downloaded from GitHub. (This is so people can easily compare recent versions without having to do much digging)

I‘ve disabled it in these initial commits because of copyright concerns, but since the binaries are free anyways (with an anonymous steam login), do we want to go ahead and enable binary artifacts?

asherkin:
My feeling is that it doesn't make sense to bring this in-tree. It makes a lot more sense in my opinion to run this as a separate community project somewhere (we could probably arrange some hosting resources if that would be beneficial though). 

It's really nice work and definitely useful, the SM source tree and/or AM oversight just don't feel like the best place for it to grow and serve the community right now.

Mooshua:
Thats fair. I will investigate avenues of creating this functionality outside the tree—hopefully that allows any solutions to be helpful to the entire SM community and not just for the core functionality.

Thanks for getting back to me! I’ll be closing this for now but the scripts are always ready if you change your mind (or anyone else changes it for you 😉)

PR #1941, opened 2023-03-03
fix mismatched delete []

sapphonie:
subject is just a `char*`


psychonic (approved):

ashort96:
If `subject` is a `char*`, isn't `delete []` correct?

sapphonie:
strdup needs to be free'd, not deleted. ugh

dvander:
Bonus points for using unique_ptr or std::string.

PR #1908, opened 2023-01-18
Improve HasEntProp performance

sirdigbot:
Implements the `HasEntProp` optimization suggested by asherkin in the discord (caching missed lookups).
I tried to make any changes as non-disruptive as possible because I barely know what I'm doing, so it could probably be optimized further.

Testing shows it to be roughly 2-3x faster:
```
100 (* 6) iterations without fix:
[HasEntProp] Average Time:1.230063ms
[HasEntProp] Average Time:1.186231ms
[HasEntProp] Average Time:1.137192ms
Average = (3.553486ms / 3) = 1.18449533ms

100 (* 6) iterations with fix:
[HasEntProp] Average Time:0.383327ms
[HasEntProp] Average Time:0.375259ms
[HasEntProp] Average Time:0.443078ms
Average = (1.201664ms / 3) = 0.4005546ms
```

Using the following code, specifically `sm_hasentprop2` (which is probably not a good test):
The props used are either common, less common, or invalid for each of Prop_Send and Prop_Data.
```sourcepawn
#include <sourcemod>
#include <profiler>

#pragma semicolon 1
#pragma newdecls required

public Plugin myinfo = {
    name        = "",
    author      = "",
    description = "",
    version     = "0.0.0",
    url         = ""
};

public void OnPluginStart()
{
    PrintToServer("Ready to test HasEntProp!");
    
    RegConsoleCmd("sm_hasentprop", Cmd_HasEntProp);
    RegConsoleCmd("sm_hasentprop2", Cmd_HasEntProp2);
}


Action Cmd_HasEntProp(int client, int args)
{
    int maxEnts = GetMaxEntities();

    Profiler p = new Profiler();
    for (int i = 1; i <= 4; ++i)
    {
        int ents;
        int name;
        int silencer;
        int speed;
        int attack;
        
        p.Start();
    
        for (int j = MaxClients + 1; j <= maxEnts; ++j)
        {
            if (!IsValidEntity(j))
                continue;
                
            ++ents;
 
            if (HasEntProp(j, Prop_Send, "m_iName"))
                ++name;

            if (HasEntProp(j, Prop_Send, "m_bSilencerOn"))
                ++silencer;
            
            if (HasEntProp(j, Prop_Send, "invalid prop"))
                PrintToChatAll("%i INVALID?! (Prop_Send)", j);
                
            if (HasEntProp(j, Prop_Data, "m_flSpeed"))
                ++speed;

[...]

PR #1943, opened 2023-03-04
Add methods for clearing netprop cache

sirdigbot:
As mentioned in PR #1908, changing the behaviour of the datamap cache would break certain extensions.
[Kenzzer suggested adding functionality for extensions to clear the cache as a fix](https://github.com/alliedmodders/sourcemod/pull/1908#issuecomment-1399471595) (though don't ask me how that works).

I never got a response in the discord if this was a good solution or not, but here's a PR to at least keep track of things.

### DIFF
--- a/core/HalfLife2.cpp
+++ b/core/HalfLife2.cpp
@@ -503,6 +503,26 @@ bool CHalfLife2::FindDataMapInfo(datamap_t *pMap, const char *offset, sm_datatab
 	return true;
 }
 
+void CHalfLife2::ClearDataTableCache()
+{
+	m_Maps.clear();
+}
+
+void CHalfLife2::ClearDataTableCache(datamap_t *pMap)
+{
+	m_Maps.removeIfExists(pMap);
+}
+
+void CHalfLife2::ClearSendPropCache()
+{
+	m_Classes.clear();
+}
+
+bool CHalfLife2::ClearSendPropCache(const char *classname)
+{
+	return m_Classes.remove(classname);
+}
+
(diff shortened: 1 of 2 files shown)


KyleSanderson:
Bumping these ifaces are painful, so while this PR does nothing right now without hacking(?). It seems ok for something that could be exposed later if there's actual pain.

Kenzzer:
While the PR exposes a clear function, this needs to go further. It'd be really sweet if the gamehelpers interface was updated to expose these new functions to extensions. A proper netprop & datamap clear (which this PR's title doesn't mention, but adds anyways) would be welcomed addition.
There are extensions/plugins out there, mine included, that create custom sendprops and datamaps for entities. Unfortunately if said extensions/plugins are unloaded/reloaded well some custom datamaps might not exist anymore or wrongly cached as non existing, which is definitively going to create a bunch of crash/errors through sourcemod's natives `SetEntPropX`.

Kenzzer:
Thank you for the changes !

> I thought netprop was the generic term for both, whereas sendprop and datamap/datatable were the specific terms?

No sendprop and netprop are synonymous. And datamaps are datamaps (aka dataprops), and datatable is their container. Just like sendtable is the container of sendprops.

peace-maker (approved):
I'm fine with either name datamap or datatable.

Kenzzer:
@sirdigbot I believe this PR has waited for long enough, would you be able to solve the conflicts so we can merge this ?

Kenzzer (approved):

PR #1944, opened 2023-03-04
Add function to find a TopMenu Item

Ilusion9:
This can be used to change the logic of TopMenu items (find item -> remove -> create again) registered by base plugins (basecommands for example). I will use this to change the logic for sm_who (instead of flags, I will display their groups) so I don't have to modify basecommands whenever its updated.


KyleSanderson:
It looks like you may have forgotten to include the header change

peace-maker:
The function needs to be added to the public ITopMenu interface too.
https://github.com/alliedmodders/sourcemod/blob/master/public/extensions/ITopMenus.h

PR #1962, opened 2023-03-29
Translation cleanup

peace-maker:
Do a clean sweep over the translation phrases, weeding out all obsolete, empty, and moved phrases in all languages.

The `"Please select a map"` phrase was present twice. In the plugin.basecommands.txt as well as in the basevotes.phrases.txt files. Only keep the basecommands one, since both plugins load that file already.

The `nextmap` plugin didn't use the phrases at all, so I've opted to start using them 😆 The "ago" phrase cannot be used as is and should be changed to include the time string so it can be reordered in the translation file.


KyleSanderson (approved):
Breaking those out into logical commits helped a lot, nice one.

PR #1972, opened 2023-04-02
Add Shuffle method to ArrayList

sirdigbot:
Needed it for a plugin and felt like it was so simple it should just exist as a stock.
Algorithm is [Durstenfeld's in-place Fisher–Yates shuffle](https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle#The_modern_algorithm) if you're curious.


Batfoxkid:
Doesn't `ArrayList.Sort(Sort_Random, X);` achieve the same effect?

PR #1973, opened 2023-04-02
Handle &datapack -> DataPack &datapack

Deathreus:


Deathreus:
Very important, must be merged post haste

peace-maker:
This was kept as `Handle` for backwards compatibility intentionally iirc. Old plugins would start throwing a tag mismatch warning when they pass a `Handle` from before `DataPack` was added. Or am I tripping?

KyleSanderson:
I don't believe you're tripping.

PR #1974, opened 2023-04-02
Update Brazilian Phrases

crashzk:
https://github.com/orgs/alliedmodders/projects/1/views/1?pane=issue&itemId=24307179

### DIFF
--- a/translations/pt/adminhelp.phrases.txt
+++ b/translations/pt/adminhelp.phrases.txt
@@ -29,5 +29,4 @@
 	{
 		"pt"		"Nenhum resultado encontrado"
 	}
-
 }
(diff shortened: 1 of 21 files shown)


psychonic (approved):

peace-maker:
@Tiagoquix Do you have concerns with these changes?

crashzk:
@Tiagoquix I understand your point, in this case it would just be a translation that I already use in my CSGO servers. In any case, I can convert it to leave it translated again as before, in which case I only sent what I already had.

Anyway, as soon as possible I'll update again and leave it in lower case.

AllTalk question, I believe we can leave it translated right then and remove (AllTalk) from the sentence, right?

crashzk:
Done, let me know if I need to change anything else.

crashzk:
Done, all right now, so I believe.

_Note: Leaving the AllTalk message translated and also putting (AllTalk) in the sentence, I believe it applies what you said, translates something and leaves others in English._

psychonic (approved):

PR #1985, opened 2023-04-15
fix: Baguette (french) translations

Rushaway:
https://github.com/orgs/alliedmodders/projects/1/views/1?pane=issue&itemId=24307187

### DIFF
--- a/translations/fr/nominations.phrases.txt
+++ b/translations/fr/nominations.phrases.txt
@@ -67,7 +67,7 @@
 
 	"Map Not In Pool"
 	{
-		"fr"		"La map '{1}' n'est pas dans la piscine de nominations."
+		"fr"		"La map '{1}' n'est pas dans la liste des nominations."
 	}
 
 }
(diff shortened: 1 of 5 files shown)


translations/fr/nominations.phrases.txt
```diff
@@ -65,4 +65,9 @@
 		"fr"		"Sélectionnée"
 	}
 
+	"Map Not In Pool"
+	{
+		"fr"		"La map '{1}' n'est pas dans la piscine de nominations."
```
Kenzzer, line 70: `Piscine` is way too of literal translation of `pool`. Consider `liste` instead or another non literal translation word that means selection/list, not a literal swimming pool.

Also is it really `de nominations`, feels like it should be `des nominations` in french. IIRC you can only use `de` for a plural word when the given word doesn't change in the singular tense.

Kenzzer (commented):
Can confirm the translations as a french individual. But I do have one nitpick about `Map Not In Pool`.

translations/fr/nominations.phrases.txt
Rushaway: It's true. Merci à vous. :)

asherkin (approved):
Thanks!

PR #1986, opened 2023-04-15
Add missing Danish translations

Rushaway:
https://github.com/orgs/alliedmodders/projects/1/views/1?pane=issue&itemId=24307184
Thanks to Hobbitten for corrections.

### DIFF
--- a/translations/da/basetriggers.phrases.txt
+++ b/translations/da/basetriggers.phrases.txt
@@ -20,6 +20,11 @@
 		"da"		"Friendly Fire er slået fra."
 	}
 
+	"Friendly Fire Percent"
+	{
+		"da"		"Fjendtligt skade er {1}%."
+	}
+
 	"Current Map"
 	{
 		"da"		"Den nuværende bane er {1}."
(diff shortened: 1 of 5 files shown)


translations/da/basetriggers.phrases.txt
```diff
@@ -20,6 +20,11 @@
 		"da"		"Friendly Fire er slået fra."
 	}
 
+	"Friendly Fire Percent"
+	{
+		"da"		"Fjendtligt skade er {1}%."
```
walliski, line 25: Similar translations are using 'Friendly Fire' instead of the translated version 'Fjendtligt skade'. Should probably keep it consistent.

translations/da/rockthevote.phrases.txt
```diff
@@ -22,7 +22,7 @@
 
 	"Already Voted"
 	{
-		"da"		"Du har allerede stemt på Rock The Vote."
+		"da"		"Du har allerede stemt på Rock The Vote. ({1} stemmer, {2} mangler)"
```
walliski, line 25: `påkrævet` could be a better word than `mangler`, as its used in other translations for the same meaning. The word `mangler` also seems to indicate "missing", which is not the case as the other number is the total required amount of votes, and not the missing part.

peace-maker (approved):

PR #1988, opened 2023-04-15
Add missing Italian translations

Zabaniya001:


translations/it/plugin.basecommands.txt
```diff
@@ -129,5 +129,9 @@
 	{
 		"it"		"Seleziona una mappa"
 	}
-
+	
+	"Choose Config"
+	{
+		"en"			"Scegli una configurazione"
```
walliski, line 135: Wrong language code

Zabaniya001: Dammit, good catch. I've proof checked it a few times but somehow missed it. Thanks.

peace-maker (approved):

PR #1993, opened 2023-04-16
Make KeyValues.Rewind traversal-stack clearing optional

sirdigbot:
As suggested by asherkin in the discord.
Hopefully implemented properly; not sure if there's other behaviour I'm not factoring in.
Tested with the following code:
```sourcepawn
char data[] = 
"\"TestRoot\""
... "{"
... "    \"node 1\""
... "    {"
... "        \"node 2\""
... "        {"
... "            \"somekey\" \"somevalue\""
... "        }"
... "    }"
... "}";

public void OnPluginStart()
{
    KeyValues kv = new KeyValues("blah");
    kv.ImportFromString(data);

    kv.JumpToKey("node 1");
    PrintToServer("1. nodes: %i", kv.NodesInStack());
    
    kv.JumpToKey("node 2");
    PrintToServer("2. nodes: %i", kv.NodesInStack());
    
    kv.JumpToKey("somekey");
    PrintToServer("3. nodes: %i", kv.NodesInStack());
    
    
    kv.Rewind();
    PrintToServer("rw1. nodes: %i", kv.NodesInStack());
    PrintToServer("goback: %i", kv.GoBack());
    

    kv.JumpToKey("node 1");
    PrintToServer("4. nodes: %i", kv.NodesInStack());
    
    kv.JumpToKey("node 2");
    PrintToServer("5. nodes: %i", kv.NodesInStack());
    
    kv.JumpToKey("somekey");
    PrintToServer("6. nodes: %i", kv.NodesInStack());
    
    
    kv.Rewind(false);
    PrintToServer("rw2. nodes: %i", kv.NodesInStack());
    PrintToServer("goback: %i", kv.GoBack());
}
```
image

### DIFF
--- a/plugins/include/keyvalues.inc
+++ b/plugins/include/keyvalues.inc
@@ -245,9 +245,9 @@ methodmap KeyValues < Handle
 	public native bool SavePosition();
 
 	// Jumps back to the previous position.  Returns false if there are no
-	// previous positions (i.e., at the root node).  This should be called
-	// once for each successful Jump call, in order to return to the top node.
-	// This function pops one node off the internal traversal stack.
+	// previous positions (i.e., at the root node with an empty traversal stack).
+	// This should be called once for each successful Jump call, in order to return
+	// to the top node.  This function pops one node off the internal traversal stack.
 	//
 	// @return              True on success, false if there is no higher node.
 	public native bool GoBack();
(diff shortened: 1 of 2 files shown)


plugins/include/keyvalues.inc
```diff
@@ -272,12 +272,14 @@ methodmap KeyValues < Handle
 	//                        thus the state is as if KvGoBack() was called.
 	public native int DeleteThis();
 
-	// Sets the position back to the top node, emptying the entire node
-	// traversal history.  This can be used instead of looping KvGoBack()
+	// Sets the position back to the top node, optionally emptying the entire
```
asherkin, line 275: Minor wordsmithing, but "Optionally" here implies not-default behaviour to me - perhaps swap it out with "By default" (or better construction of your choice.)

asherkin (commented):
One minor doc comment, then this is good to go I think. 

I'm not super familiar with this area either, but what you've implemented is what I expected it to look like at least!

plugins/include/keyvalues.inc
sirdigbot: Yeah I did think it was a bit clunky, I'm not great with docs. 
I also don't like it saying 'empty' while the parameter says "clear" because it feels mismatched, so I've fixed both now.

asherkin (approved):
LGTM

PR #1999, opened 2023-05-18
Fix flag

1mpulse:
Corrects the check for the VDECODE_FLAG_ALLOWNOTINGAME flag


KyleSanderson:
It looks like this may have been based incorrectly. Can you please update your PR?

asherkin:
The logic here looks correct to me - when you resubmit this please include a more complete explanation.

1mpulse:
> The existing logic here looks correct to me - when you resubmit this please include a more complete explanation.



> It looks like this may have been based incorrectly. Can you please update your PR?

i create new pull request

PR #2001, opened 2023-05-27
Deprecate CreateInterface

Kenzzer:
# Why this change ?

`CreateInterface_MMS` was created 15 years ago. But `smsdk_ext.cpp` was never updated to have that change, and has continued to rely on the old function `CreateInterface` for 16 years.

On this aspect alone, I think this warrants an update.

However if this isn't enough to warrant the change. It should be noted that tier1 library the various SM extensions compile against, has its own definition of  `CreateInterface` which can clash with smsdkext's `CreateInterface` and cause linkage issues. So rather than cause headaches for sm extensions authors like myself lol, let's just axe this old function.

# New SDKExtension virtual

`METAMOD_PLUGIN *SDK_OnMetamodCreateInterface(const MetamodVersionInfo *mvi, const MetamodLoaderInfo *mli)`

Which exposes to the extension the two structs passed by metamod. As well as allowing greater control should extensions want, for whatever reason, return a different `ISmmPlugin` interface.

### DIFF
--- a/public/smsdk_ext.cpp
+++ b/public/smsdk_ext.cpp
@@ -297,6 +297,12 @@ void SDKExtension::SDK_OnDependenciesDropped()
 
 #if defined SMEXT_CONF_METAMOD
 
+#if defined _MSC_VER
+	#define SMEXT_DLL_EXPORT				extern "C" __declspec(dllexport)
+#else
+	#define SMEXT_DLL_EXPORT				extern "C" __attribute__((visibility("default")))
+#endif
+
 PluginId g_PLID = 0;						/**< Metamod plugin ID */
 ISmmPlugin *g_PLAPI = NULL;					/**< Metamod plugin API */
 SourceHook::ISourceHook *g_SHPtr = NULL;	/**< SourceHook pointer */
@@ -308,27 +314,9 @@ IServerGameDLL *gamedll = NULL;				/**< IServerGameDLL pointer */
 #endif
 
 /** Exposes the extension to Metamod */
-SMM_API void *PL_EXPOSURE(const char *name, int *code)
+SMEXT_DLL_EXPORT METAMOD_PLUGIN *CreateInterface_MMS(const MetamodVersionInfo *mvi, const MetamodLoaderInfo *mli)
 {
-#if defined METAMOD_PLAPI_VERSION
-	if (name && !strcmp(name, METAMOD_PLAPI_NAME))
-#else
-	if (name && !strcmp(name, PLAPI_NAME))
-#endif
-	{
-		if (code)
-		{
-			*code = META_IFACE_OK;
-		}
-		return static_cast<void *>(g_pExtensionIface);
-	}
-
-	if (code)
-	{
-		*code = META_IFACE_FAILED;
-	}
-
-	return NULL;
+	return g_pExtensionIface;
 }
 
 bool SDKExtension::Load(PluginId id, ISmmAPI *ismm, char *error, size_t maxlen, bool late)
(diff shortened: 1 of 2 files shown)


Kenzzer:
> I have mixed feelings about this. I don't like that `CreateInterface_MMS` does not appear to be documented at all, despite it seeming to be otherwise intended for use.

Neither was CreateInterface documented as the function to define in order to expose a given binary as a metamod plugin. In fact that seems more of an attempt of Metamod to piggyback valve's interface system, most likely in attempt to keep things uniform.

However `CreateInterface_MMS` is documented here, and it clearly stated that metamod will call this function first and then `CreateInterface` if `CreateInterface_MMS` is not found, or returned `nullptr`

https://github.com/alliedmodders/metamod-source/blob/86b59173091e8ffa342942bd0bdd3806b035b5d6/core/ISmmPluginExt.h#LL159C1-L170C1

> I also don't understand without digging further, why this isn't causing an issue for first-party builds of first-party extensions that use these SMSDK helper files.

In my scenario I narrowed the why. And it turns out, SM extensions are just incredibly lucky. The static library linkage doesn't bring over everything. `CreateInterface` is defined inside `interface.o` and unless a function from this place is used, either by the extension itself, or other sections of tier1 library such as `KeyValues.o`. The symbol will not be imported, thus not cause any compile issues.

You can test this yourself by making a blank sourcemod extension, and then use `KeyValues` anywhere. You will not be able to compile. But if you define the function `HMODULE Sys_LoadLibrary( const char *pLibraryName, Sys_Flags flags ) { return (HMODULE)0; }`. `interface.o` will not be brought, and everything will compile fine.

SM Core, which is using KeyValues is not affected. Because the core defines `CreateInterface_MMS`
https://github.com/alliedmodders/sourcemod/blob/8a7f8c5cbdb549167d56ff22b0aa582f61f110b5/loader/loader.cpp#L191

> That said, there is definitely an issue here, but I think the lighter fix would be to just replace the manually-created `PL_EXPOSURE` function with `PL_EXPOSURE_FUNC(METAMOD_PLAPI_NAME, g_pExtensionIface)`. As long as `META_NO_HL2SDK` is properly defined or not, depending on the extension type, I believe that will avoid conflict.
> 
> Related: [alliedmodders/metamod-source#34](https://github.com/alliedmodders/metamod-source/pull/34)

I think this works (but will test to be sure). Cause this seems very sketchy, we end up relying on valve's macro underneath `EXPOSE_SINGLE_INTERFACE_GLOBALVAR` and I don't think that's a good idea.

asherkin:
These do appear to both be lightly documented on the MM:S wiki - it looks like the MM:S-specific loader came about as part of the 1.6 API, so wasn't really usable here generally until core-legacy was much more recently removed (we did do that, right?)

I believe `CreateInterface` was the original entrypoint so that binaries could be both server bin replacements (later VSPs) and MM:S plugins in one, to aid in the migration to MM:S. I don't think this is important for the SM ext skeleton to care about. 

Happy with either myself I think, sort of surprised the MM:S attach even takes this route. The stub and sample extensions in the MM:S tree should probably have the same treatment, whichever path is taken to fix the underlying issue.

psychonic:
> Happy with either myself I think, sort of surprised the MM:S attach even takes this route. The stub and sample extensions in the MM:S tree should probably have the same treatment, whichever path is taken to fix the underlying issue.

The stub_mm and sample_mm plugins don't have this issue because they use the `PLUGIN_EXPOSE` macro as documented in ISmmPlugin.h

asherkin:
An, yes, of course - alright, I definitely think that's the better fix here then.

public/smsdk_ext.cpp
```diff
@@ -297,6 +297,12 @@ void SDKExtension::SDK_OnDependenciesDropped()
 
 #if defined SMEXT_CONF_METAMOD
 
+#if defined _MSC_VER
+	#define SMEXT_DLL_EXPORT				extern "C" __declspec(dllexport)
```
dvander, line 301: "# define" (it's against spec to have '#' not be the first character of the line)

psychonic (approved):

PR #2005, opened 2023-06-01
Fix mismatched memory release in regex clearmatch

bottiger1:
subject is created with strdup which should be freed with C free and not C++ delete


peace-maker:
#1941 :D

peace-maker (approved):

PR #2015, opened 2023-07-04
Run basic native tests using hl2sdk-mock in CI

peace-maker:
Use [hl2sdk-mock](https://github.com/alliedmodders/hl2sdk-mock) to load SourceMod in a mock srcds and run plugins. This allows for end-to-end tests of the native API without requiring to spin up a whole game server. Only x86_64 right now until hl2sdk-mock runs on x86 again :D

This only tests the sorting.inc API similar to the existing test plugin in the plugins/testsuite folder. Tests fail if the srcds output contains "FAIL", which might be a little fragile.

The testing infrastructure can be expanded and more tests added over time. The existing tests in the [plugins/testsuite](https://github.com/alliedmodders/sourcemod/tree/master/plugins/testsuite) folder can't be used as is, but could serve as a starting point to convert to using the `testing.inc` pattern.

Ultimately something like [PawnUnit](https://forums.alliedmods.net/showthread.php?t=195717) would be desirable to organize tests and ease assertions.

### DIFF
--- a/.github/workflows/mocktest.yml
+++ b/.github/workflows/mocktest.yml
@@ -1,0 +1,104 @@
+name: hl2sdk-mock tests
+on: 
+  push:
+    branches:
+     - master
+     - '[0-9]+.[0-9]+-dev'
+  pull_request:
+    branches:
+     - master
+     - '[0-9]+.[0-9]+-dev'
+jobs:
+  mock:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


Frenzoid (approved):

dvander:
This is awesome, super excited to see this!

dvander (approved):

PR #2018, opened 2023-07-12
Add IntMap methodmap

KitRifty:
This is a continuation of #579. In addition to taking the PR comments into account, this also adds `IntMap.ContainsKey` and `IntMap.Clone` and updates the tests. Both the `StringMap` and `IntMap` tests pass.

Some other notes:
- Unlike the original PR, I've opted to not rename `CellTrie` to `StringCellTrie` and `TrieSnapshot` to `StringTrieSnapshot` for review purposes, but I can change it if requested.
- For the tests, I wasn't sure if I should be using new or old syntax, so if the tests should use the new syntax I'll be happy to update it!

Original PR message:
>  This patch makes the following changes
>
>   - Refactors StringHashMap to a generic HashMap template
>   - Adds the IntHashMap class
>   - Adds new IntMap natives
>   - Adds IntMap tests to the tries test suite
>  
>  This does not change the existing Trie API in any way so as to retain backwards compatibility, and all existing and new test suite tests pass.


Headline (approved):
Hey, cool. Thanks for picking this up.This looks good to take in.

@peace-maker can you give this a look? Seems to me this covers everything

peace-maker:
Ah sorry, this fell off the stack.
Only thing I'm noticing is the tests could be run in CI in the mock testsuite.

Alienmario:
Can I request a backport to 1.12? This could sit in 1.13 for a while..

PR #2019, opened 2023-07-18
Add more functions to ArrayStack

Mikusch:
ArrayStack can be useful at times, but the lack of functions makes it pretty limiting. This PR adds new natives for peeking at the top value of the stack without actually removing it, as well as a property to fetch the amount of elements in the stack.

Adds the following natives:
`ArrayStack.Top`
`ArrayStack.TopString`
`ArrayStack.TopArray`

And a property:
`ArrayStack.Length`

### DIFF
--- a/core/logic/smn_adt_stack.cpp
+++ b/core/logic/smn_adt_stack.cpp
@@ -240,9 +240,9 @@ static cell_t PopStackCell(IPluginContext *pContext, const cell_t *params)
 	}
 	else
 	{
-		if (idx >= array->blocksize() * 4)
+		if (idx >= array->blocksize() * sizeof(cell_t))
 		{
-			return pContext->ThrowNativeError("Invalid byte %d (blocksize: %d bytes)", idx, array->blocksize() * 4);
+			return pContext->ThrowNativeError("Invalid byte %d (blocksize: %d bytes)", idx, array->blocksize() * sizeof(cell_t));
 		}
 		*buffer = (cell_t)*((char *)blk + idx);
 	}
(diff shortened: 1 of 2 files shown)


Mooshua:
More of a bikeshed, but a term more people will probably be familiar with (and I believe better explains it's behavior) for `Top` is `Peek`.

Mikusch:
> More of a bikeshed, but a term more people will probably be familiar with (and I believe better explains it's behavior) for `Top` is `Peek`.

I thought about using `ArrayStack.Peek` but decided not to, since most implementations I looked at do use the `Top` naming, for example C++'s `std::stack` and Valve's `CUtlStack`.

core/logic/smn_adt_stack.cpp
```diff
@@ -365,6 +365,32 @@ static cell_t ArrayStack_Pop(IPluginContext *pContext, const cell_t *params)
...
+	cell_t *blk = array->at(array->size() - 1);
+	size_t idx = (size_t)params[2];
+
+	cell_t rval;
+	if (params[3] == 0) {
+		if (idx >= array->blocksize())
+			return pContext->ThrowNativeError("Invalid block %d (blocksize: %d)", idx, array->blocksize());
+		rval = blk[idx];
+	} else {
+		if (idx >= array->blocksize() * 4)
```
dvander, line 386: sizeof(cell_t) instead of 4, here and below

sapphonie:
> > More of a bikeshed, but a term more people will probably be familiar with (and I believe better explains it's behavior) for `Top` is `Peek`.
> 
> I thought about using `ArrayStack.Peek` but decided not to, since most implementations I looked at do use the `Top` naming, for example C++'s `std::stack` and Valve's `CUtlStack`.

Why not both? Make Peek an alias for Top?

peace-maker (approved):
Cool, thank you! I think keeping it close to the SDK naming is sane.

PR #2025, opened 2023-07-26
[1.11] update maxplayers define for 07-25-23 tf2 update

sapphonie:
This should really be handled per game IMO, but until then, TF2 supports 100 players now. Hooray!

### DIFF
--- a/plugins/include/clients.inc
+++ b/plugins/include/clients.inc
@@ -69,7 +69,7 @@ enum AuthIdType
  * MAXPLAYERS is a hardcoded value as an upper limit.  MaxClients changes based on the server.
  */
 
-#define MAXPLAYERS      65  /**< Maximum number of players SourceMod supports */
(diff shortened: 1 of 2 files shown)


sapphonie:
https://github.com/ValveSoftware/source-sdk-2013/blob/0d8dceea4310fde5706b3ce1c70609d72a38efdf/mp/src/game/client/game_controls/SpectatorGUI.cpp#L229-L242
"spec_target_updated"
+
"player"

image

MAX_PLAYERS == 33 here

Same function on TF2:

image


( result - 1 ) <= 100 implies 101

sapphonie:
It should be noted that several base plugins use MAXPLAYERS defines so... not sure how to handle that one, lol

psychonic:
If and when we are ready to backport this from master, we'll just cherry-pick it from there.

Issue #2031, opened 2023-07-28
Bad replay cvar name in sdktools


hydrogen-mvm:
# Help us help you
  - [ ] I have checked that my issue [doesn't exist yet](https://github.com/alliedmodders/sourcemod/issues).
  - [ ] I have tried my absolute best to reduce the problem-space and have provided the absolute smallest test-case possible.
  - [ ] I can always reproduce the issue with the provided description below.

# Environment
  * Operating System version: Debian 11
  * Game/AppID (with version if applicable): 440 (TF2)
  * Current SourceMod version: 1.11.0.6924
  * Current SourceMod snapshot:  6924
  * Current Metamod: Source snapshot: 1.11.0-dev+1148
  - [x] I have updated SourceMod to the [latest version](https://www.sourcemod.net/downloads.php) and it still happens.
  - [x] I have updated SourceMod to the [latest snapshot](https://www.sourcemod.net/downloads.php?branch=dev) and it still happens.
  - [x] I have updated SourceMM to the [latest snapshot](https://sourcemm.net/downloads.php?branch=dev) and it still happens.

# Description

From this commit: https://github.com/alliedmodders/sourcemod/commit/f9309c83d28d955a478fd804374ebfade2b34453
The replay convar was written incorrectly, it should be **replay_enable** (without the "d").

The server is still crashing as reported in this PR https://github.com/alliedmodders/sourcemod/issues/1581 . I've been trying to get rid of the FileExists hook and it looks like the botched replay cvar name is preventing the workaround from executing.

I don't have the ability to recompile Sourcemod right now, or else I would test the fix myself. If I do manage to get around it, I will submit a PR.

Thanks.

hydrogen-mvm:
Upon further probing, it seems like even fixing that spelling error doesn't correct the issue. The value of replay_enabled.GetBool() always returns false even if replay_enabled is set to 1. However, locally creating a ConVarRef object in CHookManager::NetChannelHook correctly returns the cvar value (see picture below).

![image](https://github.com/alliedmodders/sourcemod/assets/29522123/53a9a24e-fc51-4706-9573-306e8b79105f)

I don't know what's the best way to "fix" the issue but perhaps someone more knowledgeable than me can propose an appropriate remedy. In the meantime, as a workaround, I'll just replace that replay_enabled.GetBool() with true.

hydrogen-mvm:
[...]

PR #2033, opened 2023-07-29
Replay workaround fix

hydrogen-mvm:
This fixes https://github.com/alliedmodders/sourcemod/issues/2031

Tested on Debian 11 and works fine with and without replay enabled. I would like this fix to also be backported to 1.11 (stable) if possible.

Thanks.


psychonic (approved):

psychonic:
Thanks!

PR #2034, opened 2023-07-31
Added stock function `RemoveAllWhitespace`

assyrianic:
`TrimString` doesn't handle space between glyphs.

### DIFF
--- a/plugins/include/string.inc
+++ b/plugins/include/string.inc
@@ -598,11 +598,12 @@ stock int ImplodeStrings(const char[][] strings, int numStrings, const char[] jo
 stock int RemoveAllWhitespace(char[] str)
 {
 	int j;
-	for (int i=0; str[i] != 0; i++)
+	for (int i; str[i] != 0; i++)
 	{
 		if (!IsCharSpace(str[i]))
 		{
-			str[j++] = str[i];
+			str[j] = str[i];
+			j++;
 		}
 	}
 	str[j] = 0;
--- a/plugins/include/string.inc
+++ b/plugins/include/string.inc
@@ -589,7 +589,7 @@ stock int ImplodeStrings(const char[][] strings, int numStrings, const char[] jo
 	return total;
 }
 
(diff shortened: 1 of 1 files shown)


dvander:
"Whitespace" is one word, so should be capitalized as if it were one word. To differentiate from TrimString, "RemoveAllWhitespace" would be a better name.

KyleSanderson:
Regex is likely the better approach here?

assyrianic:
> "Whitespace" is one word, so should be capitalized as if it were one word. To differentiate from TrimString, "RemoveAllWhitespace" would be a better name.

agreed, done!

peace-maker:
Cool, this needs documentation in a comment like the other stocks though!

assyrianic:
> Cool, this needs documentation in a comment like the other stocks though!

done!

assyrianic:
I changed the stock so that it'd return the length of the modified string. If it was not modified, the length should be the original length.

assyrianic:
Was thinking of adding another stock where it just removes a singular character:

```c
stock int RemoveAllOfChar(char[] str, int chr)
{
	int j;
	for (int i=0; str[i] != 0; i++)
	{
		if (str[i] != chr)
		{
			str[j++] = str[i];
		}
	}
	str[j] = 0;
	return j;
}
```

Fyren:
There's nothing wrong with the implementation here, but this is also probably too narrow.  I don't see evidence of this being a pain point looking through the forum or Discord.

assyrianic:
> 

Perhaps something like this?
```cpp
stock int TrimChars(char[] str, const char[] chrs)
{
	int j = 0;
	for (int i=0; str[i] != 0; i++)
	{
		bool is_chr = false;
		for (int n=0; chrs[n] != 0; n++)
		{
			is_chr = str[i]==chrs[n] || is_chr;
		}
		if (!is_chr)
		{
			str[j] = str[i];
			j++;
		}
	}
	str[j] = 0;
	return j;
}
```

PR #2035, opened 2023-07-31
Update Github Actions using dependabot

peace-maker:
Keep the actions we use up to date automatically. This avoids us having to worry about deprecation warnings of old action versions.

https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+updates:
+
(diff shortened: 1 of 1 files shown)


dvander (approved):
But what will automatically update dependabot?

KyleSanderson:
> But what will automatically update dependabot?

gaben.

PR #2036, opened 2023-07-31
Added integer-based base-2 & base-10 Logarithm stock

assyrianic:
Added 5 integer-based logarithm stocks.
`IntLog2`, `IntLog3`, `IntLog4`, `IntLog10` and `IntLog`.

Rationale:

Using the `Logarithm` native by itself isn't bad but given that it has to execute from the C++ side as well as perform floating-point operations, it wasn't performant enough for my use-case of bit-counting a bitset that was stored per player but accessed by name via hash table.

The integer Log2 uses a "de Bruijn" sequence combined with bitwise integer rounding to achieve a fast, base-2 logarithm.

Also for a bit of a performance reason, I added `IntLog3` & `IntLog4` for use with `IntLog`. Iterating a number as Log Base-2 would require looping & dividing at most 30 times which will be significantly less performant compared to a series of bitwise operations and a multiply-shift in a table.

The same case for Log base-3 and Log base-4 is similar but not as much as Log base-2 given the range of a 32-bit signed integer. Log base-3 would require 19 loop-&-divide's whereas Log base-4 requires 15. Log base-5 wasn't included as its max loop & division range is 13.

Log base-10 was added as it may prove useful such as counting how many digits a number requires which would be useful for cases such as dynamically sizing a character array for a HUD to display a number:

```c
int num = ...;
int num_digits = IntLog10(num) + 1;
char[] display_num = new char[num_digits + 1];
IntToString(num, display_num, num_digits);
```


assyrianic:
Wanted to add, I made `IntLog2` with the purpose of reversing a bit shift:

```c
int x = i << CONSTANT;
/// how to get back constant?
int my_constant = IntLog2(x);
```

assyrianic:
Discussed this with Headline and I came up with a better, smaller level of implementation of having an arbitrary logarithm table generator that can be invoked at startup of the plugin and then used throughout the rest of the runtime as needed.

```c
stock int GeneralIntLog(int x, const int[] logN_table, const int[] powers_of_N)
{
	int y = logN_table[ IntLog2(x) ];
	return y + view_as< int >(x >= powers_of_N[y + 1]);
}

enum
{
	MAX_LOG_TABLE_SIZE = 32,
	MAX_LOG_POWER_SIZE = 21,
};
stock int GenerateIntLogarithmTables(int logN_table[MAX_LOG_TABLE_SIZE], int powers[MAX_LOG_POWER_SIZE], int base=10)
{
	if (base < 3)
	{
		return 0;
	}
	/// Generate our powers of the base.
	powers[0] = 1;
	int n = 1;
	for (int i=1; i < sizeof(powers); i++)
	{
		int prev_pow = powers[i-1];
		int next_pow = prev_pow * base;
		if (IntLog2(next_pow) <= IntLog2(prev_pow))
		{
			break;
		}
		powers[i] = next_pow;
		n++;
	}
	
	/// generate guesses table.
	float fbase = base + 0.0;
	for (int i; i < sizeof(logN_table); i++)
	{
		int r = 1 << i;
		float fr = float(r);
		if (fr < 0)
		{
			fr = -fr;
		}
		logN_table[i] = RoundToFloor(Logarithm(fr, fbase));
	}
	return n;
}

PR #2040, opened 2023-08-29
Fix sound flags enumeration

KitRifty:
This fixes the values of sound flags since everything past SND_CHANGEPITCH was wrong. 

https://github.com/ValveSoftware/source-sdk-2013/blob/0d8dceea4310fde5706b3ce1c70609d72a38efdf/sp/src/public/soundflags.h#L115C18-L115C18

This also adds three other flags from SDK2013 which are SND_IGNORE_PHONEMES, SND_IGNORE_NAME, and SND_DO_NOT_OVERWRITE_EXISTING_ON_CHANNEL.

### DIFF
--- a/plugins/include/sdktools_sound.inc
+++ b/plugins/include/sdktools_sound.inc
@@ -72,18 +72,18 @@ enum
  */
 enum
 {
-	SND_NOFLAGS = 0,                                    /** Nothing */
-	SND_CHANGEVOL = 1,                                  /** Change sound volume */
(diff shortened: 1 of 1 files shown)


peace-maker (approved):
Wow this bug is as old as it gets
https://github.com/alliedmodders/sourcemod/commit/b5831fd2636a288773d6f1ec67ecfa1c0e8bf7cb#diff-a18bdcc0bbcad4c201d48eb0458280929c1f1f537cd76096e96a9382cfe85276R53

PR #2052, opened 2023-09-27
Update SourcePawn

peace-maker:
Bring in the latest changes (again) and fix the tooling Actions artifact on macOS.

### DIFF
--- a/AMBuildScript
+++ b/AMBuildScript
@@ -773,6 +773,8 @@ if SM.use_auto_versioning():
   )
 
 class SPRoot(object):
+  def __init__(self):
+    self.generated_headers = SM.generated_headers
   # SourcePawn's build scripts are always one-offs, and attach the current target
   # to the builder, so we have to provide a shim to our StaticLibrary() method.
   def StaticLibrary(self, builder, name):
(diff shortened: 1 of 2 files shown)


peace-maker:
There are problems with the versioning changes from https://github.com/alliedmodders/sourcepawn/commit/92b70f7be87fae7e6e7b296f9b27066b6eb2cd46 @dvander 
Should the bundled sourcepawn inherit sourcemod's version or not?

dvander:
I think it'd be good to start breaking version numbers in SP. Let SP do its thing and same to SM. If it helps, I can change the SP version to something ridiculous (like 2023) to motivate this

peace-maker:
I don't see a way to let SourcePawn use its own version number without touching SourcePawn. I'd think it'd be good to have SourcePawn put its own git-rev into it's version and not piggy-back on SourceMod. Those changes seem to be done in lock-step with both repos which I'm not feeling like thinking through right now. Having the latest sourcepawn changes and fixes here has higher priority for me atm, so let's delay that separation for later. I'll open an issue so we won't forget.

PR #2042, opened 2023-09-09
Japanese translations

Teamkiller324:
Japanese translation corrections.

PR #2053, opened 2023-09-27
Fix Japanese translations

CookStar:
This fix is to revert most of #2042.

First of all, honorific expressions such as "禁止されました" are unnecessary in this case, and "蹴れました" is also a strange expression. Secondly, there is no need to replace "BAN" with "禁止" or "キック" with "蹴(れ)", for 99.9% of Japanese readers can understand "BAN" and "キック" as they are, because "BAN" and "キック" are already part of the Japanese language. (Citation: https://e-words.jp/w/BAN.html) Finally, the space between the name seems unnecessary unless it is wrapped in double quotation marks, which is also to make it consistent with other translations.

### DIFF
--- a/translations/jp/playercommands.phrases.txt
+++ b/translations/jp/playercommands.phrases.txt
@@ -17,12 +17,12 @@
 
 	"Slapped target"
 	{
-		"jp"		"{1} を叩きました。"
+		"jp"		"{1}を叩きました。"
 	}
 
 	"Slayed target"
 	{
-		"jp"		"{1} を殺害しました。"
+		"jp"		"{1}を殺害しました。"
 	}
 
 	"Name changed"
@@ -32,7 +32,7 @@
 
 	"Renamed target"
 	{
-		"jp"		"{1} をリネームしました"
+		"jp"		"{1}をリネームしました。"
 	}
 
 	"Rename player"
--- a/translations/jp/plugin.basecommands.txt
+++ b/translations/jp/plugin.basecommands.txt
@@ -7,7 +7,7 @@
 
 	"No access to cvar"
 	{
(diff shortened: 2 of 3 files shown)


peace-maker:
@Teamkiller324 can you comment please? I don't speak Japanese

faketuna:
I agree reverting theese translations.

> honorific expressions such as "禁止されました" are unnecessary in this case

Yes. in this case honorific expressions is unnecessary.

> and "蹴れました" is also a strange expression

Yes. "蹴れました" is so strange in this sentence.

> Secondly, there is no need to replace "BAN" with "禁止" or "キック" with "蹴(れ)", for 99.9% of Japanese readers can understand "BAN" and "キック" as they are, because "BAN" and "キック" are already part of the Japanese language.

Yes. Although we don't use it much in our daily lives, but no exaggeration to say that there are few people who do not know what it means on the internet.
if replace "BAN" with "禁止" makes setence more hard to read. also replace "キック" with "蹴(れ)" is same result.

> Finally, the space between the name seems unnecessary unless it is wrapped in double quotation marks, which is also to make it consistent with other translations.

In Japanese, spaces are not necessary in each word like English. but I would recommended include spaces between variables and setence for readability. e.g. "プレイヤー xx を永久BANしました。" = Permanently banned player xx.
But not applicatable to numbers like "mm分間". because theese are combined as 1 word. e.g. "dd日" = dd day(s), "hh時間" = hh hour(s)

peace-maker (approved):
Thank you. I'll merge this since you all seem to agree on the changes. @Teamkiller324 can come back and add their thoughts to this PR or open a new one if they feel strongly about something.

PR #2060, opened 2023-10-09
make RenameFile() on Windows also replace the file

rtldg:
`RenameFile` on linux uses `rename()` which will replace the destination file if it exists. `MoveFileA` doesn't replace so this is swapping it with `MoveFileExA`+`MOVEFILE_REPLACE_EXISTING` which will replace the destination.


rtldg:
accidentally made the pr on this repo instead my fork when trying to test github actions building the binary for testing. if the binary works correctly though then I'll keep the pr open

psychonic (changes requested):
Looking at the available flags, I think we should probably add `MOVEFILE_COPY_ALLOWED` as well

rtldg:
> Looking at the available flags, I think we should probably add `MOVEFILE_COPY_ALLOWED` as well

Good idea. Added it.

psychonic (approved):
Thanks

PR #2071, opened 2023-10-28
Handle `INVALID_FUNCTION` mismatches at plugin boundaries

nosoop:
Translates the following uses of `INVALID_FUNCTION` for compatibility between plugins compiled before and after the [upcoming] SourcePawn changes:

- `GetNativeFunction`
- `Call_StartFunction` (probably unnecessary since trying to call a null function shouldn't? work)
- `DataPack`s now store `INVALID_FUNCTION` as 0 internally and retrieved based on caller's context

This should make it so plugins can remain mostly compatible regardless of what value of `INVALID_FUNCTION` they are compiled with.  Function coercion to values was disallowed a while back so I'd imagine most plugins communicate their use of functions through these three avenues.

I confirmed the behavior is correct for `GetNativeFunction`; the other two are untested.

### DIFF
--- a/core/logic/smn_functions.cpp
+++ b/core/logic/smn_functions.cpp
@@ -327,14 +327,11 @@ static cell_t sm_CallStartFunction(IPluginContext *pContext, const cell_t *param
 		}
 	}
 
-	IPluginContext *pTargetContext = pPlugin->GetBaseContext();
 	cell_t funcid = params[2];
 
-	if (pContext->IsNullFunctionId(funcid))
+	if (!pContext->IsNullFunctionId(funcid))
 	{
-		s_pFunction = pTargetContext->GetFunctionById(pTargetContext->GetNullFunctionValue());
-	} else {
-		s_pFunction = pTargetContext->GetFunctionById(funcid);
+		pPlugin->GetBaseContext()->GetFunctionByIdOrNull(funcid, &s_pFunction);
 	}
 
 	if (!s_pFunction)
--- a/core/logic/smn_functions.cpp
+++ b/core/logic/smn_functions.cpp
@@ -36,6 +36,7 @@
 #include <IHandleSys.h>
 #include <IForwardSys.h>
 #include <ISourceMod.h>
+#include "LegacyPluginCompat.h"
 
 HandleType_t g_GlobalFwdType = 0;
 HandleType_t g_PrivateFwdType = 0;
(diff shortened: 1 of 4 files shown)


core/logic/smn_functions.cpp
```diff
@@ -327,7 +327,15 @@ static cell_t sm_CallStartFunction(IPluginContext *pContext, const cell_t *param
 		}
 	}
 
-	s_pFunction = pPlugin->GetBaseContext()->GetFunctionById(params[2]);
+	IPluginContext *pTargetContext = pPlugin->GetBaseContext();
+	cell_t funcid = params[2];
+
+	if (pContext->IsNullFunctionId(funcid))
+	{
+		s_pFunction = pTargetContext->GetFunctionById(pTargetContext->GetNullFunctionValue());
```
dvander, line 335: I think you can just assign nullptr here?

dvander (changes requested):
Great idea, thanks for doing this!

nosoop:
One edge case I'm starting to see:

1.  Old plugin passes a non-`INVALID_FUNCTION` funcid 0 to new plugin.
2.  New plugin retrieves it (via `GetNativeFunction` and `DataPack.ReadFunction`) as funcid 0 and ends up being treated as `INVALID_FUNCTION`.

I think the simplest workaround would be to map a valid 0th function from a legacy plugin to a special value.  Will work that out shortly.

core/logic/smn_functions.cpp
nosoop: Yeah.  I didn't look into the implementation too deeply earlier; I think it's fine just to check that it's not the caller's null function ID and then do `GetFunctionByIdOrNull()`, otherwise leave it as NULL (as that was set earlier via `ResetCall()`).  Plus some extra handling for legacy valid 0 funcids.

nosoop:
Couple of other boundaries that I need to address:

- `AddToForward` / `PrivateForward.AddFunction`
- `RemoveFromForward` / `PrivateForward.RemoveFunction`
- `GetFunctionByName`

nosoop:
Actually, can function IDs ever be zero?  Just found `PublicIndexToFuncId`, and that always sets the LSB to 1 so it sounds like a non-null function ID of 0 shouldn't exist.

dvander:
Right, zero is never valid. It's just a question of silent/intended failure vs reported error.

nosoop:
Will rebase, revert the 0 function handling, and reopen this against master shortly.

PR #2074, opened 2023-11-02
Fix missing submodules

PlayBoy31:
add --recursive for missing submodules (particulary for MM witch use submodule for hl2sdk-manifests)

### DIFF
--- a/tools/checkout-deps.ps1
+++ b/tools/checkout-deps.ps1
@@ -44,7 +44,7 @@ Function Get-Repository
 
     If (-not (Test-Path $Name -PathType Container))
     {
-        & git clone $Repo -b $Branch $Name 2>&1 | Write-Host
+        & git clone --recursive $Repo -b $Branch $Name 2>&1 | Write-Host
         If ($Origin)
         {
             Set-Location $Name
--- a/tools/checkout-deps.sh
+++ b/tools/checkout-deps.sh
@@ -95,7 +95,7 @@ fi
 checkout ()
 {
   if [ ! -d "$name" ]; then
-    git clone $repo -b $branch $name
+    git clone --recursive $repo -b $branch $name
     if [ -n "$origin" ]; then
       cd $name
       git remote set-url origin $origin


dvander (approved):

psychonic:
Out of curiosity, was anything not working before this change? SourceMod doesn't use the hl2sdk manifests, and if/when it does, I presume we'd add it as a submodule as SM itself as well, rather than borrowing from MM:S's copy.

PlayBoy31:
> Out of curiosity, was anything not working before this change? SourceMod doesn't use the hl2sdk manifests, and if/when it does, I presume we'd add it as a submodule as SM itself as well, rather than borrowing from MM:S's copy.

Hello,

I was migrating my infrastructure from Debian 11 to Debian 12 and I wanted to update my SM and MM repos to recompile my SM extensions and VSP/MM plugins with a fresh installation. 

During the compilation of VSP plugins I had an error because by using the checkout-deps.sh script, it does not download the manifests necessary for the compilation of VSP/MM.

Another point concerning the build of SM, with the current version of script checkout-deps.sh, the script will fail because python on debian 12 set a warning about "error: externally-managed-environment" during the pip install.

I have update checkout-deps.sh to create then use a pyhon venv (instead of install on user lan)d which for me the best thing that we must all do !

Here the code :

```
    echo "Installing AMBuild with a python env"
    $python_cmd -m venv venv # create venv
    . venv/bin/activate # start venv
    python -m pip install ./ambuild # install ambuild on venv
    deactivate # stop venv
```
If you want I can create a merge request regarding this.

dvander:
AMBuild installs a script in PATH, and is not super useful if that doesn't work. I'd be in favor of overriding the warning.

That said if the ambuild command continues to work in a venv, I don't really mind either way.

PlayBoy31:
> AMBuild installs a script in PATH, and is not super useful if that doesn't work. I'd be in favor of overriding the warning.

Yes, that was the point, the current version install the AMBuild to have it on the PATH but with current of python on debian 12 its not a warning but an error then the script fail and dont install "ambuild" on the .local/bin ...

Maybe we can force the pip install with a specific flag ? (Dont have checked)

> That said if the ambuild command continues to work in a venv, I don't really mind either way.

For me its the power of the venv:
1) start the venv with cmd "source venv/bin/activate"
2) do the build with cmd "ambuild" (because when you install it with the venv you have it on PATH)
3) execute cmd "deactivate" to stop the venv when its builed

Perhaps this solution is not very user friendly but as it stands the script does not work on certain distributions.

We can also add a build.sh script which does the "configure", start of venv, build and stop of venv for SM and for MM?

dvander:
I wish when starting AMBuild someone had warned me: the cosmic punishment for writing Python is then you have to deal with Python.

Anyway, you can definitely force pip to ignore that error. That's what I did on my VMs. Pip's behavior here is not user friendly. "Install a package" should be something that works, no matter where it comes from. If getting AMBuild onto pypi fixes that issue, I'm happy to do that.

Obviously "do a bunch of Python weirdness to use AMBuild" is not something we're going to accept. It's already bad enough as it is, being a very obscure build system. If that magic can be bundled into the ambuild bash/bat script itself, that works though.

PR #2075, opened 2023-11-03
Update basetriggers.sp for Dystopia compatibility

bauxiteDYS:
Dystopia has a built-in command called nextmap which breaks due to the Sourcemod nextmap command. This modification is intended to restore Dystopia compatibility.

### DIFF
--- a/plugins/basetriggers.sp
+++ b/plugins/basetriggers.sp
@@ -65,6 +65,7 @@ ConVar g_Cvar_MaxRounds;
 #define PRINT_TO_ONE			2		/* Print to a single player */
 
 bool mapchooser;
+bool isGameDystopia;
 
 int g_TotalRounds;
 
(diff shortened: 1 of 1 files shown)


sapphonie:
Calling 
```c
char folder[64];   	 
GetGameFolderName(folder, sizeof(folder));
``` every `say` command is a huge waste. it's not going to change, and you already call it once during plugin start - you could just store the bool value of `strcmp(folder, "dystopia") != 0)` in OnPluginStart as a global variable called something like `doNextmap` or something and check it that way instead of doing so much extra logic.

Alternatively, Dystopia itself could use its own command, since SM has using `nextmap` for ages and ages, but that's a tall order

bauxiteDYS:
> Calling
> 
> ```c
> char folder[64];   	 
> GetGameFolderName(folder, sizeof(folder));
> ```
> 
> every `say` command is a huge waste. it's not going to change, and you already call it once during plugin start - you could just store the bool value of `strcmp(folder, "dystopia") != 0)` in OnPluginStart as a global variable called something like `doNextmap` or something and check it that way instead of doing so much extra logic.
> 
> Alternatively, Dystopia itself could use its own command, since SM has using `nextmap` for ages and ages, but that's a tall order

Thanks, makes sense.

Unfortunately Dystopia has no devs as of recently so that probably won't be changed.

psychonic (approved):

psychonic:
Thanks!

PR #2081, opened 2023-11-07
Add void function prototypes to NativeCall and MenuHandler

MAGNAT2645:
`void` NativeCall prototype can be used for simple natives which don't return any value.
`void` MenuHandler can be used for basic menu handlers that don't return modified item styles and don't redraw menu items.

This is done to silence compiler warnings 209 (`function has explicit 'int' tag but does not return a value`) and 242 (`function "FuncName" should return an explicit value`).


psychonic (approved):

psychonic:
Thanks.

PR #2107, opened 2024-01-27
Modernize Sample Extensions's AMBuildScript

caxanga334:
# Summary
This PR aims to modernize the AMBuildScripts of the sample extensions. Bringing them closer to SM own AMBuildScript and adding support for compiling 64 bits extensions.

## Sample 
The build script is based on SM own build script. It was updated to use the HL2SDK manifests and allow compiling for multiple archs. Minimum AMBuild version increased to 2.2.

Two new arguments added to configure.py.

- --hl2sdk-manifest-path : Path to the HL2SDK manifests, this is optional if the manifests already exists in the project root. Otherwise a copy of the manifests is created on the project root folder.
- --targets : Overrides target build arch.

## Sample (NO SDK)
The build script was updated to allow compiling for multiple archs. Minimum AMBuild version increased to 2.2.

One new argument added to configure.py.

- --targets : Overrides target build arch.

### DIFF
--- a/public/sample_ext/AMBuildScript
+++ b/public/sample_ext/AMBuildScript
@@ -32,13 +32,19 @@ def SetArchFlags(compiler):
 
 hl2sdk_manifests_path = None
 
-if not builder.options.hl2sdk_manifest:
-  raise Exception('HL2SDK Manifest root path not set! (--hl2sdk-manifest-path)')
-else:
-  hl2sdk_manifests_path = os.path.join(builder.options.hl2sdk_manifest, 'SdkHelpers.ambuild')
+# First we check if the manifest exists in the current path
+hl2sdk_manifests_path = os.path.join(builder.sourcePath, 'hl2sdk-manifests/SdkHelpers.ambuild')
 
-  if not os.path.exists(hl2sdk_manifests_path):
-    raise Exception('Could not find SdkHelpers.ambuild in the given HL2SDK Manifest path!')
+if not os.path.exists(hl2sdk_manifests_path):
+  # manifests does not exists in the project file, use the path from --hl2sdk-manifest-path
+  if not builder.options.hl2sdk_manifest:
+    raise Exception('HL2SDK Manifest root path not set! (--hl2sdk-manifest-path)')
+  else:
+    hl2sdk_manifests_path = os.path.join(builder.options.hl2sdk_manifest, 'SdkHelpers.ambuild')
+
+    if not os.path.exists(hl2sdk_manifests_path):
+      raise Exception('Could not find SdkHelpers.ambuild in the given HL2SDK Manifest path!')
+
 
 SdkHelpers = builder.Eval(hl2sdk_manifests_path, {
     'Project': 'sm-extension'
--- a/public/sample_ext/AMBuildScript
+++ b/public/sample_ext/AMBuildScript
@@ -1,47 +1,5 @@
 # vim: set sts=2 ts=8 sw=2 tw=99 et ft=python:
-import os, sys
-
-# Simple extensions do not need to modify this file.
-
-class SDK(object):
-  def __init__(self, sdk, ext, aDef, name, platform, dir):
-    self.folder = 'hl2sdk-' + dir
-    self.envvar = sdk
-    self.ext = ext
-    self.code = aDef
-    self.define = name
-    self.platform = platform
-    self.name = dir
-    self.path = None # Actual path
-
-WinOnly = ['windows']
-WinLinux = ['windows', 'linux']
-WinLinuxMac = ['windows', 'linux', 'mac']
-
-PossibleSDKs = {
-  'episode1':  SDK('HL2SDK', '1.ep1', '1', 'EPISODEONE', WinLinux, 'episode1'),
-  'ep2':  SDK('HL2SDKOB', '2.ep2', '3', 'ORANGEBOX', WinLinux, 'orangebox'),
-  'css':  SDK('HL2SDKCSS', '2.css', '6', 'CSS', WinLinuxMac, 'css'),
-  'hl2dm':  SDK('HL2SDKHL2DM', '2.hl2dm', '7', 'HL2DM', WinLinuxMac, 'hl2dm'),
-  'dods': SDK('HL2SDKDODS', '2.dods', '8', 'DODS', WinLinuxMac, 'dods'),
-  'sdk2013': SDK('HL2SDK2013', '2.sdk2013', '9', 'SDK2013', WinLinuxMac, 'sdk2013'),
-  'tf2':  SDK('HL2SDKTF2', '2.tf2', '12', 'TF2', WinLinuxMac, 'tf2'),
-  'l4d':  SDK('HL2SDKL4D', '2.l4d', '13', 'LEFT4DEAD', WinLinuxMac, 'l4d'),
-  'nucleardawn': SDK('HL2SDKND', '2.nd', '14', 'NUCLEARDAWN', WinLinuxMac, 'nucleardawn'),
-  'l4d2': SDK('HL2SDKL4D2', '2.l4d2', '16', 'LEFT4DEAD2', WinLinuxMac, 'l4d2'),
-  'darkm':  SDK('HL2SDK-DARKM', '2.darkm', '2', 'DARKMESSIAH', WinOnly, 'darkm'),
-  'swarm':  SDK('HL2SDK-SWARM', '2.swarm', '17', 'ALIENSWARM', WinOnly, 'swarm'),
-  'bgt':  SDK('HL2SDK-BGT', '2.bgt', '4', 'BLOODYGOODTIME', WinOnly, 'bgt'),
-  'eye':  SDK('HL2SDK-EYE', '2.eye', '5', 'EYE', WinOnly, 'eye'),
-  'csgo': SDK('HL2SDKCSGO', '2.csgo', '22', 'CSGO', WinLinuxMac, 'csgo'),
-  'portal2':  SDK('HL2SDKPORTAL2', '2.portal2', '18', 'PORTAL2', [], 'portal2'),
-  'blade':  SDK('HL2SDKBLADE', '2.blade', '19', 'BLADE', WinLinux, 'blade'),
-  'insurgency':  SDK('HL2SDKINSURGENCY', '2.insurgency', '20', 'INSURGENCY', WinLinuxMac, 'insurgency'),
-  'contagion':  SDK('HL2SDKCONTAGION', '2.contagion', '15', 'CONTAGION', WinOnly, 'contagion'),
-  'bms':  SDK('HL2SDKBMS', '2.bms', '11', 'BMS', WinLinux, 'bms'),
-  'doi':  SDK('HL2SDKDOI', '2.doi', '21', 'DOI', WinLinuxMac, 'doi'),
-  'pvkii': SDK('HL2SDKPVKII', '2.pvkii', '10', 'PVKII', WinLinux, 'pvkii'),
-}
+import os, shutil
 
 def ResolveEnvPath(env, folder):
   if env in os.environ:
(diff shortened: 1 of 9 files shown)


public/sample_ext/AMBuilder
```diff
@@ -22,10 +19,16 @@ else:
...
-  
-  binary = Extension.HL2Config(project, projectName + '.ext.' + sdk.ext, sdk)
+  if sdk['name'] in ['mock']:
+    continue
+
+  for cxx in builder.targets:
+    if not cxx.target.arch in sdk['platforms'][cxx.target.platform]:
+      continue
+
+    binary = Extension.HL2ExtConfig(project, builder, cxx, extName + sdk['extension'], sdk)
```
FortyTwoFortyTwo, line 32: Missing a dot at `extName + sdk['extension']`? I tried to compile it, got `tank.ext2.tf2.so` instead of `tank.ext.2.tf2.so`

caxanga334:
Hi @FortyTwoFortyTwo . Yeah I indeed missed a dot. Updated the scripts. Could you please test again?

FortyTwoFortyTwo:
> Hi @FortyTwoFortyTwo . Yeah I indeed missed a dot. Updated the scripts. Could you please test again?

All works fine for Linux, though trying in windows gives me `LINK : warning LNK4044: unrecognized option`, ignoring four TF2 windows libs. Not sure if it's my fault or it's the build script.

caxanga334:
> All works fine for Linux, though trying in windows gives me `LINK : warning LNK4044: unrecognized option`, ignoring four TF2 windows libs. Not sure if it's my fault or it's the build script.

I was unable to reproduce this when compiling the **sample_ext**. Tried both x86 and x86_64 MSVC.

bottiger1:
Thanks for the update. However I noticed that when you use -static-libstdc++ it crashes when you pipe a number bigger than a char into iostream like this.

`std::cout << (uintptr64_t)123;`

```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007f1db217b09b in char32_t std::(anonymous namespace)::read_utf8_code_point<char>(std::(anonymous namespace)::range<char const, true>&, unsigned long)
#1  0x00007f1db217cf3f in std::codecvt_base::result std::(anonymous namespace)::utf16_in<char, char16_t>(std::(anonymous namespace)::range<char const, true>&, std::(anonymous namespace)::range<char16_t, true>&, unsigned long, std::codecvt_mode, std::(anonymous namespace)::surrogates) ()
#2  0x00007f1db217d0a4 in std::codecvt<char16_t, char, __mbstate_t>::do_in(__mbstate_t&, char const*, char const*, char const*&, char16_t*, char16_t*, char16_t*&) const ()
#3  0x00007f1db21df14a in std::ostream& std::ostream::_M_insert<unsigned long>(unsigned long) ()
#4  0x00007f1db2107c1d in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > PLH::int_to_hex<unsigned long>(unsigned long) ()
#5  0x00007f1db20fff0e in PLH::x64Detour::hook()
```

I am not sure we can do anything about this or whether it is preferential to dynamically link to stdc++ instead. 

However if we don't statically link to stdc++ it would mean extensions wouldn't be able to load on an OS running older versions of stdc++ which seems to be quite often. stdc++ versions often change every OS release.

Kenzzer:
> I am not sure we can do anything about this or whether it is preferential to dynamically link to stdc++ instead. SRCDS appears to dynamically link.

The issue is loosely connected to alliedmodders/metamod-source#167 the fix from that part needs to be backported to metamod 1.12

And applied there as well (that's how extensions are loaded)
https://github.com/alliedmodders/amtl/blob/e38cfe3cbf2047916e4a68017840bd325a8f7080/amtl/os/am-shared-library.h#L64

bottiger1:
> > I am not sure we can do anything about this or whether it is preferential to dynamically link to stdc++ instead. SRCDS appears to dynamically link.
> 
> The issue is loosely connected to [alliedmodders/metamod-source#167](https://github.com/alliedmodders/metamod-source/pull/167) the fix from that part needs to be backported to metamod 1.12
> 
> And applied there as well (that's how extensions are loaded) https://github.com/alliedmodders/amtl/blob/e38cfe3cbf2047916e4a68017840bd325a8f7080/amtl/os/am-shared-library.h#L64

I'm assuming you mean sourcemod? It is already in metamod.

I just added RTLD_DEEPBIND, recompiled sourcemod, and it still crashes.

Kenzzer:
> I'm assuming you mean sourcemod? It is already in metamod.

I did mean metamod. Currently the fix was applied to metamod 2.0, not 1.12. I was pointing this out in the event an extension is compiled as a metamod plugin, as it will be loaded by metamod instead.

> I just added RTLD_DEEPBIND, recompiled sourcemod, and it still crashes.

Hmmm, I'll try investigating as well when I get the chance.

bottiger1:
I just checked, I'm already using 2.0, and double checked that it has the deepbind change.

 Metamod:Source Version Information
    Metamod:Source version 2.0.0-dev+1289
    Plugin interface version: 16:14
    SourceHook version: 5:5
    Loaded As: Valve Server Plugin
    Compiled on: Apr 20 2024 14:02:31
    Built from: https://github.com/alliedmodders/metamod-source/commit/b4aa055
    Build ID: 1289:b4aa055
    http://www.metamodsource.net/

public/sample_ext_nosdk/AMBuildScript
```diff
@@ -24,13 +24,45 @@ def ResolveEnvPath(env, folder):
...
     self.mms_root = None
     self.sm_root = None
+    self.all_targets = []
+    self.target_archs = set()
+
+    if builder.options.targets:
+      target_archs = builder.options.targets.split(',')
+    else:
+      target_archs = ['x86']
+      if builder.backend != 'amb2':
```
Kenzzer, line 49: https://github.com/alliedmodders/ambuild/blob/551dd3f630f90377cbe536fbaf12e0b6bca91253/ambuild2/frontend/amb2_gen.py#L40

Since you're forcing the api version to 2.2 in your configure.py file. That check is unnecessary.

public/sample_ext_nosdk/AMBuildScript
```diff
@@ -66,14 +98,32 @@ class ExtensionConfig(object):
...
+
+    for cxx in self.all_targets:
+        self.configure_cxx(cxx)
+
+  def configure_cxx(self, cxx):
+    if cxx.family == 'msvc':
+      if cxx.version < 1914 and builder.options.generator != 'vs':
+        raise Exception(f'Only MSVC 2017 15.7 and later are supported, full C++17 support is required. ({str(cxx.version)} < 1914)')
+    elif cxx.family == 'gcc':
+      if cxx.version < 'gcc-9':
```
Kenzzer, line 115: Nitpick, full C++17 support was actually achieved in gcc-8
https://gcc.gnu.org/projects/cxx-status.html

public/sample_ext_nosdk/AMBuildScript
```diff
@@ -66,14 +98,32 @@ class ExtensionConfig(object):
...
+        raise Exception('Only clang versions 5 or later are supported, full C++17 support is required.')
 
     if cxx.like('gcc'):
       self.configure_gcc(cxx)
-    elif cxx.vendor == 'msvc':
+    elif cxx.family == 'msvc':
       self.configure_msvc(cxx)
 
-    # Optimization
+    # Optimizaiton
```
Kenzzer, line 126: Please undo, lol.

public/sample_ext_nosdk/AMBuildScript
```diff
@@ -111,31 +160,39 @@ class ExtensionConfig(object):
...
       '-Wno-switch',
       '-Wno-array-bounds',
-      '-msse',
-      '-m32',
       '-fvisibility=hidden',
     ]
+    if cxx.target.arch in ['x86', 'x86_64']:
+      cxx.cflags += ['-msse']
+
+    cxx.cxxflags += ['-std=c++17']
```
Kenzzer, line 168: Merge this with the `cxx.cxxflags` right below.

public/sample_ext_nosdk/AMBuildScript
```diff
@@ -111,31 +160,39 @@ class ExtensionConfig(object):
...
       '-fvisibility-inlines-hidden',
     ]
-    cxx.linkflags += ['-m32']
 
-    have_gcc = cxx.vendor == 'gcc'
-    have_clang = cxx.vendor == 'clang'
-    if cxx.version >= 'clang-3.6':
+    have_gcc = cxx.family == 'gcc'
+    have_clang = cxx.family == 'clang'
+    if cxx.version >= 'clang-3.9' or cxx.version == 'clang-3.4' or cxx.version > 'apple-clang-6.0':
```
Kenzzer, line 180: Since we're forcing clang version 5 (and later). This check is unnecessary, also are those no-warning flags necessary to compile the sample extension. May I trouble you to just outright remove all of them, and verify if you can still compile the sample ext fine ?

public/sample_ext_nosdk/AMBuildScript
```diff
@@ -82,16 +132,15 @@ class ExtensionConfig(object):
...
+    elif cxx.target.platform == 'mac':
       self.configure_mac(cxx)
-    elif builder.target_platform == 'windows':
+    elif cxx.target.platform == 'windows':
       self.configure_windows(cxx)
 
-    # Finish up.
     cxx.includes += [
-      os.path.join(self.sm_root, 'public'),
+      os.path.join(builder.sourcePath, 'public'),
```
Kenzzer, line 143: Remove that line. This is going to create a lot of troubles.

public/sample_ext/AMBuildScript
```diff
@@ -261,27 +312,66 @@ class ExtensionConfig(object):
...
+
+  def Program(self, context, compiler, name):
+    compiler = compiler.clone()
+    SetArchFlags(compiler)
+    return self.ProgramBuilder(compiler, name)
+    
+  def StaticLibrary(self, context, compiler, name):
+    compiler = compiler.clone()
+    SetArchFlags(compiler)
+    return self.StaticLibraryBuilder(compiler, name)
```
Kenzzer, line 373: Unneeded for the sample ext script. Should be deleted.

public/sample_ext/AMBuildScript
```diff
@@ -261,27 +312,66 @@ class ExtensionConfig(object):
...
+    if '-lgcc_eh' in binary.compiler.linkflags:
+      binary.compiler.linkflags.remove('-lgcc_eh')
+    if binary.compiler.like('gcc'):
+      binary.compiler.linkflags += ['-lstdc++', '-lpthread']
+    if binary.compiler.like('msvc'):
+      binary.compiler.linkflags += ['/SUBSYSTEM:CONSOLE']
+    return binary
+    
+  def StaticLibraryBuilder(self, compiler, name):
+    return compiler.StaticLibrary(name)
```
Kenzzer, line 358: Unneeded for the sample ext script. Should be deleted.

Kenzzer (changes requested):

caxanga334:
Hi, @Kenzzer 

I've made the changes you requested. Please review and see if I didn't forgot anything.

Both scripts compiled successfully on Debian with Clang 11.

public/sample_ext/AMBuildScript
```diff
@@ -63,77 +21,149 @@ def ResolveEnvPath(env, folder):
...
+    self.sdk_manifests = SdkHelpers.sdk_manifests
+    self.sdk_targets = SdkHelpers.sdk_targets
 
     if builder.options.mms_path:
       self.mms_root = builder.options.mms_path
     else:
       self.mms_root = ResolveEnvPath('MMSOURCE112', 'mmsource-1.12')
       if not self.mms_root:
-        self.mms_root = ResolveEnvPath('MMSOURCE', 'metamod-source')
+        self.mms_root = ResolveEnvPath('MMSOURCE_DEV', 'metamod-source')
```
Kenzzer, line 122: Accidental change ? Shouldn't it be `MMSOURCE` ?

Kenzzer (changes requested):
Thanks for the changes! I spotted one more issue, my apologies for not catching it in the earlier review. Otherwise I believe this is good to merge.

public/sample_ext/AMBuildScript
caxanga334: No. It comes from SM's AMBuildScript.

https://github.com/alliedmodders/sourcemod/blob/master/AMBuildScript#L133

Kenzzer: Then you should also match that change with your second AMBuildScript file.
https://github.com/alliedmodders/sourcemod/blob/9da7e650e63c629eb20af4dd50e6bcdc071be1a1/public/sample_ext_nosdk/AMBuildScript#L90

Kenzzer (approved):

public/sample_ext/AMBuildScript
caxanga334: Done.

Kenzzer:
I'm ready to take this in. However I'd rather not shake up the build environments for current 3rd-party extensions, and introduce two competing standards, so I'm adding extra env var checks. Anyways, thanks for your work!

PR #2108, opened 2024-01-27
Fixup entinfo offset

Kenzzer:
Did an oopsie, ofc the vtable size grew on 64bits. I unfortunately am not lucky enough to be doing `linux gaming` so I never joined my tf2 server instance, and therefore never tripped the code handling this. That could have been avoided by running a plugin that spawns entity however, my bad!


Kenzzer:
And I somehow missed to commit the tf gamedata ? Well here it is !

psychonic (approved):

PR #2118, opened 2024-02-16
Enable /WX for msvc + Fix the various new warnings that appear for x64 compilation

Kenzzer:
This PR aims at bringing consistent compilation flags between linux and windows. Namely the use of `-Werror` equivalent `/WX` for msvc. When enabled, this will make sourcemod fail compilation for 64 bits (it compiles fine but with warnings otherwise). And I don't think we should ignore those warnings, especially if those locations involve code that's about messing with memory and pointers. Putting that aside, there are so many warnings that this clutter the console and makes it really annoying to develop anything, and while we could decrease the warning level or outright silence them like on linux with something similar to `-Wno-narrowing`, I don't think we should enable that behaviour any longer since those can be very useful warnings when compiling for two different targets.

PR is marked as draft until alliedmodders/amtl#128 & alliedmodders/sourcepawn#950 have landed, otherwise the github CI will fail. (I've currently updated the submodules to target my repo so the CI doesn't fail, this will be reverted)


Kenzzer:
Will re-make this PR eventually, but this time to fix warnings into smaller batches.

PR #2119, opened 2024-02-24
[Linux x64] Fix variant_t usage by sdktools and bintools

Malifox:
[Disclaimer: I'm not familiar with c++]
On TF2 64-bit Linux, a plugin calling AcceptEntityInput will lead to a null pointer segfault such as with the following:
```c++
SetVariantString(g_strItemModel[iItem]);
AcceptEntityInput(client, "SetCustomModel");
```
jit_call_x64 will classify g_Variant_t as ObjectClass::Memory, but later does not handle that case resulting in a default Assert(no --enable-optimize build) or a null pointer crash(--enable-optimize). I just placed it in the same if statement with ObjectClass::Pointer and that seemed to fix the crash issue, but AcceptEntityInput did not function.

g_Variant_t seems to be an array of bytes representation of the real variant_t object, but the size of variant_t changes between 32/64 bit due to memory alignment. x64 adds 4 bytes of padding after the union in variant_t which changes the total size to 24. I changed #define SIZEOF_VARIANT_T to 20/24 depending on architecture when compiling and accounted for the padding. I also heard longs are evil, so I replaced those as well.

This was only tested on Linux x64 with [alliedmodders/hl2sdk/pull/198](https://github.com/alliedmodders/hl2sdk/pull/198), I'm guessing Win64 would behave the same regarding alignment.

### DIFF
--- a/extensions/sdktools/variant-t.cpp
+++ b/extensions/sdktools/variant-t.cpp
@@ -48,7 +48,7 @@ class VariantFirstTimeInit
 public:
 	VariantFirstTimeInit()
 	{
-		*(uint32_t *)(&g_Variant_t[12 + VARIANT_T_PADDING]) = INVALID_EHANDLE_INDEX;
+		((CBaseHandle *)(&g_Variant_t[sizeof(int32_t)*3 + VARIANT_T_PADDING]))->Set(nullptr);
 	}
 } g_VariantFirstTimeInit;
 
@@ -159,13 +159,11 @@ static cell_t SetVariantEntity(IPluginContext *pContext, const cell_t *params)
 {
 	CBaseEntity *pEntity;
 	unsigned char *vptr = g_Variant_t;
-	CBaseHandle bHandle;
 
 	ENTINDEX_TO_CBASEENTITY(params[1], pEntity);
-	bHandle = reinterpret_cast<IHandleEntity *>(pEntity)->GetRefEHandle();
 
 	vptr += sizeof(int32_t)*3 + VARIANT_T_PADDING;
-	*(uint32_t *)vptr = (uint32_t)(bHandle.ToInt());
+	((CBaseHandle*)vptr)->Set(reinterpret_cast<IHandleEntity*>(pEntity));
 	vptr += sizeof(CBaseHandle);
 	*(fieldtype_t *)vptr = FIELD_EHANDLE;
 
--- a/extensions/sdktools/variant-t.cpp
+++ b/extensions/sdktools/variant-t.cpp
@@ -164,9 +164,9 @@ static cell_t SetVariantEntity(IPluginContext *pContext, const cell_t *params)
 	ENTINDEX_TO_CBASEENTITY(params[1], pEntity);
 	bHandle = reinterpret_cast<IHandleEntity *>(pEntity)->GetRefEHandle();
 
-	vptr += sizeof(int)*3;
-	*(size_t *)vptr = (size_t)(bHandle.ToInt());
-	vptr += sizeof(size_t);
+	vptr += sizeof(int)*2 + sizeof(size_t);
+	*(uint32_t *)vptr = (uint32_t)(bHandle.ToInt());
+	vptr += sizeof(uint32_t);
 	*(fieldtype_t *)vptr = FIELD_EHANDLE;
 
 	return 1;
--- a/extensions/sdktools/variant-t.cpp
+++ b/extensions/sdktools/variant-t.cpp
@@ -48,7 +48,7 @@ class VariantFirstTimeInit
 public:
 	VariantFirstTimeInit()
 	{
-		*(unsigned int *)(&g_Variant_t[12]) = INVALID_EHANDLE_INDEX;
+		*(uint32_t *)(&g_Variant_t[12 + VARIANT_T_PADDING]) = INVALID_EHANDLE_INDEX;
 	}
 } g_VariantFirstTimeInit;
 
@@ -57,7 +57,7 @@ static cell_t SetVariantBool(IPluginContext *pContext, const cell_t *params)
 	unsigned char *vptr = g_Variant_t;
 
 	*(bool *)vptr = (params[1]) ? true : false;
-	vptr += sizeof(int)*3 + sizeof(size_t);
+	vptr += sizeof(int32_t)*3 + VARIANT_T_PADDING + sizeof(CBaseHandle);
 	*(fieldtype_t *)vptr = FIELD_BOOLEAN;
 
 	return 1;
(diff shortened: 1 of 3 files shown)


Kenzzer:
>  I also heard longs are evil, so I replaced those as well.

Those comments were in regard to tf2 hl2sdk. Using `unsigned long` is fine so as long you remember it changes the size depending on which architecture you target. Replacing them with `size_t` has no effect, since it also changes size depending on the architecture.
Alignment hasn't changed on windows according to PR #2117, could you should me what allowed you to draw this conclusion ?

Finally if you're going to edit all the instances where we offset into variant_t, could you instead do this : 
```cpp
vptr += sizeof(int32_t) * 3 + VARIANT_T_PADDING + sizeof(CBaseHandle);
*(fieldtype_t *)vptr = FIELD_BOOLEAN;
```
It would be much more explicit than `+ sizeof(size_t)` or `+ sizeof(unsigned long)` which gives us the right offset for the wrong reasons.

Kenzzer:
Answering my own question, the padding for the union is real, disassemble `CVariantSaveDataOps::IsEmpty(SaveRestoreFieldInfo_t const&)`
```
mov     ecx, [rax+14h]
test    ecx, ecx
jz      short loc_D934F0
```
0x14 is the position of `fieldtype_t fieldType;`. Accounting for its size, this indeed gives us a size of 24 bytes for variant_t on linux.

And disassembling `variant_t::SetEntity` yields that `CHandle<CBaseEntity> eVal` is at 0x10, confirming that the extra 4 bytes are indeed between it and the union.

Malifox:
I'd seen mention that `unsigned long` has odd behavior in an x64 environment with it becoming 8 bytes on Linux, but remaining 4 bytes on windows, so I used size_t. It solves the issue, but does make the reason why unclear and harder to read. I do agree that making it more explicit would be a better route and I've updated to reflect that using your suggested format.

dvander:
Yes, Windows is LLP64 and Linux is LP64: https://en.wikipedia.org/wiki/64-bit_computing#64-bit_data_models

Kenzzer:
I didn't know about this behaviour discrepancy between linux and windows. All the more reason to avoid `long` in general then.

extensions/sdktools/variant-t.cpp
```diff
@@ -164,9 +164,9 @@ static cell_t SetVariantEntity(IPluginContext *pContext, const cell_t *params)
 	ENTINDEX_TO_CBASEENTITY(params[1], pEntity);
 	bHandle = reinterpret_cast<IHandleEntity *>(pEntity)->GetRefEHandle();
```
Kenzzer, line 165:
```suggestion
```

extensions/sdktools/variant-t.cpp
```diff
@@ -164,9 +164,9 @@ static cell_t SetVariantEntity(IPluginContext *pContext, const cell_t *params)
...
 
-	vptr += sizeof(int)*3;
-	*(unsigned long *)vptr = (unsigned long)(bHandle.ToInt());
-	vptr += sizeof(unsigned long);
+	vptr += sizeof(int32_t)*3 + VARIANT_T_PADDING;
+	*(uint32_t *)vptr = (uint32_t)(bHandle.ToInt());
```
Kenzzer, line 168:
```suggestion
	((CBaseHandle*)vptr)->Set(pEntity);
```

extensions/sdktools/variant-t.h
```diff
@@ -47,10 +52,10 @@ inline void _init_variant_t()
 {
 	unsigned char *vptr = g_Variant_t;
 
-	*(int *)vptr = 0;
-	vptr += sizeof(int)*3;
-	*(unsigned long *)vptr = INVALID_EHANDLE_INDEX;
-	vptr += sizeof(unsigned long);
+	*(size_t *)vptr = 0;
+	vptr += sizeof(int32_t)*3 + VARIANT_T_PADDING;
+	*(uint32_t *)vptr = INVALID_EHANDLE_INDEX;
```
Kenzzer, line 57:
```suggestion
	((CBaseHandle *)vptr)->Set(nullptr);
```

Kenzzer (commented):
A thought crossed my mind, since I suggested you use `sizeof(CBaseHandle)` it would make more sense that we treat the `vptr` as a `CBaseHandle` as well, much like `SetEntPropEnt` and `GetEntPropEnt` natives do.

extensions/sdktools/variant-t.h
Malifox: I guess the only nagging thought I have on this is that `Set(nullptr)` Is doing the same thing, but with an added if check. Depending on how I look at it, uint32_t can seem more straightforward as INVALID_EHANDLE_INDEX is the value always used here, but I also see that it breaks the theme when CBaseHandle is being used everywhere else.

extensions/sdktools/variant-t.cpp
Kenzzer: yeah that works too!

extensions/sdktools/variant-t.h
Kenzzer: It's more for the 'straightforwardness' of it all than anything. If we treat this section of the vptr as a `CBaseHandle` elsewhere but not here, that just makes confusing. Alternatively, different source games might differ in their definition of `CBaseHandle` and it won't always be a `uint32_t`

Malifox: That's true, it would for sure be safer than to always assume.

psychonic:
Is there a reason we're not just including variant_t.h and doing `sizeof(variant_t)`?

Kenzzer:
> Is there a reason we're not just including variant_t.h and doing `sizeof(variant_t)`?

I assumed there was, and didn't mention anything. Instead hoping you'd comment on it lol. 

I agree that if we were to include `variant_t.h` header, all of this code could be much much simpler. Remains to see if this brings any compilation warnings/errors from the hl2sdk. Hopefully not.

Malifox:
I had wondered the same too. I never really learned any languages outside of sourcepawn and this issue is quite a leap from that, so I largely just kept to the existing format and just assumed there was a reason for everything. Based on things I read at the time, I think the rules for memory alignment are not standardized and is up to each compiler for how much and where padding will be placed. In that case, using `sizeof(variant_t)` would be based on how the compiler you are using interprets that rather than the real variant_t that is being emulated with this. I would guess it's highly unlikely for there to ever be a size disagreement between gcc/clang with past/future versions here and with what the many games using this expects, but maybe adds that 0.01% chance for unforeseen consequences? That also makes me wonder if the position where padding for x64 is currently added could ever change and variant_t is just cursed.

bottiger1:
I tested this on Linux and it appears to work.

psychonic:
Thanks for getting this started. We've now merged the superseding version in #2149

PR #2121, opened 2024-03-04
fix: sorted alphabetically, remove duplicate 'ko'

Rushaway:


asherkin (changes requested):
`Korean` and `KoreanA` are both valid `cl_language` values for Korean - the entries were not duplicates.

See https://bugs.alliedmods.net/show_bug.cgi?id=4667

psychonic (approved):

PR #2133, opened 2024-04-18
Prevent recursion in sm_ServerCommandEx (fixes #967)

### LINKED ISSUE #967: ServerCommandEx() can't grab `stats` output
# Help us help you
  - [X] I have checked that my issue [doesn't exist yet](https://github.com/alliedmodders/sourcemod/issues).
  - [X] I have tried my absolute best to reduce the problem-space and have provided the absolute smallest test-case possible.
  - [X] I can always reproduce the issue with the provided description below.

# Environments
|Parameter         |Windows   |Linux          |
|------------------|----------|---------------|
|Game Client App ID|440       |440|
|Game Server App ID|232250    |232250|
|SourceMod Version |1.9.0.6275|1.9.0.6237|
|MetaMod Version   |1.10.7-dev|1.11.0-dev+1105|
  - [X] I have updated SourceMod to the [latest version](https://www.sourcemod.net/downloads.php) and it still happens.
  - [X] I have updated SourceMod to the [latest snapshot](https://www.sourcemod.net/downloads.php?branch=dev) and it still happens.
  - [X] I have updated SourceMM to the [latest snapshot](https://sourcemm.net/downloads.php?branch=dev) and it still happens.

# Description
It looks like `ServerCommandEx()` can't grab output for any command (if executed via server console), if this call placed in command handler, and can "kill" server if used in another subcommand with `ServerCommandEx()`.

  ```sourcepawn
#include <sourcemod>

public void OnPluginStart()
{
    RegConsoleCmd("sm_stats", CmdStats);
}

public Action CmdStats(int iClient, int iArgC)
{
    char szBuffer[512];
    ServerCommandEx(szBuffer, sizeof(szBuffer), "stats");
    
    ReplyToCommand(iClient, "Command length response: %d", strlen(szBuffer));
    ReplyToCommand(iClient, "Command output:\n%s", szBuffer);
    return Plugin_Handled;
}
  ```
1. Join to server with admin with access to `sm_rcon`.
2. Execute `sm_rcon stats`. Receive output (all ok).
3. Execute `sm_stats`. Receive output (all ok).
Server runs ok
4. Execute `sm_rcon sm_stats`, get a server freeze. Fatal freeze. He don't unfreeze.
In Task Manager, SRCDS uses fully one server CPU.
Server Freeze

ojlanders:
This fixes the issue with recursive `ServerCommandEx` in old engine (pre-Nuclear Dawn). The issue was that if `ServerCommandEx` is called recursively, `sm_conhook_start` and `sm_conhook_stop` will be added to the console buffer again. When that happens, it will set `g_OriginalSpewOutputFunc` to `SourcemodSpewOutputFunc` because the earlier call set the spew func to `SourcemodSpewOutputFunc`, and then when a line is printed, `SourcemodSpewOutputFunc` will call `SourcemodSpewOutputFunc` indefinitely until the stack overflows (in debug) or you get bored (in release). This patch fixes that by checking if we're already catching spew before adding `sm_conhook_start`/`stop` again.

### DIFF
--- a/core/smn_console.cpp
+++ b/core/smn_console.cpp
@@ -979,11 +979,15 @@ static cell_t sm_ServerCommandEx(IPluginContext *pContext, const cell_t *params)
 	}
 
 	engine->ServerExecute();
-
-	g_ShouldCatchSpew = true;
-	engine->ServerCommand("sm_conhook_start\n");
-	engine->ServerCommand(buffer);
-	engine->ServerCommand("sm_conhook_stop\n");
+	if (!g_ShouldCatchSpew) {
+		g_ShouldCatchSpew = true;
+		engine->ServerCommand("sm_conhook_start\n");
+		engine->ServerCommand(buffer);
+		engine->ServerCommand("sm_conhook_stop\n");
+	}
+	else {
+		engine->ServerCommand(buffer);
+	}
 
 	engine->ServerExecute();
 
--- a/core/smn_console.cpp
+++ b/core/smn_console.cpp
@@ -984,8 +984,7 @@ static cell_t sm_ServerCommandEx(IPluginContext *pContext, const cell_t *params)
 		engine->ServerCommand("sm_conhook_start\n");
 		engine->ServerCommand(buffer);
 		engine->ServerCommand("sm_conhook_stop\n");
-	}
-	else {
+	} else {
 		engine->ServerCommand(buffer);
 	}
 


core/smn_console.cpp
```diff
@@ -979,11 +979,15 @@ static cell_t sm_ServerCommandEx(IPluginContext *pContext, const cell_t *params)
...
-	engine->ServerCommand("sm_conhook_start\n");
-	engine->ServerCommand(buffer);
-	engine->ServerCommand("sm_conhook_stop\n");
+	if (!g_ShouldCatchSpew) {
+		g_ShouldCatchSpew = true;
+		engine->ServerCommand("sm_conhook_start\n");
+		engine->ServerCommand(buffer);
+		engine->ServerCommand("sm_conhook_stop\n");
+	}
+	else {
```
asherkin, line 988:
```suggestion
	} else {
```

Very minor nit, but prevailing style here is else on the same line as the closing brace.

asherkin (approved):
Thank you for the PR! It's really nice to see this bug fixed.

core/smn_console.cpp
ojlanders: Fixed, thank you.

psychonic:
Thanks!

PR #2136, opened 2024-04-19
Handle `INVALID_FUNCTION` mismatches at plugin boundaries, redux

adriansmares:
References https://github.com/alliedmodders/sourcemod/pull/2071

This PR picks back from the closure of the original one. The commits have been cherry picked on top of the current HEAD and the special handling of pre-1.12-zero-function-IDs has been dropped completely, as zero function IDs were invalid before 1.12.

The `INVALID_FUNCTION` is stored as a zero in the data packs, and it is converted to the appropriate (`-1`/`0`) function ID on load.

There are some explicit checks for non-null-but-negative-or-zero function IDs which are extra compared to the original code and the original PR. These are added under the assumption that it's better to have an explicit error than propagating 'wrong' function IDs around. Do let me know if these should be removed or not.


dvander (approved):
thanks for doing this. lgtm

PR #2138, opened 2024-04-21
[BMS] Fix SDKHooks IEntityListener

### LINKED ISSUE #2132: [BMS] Gamedata update
All plugins using SDKHooks cause the server crash after srcds update

Alienmario:
Fixes #2132 SDKHooks crashing after recent BlackMesa update.
Tested on Linux.

Thanks goes to futureunseen and Josh Hubi from Crowbar for tracking it down.


psychonic (approved):
Nice

PR #2141, opened 2024-04-26
64bit gamedata fixes

artvin01:


gamedata/core.games/engine.ep2valve.txt
```diff
@@ -34,8 +36,15 @@
 		{
 			"LevelShutdown"
 			{
+				/* not confirmed to be : CServerGameDLL:: LevelShutdown */
```
psychonic, line 39: Just a note for context...

The name of the block is misleading. `LevelShutdown` was initially used to find `gEntList` when logical entity support was first added. However, due to code or compilation differences in some games or over time, other functions have been used in it's place.

This is only used to help gEntList. Any unique signature can be used, not even having to be at the start of any specific function. All that matters is that the offset from the function is accurate for finding gEntList. It's referenced in multiple functions. I don't know what function we actually

psychonic (approved):

Issue #2145, opened 2024-05-01
Using an enum as an enum struct array member size causes compiler to crash


Bradasparky:
Version: Sourcemod Dev 7125

The compiler crashes no matter what the value of `A` is in this test case
```c++
enum 
{ 
    A = 2, 
}

enum struct B 
{ 
    int i[A]; 
}
```

Fyren:
Should be fixed by the above PR.

PR #2146, opened 2024-05-01
Update sourcepawn submodule

tsuza:
The error reported in https://github.com/alliedmodders/sourcemod/issues/2145 has been fixed in https://github.com/alliedmodders/sourcepawn/commit/9594f99766f9326f336a0307e1887e1317f18b48. 
This PR just updates the submodule to the latest remote commit ( I have not personally compiled & tested it ).

### DIFF
--- a/plugins/include/sorting.inc
+++ b/plugins/include/sorting.inc
@@ -95,7 +95,7 @@ native void SortStrings(char[][] array, int array_size, SortOrder order = Sort_A
  *                      0 if first is equal to second
  *                      1 if first should go after second
  */
-typedef SortFunc1D = function int (int elem1, int elem2, const int[] array, Handle hndl);
+typedef SortFunc1D = function int (any elem1, any elem2, const any[] array, Handle hndl);
 
 /**
  * Sorts a custom 1D array.  You must pass in a comparison function.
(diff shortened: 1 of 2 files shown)


dvander (approved):

dvander:
Looks like there are some build errors, I'll try to address those in SP this week.

dvander:
Oh, you just need to bump the AMTL version in your PR. That should fix it.

tsuza:
Sorry for the late commit. I hadn't realized that you had responded until now.

tsuza:
The mock test is failing due to it using "any". If we were to change it to "int", it'd compile & correctly sort the array.

However, I've noticed that the with the latest spcomp the plugin prints gibberish.
image (87)

dvander:
Looks like there is definitely some kind of breakage here, I'll take a look.

dvander:
Sorry this took so long. This was a real bug, that is now fixed on master. To move this PR forward, you should just need to:

(1) Bring sourcepawn to the latest changeset (08d4d0c8b0a54d5778a643c6e16e1cd5b093bcc6) and then make your suggested change to sorting.inc (changing the "int" parameter to "any" for SortCustom1D).

tsuza:
No worries.

I accidentally committed a wrong fix to test_sorting that I made as a test a while ago, which I've reverted.

I've applied the fix to `sorting.inc`. Though, shouldn't we also update
https://github.com/alliedmodders/sourcemod/blob/758714f4241204c4e0feed2cedeffda78ed3c5b8/plugins/include/sorting.inc#L98 and https://github.com/alliedmodders/sourcemod/blob/758714f4241204c4e0feed2cedeffda78ed3c5b8/plugins/include/sorting.inc#L122-L126?

dvander:
Yeah, it should probably be updated as well.

tsuza:
Done 👍

PR #2149, opened 2024-05-04
AcceptEntityInput 64bit fix

bottiger1:
Fixes AcceptEntityInput by using the struct definition instead of messy pointer math.

I could not add variant_t.cpp from the sdk because including cbase.h causes a large number of compile errors so I just pasted the 1 method into the sdktools file. If anyone wants to play around with making it work, you can uncomment the file include in the AMBuildScript file in the sdktools folder.

Includes a fix to bintools from malifox which is supposedly related, but I have no idea what it does to be honest.

https://github.com/alliedmodders/sourcemod/pull/2119

I only tested and confirmed this works on Linux and only on int and string variants.

### DIFF
--- a/extensions/sdktools/AMBuilder
+++ b/extensions/sdktools/AMBuilder
@@ -48,7 +48,12 @@ for sdk_name in SM.sdks:
     binary.compiler.cxxincludes += [
       os.path.join(builder.sourcePath, 'public', 'jit'),
       os.path.join(builder.sourcePath, 'public', 'jit', 'x86'),
+      os.path.join(builder.options.hl2sdk_root, sdk['path'], 'game', 'server'),
+      os.path.join(builder.options.hl2sdk_root, sdk['path'], 'game', 'shared'),      
     ]
+
+    #binary.sources += [os.path.join(builder.options.hl2sdk_root, sdk['path'], 'game', 'server', 'variant_t.cpp')]
+
     binary.compiler.defines += ['HAVE_STRING_H']
 
     if sdk['name'] != 'episode1':
--- a/extensions/sdktools/AMBuilder
+++ b/extensions/sdktools/AMBuilder
@@ -48,9 +48,16 @@ for sdk_name in SM.sdks:
     binary.compiler.cxxincludes += [
       os.path.join(builder.sourcePath, 'public', 'jit'),
       os.path.join(builder.sourcePath, 'public', 'jit', 'x86'),
-      os.path.join(builder.options.hl2sdk_root, sdk['path'], 'game', 'server'),
-      os.path.join(builder.options.hl2sdk_root, sdk['path'], 'game', 'shared'),      
     ]
+
+    if sdk['name'] in ('episode1', 'darkm'):
+      binary.compiler.cxxincludes += [os.path.join(builder.options.hl2sdk_root, sdk['path'], 'game_shared')]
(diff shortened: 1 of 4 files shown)


Kenzzer:
Thank you for continuing PR. Looks like this is going to require some fixing on the hl2sdks before the CI can pass.

bottiger1:
It looks like the problem with the CI is because it is not finding variant_t.h from the SDK for some reason.

```
-I /home/runner/work/sourcemod/sourcemod/dependencies/hl2sdk-episode1/game/server 
-I /home/runner/work/sourcemod/sourcemod/dependencies/hl2sdk-episode1/game/shared 

-MD -MF inputnatives.d -c /home/runner/work/sourcemod/sourcemod/sourcemod/extensions/sdktools/inputnatives.cpp -o inputnatives.o
/home/runner/work/sourcemod/sourcemod/sourcemod/extensions/sdktools/variant-t.h:35:10: fatal error: 'variant_t.h' file not found
```

The include output looks correct except for the fact that game/shared doesn't exist in episode1, it is called game_shared, However game/server/variant_t.h does exist but the compiler says it still can't be found. 

I'm going to chalk this up to a CI or ambuild problem.

bottiger1:
I think I fixed episode1, but now it is complaining about the SDK. So I believe it's up to psychonic whether we ignore this error or edit all the SDKs. I think I've done all I can here.

```
/home/runner/work/sourcemod/sourcemod/dependencies/hl2sdk-csgo/game/server/variant_t.h:42:16: error: field 'fieldType' will be initialized after field 'iVal' [-Werror,-Wreorder-ctor]
        variant_t() : fieldType(FIELD_VOID), iVal(0) {}
                      ^~~~~~~~~~~~~~~~~~~~~  ~~~~~~~
                      iVal(0)                fieldType(FIELD_VOID)
```

extensions/sdktools/AMBuilder
```diff
@@ -49,9 +49,13 @@ for sdk_name in SM.sdks:
       os.path.join(builder.sourcePath, 'public', 'jit'),
       os.path.join(builder.sourcePath, 'public', 'jit', 'x86'),
       os.path.join(builder.options.hl2sdk_root, sdk['path'], 'game', 'server'),
-      os.path.join(builder.options.hl2sdk_root, sdk['path'], 'game', 'shared'),      
     ]
 
+    if sdk['name'] == 'episode1':
```
psychonic, line 54: `'darkm'` will need the same treatment

psychonic:
> So I believe it's up to @psychonic whether we ignore this error or edit all the SDKs. I think I've done all I can here. Strangely I don't get this error when I compile.
> 
> ```
> /home/runner/work/sourcemod/sourcemod/dependencies/hl2sdk-csgo/game/server/variant_t.h:42:16: error: field 'fieldType' will be initialized after field 'iVal' [-Werror,-Wreorder-ctor]
>         variant_t() : fieldType(FIELD_VOID), iVal(0) {}
>                       ^~~~~~~~~~~~~~~~~~~~~  ~~~~~~~
>                       iVal(0)                fieldType(FIELD_VOID)
> ```

I've now updated the SDK branches that had the bad initialization order.

psychonic (approved):

psychonic:
Thanks for tackling the rest of this

PR #2151, opened 2024-05-06
Add OnServerHibernationUpdate forward (closes #1483)

### LINKED ISSUE #1483: Add OnServerHibernate() and OnServerHibernate_Post() forwards
Hello,

Can some implement a forward to be notified when server goes on hibernate state please?

Cases for exemple:
- setup plugins/scripts just before hibernate
- make sql queries just before the sleep (because we need to use the main thread before it sleep, otherwise it can cause troubles)
- others actions that must be done when the server goes in this state 

2 forwards like OnServerHibernate() and OnServerHibernate_Post() will be very nice !

ojlanders:
Adds the OnServerHibernationUpdate forward. OnServerHibernationUpdate_Post() isn't possible to implement because there would be no place for it to be called from after SV_Frame stops executing.
Link to Discord discussion:
https://discord.com/channels/335290997317697536/335290997317697536/1236880595448233995

### DIFF
--- a/plugins/include/clients.inc
+++ b/plugins/include/clients.inc
@@ -216,6 +216,15 @@ forward Action OnClientPreAdminCheck(int client);
 forward void OnClientPostAdminFilter(int client);
 
 /**
+ * Called directly before the server hibernation state changes.
+ * This is your last chance to do anything in the plugin before
+ * hibernation occurs, as SV_Frame will no longer be called.
+ *
+ * @param state         The new hibernation state.
+ */
+forward void OnServerHibernationUpdate(bool state);
+
+/**
  * Called once a client is authorized and fully in-game, and 
  * after all post-connection authorizations have been performed.  
  *
--- a/plugins/include/clients.inc
+++ b/plugins/include/clients.inc
@@ -216,13 +216,16 @@ forward Action OnClientPreAdminCheck(int client);
 forward void OnClientPostAdminFilter(int client);
 
 /**
- * Called directly before the server hibernation state changes.
+ * Called directly before the server enters hibernation.
  * This is your last chance to do anything in the plugin before
  * hibernation occurs, as SV_Frame will no longer be called.
- *
- * @param state         The new hibernation state.
  */
-forward void OnServerHibernationUpdate(bool state);
+forward void OnServerEnterHibernation();
+
+/**
+ * Called directly before the server leaves hibernation.
+ */
+forward void OnServerExitHibernation();
 
 /**
  * Called once a client is authorized and fully in-game, and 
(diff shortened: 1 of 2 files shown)


ojlanders:
Test script:
```
#include <sourcemod>

public void OnServerHibernationUpdate(bool state)
{
    PrintToServer("Server hibernation state changed to: %s", state ? "true" : "false");
}
```
here's the output:
```
sm_cvar sv_hibernate_when_empty 0
[SM] Changed cvar "sv_hibernate_when_empty" to "0".
L 05/05/2024 - 23:11:48: [basecommands.smx] "Console<0><Console><Console>" changed cvar (cvar "sv_hibernate_when_empty") (value "0")
Server waking up from hibernation
Server hibernation state changed to: false
sm_cvar sv_hibernate_when_empty 1
[SM] Changed cvar "sv_hibernate_when_empty" to "1".
L 05/05/2024 - 23:11:49: [basecommands.smx] "Console<0><Console><Console>" changed cvar (cvar "sv_hibernate_when_empty") (value "1")
Server is hibernating
Server hibernation state changed to: true
```

plugins/include/clients.inc
```diff
@@ -215,6 +215,15 @@ forward Action OnClientPreAdminCheck(int client);
  */
 forward void OnClientPostAdminFilter(int client);
 
+/**
+ * Called directly before the server hibernation state changes.
+ * This is your last chance to do anything in the plugin before
+ * hibernation occurs, as SV_Frame will no longer be called.
+ *
+ * @param state         The new hibernation state.
```
Headline, line 223: new state? we should mention what true means here so it's clear we have begun hibernating, does false mean we have just awakened? Name it isHibernating

plugins/include/clients.inc
```diff
@@ -215,6 +215,15 @@ forward Action OnClientPreAdminCheck(int client);
...
 forward void OnClientPostAdminFilter(int client);
 
+/**
+ * Called directly before the server hibernation state changes.
+ * This is your last chance to do anything in the plugin before
+ * hibernation occurs, as SV_Frame will no longer be called.
+ *
+ * @param state         The new hibernation state.
+ */
+forward void OnServerHibernationUpdate(bool state);
```
Headline, line 225: OnHibernationChanged? Update implies to me we're getting this every frame, but this could just be me..

Headline (changes requested):
Thanks! - I just have some dev-user clarifications from me

plugins/include/clients.inc
ojlanders: added a bit to explain thsi

plugins/include/clients.inc
ojlanders: in general there's not a better way to do this: the function name "ServerHibernationUpdate" is well known, and the proper nomenclature used both in the engine and in everywhere i've seen it (like master server comms) is "hibernation state" unfortunately

Headline (approved):

psychonic (approved):
Thanks

PR #2152, opened 2024-05-09
Fix more sdktools crash on 64 bits

bottiger1:
Found some more pointer arithmetic in sdktools assuming 32bits.

### DIFF
--- a/extensions/sdktools/hooks.cpp
+++ b/extensions/sdktools/hooks.cpp
@@ -584,7 +584,7 @@ bool CHookManager::SendFile(const char *filename, unsigned int transferID)
 #if !defined CLIENTVOICE_HOOK_SUPPORT
 bool CHookManager::ProcessVoiceData(CLC_VoiceData *msg)
 {
-	IClient *pClient = (IClient *)((intptr_t)(META_IFACEPTR(IClient)) - 4);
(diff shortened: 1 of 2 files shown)


bottiger1:
> is there still - 4 to get IClient?

I did grep "+ 4" and "- 4" in the sdktools directory. I did not check any other directory.

asherkin:
While I understand it seems to be the same in practice, `sizeof(intptr_t)` feels wrong here - I would expect `sizeof(void *)` for this usage.

assyrianic:
> While I understand it seems to be the same in practice, `sizeof(intptr_t)` feels wrong here - I would expect `sizeof(void *)` for this usage.

Any particular reason why? Most of the code itself is already using `[u]intptr_t` anyway.

asherkin:
> Any particular reason why? Most of the code itself is already using `[u]intptr_t` anyway.

Which is obviously correct for the math, but the spec only appears to guarantee `sizeof(intptr_t) >= sizeof(void *)` - and what we want here is the size of a (specific) pointer, not the size of the integer type to represent any arbitrary pointer.

bottiger1:
We don't need to have a long discussion about this. If you want to change it to void* for a hypothetical architecture where the fastest integer type is somehow larger than the pointer type, it doesn't matter to me. I just want these crashes gone.

dvander:
static_assert() that your intention is true and move on, I say.

psychonic (approved):

PR #2160, opened 2024-05-19
Add gagged and muted reminders.

speedvoltage:
Players could be reminded that they are either gagged or muted, which could be useful if they are in either or both state for a prolonged amount of time.


Rushaway:
SourceMod using translations on all base plugins.
Your changes request to update translations for all supported languages. 

Your change also have file format issue.

PR #2161, opened 2024-05-19
Fix rare crash with logs

Kenzzer:
Issue mentioned to Asherkin or Psychonic.

Long story short, if a metamod plugin happens to be detouring some game functions that SM unfortunately calls while initialising, and decides to have the console print something. It will trigger `FileFileNatives::LogPrint`, however by that point `g_pLogHook` has not yet been initialised, it's still a nullptr. This of course ends up with a nullpointer crash.

Let's ~~add a member property that tracks whether or not we've been initialised.~~ nullcheck `g_pLogHook`

### DIFF
--- a/core/logic/smn_filesystem.cpp
+++ b/core/logic/smn_filesystem.cpp
@@ -266,7 +266,7 @@ class FileNatives :
 	public IPluginsListener
 {
 public:
-	FileNatives() : m_initialised(false)
+	FileNatives()
 	{
 	}
 	virtual void OnSourceModAllInitialized()
@@ -276,7 +276,6 @@ public:
 		g_ValveDirType = handlesys->CreateType("ValveDirectory", this, 0, NULL, NULL, g_pCoreIdent, NULL);
 		g_pLogHook = forwardsys->CreateForwardEx(NULL, ET_Hook, 1, NULL, Param_String);
 		pluginsys->AddPluginsListener(this);
-		m_initialised = true;
 	}
 	virtual void OnSourceModShutdown()
 	{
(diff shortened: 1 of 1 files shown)


core/logic/smn_filesystem.cpp
```diff
@@ -290,6 +291,10 @@ class FileNatives :
 	}
 	virtual void OnHandleDestroy(HandleType_t type, void *object)
 	{
+		if (!m_initialised) {
```
asherkin, line 294: I don't think this check should be here - this is a SM callback.

core/logic/smn_filesystem.cpp
```diff
@@ -309,19 +314,32 @@ class FileNatives :
...
+			return false;
+		}
+
 		cell_t result = 0;
 		g_pLogHook->PushString(msg);
 		g_pLogHook->Execute(&result);
 		return result >= Pl_Handled;
 	}
+	bool m_initialised;
```
asherkin, line 342: I think all the uses of this should just be checking `g_pLogHook`?

(Which is unfortunate that it's a global rather than part of the class, but not something to refactor to fix the bug.)

asherkin (changes requested):

core/logic/smn_filesystem.cpp
Kenzzer: I'll replace with a null check

psychonic (approved):

asherkin (approved):
LGTM - I think checking Add/RemoveLogHook was sane too, but is indeed unlikely to be necessary.

PR #2163, opened 2024-05-21
Move Safetyhook mess into a mirror repo

Headline:
This patch reimplements all of the modifications made to safetyhook to make it compatible with SourceMod, but does so in the actual source/inc files so we can possibly take upstream changes in the future and make the amalgamation scripts a part of our build process. 

I didn't attempt to call the amalgamation scripts in our builds yet - they're hard-pathed still (see https://github.com/alliedmodders/safetyhook/commit/0c3a4c0364272c33eaf7d7d829f2bd3910d5b7e0), but this gets us most of the way there.

Someone should test that this stuff still works before we take it in

EDIT: I haven't compiled this for linux, either, so builds may fail with safetyhook errors


Headline:
Assuming those linux builders pass we can take this in if it tests ok with the TODO of: 
1) Generate the amalgam stuff with ambuild
2) build this into a lib and link instead of including this in everything

GAMMACASE:
Providing some input on testing this fork, works great rn for me on x64 platforms on windows and linux.

rowedahelicon:
Tested this on Debian 12, server boots up fine and a detour I use for testing worked out without problems.

Issue #2172, opened 2024-06-01
ParseSMCFile crash on x64 when gamedata file uses windows line endings


bottiger1:
Only happens on 64 bits and  only happens if the file uses windows instead of linux line endings

[misc.txt](https://github.com/user-attachments/files/15523654/misc.txt)

```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007fc29f81edb2 in TextParsers::ParseStream_SMC (this=<optimized out>, stream=0x7cb6d40, srdr=0x7fc29f81fa50 <FileStreamReader(void*, char*, unsigned long, unsigned int*)>, smc=0xf655fb0, pStates=0x7fff110fe170) at /home/tf2/extensions/sdk/sourcemod/core/logic/TextParsers.cpp:457
457                                             if ((&parse_point[i] != in_buf) && c == '"' && parse_point[i-1] != '\\')
[Current thread is 1 (Thread 0x7fc2b4d1b740 (LWP 38868))]
#0  0x00007fc29f81edb2 in TextParsers::ParseStream_SMC (this=<optimized out>, stream=0x7cb6d40, srdr=0x7fc29f81fa50 <FileStreamReader(void*, char*, unsigned long, unsigned int*)>, smc=0xf655fb0, pStates=0x7fff110fe170) at /home/tf2/extensions/sdk/sourcemod/core/logic/TextParsers.cpp:457
#1  0x00007fc29f81e678 in TextParsers::ParseSMCFile (this=0x7fc29f977d38 <g_TextParser>, file=<optimized out>, smc_listener=0x7fff110fcf31, states=0x7fff110fe170, buffer=0x7fff11103550 "Unknown error", maxsize=256) at /home/tf2/extensions/sdk/sourcemod/core/logic/TextParsers.cpp:142
#2  0x00007fc29f83ac0f in CGameConfig::EnterFile (this=0xf655fb0, file=<optimized out>, error=0x7fff11103550 "Unknown error", maxlength=0) at /home/tf2/extensions/sdk/sourcemod/core/logic/GameConfigs.cpp:962
[...]

PR #2173, opened 2024-06-02
fix 64 bit crash on large gamedata files

bottiger1:
As explained here, there was a crash here caused by parsing a large gamedata file on 64 bits (it doesn't crash on 32 bits). It was fixed by changing the line endings from windows to linux.

https://github.com/alliedmodders/sourcemod/issues/2172

Increasing a buffer size fixes the issue so the line ending change isn't needed. I suspect there's probably some pointer math error that is avoided when the buffer is large enough.

### DIFF
--- a/core/logic/TextParsers.cpp
+++ b/core/logic/TextParsers.cpp
@@ -342,6 +342,7 @@ SMCError TextParsers::ParseStream_SMC(void *stream,
 	SMCResult res;
 	SMCStates states;
 	char c;
+	bool end_of_last_buffer_was_backslash = false;
 
 	StringInfo strings[3];
 	StringInfo emptystring;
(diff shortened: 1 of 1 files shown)


asherkin:
I think what you're seeing there happens when a read chunk starts exactly on a double quote. Increasing the buffer size would paper over it for your specific input file only.

I think the correct fix is to keep the previous character in a variable for the escape character check, rather than attempting to read backwards from the pointer. The other easy option would be overlapping the reads.

peace-maker:
It looks like unrelated updates to amtl and sourcepawn submodules found their way in here. Is this intentional?

bottiger1:
no it's not intentional. I had to shuffle a lot of stuff around since my pull requests needed to make 64 bit not crash aren't getting merged and I have to update amtl and sourcemod as well.

psychonic (approved):

psychonic:
Thanks

PR #2178, opened 2024-06-10
Fix SDKCall NULL_STRING

fdxx:
```sourcepawn
PrepSDKCall_AddParameter(SDKType_String, SDKPass_Pointer, VDECODE_FLAG_ALLOWNULL);
SDKCall(xx, NULL_STRING);
```

Expected to pass `nullptr`. But in fact `addr[0] == '\0'`, not `nullptr`.


Alienmario:
We have also ran into this.

The current workaround is to pass 0 with the parameter defined as
```PrepSDKCall_AddParameter(SDKType_PlainOldData, SDKPass_Plain);```

ampreeT:
Would love to see this merged. We have a use case for passing `nullptr` into the second argument of `CBaseEntity::SetNextThink` but we ended up passing the argument as POD set to `0` like @Alienmario said.

psychonic (approved):
Thanks!

PR #2190, opened 2024-07-11
Change FunCommands defaults

Tiagoquix:
I always thought it was funny that SourceMod had the ability to blow everyone up with the timebomb, so I think it would be cool if this was enabled by default.

I also changed the ticks down to 5 from 10, and increased burn duration to 30.


psychonic:
Thanks for the contribution, however, we're going to keep these values as-is for now.

PR #2192, opened 2024-07-15
Update libmaxminddb and MySQL 5.7

accelerator74:
Suggestion to use MySQL 5.7 as default for `checkout-deps.sh` on Linux. Will be useful after merging this PR: https://github.com/alliedmodders/sourcemod/pull/2189

In addition, an update to the libmaxminddb library.

### DIFF
--- a/extensions/geoip/data-pool.c
+++ b/extensions/geoip/data-pool.c
@@ -1,11 +1,13 @@
+#ifndef _POSIX_C_SOURCE
+#define _POSIX_C_SOURCE 200809L
+#endif
+
 #include "data-pool.h"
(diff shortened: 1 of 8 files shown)


Rushaway:
Can't directly move to MySQL 8 ? 
MySQL 5.7 EOL was 20 Sep 2023.

accelerator74:
> Can't directly move to MySQL 8 ? MySQL 5.7 EOL was 20 Sep 2023.

This seems to be more problematic. The MySQL 8.0 driver simply won't load.

psychonic (approved):
Thanks!

PR #2159, opened 2024-05-17
Attempt to add 64 bit support to sdktools

bottiger1:

PR #2196, opened 2024-07-27
Introduce MemoryPointer

Kenzzer:
_This whole PR is still an heavy draft, nothing has been field tested._

## Introduction

As of today no immediate solution to `PseudoAddress` has been explored. Only third-party solutions like [NotnHeavy/SM-Address64](https://github.com/NotnHeavy/SM-Address64) & [skial-com/port64](https://github.com/skial-com/port64) have been made, but no major adoption so far. And while those solutions can work, I believe the friction for plugin authors is quite high. Now having to worry about the server architecture, and the confusing concept of treating ptrs as array of int/any.

## Explanation

Given that there's no timeline as for when enum struct natives will ever be a thing. I propose instead to expose a new handle type that shall act as an opaque pointer towards abitrary memory regions. The underlying C++ class `IMemoryPointer` contained by the handle type is exposed as a public header. Allowing third-party extensions to wrap their pointers into it, no longer having to rely and depend on `cell_t` capacity to store the pointer's value on 32 bits.

`IMemoryPointer` shall become the defacto standard to pass abitrary pointers around for plugins & extensions.

### List of natives

`MemoryPointer.MemoryPointer(int size)`
Allocates an abitrary amount of memory.

`MemoryPointer.Load`
Similar to `LoadFromAddress` native.

`MemoryPointer.Store`
Similar to `StoreToAddress` native.

`MemoryPointer.Offset`
Offsets the base pointer, and creates a new MemoryPointer handle.

`MemoryPointer.LoadMemoryPointer`
Interprets the loaded data as an address and wraps it into a new MemoryPointer handle.

`MemoryPointer.StoreMemoryPointer`
Retrieves the pointer's value wrapped inside the MemoryPointer handle, and stores it.

`MemoryPointer.LoadEntityFromHandle`
Similar to `LoadEntityFromHandleAddress` native.

`MemoryPointer.StoreEntityToHandle`
Similar to `StoreEntityToHandleAddress` native.

`GameData.GetMemSigEx`
Similar to `GameData.GetMemSig` but wraps the pointer into a `MemoryPointer` handle.

`GameData.GetAddressEx`
Similar to `GameData.GetAddress` but wraps the pointer into a `MemoryPointer` handle.

`GetEntityMemoryPointer`
Similar to `GetEntityAddress` but wraps the entity pointer into a `MemoryPointer` handle.

`SDKCall_MemoryPointer`
New sdkcall type to allow member function calls from the ptr contained by a MemoryPointer handle.

`SDKType_MemoryPointer`
New sdkparam type, to retrieve the abitrary ptr contained inside a `MemoryPointer` handle. `INVALID_HANDLE` / `null` are allowed values to represent nullptr, if allow null flag was passed.

`PrepSDKCall_SetAddressFromMemoryPointer`
Similar to `PrepSDKCall_SetAddress` native.

Ultimately this PR aims to supersede #2159 & #2112

### DIFF
--- a/core/logic/MemoryPointer.cpp
+++ b/core/logic/MemoryPointer.cpp
@@ -50,6 +50,11 @@ MemoryPointer::~MemoryPointer()
 	}
 }
 
+void MemoryPointer::Delete()
+{
+	delete this;
+}
+
 void* MemoryPointer::Get()
 {
 	return m_ptr;
--- a/core/logic/MemoryPointer.cpp
+++ b/core/logic/MemoryPointer.cpp
@@ -1,0 +1,61 @@
+/**
+ * vim: set ts=4 sw=4 tw=99 noet :
+ * =============================================================================
+ * SourceMod
+ * Copyright (C) 2024 AlliedModders LLC.  All rights reserved.
+ * =============================================================================
+ *
+ * This program is free software; you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License, version 3.0, as published by the
+ * Free Software Foundation.
+ * 
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * As a special exception, AlliedModders LLC gives you permission to link the
+ * code of this program (as well as its derivative works) to "Half-Life 2," the
+ * "Source Engine," the "SourcePawn JIT," and any Game MODs that run on software
+ * by the Valve Corporation.  You must obey the GNU General Public License in
+ * all respects for all other code used.  Additionally, AlliedModders LLC grants
+ * this exception to all derivative works.  AlliedModders LLC defines further
+ * exceptions, found in LICENSE.txt (as of this writing, version JULY-31-2007),
+ * or <http://www.sourcemod.net/license.php>.
+ *
+ * Version: $Id$
+ */
+
+#include "MemoryPointer.h"
+#include <sourcehook.h>
+#include <sh_memory.h>
+#include <algorithm>
+
+MemoryPointer::MemoryPointer(cell_t size) : m_ptr(malloc(size)), m_owned(true), m_size(size)
+{
+}
+
+MemoryPointer::MemoryPointer(void* ptr, cell_t size) : m_ptr(ptr), m_owned(false), m_size(size)
+{
+}
+
+MemoryPointer::~MemoryPointer()
+{
+	if (m_owned)
+	{
+		free(m_ptr);
+	}
+}
+
+void* MemoryPointer::Get()
+{
+	return m_ptr;
+}
+
+cell_t MemoryPointer::GetSize()
+{
+	return m_size;
+}
--- a/core/logic/MemoryPointer.cpp
+++ b/core/logic/MemoryPointer.cpp
@@ -44,9 +44,10 @@ MemoryPointer::MemoryPointer(void* ptr, cell_t size) : m_ptr(ptr), m_owned(false
 
 MemoryPointer::~MemoryPointer()
 {
-	if (m_owned)
+	if (m_owned && m_ptr)
 	{
 		free(m_ptr);
+		m_ptr = nullptr;
 	}
 }
 
--- a/core/logic/smn_core.cpp
+++ b/core/logic/smn_core.cpp
@@ -88,11 +88,13 @@ public:
 		g_PlIter = handlesys->CreateType("PluginIterator", this, 0, NULL, NULL, g_pCoreIdent, NULL);
 		g_FrameIter = handlesys->CreateType("FrameIterator", this, 0, NULL, NULL, g_pCoreIdent, NULL);
 
-		HandleAccess security;
-		security.access[HandleAccess_Read] = 0;
-		security.access[HandleAccess_Delete] = HANDLE_RESTRICT_OWNER;
-		security.access[HandleAccess_Clone] = HANDLE_RESTRICT_IDENTITY | HANDLE_RESTRICT_OWNER;
-		g_MemoryPtr = handlesys->CreateType("MemoryPointer", this, 0, NULL, &security, g_pCoreIdent, NULL);
+		HandleAccess mp_hacc;
+		TypeAccess mp_tacc;
+		mp_hacc.access[HandleAccess_Read] = 0;
+		mp_hacc.access[HandleAccess_Delete] = HANDLE_RESTRICT_OWNER;
+		mp_hacc.access[HandleAccess_Clone] = HANDLE_RESTRICT_IDENTITY | HANDLE_RESTRICT_OWNER;
+		mp_tacc.access[HTypeAccess_Create] = true;
+		g_MemoryPtr = handlesys->CreateType("MemoryPointer", this, 0, &mp_tacc, &mp_hacc, g_pCoreIdent, NULL);
 
 		g_OnLogAction = forwardsys->CreateForward("OnLogAction", 
 			ET_Hook, 
--- a/core/logic/smn_core.cpp
+++ b/core/logic/smn_core.cpp
@@ -42,6 +42,7 @@
 #include <ITranslator.h>
 #include <DebugReporter.h>
 #include <FrameIterator.h>
+#include <MemoryPointer.h>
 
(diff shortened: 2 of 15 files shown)


NotnHeavy:
I think this is a decent work-around, considering the current `cell_t` limitations of SourcePawn and the numerous extensions that currently exist as a result of 64-bit address workarounds - including mine (which I hoped would've been the standard until an official SM native of some kind came around).

However, I feel like it's still missing out on some features, such as being able to wrap around existing addresses rather than only being able to use the handle to allocate a new block of memory. 

Perhaps natives for converting between pseudo addresses and `MemoryPointer` would also be ideal?

In the end as well this obviously creates a new issue of having to `delete` your `MemoryPointer` objects every time you use them as well, but I suppose active use of it wouldn't necessarily be encouraged.

KaelaSavia:
> In the end as well this obviously creates a new issue of having to `delete` your `MemoryPointer` objects every time you use them as well, but I suppose active use of it wouldn't necessarily be encouraged.

Perhaps it could be possible to have argument in MemoryPointer constructor that would decide whether pointer is freed on plugin end, or requires manual freeing. 

Also, would it be perhaps possible for "reasons" to have those two natives added?
- For converting two 32-bit ints (array) into MemoryPointer
- For converting MemoryPointer into array of two 32-bits


Other than that, current solution is pretty decent, people should be able to easily adapt. Although having wiki page just like we had for transistional syntax will make things even smoother, but thats not priority for now imo.

Kenzzer:
> Perhaps it could be possible to have argument in MemoryPointer constructor that would decide whether pointer is freed on plugin end, or requires manual freeing.

Purposefully leaking memory isn't something I'd advise. The standard in Sourcemod is that all plugin-instanced ressources are freed on unloading, this shouldn't be an exception. If a plugin or extension wants to keep a memorypointer around for longer, they're free to create their own and copy the data over.

plugins/include/sourcemod.inc
```diff
@@ -113,13 +113,28 @@ methodmap GameData < Handle
...
 	// @return              An address calculated on success, or 0 on failure.
+	#pragma deprecated Use GameData.GetAddressEx instead
 	public native Address GetAddress(const char[] name);
 
+	// Finds an address calculation in a GameConfig file,
+	// performs LoadFromAddress on it as appropriate, then returns the final address as a new MemoryPointer handle.
+	//
+	// @param name          Name of the property to find.
+	// @return              New MemoryPointer handle containing the address calculated on success, or null on failure.
+	public native Address GetAddressEx(const char[] name);
```
FortyTwoFortyTwo, line 124: Missed here and `GetMemSigEx` having return field as `MemoryPointer`?

psychonic (approved):

Headline (changes requested):
Most is nit picky or have already been pointed out

core/logic/MemoryPointer.cpp
```diff
@@ -0,0 +1,66 @@
...
+MemoryPointer::MemoryPointer(void* ptr, cell_t size) : m_ptr(ptr), m_owned(false), m_size(size)
+{
+}
+
+MemoryPointer::~MemoryPointer()
+{
+	if (m_owned)
+	{
+		free(m_ptr);
+	}
```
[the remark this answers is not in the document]
Kenzzer: This is in the dtor, so m_ptr is never gonna be used again but I can fix it.

core/logic/smn_core.cpp
```diff
@@ -100,7 +108,11 @@ class CoreNativeHelpers :
 	}
 	void OnHandleDestroy(HandleType_t type, void *object)
 	{
-		if (type == g_FrameIter)
+		if (type == g_MemoryPtr)
+		{
+			((IMemoryPointer *)object)->Delete();
```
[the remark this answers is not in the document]
Kenzzer: Given the nature of `IMemoryPointer` I cannot guarantee the method by which it was allocated in third party extensions. So calling `delete (IMemoryPointer*)object` would be ill advised, hence a delete method.

core/logic/smn_gameconfigs.cpp
```diff
@@ -194,6 +195,78 @@ static cell_t smn_GameConfGetMemSig(IPluginContext *pCtx, const cell_t *params)
...
+	sec.pIdentity = g_pCoreIdent;
+
+	if ((herr=handlesys->ReadHandle(hndl, g_GameConfigsType, &sec, (void **)&gc))
+		!= HandleError_None)
+	{
+		return pCtx->ThrowNativeError("Invalid game config handle %x (error %d)", hndl, herr);
+	}
+
+	char *key;
+	void* val;
```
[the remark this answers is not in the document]
Kenzzer: This is just a straight copy of the original code. I assume if it passed review the first time around that yes `GetAddress` would initialise `val`.

Headline (approved):

NotnHeavy:
There's issues with using `SDKType_MemoryPointer`. These are addressed in [a PR I made to Kenzzer's PR](https://github.com/Kenzzer/sourcemod/pull/1).

Kenzzer:
We should be good to merge this time.

dvander:
It'll be a while. It's a slog if there ever was a slog. I'd branch if you need to.On Aug 22, 2024 10:28 AM, NotnHeavy ***@***.***> wrote:
The thing with 1.12 at the moment is that dvander is currently refactoring the SourcePawn compiler completely, to the point that int64 hypothetically may be introduced. I'm not sure if it will make sense to start 1.13 now.

Kenzzer:
> The thing with 1.12 at the moment is that dvander is currently refactoring the SourcePawn compiler completely, to the point that `int64` hypothetically may be introduced. I'm not sure if it will make sense to start 1.13 now.

Branching makes the most sense, because we can make breaking changes on the dev branch. This was suggested because `int64` on SP is a possibility. And when it is, we can destroy `MemoryPointer` without really breaking bcompat, because its dev branch.

NotnHeavy:
Ah, I supposed I haven't really thought about that. That makes more sense then.

extensions/sdktools/vdecoder.cpp
```diff
@@ -276,6 +318,37 @@ DataStatus EncodeValveParam(IPluginContext *pContext,
...
+			auto ptr = *(void **)buffer;
+
+			if (ptr == nullptr)
+			{
+				*addr = 0;
+				return Data_Okay;
+			}
+
+			HandleError err = HandleError_None;
+			Handle_t hndl = handlesys->CreateHandle(g_MemPtrHandle, new ForeignMemoryPointer(ptr), pContext->GetIdentity(), myself->GetIdentity(), &err);
```
Kenzzer, line 342: Don't leak memory.

extensions/sdktools/vdecoder.cpp
```diff
@@ -579,6 +652,39 @@ DataStatus DecodeValveParam(IPluginContext *pContext,
...
+			Handle_t hndl = (Handle_t)*addr;
+
+			if (hndl == 0)
+			{
+				if (data->decflags & VDECODE_FLAG_ALLOWNULL)
+				{
+					*(void **)buffer = nullptr;
+					return Data_Okay;
+				}
+				pContext->ThrowNativeError("Null/Invalid Handle MemoryPointer isn't allowed");
```
Kenzzer, line 674: Remove "invalid", we can never branch here with an invalid handle. And it'd be weird to imply invalid handles can be used as "nullptr".

plugins/include/sourcemod.inc
```diff
@@ -752,8 +769,57 @@ native any LoadFromAddress(Address addr, NumberType size);
...
+	// Wraps the data loaded at the provided offset into a new MemoryPointer handle.
+	// @param offset          Offset in bytes from the start.
+	// @return                New Handle to a memory pointer.
+	public native MemoryPointer LoadMemoryPointer(int offset = 0);
+
+	// Creates a new MemoryPointer handle by offsetting the base pointer.
+	// @param size          Offset from base pointer in bytes.
+	// @return              New Handle to a memory pointer.
+	public native MemoryPointer Offset(int offset);
+};
```
Kenzzer, line 821: Missing equal native; i.e if two MemoryPointer handles hold the same pointer value.

Kenzzer (commented):
Leaving a self-review; before I forget again what I was missing on this PR.
I plan on fixing all of this later this week.

NotnHeavy:
[I decided to work on these for you](https://github.com/NotnHeavy/sourcemod/commit/7b1ae731f5c483cc268cc99e24e3d48740edf72e). My code should be fine, although I can't test on my laptop because both my laptop and internet are horrible in the place I am currently staying at.

Kenzzer:
Closing in favor of #2226 . The consensus internally seems to be that we avoid introducing a new handle type.

Will re-open if cited PR fails.

PR #2203, opened 2024-08-31
Update gamedata engine.ep2v.txt for 64 bits

Mentrillum:
GetMaxHealth in this file has the Windows 64 bit signature wrong using the Linux 32 bit/64 bit signature resulting in crashes.


psychonic (approved):
Thanks

PR #2213, opened 2024-10-29
Setup enum-struct address

Kenzzer:
This PR attempts to transform `Address` into an enum struct, but ultimately I believe the PR serves as a good showcase as to why we shouldn't...

## Primary goal

Avoid deprecation of 5-10 natives, whilst also keeping memory manipulation in plugin 'simple'.

## Backward Compatibility

We're still runtime backward compatible. However we're not source backward compatible, operations like `view_as<int>` or `++` will no longer work as they do not exist for enum structs. This is going to break compilation (to the best of our knowledge) of at least 200 plugins.

## What about PR #2196 ?

This is ultimately down to preferences. But after working on this, I believe it would be better to turn `Address` into an handle type, and overload the math operators. This might create problems regarding freeing `Address`, as plugin authors will get confused looking at older plugin source and not see in use the `delete` operator. But that outcome is still better, than wrecking havok on in the C++ api with an enum struct type that cannot be passed around extensions as easily as if it was an handle type. Something I could not forsee prior working on this.

### DIFF
--- a/core/logic/smn_core.cpp
+++ b/core/logic/smn_core.cpp
@@ -863,11 +863,22 @@ enum NumberType
 
 static cell_t LoadFromAddress(IPluginContext *pContext, const cell_t *params)
 {
+	void* addr = nullptr;
+	if (g_pSM->IsUsingPluginAddress(pContext))
+	{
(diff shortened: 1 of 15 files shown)


Kenzzer:
Closing in favor of #2226

PR #2216, opened 2024-11-03
Add new options to ArrayList FindString, FindValue

Alienmario:
Adds new optional parameters to ArrayList find natives.

- FindString: start index, reverse search, case sensitivity
- FindValue: start index, reverse search

The intention is to cover most remaining use cases of easily and efficiently iterating over enum struct ArrayLists.
Test is included in PR, tested on linux, passed.

_I am still learning, so excuse any newbie mistakes. I'm open to suggestions._


psychonic (approved):
Thanks

PR #2218, opened 2024-11-10
Add Cookie.IsSet()

Alienmario:
Adds a simple QoL method that checks whether a cookie has been set.

### DIFF
--- a/plugins/include/clientprefs.inc
+++ b/plugins/include/clientprefs.inc
@@ -157,7 +157,6 @@ methodmap Cookie < Handle {
 		this.Set(client, sValue);
 	}
 
-
 	// Retrieve the value of a Client preference cookie.
 	//
 	// @param client        Client index.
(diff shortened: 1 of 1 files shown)


asherkin:
This feels like a responsibility that should possibly belong to `Get`. A
non-breaking enhancement could be something like:

```
public native bool Get(int client, char[] buffer, int maxlen, const char[]
defaultValue = "");
```

Where the impl returns `true` if it was truly set and buffer has been
populated with the set value, and `false` if it has come from the passed in
default. This provides correct logic for plugins where empty might be a
desired actual value, and simplifies cases where plugins want to just get a
reasonable value and don't care where it has come from.

I don't think there's much value in a `Has` if it has to get the value
internally anyway.

Alienmario:
This is just a simple stock I've been using that helps keep client code clear of char arrays. I've assumed that "unset" is equal to "empty string", since plugins have had no way distinguish so far.

Anyway, since GetFloat/GetInt has been added, I've had no use for the original Get.
I think there's room for both changes. One could prompt new players to set their preferences, for example.

Headline:
I agree with asherkin's take on this. Closing this impl feel free to update with her suggestions.

Thank you!

PR #2222, opened 2024-11-18
Update Half-Life 2: Deathmatch Gamedata

caxanga334:
This PR updates the gamedata files of SDKTools and SDKHooks for Half-Life 2 Deathmatch.

Offsets obtained via Asherkin's vtable dumper using the Linux binary from the beta branch of the dedicated server.

I'm also removing the gamedata for mac since Sourcemod no longer supports it.

### DIFF
--- a/gamedata/sdkhooks.games/engine.ep2v.txt
+++ b/gamedata/sdkhooks.games/engine.ep2v.txt
@@ -203,147 +203,123 @@
 			}
 			"EndTouch"
 			{
-				"windows"	"100"
-				"linux"		"101"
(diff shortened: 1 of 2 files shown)


caxanga334:
Marking as ready for review now that HL2DM has been updated and the servers works now.

Did some basic tests on a Windows DS. Some commands from [Silver's dev cmds plugin](https://forums.alliedmods.net/showthread.php?p=1729015) that were crashing are now working as expected.

psychonic (approved):
Thanks!

PR #2226, opened 2024-11-28
Overhaul Address

Kenzzer:
Another Address PR!

The idea and implementation is primarily @dvander's, this PR just makes the bridge to SM.
In order to use the new virtual address :
```
#define VIRTUAL_ADDRESS
#include <sourcemod>
#include <sdktools>
```
If enabled, plugins need to ensure their SDKCalls are using the new `SDKType_VirtualAddress`, whether they're on 32bits or 64bits. Otherwise the plugin will most definitively crash the server.

Marking as draft for now until I can get a few server operators to try this out. Extra natives idea are welcomed !

### DIFF
--- a/extensions/sdktools/vcaller.cpp
+++ b/extensions/sdktools/vcaller.cpp
@@ -172,11 +172,10 @@ static cell_t PrepSDKCall_SetSignature(IPluginContext *pContext, const cell_t *p
 
 static cell_t PrepSDKCall_SetAddress(IPluginContext *pContext, const cell_t *params)
 {
-#ifdef PLATFORM_X86
-	s_call_addr = reinterpret_cast<void *>(params[1]);
-#else
-	s_call_addr = g_pSM->FromPseudoAddress(params[1]);
-#endif
+	s_call_addr = reinterpret_cast<void*>(params[1]);
+	if (pContext->GetRuntime()->FindPubvarByName("__Virtual_Address__", nullptr) != SP_ERROR_NONE) {
+		s_call_addr = g_pSM->FromPseudoAddress(params[1]);
+	}
 
 	return (s_call_addr != NULL) ? 1 : 0;
 }
--- a/extensions/sdktools/vdecoder.cpp
+++ b/extensions/sdktools/vdecoder.cpp
@@ -164,6 +164,21 @@ size_t ValveParamToBinParam(ValveType type,
 				return sizeof(float);
 			}
 		}
+	case Valve_VirtualAddress:
+	    {
+			info->flags = flags;
+			if (flags & PASSFLAG_ASPOINTER)
+			{
+				needs_extra = true;
+				info->type = PassType_Basic;
+				info->size = sizeof(void**);
+				return sizeof(void**) + sizeof(void*);
+			} else {
+				info->type = PassType_Basic;
+				info->size = sizeof(void*);
+				return sizeof(void*);
+			}
+		}
 	}
 
 	return 0;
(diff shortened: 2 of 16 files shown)


plugins/include/sdktools.inc
```diff
@@ -88,7 +88,10 @@ enum SDKType
 	SDKType_Float,          /**< Float (any) */
 	SDKType_Edict,          /**< edict_t (always as pointer) */
 	SDKType_String,         /**< NULL-terminated string (always as pointer) */
-	SDKType_Bool            /**< Boolean (any) */
+	SDKType_Bool,           /**< Boolean (any) */
+#if defined VIRTUAL_ADDRESS
```
rtldg, line 92: I think it's a good idea to have `SDKType_VirtualAddress` always in the enum (and not in an ifdef). It'll probably cause problems otherwise if anything is ever added to `SDKType`.

rtldg:
Can you also add this:
```c++
		} else if (vc->retinfo->vtype == Valve_VirtualAddress) {
			void *addr = (void *)vc->retbuf;
			return g_pSM->ToPseudoAddress(addr);
		} else {
```
to here
https://github.com/alliedmodders/sourcemod/blob/b71d3c0174b9fd008a7dd4d523c06b75b811f49d/extensions/sdktools/vcaller.cpp#L532-L545

This would allow things like calling `CreateInterface` with `SDKCall()` and actually having a usable result.

Kenzzer:
I think this is great, and indeed much better than the define.

Kenzzer:
A big thanks to @Malifox for taking the first step into trying this PR with a well established plugin, which helped me hunt down a ton of bugs.

psychonic (approved):

extensions/sdktools/vdecoder.cpp
```diff
@@ -276,6 +291,16 @@ DataStatus EncodeValveParam(IPluginContext *pContext,
 
 			*addr = *(bool *)buffer ? 1 : 0;
 
+			return Data_Okay;
+		}
+	case Valve_VirtualAddress:
+	    {
+			if (data->flags & PASSFLAG_ASPOINTER)
```
Malifox:
```suggestion
			cell_t *addr;
			pContext->LocalToPhysAddr(param, &addr);

			if (data->flags & PASSFLAG_ASPOINTER)
```

extensions/sdktools/vdecoder.cpp
```diff
@@ -276,6 +291,16 @@ DataStatus EncodeValveParam(IPluginContext *pContext,
...
 
+			return Data_Okay;
+		}
+	case Valve_VirtualAddress:
+	    {
+			if (data->flags & PASSFLAG_ASPOINTER)
+			{
+				buffer = *(void **)buffer;
+			}
+			param = g_pSM->ToPseudoAddress((void*)buffer);
```
Malifox:
```suggestion
			*addr = g_pSM->ToPseudoAddress(*(void **)buffer);
```

Malifox:
```suggestion
			cell_t *addr;
			pContext->LocalToPhysAddr(param, &addr);

			if (data->flags & PASSFLAG_ASPOINTER)
			{
				buffer = *(void **)buffer;
			}

			*addr = g_pSM->ToPseudoAddress(*(void **)buffer);
```

extensions/sdktools/vcaller.cpp
Malifox:
```suggestion
			|| vtype == Valve_Bool
			|| vtype == Valve_VirtualAddress)
```

extensions/sdktools/vdecoder.cpp
```diff
@@ -585,6 +610,18 @@ DataStatus DecodeValveParam(IPluginContext *pContext,
 			*(char **)buffer = addr;
 			return Data_Okay;
 		}
+	case Valve_VirtualAddress:
+	    {
+			void* addr = g_pSM->FromPseudoAddress(param);
```
Malifox:
```suggestion
			if (data->decflags & VDECODE_FLAG_BYREF)
			{
				cell_t *addr;
				pContext->LocalToPhysAddr(param, &addr);
				param = *addr;
			}

			void* addr = g_pSM->FromPseudoAddress(param);
```

Issue #2225, opened 2024-11-28
[Synergy] Gamedata update


yarik2720:
# Help us help you
  - I have checked that my issue [doesn't exist yet](https://github.com/alliedmodders/sourcemod/issues).
  - I have tried my absolute best to reduce the problem-space and have provided the absolute smallest test-case possible.
  - I can always reproduce the issue with the provided description below.

# Environment
  * Operating System version: Ubuntu Server 20.04 LTS ESR
  * Game/AppID (with version if applicable): 17520
  * Current SourceMod version: 1.12
  * Current SourceMod snapshot: 7163, 7169
  * Current Metamod: Source snapshot: 1.11.1156
  - [I have updated SourceMod to the [latest version](https://www.sourcemod.net/downloads.php) and it still happens.
  - I have updated SourceMod to the [latest snapshot](https://www.sourcemod.net/downloads.php?branch=dev) and it still happens.
  - I have updated SourceMM to the [latest snapshot](https://sourcemm.net/downloads.php?branch=dev) and it still happens.

# Description
Synergy got a public update. The latest Sourcemod causes server crash
`[60981687.425874] srcds_linux[1198924]: segfault at f5b89000 ip 00000000f7bc5235 sp 00000000ffaaa918 error 4 in libc-2.31.so[f7a8f000+15b000] `

With Sourcemod version 1.10.6503 server starts, but gamedata offsets are outdated causes plugins error with `[SDKTOOLS] Sigscan for FireOutput failed`

# Logs
* Please attach in separate files: game output, library logs, kernel logs, and any other supporting information.
* In case of a crash, please attach minidump or dump analyze output.
[crash_20241127195640_14.dmp](https://github.com/user-attachments/files/17942916/crash_20241127195640_14.dmp)

PR #2228, opened 2024-12-07
Update SDKTools & SDKHooks Gamedata for Synergy

Balimbanana:
Update Synergy offsets and signatures for latest Synergy. There are still updates happening, so these may become incorrect in the coming weeks.

Resolves: https://github.com/alliedmodders/sourcemod/issues/2225


psychonic (approved):
Thanks!

Issue #2140, opened 2024-04-22
TF2_OnPlayerTeleport isn't working anymore on Linux


StShAzzy:
# Help us help you
  - [X] I have checked that my issue [doesn't exist yet](<https://github.com/alliedmodders/sourcemod/issues>).
  - [X] I have tried my absolute best to reduce the problem-space and have provided the absolute smallest test-case possible.
  - [X] I can always reproduce the issue with the provided description below.

# Environment
  * Operating System version: Linux
  * Game/AppID (with version if applicable): 440
  * Current SourceMod version: 1.12.0.7121
  * Current SourceMod snapshot: 7121:7d78c141
  * Current Metamod: Source snapshot: 1.12.0-dev+1192V
  - [ ] I have updated SourceMod to the [latest version](<https://www.sourcemod.net/downloads.php>) and it still happens.
  - [ ] I have updated SourceMod to the [latest snapshot](<https://www.sourcemod.net/downloads.php?branch=dev>) and it still happens.
  - [ ] I have updated SourceMM to the [latest snapshot](<https://sourcemm.net/downloads.php?branch=dev>) and it still happens.

# Description
TF2_OnPlayerTeleport isn't fired anymore since the update of 18/04/2024
both on x86 and x64

# Problematic Code (or Steps to Reproduce)
```cpp
#include <tf2_stocks>

public Action TF2_OnPlayerTeleport(int client, int teleporter, bool& result) {
	PrintToChatAll("[DEBUG]: TF2_OnPlayerTeleport Fired! Player Should Teleport Through!");
	result = true;
	return Plugin_Changed;
}
```

StShAzzy:
fixed with https://github.com/alliedmodders/sourcemod/pull/2234

PR #2214, opened 2024-11-02
- Fixes for OnPlayerTeleport hook for linux and linux64 systems

MrD4rk5oul:
- Changed signature call of `CanPlayerTeleport` for linux/linux64 to current used name
  - Applied fix at detour class `CanPlayerBeTeleported` to use standard parameter proceedings for linux

Fixes issue #2140 

Fix tested on:
  - Linux Mint 21.3, built in Ubuntu 20.04, working properly
  - Windows 10, working properly

PR #2234, opened 2024-12-26
Fixes for OnPlayerTeleport hook for linux and linux64 systems

MrD4rk5oul:
Merge #2214 failed due to pointer misalignment which was fixed in safetyhook and a [revision commit](https://github.com/alliedmodders/sourcemod/commit/4004471d8dde9ae3523cc61cb98b954dee545a41) was made, making the process of hooking work normally again.

Tests were made on Linux with bare MM+SM installations, MM+SM+many dummy plugins using different hooks and both mock and production servers , and no recurring crashes were reported.

Tests were also made on Windows, both with MM+SM installations and MM+SM+dummy hook plugins.

Pull request fixes issue #2140.

### DIFF
--- a/extensions/tf2/teleporter.cpp
+++ b/extensions/tf2/teleporter.cpp
@@ -37,7 +37,18 @@ IForward *g_teleportForward = NULL;
 
 class CTFPlayer;
 
+#if defined(__linux__) && defined(__i386__)
+class CanPlayerBeTeleportedClass
+{
+public:
+	__attribute__((regparm(2))) bool CanPlayerBeTeleported(CTFPlayer * pPlayer); 
+	static __attribute__((regparm(2))) bool (CanPlayerBeTeleportedClass::* CanPlayerBeTeleported_Actual)(CTFPlayer *);
+};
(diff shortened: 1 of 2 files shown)


psychonic (changes requested):
Thanks for following up on this! Can we change the gamedata update to instead use a new name for this, leaving the original one as-is? Otherwise, if we push the updated gamedata to the gamedata updater without users necessarily having manually updated SourceMod, there will be a mismatch between what function variant the code is targeting versus what the gamedata is targeting.

psychonic:
> Thanks for following up on this! Can we change the gamedata update to instead use a new name for this, leaving the original one as-is? Otherwise, if we push the updated gamedata to the gamedata updater without users necessarily having manually updated SourceMod, there will be a mismatch between what function variant the code is targeting versus what the gamedata is targeting.

Should also add a comment to note why it was done, noting that the older one can be nuked later.

MrD4rk5oul:
Added a new registry for the teleportation checking and marked the old one as Obsolete.

psychonic (approved):
That's perfect. Thanks!

PR #2238, opened 2024-12-30
Add support for radio menu and HUD message for Dystopia

bauxiteDYS:
Radio menus such as those created with `sm_vote` don't show up in the game, but the game seems to support them once the keys in `common.games.txt` are updated. Similarly HUD messages seem to work fine in maps with the `game_text` entity.

radio_menu


psychonic (approved):
Sweet! The omission of these likely stems from when the mod was still on the 2006 SDK.

PR #2239, opened 2024-12-30
Allow sending panels to SourceTV client via settings

### LINKED ISSUE #1563: Sourcemod menus do not show for Source TV (L4D2)

A1mDev:
Code tested in game Left4Dead2, menu successfully sent to SourceTV client.

Close #1563 
Related PR #1775

### DIFF
--- a/core/MenuStyle_Base.cpp
+++ b/core/MenuStyle_Base.cpp
@@ -419,7 +419,19 @@ bool BaseMenuStyle::DoClientMenu(int client, IMenuPanel *menu, IMenuHandler *mh,
 		time);
 #endif
 	CPlayer *pPlayer = g_Players.GetPlayerByIndex(client);
-	if (!pPlayer || pPlayer->IsFakeClient() || !pPlayer->IsInGame())
+	if (!pPlayer || !pPlayer->IsInGame())
+	{
+		return false;
+	}
+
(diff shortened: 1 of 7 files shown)


asherkin:
This seems like too much faff around it, based on the discussion in the previous PR just removing the `pPlayer->IsFakeClient()` check seems good enough? The client index is passed in, I don't think we need an additional param of "yes, I really meant to pass _that_ client index in".

A1mDev:
I think it would be nice to keep the default behavior?

A1mDev:
It's just that the old PR was never accepted and I thought it wasn't worth changing the default behavior, of course I can remove the extra code if you say so.

A1mDev:
Done, I made one check for SourceTV, I don't know if it is needed for other bots, let someone else make a PR if it is needed for ReplayClient.

asherkin:
Presumably nothing breaks if sent to other fake clients? We should err on trusting the plugin that it is doing what it wants.

A1mDev:
Done

asherkin (approved):
LGTM, @psychonic ?

psychonic (approved):

PR #2256, opened 2025-02-06
Revert modtime==0 change due to new config issues

Headline:
Spookmaster reported new issues on latest SM, alleged reparse invaliding a string. Just a quick revert while this gets handled


Fyren:
#2257 was merged instead.

PR #2257, opened 2025-02-07
Fix storing garbage strings

### LINKED ISSUE #2193: SetVariantString uses string by reference
# Help us help you

SourceMod Version: 1.12.0.7031

# Description
SetVariantString will keep a reference to Sourcepawn string variable, instead of copying its value.
Not sure if this is intentional, if it is, should probably be specified in the docs.

The following results in both the **input** and **variant** being **"IgniteLifetime"**
The expected behavior is **variant** being **"2"**.
  ```
char buffer[128];
buffer = "2";
SetVariantString(buffer);
buffer = "IgniteLifetime";
AcceptEntityInput(entity, buffer);
  ```

Kenzzer:
As of #2253 plugins/extensions might reload sdktools gamedata, which turns all the strings globally stored into garbage.
For example :
https://github.com/alliedmodders/sourcemod/blob/f6737a4760e4bd6226f327f0e08ec5d8c7f6a4e4/extensions/sdktools/gamerulesnatives.cpp#L40

I've scoured the codebase for the usage of `GetKeyValue` and implemented a string copy of the returned value if it happens to be stored into a global variable. Verdict is that only sdktools and cstrike extension seem affected, hopefully that's all the instances of this issue.

Also closes #2193

### DIFF
--- a/extensions/sdktools/gamerulesnatives.cpp
+++ b/extensions/sdktools/gamerulesnatives.cpp
@@ -33,11 +33,15 @@
 #include "gamerulesnatives.h"
 #include "vglobals.h"
 
-const char *g_szGameRulesProxy;
+char g_szGameRulesProxy[64];
 
 void GameRulesNativesInit()
 {
-	g_szGameRulesProxy = g_pGameConf->GetKeyValue("GameRulesProxy");
+	auto key = g_pGameConf->GetKeyValue("GameRulesProxy");
+	if (key)
+	{
+		strcpy(g_szGameRulesProxy, key);
+	}
 }
 
 static CBaseEntity *FindEntityByNetClass(int start, const char *classname)
@@ -202,7 +206,7 @@ static cell_t GameRules_GetProp(IPluginContext *pContext, const cell_t *params)
 
 	void *pGameRules = GameRules();
 
-	if (!pGameRules || !g_szGameRulesProxy || !strcmp(g_szGameRulesProxy, ""))
+	if (!pGameRules || !g_szGameRulesProxy[0])
 		return pContext->ThrowNativeError("Gamerules lookup failed.");
 
 	pContext->LocalToString(params[1], &prop);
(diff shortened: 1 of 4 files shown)


Headline:
Has this been tested to fix the issue? Not sure if we have reliable repro

Kenzzer:
> Has this been tested to fix the issue? Not sure if we have reliable repro

Not yet no, but I can test it and maybe get a few server operators to try out too. A reliable way of testing this, is to either load my extension [CBaseNPC](https://github.com/TF2-DMB/CBaseNPC) or make a plugin that just loads sdktools gamedata. Either will trigger a reparse and cause the issue to appear, but with this PR's fix it should not.

extensions/sdktools/gamerulesnatives.cpp
```diff
@@ -33,11 +33,15 @@
...
-const char *g_szGameRulesProxy;
+char g_szGameRulesProxy[64];
 
 void GameRulesNativesInit()
 {
-	g_szGameRulesProxy = g_pGameConf->GetKeyValue("GameRulesProxy");
+	auto key = g_pGameConf->GetKeyValue("GameRulesProxy");
+	if (key)
+	{
+		strcpy(g_szGameRulesProxy, key);
```
Fyren, line 43: Use `strncpy` here and for `m_iScore`.

extensions/sdktools/gamerulesnatives.cpp
```diff
@@ -33,11 +33,15 @@
 #include "gamerulesnatives.h"
 #include "vglobals.h"
 
-const char *g_szGameRulesProxy;
+char g_szGameRulesProxy[64];
```
Fyren, line 36: Make sure this doesn't stay un-init'd if the key doesn't exist in the conf.

psychonic (approved):

Kenzzer:
Given @psychonic approved, and internal testing hasn't reported any issues. I'm merging this, I'm currently producing unofficial builds of sourcemod for that new source engine update, and it's important that fix is included otherwise I'm getting false reports.

PR #2268, opened 2025-02-19
Tentative compilation fix for css + tf2

Kenzzer:
Same as alliedmodders/hl2sdk-manifests#9

Special note, the define `PLATFORM_X86` is now used by valve to refer to both x86 and x86_64. It holds either 32 or 64 value.
So instead of potentially conflicting with whatever valve decides to do with its preprocessor variable, let's just make sure our variable well distinct.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     name: ${{ matrix.os_short }}-${{ matrix.compiler_cc }}
     env:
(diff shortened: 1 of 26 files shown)


peace-maker:
We could use the `KE_ARCH` defines defined by amtl.
https://github.com/alliedmodders/amtl/blob/2d3b1a3378a3728637f26660c9ffc2df3189cf62/amtl/am-platform.h#L69

Kenzzer:
Ah yes! That makes perfect sense

psychonic (approved):

PR #2269, opened 2025-02-19
Update TF2 & CSS gamedata

Kenzzer:
Update the gamedata for the games CSS and TF2

TF2 is 100% updated
CSS is 90% updated, cstrike extension gamedata is still left to do. But being unfamiliar with css and lack of documentation in the gamedata leaves me guessing as to where I can find each offset.


xen-000 (commented):
This should be "cstrike"

gamedata/sdkhooks.games/engine.ep2v.txt
```diff
@@ -474,4 +474,186 @@
 			}
 		}
 	}
+
+	"css"
```
xen-000, line 478: This should be "cstrike"

gilb0106:
thank you for your hard work

psychonic (approved):

gilb0106:
this is excellent, it fixed revive, give weapons and bot control and most player entities, i will investigate source code of things that move players between teams as this is still non functional in css

naydef:
The new gamedata for `RemoveDisguise` for 32-bit Windows TF2 server is wrong and the new signature actually is for `CTFWeaponBase::ApplyOnHitAttributes`, thus the server is crashing.

The old signature works fine (at least for 32-bit Windows server).

Issue #2148, opened 2024-05-04
admin-sql-threaded: SQL Query has improper syntax and is truncated when called in FetchUser()


ArcalaAlien:
# Help us help you
  - [X] I have checked that my issue [doesn't exist yet](https://github.com/alliedmodders/sourcemod/issues).
  - [X] I have tried my absolute best to reduce the problem-space and have provided the absolute smallest test-case possible.
  - [X] I can always reproduce the issue with the provided description below.

# Environment
  * Operating System version: ubuntu 20.04
  * Game/AppID (with version if applicable):  
  * Current SourceMod version: 1.12
  * Current SourceMod snapshot: 7125
  * Current Metamod: Source snapshot: 1.12.0-dev+1192
  - [ ] I have updated SourceMod to the [latest version](https://www.sourcemod.net/downloads.php) and it still happens.
  - [X] I have updated SourceMod to the [latest snapshot](https://www.sourcemod.net/downloads.php?branch=dev) and it still happens.
  - [X] I have updated SourceMM to the [latest snapshot](https://sourcemm.net/downloads.php?branch=dev) and it still happens.

# Description
When using admin-sql-threaded.smx, the error in the picture is thrown when an admin joins a server.

The SQL Query in the SELECT statement in lines 510-529 has incorrect syntax. COUNT() is an aggregate function, and so all values that aren't in an aggregate function must be in the GROUP BY clause. However, adding all the additional columns needed causes the SQL query to get truncated so the length of the array should be increased.

# Problematic Code (or Steps to Reproduce)
Lines 510-529

```
// Increase the char buffer's size or the query gets truncated with the new GROUP BY clause
char query[512];
int len = 0;
	
len += Format(query[len], sizeof(query)-len, "SELECT a.id, a.authtype, a.identity, a.password, a.flags, a.name, COUNT(ag.group_id), immunity");
len += Format(query[len], sizeof(query)-len, " FROM sm_admins a LEFT JOIN sm_admins_groups ag ON a.id = ag.admin_id WHERE ");
len += Format(query[len], sizeof(query)-len, " (a.authtype = 'ip' AND a.identity = '%s')", ipaddr);
len += Format(query[len], sizeof(query)-len, " OR (a.authtype = 'name' AND a.identity = '%s')", safe_name);
if (steamid2[0] != '\0' && steamid3[0] != '\0' && steamid64[0] != '\0')
{
	strcopy(steamid2alt, sizeof(steamid2alt), steamid2);
[...]

PR #2270, opened 2025-02-20
Fixed missing columns in GROUP BY statement in admin-sql-threaded

ArcalaAlien:
Related to [#2148](https://github.com/alliedmodders/sourcemod/issues/2148):

* Added columns from the SELECT clause that were missing from the GROUP BY clause on line 529 that was causing the query to fail.
* Increased query buffer size on line 513 from 512 to 544 to accommodate the extra characters needed.


psychonic (approved):
Thanks!

PR #2274, opened 2025-02-20
Update Day of Defeat: Source SDKHooks and SDKTools Gamedata

caxanga334:
Offsets taken from Asherkin's vtable dumper using the Linux binary from the dedicated server.

Most offsets were tested on Windows x86_64 and are functional.

One issue: `CommitSuicide` isn't working. I can't figure out why.


caxanga334:
The offset for `CommitSuicide` is correct and functional.

The issues is apparently `CommitSuicide` doesn't work if you're inside spawn.

psychonic:
Thanks!

PR #2277, opened 2025-02-21
RemoveDisguise GameData fix

M60TM:
https://crash.limetech.org/gvjiwehomlvu
`RemoveDisguise` still have crash issue after #2273. I have already tested this and confirmed that there are no issues.


psychonic (approved):
Thanks!

PR #2288, opened 2025-02-28
Fix a Windows64 crash when unloading a plugin that DHooked something

rtldg:
(I was using [this](https://github.com/xen-000/SuppressViewpunch/pull/2) to test the crash)


rtldg:
@Kenzzer 🕵️‍♀️ i summon thee

Kenzzer (approved):
Thanks for getting around to do this. I was going to commit that change myself after a talk on discord with another user, but I forgot... so my apologies for that and thanks again for taking care of it.

PR #2292, opened 2025-03-02
DHooks x64 Linux enabler

rtldg:
Companion to https://github.com/alliedmodders/metamod-source/pull/212

For `x86_64SystemVDefault.cpp`, I copied `x86_64MicrosoftDefault.cpp` and then replaced things so you can see the diff in the second commit ([link](https://github.com/alliedmodders/sourcemod/commit/dfa865906248d0912174e30173c04bd2788e0177)) on this branch.

Currently includes the changes from https://github.com/alliedmodders/sourcemod/pull/2290

## TODO:
- Handle a TODO note in `x86_64SystemVDefault.cpp`
- More testing...
- Even more testing...

### DIFF
--- a/extensions/dhooks/DynamicHooks/conventions/x86_64SystemVDefault.cpp
+++ b/extensions/dhooks/DynamicHooks/conventions/x86_64SystemVDefault.cpp
@@ -23,3 +23,259 @@
 *
 * 3. This notice may not be removed or altered from any source distribution.
 */
+
+// ============================================================================
+// >> INCLUDES
+// ============================================================================
+#include "x86_64MicrosoftDefault.h"
+#include <smsdk_ext.h>
+
+// ============================================================================
+// >> CLASSES
+// ============================================================================
+x86_64MicrosoftDefault::x86_64MicrosoftDefault(std::vector<DataTypeSized_t> &vecArgTypes, DataTypeSized_t returnType, int iAlignment) :
+	ICallingConvention(vecArgTypes, returnType, iAlignment),
+	m_stackArgs(0)
+{
+	const Register_t params_reg[] = { RCX, RDX, R8, R9 };
+	const Register_t params_floatreg[] = { XMM0, XMM1, XMM2, XMM3 };
+	//const char* regNames[] = { "RCX or XMM0", "RDX or XMM1", "R8 or XMM2", "R9 or XMM3"};
+	const std::uint8_t num_reg = sizeof(params_reg) / sizeof(Register_t);
+
+	bool used_reg[] = { false, false, false, false };
+
+	// Figure out if any register has been used
+	auto retreg = m_returnType.custom_register;
+	used_reg[0] = (retreg == RCX || retreg == XMM0);
+	used_reg[1] = (retreg == RDX || retreg == XMM1);
+	used_reg[2] = (retreg == R8 || retreg == XMM2);
+	used_reg[3] = (retreg == R9 || retreg == XMM3);
+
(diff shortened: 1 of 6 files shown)


rtldg:
Seems like the tests are going to be failing until https://github.com/alliedmodders/metamod-source/pull/212 is merged :(

extensions/dhooks/DynamicHooks/conventions/x86_64SystemVDefault.cpp
```diff
@@ -22,4 +22,288 @@
...
+	return registers->m_rax->m_pAddress;
+}
+
+void x86_64SystemVDefault::ReturnPtrChanged(CRegisters* pRegisters, void* pReturnPtr)
+{
+}
+
+void x86_64SystemVDefault::SaveReturnValue(CRegisters* registers)
+{
+	// It doesn't matter what the return value is, it will always be fitting on 8 bytes (or less)
```
Kenzzer, line 270: That is definitively not true for SystemV, return value can get exploded on two registers.

extensions/dhooks/DynamicHooks/conventions/x86_64SystemVDefault.cpp
```diff
@@ -22,4 +22,288 @@
...
+	std::unique_ptr<uint8_t[]> savedReturn = std::make_unique<uint8_t[]>(8);
+	memcpy(savedReturn.get(), GetReturnPtr(registers), 8);
+	m_pSavedReturnBuffers.push_back(std::move(savedReturn));
+}
+
+void x86_64SystemVDefault::RestoreReturnValue(CRegisters* registers)
+{
+	// Like stated in SaveReturnValue, it will always fit within 8 bytes
+	// the actual underlining type does not matter
```
Kenzzer, line 279: Ditto, there's more to investigate

Kenzzer (commented):
Not much to say at the moment, just that any hooks with 'vector' is going to make all of this explode.

extensions/dhooks/DynamicHooks/conventions/x86_64SystemVDefault.cpp
```diff
@@ -22,4 +22,288 @@
...
+		return reg->m_pAddress;
+	}
+
+	// Return address
+	size_t offset = 8;
+	for (unsigned int i = 0; i < index; i++)
+	{
+		if (m_vecArgTypes[i].custom_register == None)
+		{
+			// No matter what, the stack is allocated in slices of 8 bytes
```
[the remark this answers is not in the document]
rtldg: Turns out to not be the case because (non-pointer trivial) objects can be passed inline on the stack which is really unfortunate.

extensions/dhooks/DynamicHooks/conventions/x86_64SystemVDefault.cpp
rtldg: > That is definitively not true for SystemV, return value can get exploded on two registers.

This should only matter in DHooks for `ReturnType_Vector` being split into XMM0/XMM1, right?

Kenzzer: Given our present situation, yes it's not critical to handle all use-case. But it does need to be handled for vector return...

rtldg:
This was going to be glue & bandaids in the short-term but I burned out. Kenzzer's work on nuking sourcehook/related in mms & sm to allow something to completely replace the systems (khook) is a better way forward.

PR #2298, opened 2025-03-06
Update engine.ep2v.txt - DOD:S Touch offset

ambaca:
According this forum post
https://forums.alliedmods.net/showpost.php?p=2835080&postcount=2

...and I encounter in same problem while doing plugin for DOD:S game, SDKHooks "Touch" did not work.
And seems hl2mp and cstrike have very same offsets in "EndTouch" "StartTouch" so this should be correct for "Touch".

- I have tested on Windows, srcds.exe and srcds_win64.exe


psychonic (approved):
Thanks

PR #2300, opened 2025-03-09
Update HL2DM x86-64 SDKTools Gamedata

caxanga334:
Adds missing offsets for x86-64 HL2DM's SDKTools gamedata.


psychonic (approved):
Thanks!

PR #2310, opened 2025-04-19
Update SDKTools for Synergy

Balimbanana:
Update IgniteEntity function for an added parameter for Synergy.
Updated Linux SDKTools gamedata signatures to use the named format. Note that I don't have the latest Windows signatures.


psychonic (approved):
Thanks

PR #2311, opened 2025-04-25
Fix NULL Crash on ReferenceToIndex

caxanga334:
# Description

This PR fixes a crash caused by passing a bad reference value to `CHalfLife2::ReferenceToIndex`.

Example crash: https://crash.limetech.org/xyv66kz6dkvd

# Code to Reproduce

```
#include <sourcemod>

#pragma newdecls required
#pragma semicolon 1

const int ENTREF_MASK = (1 << 31);

public void OnMapStart()
{
	int ref = 9000 | ENTREF_MASK;
	int index = EntRefToEntIndex(ref);
	PrintToServer("%i", index);
}
```


psychonic (approved):
Thanks

Issue #2110, opened 2024-02-02
format functions


0x00000ED:
# Description
Why all functions that have a format argument dont indicate that the number is infinite?

# Problematic Code (or Steps to Reproduce)
  ```cpp
public void OnPluginStart()
{
    float nan = 0.0/0.0;
    float inf = 1.0/0.0;

    PrintToServer("PrintToServer %f %f %f", nan, inf, -inf);

    char buff[128];
    Format(buff, sizeof(buff), "Format %f %f %f", nan, inf, -inf);
    PrintToServer("%s", buff);

    char fts1[10];
    FloatToString(nan, fts1, sizeof(fts1));
    char fts2[10];
    FloatToString(inf, fts2, sizeof(fts2));
    char fts3[10];
    FloatToString(-inf, fts3, sizeof(fts3));

    PrintToServer("FloatToString %s %s %s", fts1, fts2, fts3);
}
  ```

# Logs
// PrintToServer NaN 0.000000 -0.000000
// Format NaN 0.000000 -0.000000
// FloatToString -nan(ind) inf -inf

F1F88:
I checked the source code of SourceMod:

- [FloatToString](https://github.com/alliedmodders/sourcemod/blob/7e784d5453a5adb751c8f6fc7bff34b9e8827042/core/logic/smn_string.cpp#L206) uses [_vsnprintf](https://github.com/alliedmodders/amtl/blob/ac9ed94208f06bfe1a2132d54fef4bd4ddaf3395/amtl/am-string.h#L67) or [vsnprintf](https://github.com/alliedmodders/amtl/blob/ac9ed94208f06bfe1a2132d54fef4bd4ddaf3395/amtl/am-string.h#L69) to format float values

- [PrintToServer](https://github.com/alliedmodders/sourcemod/blob/7e784d5453a5adb751c8f6fc7bff34b9e8827042/core/logic/smn_console.cpp#L96) and [Format](https://github.com/alliedmodders/sourcemod/blob/7e784d5453a5adb751c8f6fc7bff34b9e8827042/core/logic/smn_string.cpp#L224) uses [atcprintf](https://github.com/alliedmodders/sourcemod/blob/7e784d5453a5adb751c8f6fc7bff34b9e8827042/core/logic/sprintf.cpp#L1016) to format float values

[...]

PR #2324, opened 2025-05-30
Supports "%f" formatting value as inf float

F1F88:
Related to: #2110 

Added special handling for float values ​​that are inf (formatted as "Inf" string instead of "0.0")

### DIFF
--- a/core/logic/sprintf.cpp
+++ b/core/logic/sprintf.cpp
@@ -254,6 +254,12 @@ void AddFloat(char **buf_p, size_t &maxlen, double fval, int width, int prec, in
 		return;
 	}
 
+	if (std::isinf(fval))
+	{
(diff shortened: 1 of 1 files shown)


core/logic/sprintf.cpp
```diff
@@ -254,6 +254,12 @@ void AddFloat(char **buf_p, size_t &maxlen, double fval, int width, int prec, in
 		return;
 	}
 
+	if (std::isinf(fval))
```
Headline, line 257: ke::IsNaN

F1F88: Sorry, I'm a bit confused.
Based on the function name, I assumed that ke::IsInfinite() does the same thing as std::isinf(), but when I tested it with the test case from #2110, the output was different.
And testing **std::isinf(inf)** returns **1**
while **ke::IsInfinite(inf)** returns **0**

psychonic (approved):

Issue #2328, opened 2025-05-31
[Bug]: When using "%0[width]d" to format a negative integer, the minus sign is placed incorrectly


F1F88:
### Prerequisites

- [x] I have checked that my issue doesn't exist yet in the [issue tracker](https://github.com/alliedmodders/sourcemod/issues)

### Operating System and Version

Windows 11 24H2

### Game / AppID and Version

NMRIH version 1.14.2

### SourceMod Version

1.12.0-git7197

### Metamod:Source Version

1.12.0-git1219

### Version Verification

- [x] I have updated SourceMod to the [latest version](https://www.sourcemod.net/downloads.php) and the issue persists
- [x] I have updated SourceMod to the [latest snapshot](https://www.sourcemod.net/downloads.php?branch=dev) and the issue persists
- [x] I have updated Metamod:Source to the [latest snapshot](https://sourcemm.net/downloads.php?branch=dev) and the issue persists

### Updated SourceMod Version

1.13.0-git7241

### Updated Metamod:Source Version

2.0.0-1350

### Description

When using **"%0[width]d"** to format a **negative integer**, the **minus sign** is formatted before the integer value but after the filler '0'.

### Steps to Reproduce

```sourcepawn
#include <sourcemod>

public void OnPluginStart()
{
    PrintToServer("Int: (%d)", -123);
    PrintToServer("Int: (%7d)", -123);
    PrintToServer("Int: (%07d)", -123);
    PrintToServer("Int: (%-07d)", -123);
}
```

### Relevant Log Output

**Expected Output:**

```shell
Int: (-123)
Int: (   -123)
Int: (-000123)
Int: (-123000)
```

**Actual Output:**

```shell
Int: (-123)
Int: (   -123)
Int: (000-123)
Int: (-123000)
```

PR #2329, opened 2025-05-31
Fixed "0[width]%d" formatting negative integers

F1F88:
Related to: #2328 

Fixed when formatting **negative integers** with **"%0[width]d"**, the **minus sign** should be formatted before the padding '0' instead of after the padding '0'.


psychonic (approved):
Thanks

PR #2335, opened 2025-06-06
fix geoip chi and zho not return chinese translation

Lxeon:
GeoLite2-City.mmdb supports zh-CN, but does not support chi and zho in sourcemod

The Chinese translation cannot be returned.
图片

This is a fix when GetLanguageInfo got chi or zho, convert to zh-CN to match GeoLite2-City.mmdb

### DIFF
--- a/extensions/geoip/geoip_util.cpp
+++ b/extensions/geoip/geoip_util.cpp
@@ -183,6 +183,9 @@ const char *getLang(int target)
 
 		if (translator->GetLanguageInfo(langid, &code, NULL))
 		{
+			if (strcmp(code, "chi") == 0 || strcmp(code, "zho") == 0) {
+				code = "zh-CN";
+			}
 			for (size_t i = 0; i < mmdb.metadata.languages.count; i++)
 			{
 				if (strcmp(code, mmdb.metadata.languages.names[i]) == 0)


peace-maker:
Good catch! Can we be sure that the mmdb always contains the zh-CN locale? It might be safer to check that and fallback properly if not.

F1F88:
> sourcemod default mmdb version didnt contain zh-TW, so it will cause error.
> 
> i not sure if mmdb is customizable.

This does not cause an error, but returns the default language "en".

See line 196:

https://github.com/alliedmodders/sourcemod/blob/20642a8518904ccfa0d36905471a313fcca2db7a/extensions/geoip/geoip_util.cpp#L186-L197

I checked GeoLite2-City_20250610 and it doesn't include zh-TW either.

~~When I asked chatgpt what languages ​​geoip returns, gpt's answer included zh-TW. Sorry, I should have verified this answer myself first.~~

Lxeon:
> > sourcemod default mmdb version didnt contain zh-TW, so it will cause error.
> > i not sure if mmdb is customizable.
> 
> This does not cause an error, but returns the default language "en".
> 
> See line 196:
> 
> https://github.com/alliedmodders/sourcemod/blob/20642a8518904ccfa0d36905471a313fcca2db7a/extensions/geoip/geoip_util.cpp#L186-L197
> 
> I checked GeoLite2-City_20250610 and it doesn't include zh-TW either.
> 
> ~When I asked chatgpt what languages ​​geoip returns, gpt's answer included zh-TW. Sorry, I should have verified this answer myself first.~

ye i miss the final return. and i think mmdb must include zh-cn.

PR #2345, opened 2025-07-31
Korean translation 250731

garrettjavalia:


Rushaway:
You are using "en" instead of "ko".

peace-maker:
Thank you! The '#format" keys are only allowed in the main English file and shouldn't be present in the other languages. Please remove them.

peace-maker:
Thank you! Since there were no complaints I'll assume this is fine :)

PR #2347, opened 2025-08-07
Push build for safetyhook submodule update

Headline:
Needed to trigger builders, doesn't seem like the submodule update did the trick


Kenzzer (approved):
I think the duck is asleep. So I'll merge this one, otherwise I fear the message in pushbuild.txt will become real

PR #2355, opened 2025-09-20
Updated sdktools gamedata offsets for Dystopia

bauxiteDYS:
Updated sdktools gamedata offsets for Dystopia


psychonic (approved):
Thanks! For what it's worth, the DispatchKeyValues offsets are unlikely to actually be used due to Dystopia having a new enough version of the IServerTools interface

Issue #2373, opened 2025-11-04
Admin plugin postgresql script should use `IF NOT EXISTS` on tables


solidDoWant:
## Help us help you
  - [x] I have checked that my issue [doesn't exist yet](https://github.com/alliedmodders/sourcemod/issues).

## Is your feature request related to a problem? Please describe.
I'd like to be able to apply DB migrations upon every server restart for all plugins. This works for the clientprefs plugins out of the box, but not the admins plugin because it is missing `IF NOT EXISTS` on the tables.

## Describe the solution you'd like
Add `IF NOT EXISTS` to `CREATE  TABLE` statements in the postgresql for the admin plugin specifically

## Describe alternatives you've considered
Copy the script, add it myself, and manage updates to it every time sourcemod changes

## Additional context
Should be a simple 8 line change. I can open a PR for this if desired. Here's what this looks like with the current version:

```sql
CREATE TABLE IF NOT EXISTS sm_admins (
  id serial,
  authtype varchar(6) NOT NULL,
  CHECK (authtype in ('steam', 'name', 'ip')),
  identity varchar(65) NOT NULL,
  password varchar(65),
  flags varchar(30) NOT NULL,
  name varchar(65) NOT NULL,
  immunity int NOT NULL,
  PRIMARY KEY (id)
);

CREATE TABLE IF NOT EXISTS sm_groups (
  id serial,
  flags varchar(30) NOT NULL,
  name varchar(120) NOT NULL,
  immunity_level int NOT NULL,
  PRIMARY KEY (id)
);

CREATE TABLE IF NOT EXISTS sm_group_immunity (
  group_id int NOT NULL,
  other_id int NOT NULL,
  FOREIGN KEY (group_id) REFERENCES sm_groups(id) ON DELETE CASCADE,
  FOREIGN KEY (other_id) REFERENCES sm_groups(id) ON DELETE CASCADE,
  PRIMARY KEY (group_id, other_id)
);

CREATE TABLE IF NOT EXISTS sm_group_overrides (
  group_id int NOT NULL,
  FOREIGN KEY (group_id) REFERENCES sm_groups(id) ON DELETE CASCADE,
  type varchar(10) NOT NULL,
  CHECK (type in ('command', 'group')),
  name varchar(32) NOT NULL,
  access varchar(5) NOT NULL,
  CHECK (access in ('allow', 'deny')),
  PRIMARY KEY (group_id, type, name)
);

[...]

PR #2376, opened 2025-11-10
Make PostgreSQL setup scripts idempotent

solidDoWant:
This makes the PostgreSQL setup scripts idempotent, so that they can be repeatedly ran against a database without failing. This is useful for applying database migrations upon server restart.

By definition this should have zero effect on existing installs, and is therefore 100% backwards compatible.

Closes https://github.com/alliedmodders/sourcemod/issues/2373


peace-maker (approved):
Thanks. Postgres supports the IF NOT EXISTS syntax since 9.1 which is so old now that we don't have to worry about it anymore.

Issue #950, opened 2019-01-14
Feature Request: [basebans] Ability to omit ban time for optional default/permanent setting


ghost:
# Environment
Current stable versions of Sourcemod & Metamod

# Description
When banning somebody using sm_ban from the basebans plugin, the plugin recognizes when a bantime is not specified, and correctly assumes it should be a permanent ban in this case. However, the ban reason string doesn't seem to work if you ban this way.

# Examples

`/sm_ban Theirname 0 BanReason`
This logs the ban reason, displays the ban reason to the person who got banned, and permanently bans the person from your server since 0 is interpreted as a permanent ban.

`/sm_ban Theirname BanReason`
This bans the person permanently, but does not log the ban reason or display the ban reason to the player.

For example, when banning "Someone" via the command `/sm_ban Someone TestReason`, the following gets logged (no reason)

`L 01/13/2019 - 05:00:33: [basebans.smx] "Console<0><Console><Console>" banned "Someone<115><STEAM_1:1:REMOVEDID><>" (minutes "0") (reason "")`

# Problematic Code (or Steps to Reproduce)

I'm not sure where exactly this is happening or how to fix it, however, according to the following line in ban.sp:
```
LogAction(client, target, "\"%L\" banned \"%L\" (minutes \"%d\") (reason \"%s\")", client, target, time, reason);
  ```
Based on the resulting sourcemod log of the reason being "", the 'reason' character array is blank when ban time is omitted.

**Comment by asherkin** [comment]
The ban time is not optional.

```
[SM] Usage: sm_ban <#userid|name> <minutes|0> [reason]
```

What is happening here is that your "reason" is being converted to an integer, which results in 0, and then no reason is provided (which is valid as the reason **is** optional).

This is mostly just classic bad integer conversion handling.

Greg48304:
Hmm, that makes sense! I guess I'll try to modify mine to always do this when I use the ban then. If I could suggest the ability to omit the ban time or a setting in sourcemod to allow this for permanent bans I think it would be helpful, I would assume most cases of people calling sm_ban are for permanent bans 

Thanks for your help

PR #2383, opened 2025-12-31
Fix a bug where ban reason is used as ban duration

luki1412:
In basebans there are 3 commands for banning: sm_banip, sm_addban, sm_ban
They have required arguments for ban duration and optional arguments for ban reason. However the functions that process the duration arguments, do not properly validate the duration and if the reason/text is provided instead of the duration, the functions just convert that string into an integer and use it as the ban duration. This usually results in a permaban with an empty reason. This commit adds a loop for checking duration string chars to be numeric before converting them to an int.

Also changing `<time>` to `<minutes|0>` for sm_banip and sm_addban to be consistent with sm_ban

Related to issue #950

Tested with sm_banip and sm_addban. Cannot test sm_ban

### DIFF
--- a/plugins/basebans.sp
+++ b/plugins/basebans.sp
@@ -10,7 +10,7 @@
  * This program is free software; you can redistribute it and/or modify it under
  * the terms of the GNU General Public License, version 3.0, as published by the
  * Free Software Foundation.
- *
+ * 
  * This program is distributed in the hope that it will be useful, but WITHOUT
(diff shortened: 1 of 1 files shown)


Kenzzer:
Nice work, thanks ! There's a lot of unnecessary space formatting changes in those files, this is going to make git diffing messy. If you could undo those, then this can be merged.

PR #2393, opened 2026-01-21
Update libmaxminddb

accelerator74:
Fixed memory leaks in MMDB_open() https://github.com/maxmind/libmaxminddb/pull/356


extensions/geoip/extension.cpp
```diff
@@ -55,44 +55,53 @@ bool GeoIP_Extension::SDK_OnLoad(char *error, size_t maxlength, bool late)
...
-					break;
 				}
 			}
 			dir->NextEntry();
 		}
 		libsys->CloseDirectory(dir);
 	}
 
-	if (!hasEntry)
+	if (m_GeoipDb[0] != '\0')
```
psychonic, line 94: We need to 0-initialize `m_GeoipDb`, else if the directory isn't found or otherwise fails to open, garbage data may still cause this check to still pass.

psychonic: (Looks good otherwise)

psychonic (approved):

PR #2394, opened 2026-01-21
Always up-to-date version of SQLite3

accelerator74:
I don't know how good this solution is, but it allows you to always keep SQLite3 up to date.


accelerator74:
I tried compiling SQLite3 from `.c` files in the `src` directory. This is impossible without running tcl script. At a minimum, sqlite.h.in is raw and requires placeholders to be filled. Some `.c` files are missing entirely (they are generated by the same tcl script). I don't think it's wise to try anything here and just continue using amalgation.

Kenzzer:
Hmmm this is really unfortunate, still thanks for looking into the matter it's highly appreciated ❤️

PR #2397, opened 2026-01-25
Guard IClient Netchan Path by GamedData

heymykola:
Prevent crash on Source 2007 forks where GetPlayerNetInfo()->GetMsgHandler() is not compatible with IClient.

Only attempt the netchan/IClient path for known-safe game folders; otherwise leave m_pIClient null.

PR #2401, opened 2026-01-31
Guard SetCSteamID When Auth String Is Invalid

heymykola:
This is a follow-up to PR #2397. While the IClient/netchan guard resolved the initial crash, the server is still crashing during CPlayer::Initialize() at the UpdateAuthIds() call on a non-stock Source 2007 engine fork.

This change simply adds a small defensive guard to the reads performed in that path so they are treated as unset instead of triggering an access violation.

### DIFF
--- a/gamedata/core.games/common.games.txt
+++ b/gamedata/core.games/common.games.txt
@@ -405,7 +405,6 @@
 		"Keys"
 		{
 			"RadioMenuClosesOnInvalidSlot"	"yes"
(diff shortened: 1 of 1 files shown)


Kenzzer:
I think It's best if we circle back on this crash from the beginning. I'm closing this PR and reverting #2397 (given that's what this new PR does anyways). Until a thorough investigation is conducted no subsequent PRs will be merged.

PR #2403, opened 2026-02-26
Replace `SourceHook` and `CDetour` with `KHook`

Kenzzer:
In the continuity of alliedmodders/metamod-source#223. This PR strips `SourceHook` and `CDetour` from SourceMod.

_Marked as a Draft until I finish writing the x86 assembly for dhooks on windows, and I'd like to have some unofficial builds tested around first. Local testing looks good though!_

## Breaking changes
- SourceMod (SM) 1.13 onwards will only be compatible with **MetaMod:Source (MM:S) 2.0+**.
- Extensions that make use of SourceHook will fail loading and compilation against SM 1.13 headers will fail.
- Extensions that make use of CDetour only will continue to load but fail compilation against SM 1.13 headers.

## Benefits
- DHooks extension finally supports x86_64.
- DHooks plugins can now detour the same function without crashing.
- Various other bug/crash fixes with DHooks.
- Extensions that make use of CDetour and hook the same function will no longer crash.

Those breaking changes are definitively not ideal, and I went in working on this knowing full well the PR might die at the finish line.

In the PR's defense, I have worked on this x86_64 support project (MM:S, SM, KHook) on and off for nearly 2 years, and in that timespan nobody has offered something close to a working alternative. I believe this is as good as it will ever get..


Kenzzer:
> Are we still concerned about using STL on public interfaces? things will get weird with gcc vs clang extensions.

That's a good point. I've mainly stripped every instances of the word `SourceHook` without auditing the consequences.

That said, looking at the two interfaces affected `IExtensionSys` & `IScriptManager` these two are bridge-only interfaces and don't get leaked to extensions. There's a public header called `IExtensionSys.h` but it exposes `IExtensionManager` not `IExtensionSys`. This can be further verified by the fact the `/public` dir's includes have not been modified by this PR, and none of the public interfaces make use of STL containers (or sourcehook's for that matter).

Core & bridge are built as a package deal so we probably don't have to worry about diverging STL implems. Unless we want to preserve that flexibility in hot swapping core and bridge libs, in which case I can just bring `sh_vector.h` into core only.

core/logic/ExtensionSys.cpp
```diff
@@ -394,7 +402,12 @@ bool CLocalExtension::IsLoaded()
 
 void CExtension::AddDependency(const IfaceInfo *pInfo)
 {
-	if (m_Deps.find(*pInfo) == m_Deps.end())
+	auto iter = m_Deps.begin();
+	while (iter != m_Deps.end() && !((*iter) == (*pInfo)))
```
Headline, line 406: why not std::find for this and m_Dependents

extensions/bintools/CallMaker.cpp
```diff
@@ -30,53 +30,82 @@
...
-	if (type < SourceMod::PassType_Basic || type > SourceMod::PassType_Object)
+	void SetReturnType(size_t size, PassType type, int flags,
+		void *pNormalCtor, void *pCopyCtor, void *pDtor, void *pAssignOperator)
 	{
-		return SourceHook::PassInfo::PassType_Unknown;
+		if (pNormalCtor)
+			flags |= PASSFLAG_OCTOR;
+
+		//if (pCopyCtor)
+		//	flags |= PassInfo::PassFlag_CCtor;
```
Headline, line 64: is this not needed?

Headline (approved):
After more carefully reading through the changes to MM:S I skimmed this SM side of resulting changes (and i skipped dhooks changes entirely) - I feel this slotted in pretty nicely all things considered. Any comments I felt I could make are trivial overall.

For such a massive change it's obviously difficult to review, but it's clearly matured and is obviously the next step

extensions/bintools/CallMaker.cpp
Mikusch: As far as I can tell this was not needed even with SH since all call sites passed NULL. Maybe this is a good opportunity to clean it up though, since the other 3 params are effectively dead too. Callers set these passflags by hand on the `PassInfo`. Could slim down the signatures a bit:
```cpp
void SetReturnType(size_t size, PassType type, int flags);
void AddParam(size_t size, PassType type, int flags);
```

core/logic/ExtensionSys.cpp
Kenzzer: Going to be completely honest, I was not thinking when doing these changes. Just whatever got me to the finish line of stripping SH containers haha. Anyways yeah thanks, I'll apply the suggested change

extensions/bintools/CallMaker.cpp
Kenzzer: This got answered, but yes. When I was making those changes to bintools, I realised a few statements were in reality unnecessary.

But I did not do any cleanups afterwards.

Issue #2170, opened 2024-05-26
Extensions don't trigger sourcepawn OnLibraryAdded unless you reload them


bottiger1:
Extensions that call sharesys->RegisterLibrary don't trigger OnLibraryAdded unless they are reloaded.

SourceMod Version Information:
    SourceMod Version: 1.12.0.7140
    SourcePawn Engine: 1.12.0.7140, interp-x86 (build 1.12.0.7140 NO JIT)
    SourcePawn API: v1 = 5, v2 = 16
    Compiled on: May 21 2024 12:31:10
    Built from: https://github.com/alliedmodders/sourcemod/commit/0f588b87
    Build ID: 7140:0f588b87
    http://www.sourcemod.net/

Example
```
public void OnLibraryAdded(const char[] name)
{
  PrintToServer("OnLibraryAdded %s", name);
}

public void OnAllPluginsLoaded()
{
  PrintToServer("loaded: %i", LibraryExists("testlib")); // always works
}

void Sample::SDK_OnAllLoaded()
{
    sharesys->RegisterLibrary(myself, "testlib"); // doesn't matter if you put this here or on SDK_OnLoad
    sharesys->AddNatives(myself, g_natives);
}

```

Kenzzer:
Is this a bug report or a feature request ? Is the behaviour you're describing of the forward different if you were to have used a plugin library as dependency instead ?

MAGNAT2645:
I assume this is a bug report because I also faced this problem. I wanted to use SteamWorks as an optional feature in my plugin so I added `#undef REQUIRE_EXTENSIONS` before including SteamWorks and added global boolean to track its state (so if it's `true` I can call its natives). But SteamWorks does not call `OnLibraryAdded` on first launch.

So in addition to `OnLibraryAdded` I had to add `LibraryExists` check in `OnAllPluginsLoaded` because at this point all extensions are loaded.

Kenzzer:
Right, but this might just be a load order issue, rather than a bug. If plugin A has X extension as dependency, and plugin B has X extension optional dependency. If plugin A loads before plugin B, then X extension has already been loaded, and `OnLibraryAdded` will not fire for subsequent loaded plugins. Hence why I'm asking if that behaviour is consistent with plugins that have other plugin library as dependencies.

MAGNAT2645:
Yeah, but should this happen on first server start (or restart) when all plugins are loading at the same time, not late ones?

bottiger1:
[...]

PR #2417, opened 2026-03-24
Fire OnLibraryAdded for extension libraries when plugins load.

bottiger1:
#2170


Kenzzer (dismissed):
Seems sane to me, matches already existing logic for plugin-libs. But @KyleSanderson what are your thoughts on this, it seems you were against this in the original issue - are you still against it now ?

PR #2418, opened 2026-03-24
Fix sm exts reload crashing because it doesn't update native pointers

bottiger1:
Automatically reloads dependent plugins.

### DIFF
--- a/core/logic/PluginSys.h
+++ b/core/logic/PluginSys.h
@@ -436,6 +436,17 @@ public:
 	void _SetPauseState(CPlugin *pPlugin, bool pause);
 
 	void ForEachPlugin(ke::Function<void(CPlugin *)> callback);
+
+	/**
+	 * Batch-loads plugins using the two-pass approach (compile all, then
+	 * run second pass for all) so that inter-plugin dependencies — including
+	 * circular ones — resolve the same way they do during initial load.
+	 *
+	 * Returns a vector of CPlugin pointers in the same order as the input.
+	 * Entries are nullptr for plugins that failed to load.
+	 */
+	std::vector<CPlugin *> LoadPluginBatch(
+		const std::vector<std::pair<std::string, PluginType>> &plugins);
 private:
 	LoadRes LoadPlugin(CPlugin **pPlugin, const char *path, bool debug, PluginType type);
 
--- a/core/logic/ExtensionSys.cpp
+++ b/core/logic/ExtensionSys.cpp
@@ -32,6 +32,11 @@
 #include <stdlib.h>
 
 #include <memory>
+#include <algorithm>
+#include <string>
+#include <unordered_map>
+#include <unordered_set>
+#include <vector>
 
 #include "ExtensionSys.h"
 #include <ILibrarySys.h>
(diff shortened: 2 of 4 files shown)


core/logic/PluginSys.h
```diff
@@ -436,6 +436,17 @@ class CPluginManager :
...
+
+	/**
+	 * Batch-loads plugins using the two-pass approach (compile all, then
+	 * run second pass for all) so that inter-plugin dependencies — including
+	 * circular ones — resolve the same way they do during initial load.
+	 *
+	 * Returns a vector of CPlugin pointers in the same order as the input.
+	 * Entries are nullptr for plugins that failed to load.
+	 */
+	std::vector<CPlugin *> LoadPluginBatch(
```
Headline: Consider ABI issues that would emerge from this

Headline (changes requested):

Kenzzer (changes requested):
I haven't gauged the ramifications of those changes. If I take the PR at face value, and the bug it claims to fix, I think this is a good PR.

I'll provide an in-depth review after @Headline's comment is addressed.
One request change though in the meanwhile.

dvander:
If you don't see this code as complex, something is wrong. It includes like five new data structures.

"working" is generous. The plugin implementation relies on authors confirming to a specification that is not easy to test. There are known bugs, that I gave up trying to fix.

c# only supports unloading at the assembly load context or app domain boundary, which is similar to our per plugin isolation.

Kenzzer:
> I'm fine with this going in, I'll leave that to others. But the consequences of having this functionality are big. If it were me, I would spend time thinking about how to not rely on it, since there might come a time where all plugins share a single heap, and unloading those individually will not be possible. And at that point, it's the same as meta reload.

Except for a few things, I would say the 'now' is more important than the future. If fixing this crash helps sourcemod users in their development cycle, then let them have it. Accepting this PR won't mean you can't backtrack the whole unload/reload/load feature later.

bottiger1:
> For the vast majority, if updating an extension in a production environment is necessary, I believe the safest approach will always be to script a full reload of the dependency chain, like so:

This is what my pull request does, so you don't have to do all that yourself. 

And with sm exts reload you are reloading extensions you know to be safe to reload. For a system based around plugins, not having the option to reload them seems crazy. Reloading is not only valuable to a few developers, it's also good to be able to unload leaking/laggy plugins without having to restart the server. These days, especially in older games, shutting down your server means the server is empty until the next day. 

If you are doing something insane like shared memory between extension and plugin... then don't reload that extension! Why make it impossible to reload just because some extensions or plugins aren't designed for it?

Kenzzer (commented):
To re-assure @bottiger1 I am keeping an eye on this PR, since there's at least me convinced to merge a fix for this. I dont believe there's anything wrong in bringing parity with plugins reload. If we have to break reload in the future well we will, but in the meantime I don't see any reason not to fix a crash for an helpful dev command, even if in your case its not the intention.

I'll be waiting on headline's thoughts, and of course psychonic's.

core/logic/PluginSys.h
Headline: Thought this was being exposed through a public iface- nvm

Headline:
> "meta reload sourcemod" is not viable because reloading everything just to test or update 1 extension is not something people do. Many extensions also have bugs that prevent them from being unloaded without crashing. But some of them don't. I take special care to make sure my extensions without natives are always reloadable.

"It's not viable" because "people don't do it" is not sufficient justification. The fact that your mention of many extensions crashing on reload indicates this feature isn't even being respected/cared for by most extensions either.

All in all I don't have very strong feelings on this matter, happy to take this for this SM version with slated removal in the future if we really want to tidy up.

Headline (approved):
Code itself looks fine, can't approve/validate/confirm edge cases. Also assuming this was tested with all extensions SM ships

Issue #2413, opened 2026-03-13
[Feature request] int64 to string conversion


godrider:
The `int64` type was recently introduced, but there currently does not appear to be a way to convert an `int64` value to a string.

Being able to convert an `int64` value to a string would be very useful. It would allow `int64` values to be displayed to players, used in menus, written to logs, or passed to anything that accepts string parameters. It would also make debugging a lot easier.

A more specific use case is formatting large numbers to make them easier to read. For example, the code below converts an `int` to a string and inserts decimal separators. A simple `int64` to string conversion, followed by a function like this, would make large `int64` values much easier to read.
```
stock void AddDecimalSeparator(int num, char[] buf, int maxlen)
{
	FormatEx(buf, maxlen, "%u", num);
	int pos = strlen(buf);
	int count = 1;
	
	while (pos > 0)
	{
		if (count == 4)
		{
			count = 0;
			int len = strlen(buf);
			for (int i = len; i >= pos; i--)
			{
				buf[i + 1] = buf[i];
			}
			buf[pos] = ',';
			buf[len + 2] = 0;
			pos++;
		}
		count++;
		pos--;
	}
}
```

PR #2421, opened 2026-03-26
Add direct int64-to-string support

Ayrton09:
## Summary

Add direct string conversion support for the newer `int64` type.

## Why

Issue #2413 points out that while `int64` now exists, there is no direct way to convert it to a string. This makes it harder to display values in menus, chat, logs, and debugging output.

## Changes

* add a direct conversion native for `int64`
* keep the legacy `Int64ToString(const int num[2], ...)` API unchanged
* clarify the legacy `Int64ToString` documentation in `string.inc`

## Notes

This is intentionally kept small and additive for backwards compatibility.

Draft PR for early feedback on API naming and parameter handling.

Refs #2413

### DIFF
--- a/core/logic/smn_string.cpp
+++ b/core/logic/smn_string.cpp
@@ -175,6 +175,18 @@ static cell_t Int64ToString(IPluginContext *pCtx, const cell_t *params)
 	return static_cast<cell_t>(res);
 }
 
+static cell_t Int64ToStringEx(IPluginContext *pCtx, const cell_t *params)
+{
+	char *str;
+	pCtx->LocalToString(params[3], &str);
+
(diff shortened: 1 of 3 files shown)


Kenzzer (changes requested):
Thank you for doing the work! There's one tiny mistake with the native implem.


Alternatively, If you feel like shouldering the burden, it would be great to add a formatting key for `int64`

E.g : `Format(string, sizeof(string), "%li", i64)`

You could then rewrite `native int Int64ToStringEx(int64 num, char[] str, int maxlength);` to
```sourcepawn
stock int Int64ToStringEx(int64 num, char[] str, int maxlength) {
    return Format(str, maxlength, "%li", num);
}
```

Kenzzer (approved):
Very nice work. Make sure to test changes and we can merge this.

Issue #799, opened 2018-04-03
Feature Request: Add/Remove ConVars using AutoExecConfig()


stickz:
I would like to request a feature which adds newly created plugin ConVars to sourcemod cfg files. Currently, the cfg file needs to be manually regenerated every time a new plugin ConVar is added.

A default boolean argument with a value set to false, could maintain backwards compatibility.  A plugin developer may not intend to use this behavior, knowing ConVars could be removed in the near future.

Also, it would be great anther default boolean argument with a false value, could be added to the function to remove old ConVars. This would keep configuration files cleaner when updating plugins.

stickz:
I have a plugin for Nuclear Dawn with 24 ConVars. Soon it's going to have 35 ConVars. The cfg files for sourcemod are pain to manage, when there's four different servers, some running different settings.
https://github.com/stickz/Redstone/blob/master/addons/sourcemod/scripting/nd_damage/convars.sp

stickz:
That wrapper is really useful for a handful of complex plugins. However, I still have about thirty other plugins with config files are a mess. It would be impractical to rewrite everything to use that wrapper.

Kenzzer:
Completed in #2436

PR #2436, opened 2026-04-09
Refactor AutoExecConfig to support new cvars

Dolly132:
Solves https://github.com/alliedmodders/sourcemod/issues/2427 & https://github.com/alliedmodders/sourcemod/issues/799

- This PR will automatically add any cvar that doesn't exist in the config and does exist in the plugin by deleting the generated file and generate a new one
- This PR will also automatically recreate the cfg file if it has a cvar that doesn't exist in the plugin's convars list


Dolly132:
I have made this PR draft until I safely recreate the .cfg file incase of a cvar mismatch

core/CoreConfig.cpp
```diff
@@ -394,90 +639,53 @@ bool SM_ExecuteConfig(IPlugin *pl, AutoConfig *cfg, bool can_create)
...
-#endif
-					{
-						continue;
-					}
-
-					char descr[255];
-					char *dptr = descr;
+	if (!will_create)
+	{
+		SM_ExecuteConfigFile(local);
```
psychonic: This seems to still be attempting to execute even if it doesn't exist.

core/CoreConfig.cpp
```diff
@@ -394,90 +639,53 @@ bool SM_ExecuteConfig(IPlugin *pl, AutoConfig *cfg, bool can_create)
...
-							}
-							next_ptr++;
-						}
-						fprintf(fp, "// %s\n", dptr);
-						dptr = next_ptr;
-					}
+	List<const ConVar *> *convars = NULL;
+	if (!pl->GetProperty("ConVarList", (void **)&convars, false) || !convars)
+	{
+		SM_ExecuteConfigFile(local);
```
psychonic: Same here.

core/CoreConfig.h
```diff
@@ -71,6 +71,18 @@ class CoreConfig :
...
 
+struct StringPolicy
+{
+	static inline uint32_t hash(const std::string &key)
+	{
+		return ke::FastHashCharSequence(key.c_str(), key.length());
+	}
+	static inline bool matches(const char *find, const std::string &key)
+	{
+		return key.compare(find) == 0;
```
psychonic: We need to be careful with case-sensitivity. IIRC, different versions of the engine have different handling in that regard. I don't remember offhand exactly what SM does for them in other areas.

core/CoreConfig.cpp
```diff
@@ -394,90 +639,53 @@ bool SM_ExecuteConfig(IPlugin *pl, AutoConfig *cfg, bool can_create)
...
+		return can_create;
 	}
 
-	if (file_exists)
+	/* File exists and misses cvars between the config file and the plugin's convars list
+	 * Remove the existing config file and generate a new one with the existing values in the old config file
+	*/
+	if (remove(file) != 0)
```
psychonic: If SM_GenerateConfigFile fails, we will have lost any existing custom values. Granted, I can't think of a _realistic_ case where that would happen where we could delete the file but not recreate it, but it may be worth having the new config written to a temp location and then moving over top of the existing one, rather than delete + regen.

psychonic:
Thanks for working on this. Concept looks good

core/CoreConfig.cpp
Dolly132: Fixed.

core/CoreConfig.cpp
Dolly132: Fixed.

core/CoreConfig.h
Dolly132: I have used strcasecmp instead

core/CoreConfig.cpp
Dolly132: I have changed the logic so it creates a new temporary file and overrides it to the existing file instead of removing the original one and regenerating it with the same file name.
I also made sure it works perfectly on windows with the MoveFileEx function and rename for Linux or any other os

Dolly132:
> Thanks for working on this. Concept looks good

Thank you so much for your review
I have made a commit to resolve everything you have reviewed, please let me know if there is anything else that needs to be done!

core/CoreConfig.h
```diff
@@ -71,6 +71,18 @@ class CoreConfig :
 	StringHashMap<std::string> m_KeyValues;
 };
 
+struct StringPolicy
+{
+	static inline uint32_t hash(const std::string &key)
+	{
+		return ke::FastHashCharSequence(key.c_str(), key.length());
```
psychonic: The hashing needs to have casing normalized. Something like:

```cpp
static inline uint32_t hash(const std::string &key) {
    std::string lower = ke::Lowercase(key.c_str());
    return ke::FastHashCharSequence(lower.c_str(), lower.length());
}
```

core/CoreConfig.cpp
```diff
@@ -318,6 +318,288 @@ inline bool IsPathSepChar(char c)
...
+#if SOURCE_ENGINE >= SE_ORANGEBOX
+		if (cvar->IsFlagSet(FCVAR_DONTRECORD))
+#else
+		if (cvar->IsBitSet(FCVAR_DONTRECORD))
+#endif
+		{
+			continue;
+		}
+
+		if (strcmp(cvar_name, cvar->GetName()) == 0)
```
psychonic: Switch this to strcasecmp now to match the `matches` func.

psychonic (changes requested):

core/CoreConfig.h
Dolly132: Done

core/CoreConfig.cpp
Dolly132: Done

psychonic (approved):

core/CoreConfig.cpp
```diff
@@ -318,6 +318,284 @@ inline bool IsPathSepChar(char c)
...
+		}
+
+		// Ignore comments
+		if (*ptr == '/' && *(ptr + 1) == '/')
+		{
+			continue;
+		}
+
+		size_t i = 0;
+		while (*ptr != '\0' && *ptr != '\n' && *ptr != '\r' && *ptr != ' ' && *ptr != '\t' && i < (sizeof(cvar_name) - 1))
```
Headline, line 349: This is all pretty tedious - as far as I can tell it's correct, but

Headline (approved):

Rushaway:
Congrats for your commit @Dolly132 👏

said72006:
Good work

PR #2437, opened 2026-04-10
Fix return values of some natives.

dvander:
These are listed as returning void, but actually have non-void returns in their implementation and pre-methodmap signatures.


xerox8521:
Shouldn't the return value be int? The return description says the return the number of bytes / characters written.

dvander:
Yes, you're right, good catch.

PR #2445, opened 2026-04-22
Fix out-of-bounds crash in PluginIterator_Next

### LINKED ISSUE #1880: PluginIterator crash for do-while
# Help us help you
  - [x ] I have checked that my issue [doesn't exist yet](https://github.com/alliedmodders/sourcemod/issues).
  - [x ] I have tried my absolute best to reduce the problem-space and have provided the absolute smallest test-case possible.
  - [x ] I can always reproduce the issue with the provided description below.

# Description
PluginIterator methodmap designed differently from normal functions, requiring Next to be called prior to accessing Plugin. This is not required for normal functions, so code needs to be changed from do-while to while. Using do-while causes crashing, but should instead throw error.

  ```
  PluginIterator iter = new PluginIterator();
  do {
    Handle p = iter.Plugin; /// <------- crashing here.
    ...
  } while( iter.Next() );
  delete iter;
  ```

# Logs
* Please attach in separate files: game output, library logs, kernel logs, and any other supporting information.
* In case of a crash, please attach minidump or dump analyze output.
Crash

SCP-076:
fixes [#1880](https://github.com/alliedmodders/sourcemod/issues/1880)
Previously, `PluginIterator_Next` checked `MorePlugins()` before calling `NextPlugin()`. This flawed logic caused the native to return true even when `NextPlugin()` pushed the internal iterator to the `end()` of the list. 

This commit swaps the execution order: the iterator is now advanced first, and then validated. This ensures `Next()` correctly returns false when reaching the end of the plugin list.

### DIFF
--- a/core/logic/smn_core.cpp
+++ b/core/logic/smn_core.cpp
@@ -151,6 +151,8 @@ public:
 	}
 	virtual IPlugin *GetPlugin() override
 	{
+		if (m_current == m_list.end())
+			return nullptr;
 		return *m_current;
 	}
 	virtual void NextPlugin() override
@@ -161,7 +163,8 @@ public:
 			return;
 		}
 
-		m_current++;
+		if (m_current != m_list.end())
+			m_current++;
 	}
 	virtual void Release() override
 	{
@@ -171,7 +174,7 @@ public:
 public:
 	virtual void OnPluginDestroyed(IPlugin *plugin) override
 	{
-		if (*m_current == plugin)
+		if (m_current != m_list.end() && *m_current == plugin)
 			m_current = m_list.erase(m_current);
 		else
 			m_list.remove(static_cast<SMPlugin *>(plugin));
@@ -402,12 +405,12 @@ static cell_t PluginIterator_Next(IPluginContext *pContext, const cell_t *params
 		return pContext->ThrowNativeError("Could not read Handle %x (error %d)", hndl, err);
 	}
 	
+	if(!pIter->MorePlugins())
+		return pContext->ThrowNativeError("PluginIterator %x is exhausted.", hndl);
+	
 	pIter->NextPlugin();
 	
-	if(!pIter->MorePlugins())
-		return 0;
-	
-	return 1;
+	return pIter->MorePlugins() ? 1 : 0;
 }
 
 static cell_t PluginIterator_Plugin_get(IPluginContext *pContext, const cell_t *params)
--- a/core/logic/smn_core.cpp
+++ b/core/logic/smn_core.cpp
@@ -401,11 +401,12 @@ static cell_t PluginIterator_Next(IPluginContext *pContext, const cell_t *params
 	{
 		return pContext->ThrowNativeError("Could not read Handle %x (error %d)", hndl, err);
 	}
-
+	
+	pIter->NextPlugin();
+	
 	if(!pIter->MorePlugins())
 		return 0;
-
-	pIter->NextPlugin();
+	
 	return 1;
 }
 
--- a/plugins/include/sourcemod.inc
+++ b/plugins/include/sourcemod.inc
@@ -262,7 +262,7 @@ methodmap PluginIterator < Handle
 	// Advances the iterator. Returns whether there are more plugins available in the iterator.
 	//
 	// @return              True on more plugins, false otherwise.
-	// @error               Invalid Handle.
+	// @error               Invalid Handle, or called when the iterator is already exhausted.
 	public native bool Next();
 
 	// Returns the current plugin in the iterator.


core/logic/smn_core.cpp
```diff
@@ -401,11 +401,12 @@ static cell_t PluginIterator_Next(IPluginContext *pContext, const cell_t *params
...
+	
+	pIter->NextPlugin();
+	
 	if(!pIter->MorePlugins())
 		return 0;
-
-	pIter->NextPlugin();
+	
 	return 1;
 }
```
Kenzzer, line 410:
```suggestion
	if(!pIter->MorePlugins())
		return pContext->ThrowNativeError("PluginIterator %x is exhausted.");
	
	pIter->NextPlugin();
	
	return pIter->MorePlugins() ? 1 : 0;
```

Kenzzer (changes requested):
Good catch, but this is still going to create undefined behaviour/crash if the iterator is advanced again after being exhausted.

core/logic/smn_core.cpp
SCP-076: Great catch! I applied your suggestion to throw a NativeError. To provide defense in depth, I also added bounds checks to `CMMPluginIterator`—this strictly prevents any out-of-bounds crashes from legacy API `ReadPlugin()` misuse or concurrent plugin unloads.

Kenzzer (approved):
This is much needed sanity around the plugin iterator, thanks for the work. Will let this sit for a few days to give the others a chance to comment if they want to.

SCP-076:
> This is much needed sanity around the plugin iterator, thanks for the work. Will let this sit for a few days to give the others a chance to comment if they want to.

Thanks. Just to be clear, the legacy iterator from `GetPluginIterator()` can actually still cause crashes; this PR strictly focuses on making the long-broken Methodmap version fully usable and safe. 

The current mix of the Methodmap implementation and its base class is quite confusing and definitely calls for a proper refactor someday.

~~BTW, One quick thought: since its traversal logic differs entirely from the legacy one, should we also check `m_hasStarted` in `PluginIterator_Plugin_get` and throw an error to enforce that `Next()` is called at least once? **Otherwise, if someone uses a `do-while` loop for traversal, the first two plugin objects retrieved will be duplicates.** I suppose this might be outside the scope of this PR, though.~~
_Actually, never mind. With the current `g_PlIter` mix, doing this would require either breaking the ABI or using `dynamic_cast`. It's a dead end for now._

SCP-076:
To prevent legacy iterator crashes without altering its expected behavior, we could add similar bounds checking to the underlying core `IPluginIterator`. However, I'm hesitant to touch the core to avoid unintended side effects. A safer compromise is to add the check directly inside the `ReadPlugin` native:

```c++
static cell_t ReadPlugin(IPluginContext *pContext, const cell_t *params)
{
	Handle_t hndl = (Handle_t)params[1];
	HandleError err;
	IPluginIterator *pIter;

	HandleSecurity sec;
	sec.pIdentity = g_pCoreIdent;
	sec.pOwner = pContext->GetIdentity();

	if ((err=handlesys->ReadHandle(hndl, g_PlIter, &sec, (void **)&pIter)) != HandleError_None)
	{
		return pContext->ThrowNativeError("Could not read Handle %x (error %d)", hndl, err);
	}
	
	// Check at the native level without modifying IPluginIterator->GetPlugin()
	if(!pIter->MorePlugins())
		return BAD_HANDLE;
	
	IPlugin *pPlugin = pIter->GetPlugin();
	if (!pPlugin)
	{
		return BAD_HANDLE;
	}

	pIter->NextPlugin();

	return pPlugin->GetMyHandle();
}
```
Let me know if you'd like me to include this fix to make the legacy iterator completely crash-proof from any SP calls.

core/logic/smn_core.cpp
```diff
@@ -401,12 +404,13 @@ static cell_t PluginIterator_Next(IPluginContext *pContext, const cell_t *params
 	{
 		return pContext->ThrowNativeError("Could not read Handle %x (error %d)", hndl, err);
 	}
-
+	
 	if(!pIter->MorePlugins())
-		return 0;
-
+		return pContext->ThrowNativeError("PluginIterator %x is exhausted.", hndl);
```
psychonic, line 409: I don't know that throwing an error here is the right answer. At minimum, it's a potentially breaking change. Is there harm in continuing to return 0/false?

psychonic: (The rest of the change seems okay)

SCP-076: > I don't know that throwing an error here is the right answer. At minimum, it's a potentially breaking change. Is there harm in continuing to return 0/false?

In practice, hitting this error should be extremely rare. It only triggers if Next() is called again after it has already returned false and terminated the loop.

For example:
```sourcepawn
while(iter.Next())
 {

 }
iter.Next(); // This would throw the error
```

I can't think of a valid use case for doing this. Given that this Methodmap feature has been broken/crashing for a long time, its actual usage is likely very low. Adding a protective error to catch objectively flawed logic seemed worthwhile.

That said, I'm completely fine with changing it to just return `false` if you prefer to strictly avoid a breaking change.

psychonic: We can leave it as a breaking change, and reevaluate that if/when backporting the change to the stable branch.

psychonic (approved):

core/logic/smn_core.cpp
Kenzzer: In my initial review I internally debated whether or not an exception should be thrown. And I landed on the answer that it's fine to throw one, because the alternative (old behaviour) is to crash the server.

So at worst I doubt this will be breaking any plugins, and at best this is a nice slap on the wrist for plugin makers not checking the return value of `.Next`.

psychonic: It wasn't crashing in all cases before though, and the crash is able to be fixed without throwing the exception. Either way, it's still fine enough, especially for dev branch.

PR #2446, opened 2026-04-24
Fix the funcommands plugin resetting all players' color and transparency on round end

RealBerke:
The code does not check if the command is active before removing its effects. Since these reset functions are also ran at the end of the round, they cause every player's set color and transparency to be reset when round ends even if the command was not activated and it was set by other means such as other plugins or maps. This goes against the default behaviour where player color and transparency persists between round resets, some maps may prefer to keep these between rounds.

Also removed the unnecessary color and transparency reset in beacon code as this plugin never sets those, so no need to reset it.

### DIFF
--- a/plugins/funcommands/beacon.sp
+++ b/plugins/funcommands/beacon.sp
@@ -44,11 +44,6 @@ void CreateBeacon(int client)
 void KillBeacon(int client)
 {
 	g_BeaconSerial[client] = 0;
-
-	if (IsClientInGame(client))
-	{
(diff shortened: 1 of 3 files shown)


psychonic (approved):
Thanks!

The beacon changes looked good. For firebomb/timebomb, I think it was only fixing part of the issue. I've pushed a further fix for that. I also pushed a fix for an existing issue in firebomb where it was referencing the triggering client rather than the targeted one.

PR #2449, opened 2026-04-26
Add native StringToInt64Ex

### LINKED ISSUE #2441: [Feature request] String to int64 conversion
## Help us help you

## Is your feature request related to a problem? Please describe.
Recently #2421 was merged, which converts the new `int64` type to a string. I would like to use a function that converts a string back to `int64`, just like the existing `StringToInt64` function but for the real `int64` type.

## Describe the solution you'd like
A function such as `int StringToInt64Ex(const char[] str, int64 result, int nBase)` would be nice. This means we can keep `StringToInt64` for backward compatibility. Even better would be to return the result without passing it as a parameter like the `StringToInt` function. Either way works for me.

F1F88:
Close #2441


core/logic/smn_string.cpp
```diff
@@ -155,6 +155,25 @@ static cell_t StringToInt64(IPluginContext *pCtx, const cell_t *params)
...
+	if (int err = pCtx->LocalToString(params[2], &str); err != SP_ERROR_NONE)
+		return pCtx->ThrowNativeErrorEx(err, "Invalid str (error %d)", err);
+
+	cell_t *bytes;
+	if (int err = pCtx->LocalToPhysAddr(params[3], &bytes); err != SP_ERROR_NONE)
+		return pCtx->ThrowNativeErrorEx(err, "Invalid bytes (error %d)", err);
+
+	*reinterpret_cast<int64_t*>(out) = strtoll(str, &dummy, params[4]);
+	*bytes = dummy - str;
+	return 0;
```
Headline, line 174: everything here it's pretty inconsistent with everything else, nothing here error checks, init-statement use

functionally looks good otherwise

Headline:
I see it appears to be tested given the presence of tests

core/logic/smn_string.cpp
F1F88: I'm a bit confused. Could you give a more specific description of the problem?

Regarding the error checking of the `pCtx->LocalTo***` part, I referenced: https://github.com/alliedmodders/sourcepawn/blob/4e87a849f092e9bb3e70e50f6a6f689e963974f0/vm/shell/shell.cpp#L127

I think this way is safer, and those few condition checks won't really slow things down.

Regarding the usage of the `strtoll` part, I referenced: https://github.com/alliedmodders/sourcemod/blob/87d668afce37caea584886b29fa013a6d1b27767/core/logic/smn_string.cpp#L150

F1F88:
> I see it appears to be tested given the presence of tests

Yes, the `test_string.sp` script runs the test suite on the mock server in CI

plugins/include/string.inc
```diff
@@ -212,6 +212,16 @@ native int StringToIntEx(const char[] str, int &result, int nBase=10);
...
 
+/**
+ * Converts a string to an int64 value.
+ *
+ * @param str           String to convert.
+ * @param bytes         Number of characters consumed.
+ * @param nBase         Numerical base to use.  10 is default.
+ * @return              int64 conversion of string, or 0 on failure.
+ */
+native int64 StringToInt64Ex(const char[] str, int &bytes=0, int nBase=10);
```
psychonic: I'm a little torn on the name. With the other StringTo_Ex natives, the Ex versions return the count of chars consumed and give the actual result via an out param. This does the opposite, due to it using the Ex suffix for a different reason.

That said, I'm struggling to think of a better name. Something like StringToInt64Native sounds clunky. Maybe StringToI64? Anyone else have thoughts?

plugins/testsuite/mock/test_string.sp
```diff
@@ -0,0 +1,108 @@
...
+        1234567654321);
+    AssertEq("bytes  11F71F76BB1", bytes, 14);
+
+    AssertInt64Eq(
+        "result 11F71F76BB1", 
+        StringToInt64Ex("11F71F76BB1", bytes, 16),
+        1234567654321);
+    AssertEq("bytes  11F71F76BB1", bytes, 11);
+
+    // Orther
```
psychonic: Minor typo here.

plugins/include/string.inc
F1F88: I think `StringToI64` is a good name and fits the existing StringTo_Ex natives well.

I'd also like to suggest `ParseInt64` as another candidate — it avoids the `Ex` suffix confusion entirely, and the `Parse` verb more naturally conveys "extract a value from a string" with the result returned directly. It also aligns with naming conventions in languages like `native int ParseTime(const char[] dateTime, const char[] format);`, C# (`Int64.Parse`) and Java (`Long.parseLong`), which might feel intuitive to a wider audience.

Happy to go with whichever name the team prefers!

godrider: `StringToI64` makes the most sense to me! It keeps StringTo and removes the Ex suffix to avoid confusion.

F1F88:
- `StringToInt64Ex` has been renamed to `StringToI64`.

    > Just out of curiosity, I'd like to ask what you think of the name `StringToInt64t`? 
    > `StringToI64` is a good and short name, but it doesn't seem to clearly distinguish between 64-bit and int64 types.

- The **second parameter** has been changed to `nBase`, and the **third parameter** has been changed to `nConsumed (originally bytes)`. 

    > I think this is closer to `StringToInt` than `StringToIntEx`.

- The handling of `str` value out-of-bounds errors has been explained in the documentation.

F1F88:
Strangely, it seems that `smn_string.cpp` and `string.inc` are showing as a complete rewrite of the entire file, but when I view the diff locally, the modified parts are displayed correctly.

Headline (approved):

PR #2457, opened 2026-05-12
Fix `IForward` definition

### LINKED ISSUE #2456: [Bug]: New virtual PushInt64 silently breaks every extension compiled before this
### Prerequisites

- [x] I have checked that my issue doesn't exist yet in the [issue tracker](https://github.com/alliedmodders/sourcemod/issues)

### Operating System and Version

any

### Game / AppID and Version

any

### SourceMod Version

https://github.com/alliedmodders/sourcemod/commit/59ecab555b16186384ca7ccdeb5e4071b93b6fa3

### Metamod:Source Version

2.0.0-dev+1293

### Version Verification

- [x] I have updated SourceMod to the [latest version](https://www.sourcemod.net/downloads.php) and the issue persists
- [x] I have updated SourceMod to the [latest snapshot](https://www.sourcemod.net/downloads.php?branch=dev) and the issue persists
- [x] I have updated Metamod:Source to the [latest snapshot](https://sourcemm.net/downloads.php?branch=dev) and the issue persists

### Updated SourceMod Version

_No response_

### Updated Metamod:Source Version

_No response_

### Description

Due to adding a new vritual PushInt64 to ICallable, all extensions compiled before this will have a broken forward->Execute call due to pushing the offsets of every virtual on IForward.

Is this intentional or will any attempts will be made to mitigate this problem before I recompile a bunch of extensions?

### Steps to Reproduce

_No response_

### Relevant Log Output

```shell

```

Kenzzer:
Fixes #2456 

This bug was known for a while (9th of April 2026), but unfortunately the lack of issue on the tracker made me forget how urgent it was.
https://discord.com/channels/335290997317697536/335290997317697536/1491812815860797532

This PR hopefully fixes vtable shift bugs with `IForward` forever. I have removed the inheritance of `ICallable` by `IForward`, if any extension out there relied on the ability to mix `IForward(s)` with `ICallable(s)` this is unfortunately permanently broken. However I am highly doubtful this is a feature that was desired by extensions.

An alternative fix to this PR would be to have sourcepawn's `ICallable` be turned into a new class `ICallable2`. I don't think I need to argue why this is a bad solution and avoided it.

One more fix would be to find a method so that gcc,clang,msvc doesn't merge the vtable of `IForward` with `ICallable`. However even if that is possible, this won't help already compiled extensions. So this fix is a no go as well.

### DIFF
--- a/core/logic/ForwardSys.cpp
+++ b/core/logic/ForwardSys.cpp
@@ -446,12 +446,6 @@ int CForward::PushCell(cell_t cell)
 	return SP_ERROR_NONE;
 }
 
-int CForward::PushInt64(int64_t value)
-{
-	// Not supported yet
-	return SetError(SP_ERROR_PARAM);
-}
-
 int CForward::PushFloat(float number)
 {
 	if (m_curparam < m_numparams)
--- a/core/logic/ForwardSys.h
+++ b/core/logic/ForwardSys.h
@@ -41,21 +41,19 @@ typedef ReentrantList<IPluginFunction *>::iterator FuncIter;
 
 class CForward : public IChangeableForward
 {
-public: //ICallable
-	virtual int PushCell(cell_t cell);
-	virtual int PushCellByRef(cell_t *cell, int flags);
-	virtual int PushFloat(float number);
-	virtual int PushFloatByRef(float *number, int flags);
-	virtual int PushArray(cell_t *inarray, unsigned int cells, int flags);
-	virtual int PushString(const char *string);
-	virtual int PushStringEx(char *buffer, size_t length, int sz_flags, int cp_flags);
-	virtual int PushInt64(int64_t value);
-	virtual void Cancel();
 public: //IForward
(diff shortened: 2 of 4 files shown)


Kenzzer:
Disclaimer I have not tested the following changes, everything was done in the web editor. We desperately need a test suite for this sort of things.

dvander:
I'm 80/20 convinced now is the right time to delete a ton of crap from the SP API. It's almost 20 years old and we should never have tied ICallable like this... we will need a new native signature to add anything.On May 12, 2026 2:10 AM, Benoist ***@***.***> wrote:Kenzzer left a comment (alliedmodders/sourcemod#2457)
Disclaimer I have not tested the following changes, everything was done in the web editor. We desperately need a test suite for this sort of things.

dvander:
As noted on ~~IRC~~Discord, the fix has to be upstream in SP since it also pushed the IPluginFunction vtables around.

PR #2458, opened 2026-05-13
Fix strncpy buffer overflow in OnConVarChanged

SCP-076:
strncpy does not append '\0' when the source is >= sizeof(dst). 
If a convar value is 512+ chars, PushString(newValue) leaks stack garbage into SP callbacks' newValue parameter.


Headline (approved):
Nice one

PR #2472, opened 2026-05-25
doc: clarify late load callbacks

SCP-076:
`OnMapStart` and `OnConfigsExecuted` are called during reload. 
This is slightly counterintuitive and may affect some plugins' initialization sequence. 
I suggest adding notes to clarify their late-load behavior.


psychonic (approved):
Thanks

PR #2475, opened 2026-05-27
Remove exception from Call_Finish

Kenzzer:
- Removed native throw on failure to execute forward


Malifox:
This is effectively what I did in my testing build to clear that out of the way. The main thing that gave me hesitation is that the error code information is lost when using Call_StartFunction() compared to before, but I'm not sure how practically relevant that really is. It also maybe doesn't comply with the wording mentioned here for Invoke() though: https://github.com/alliedmodders/sourcepawn/blob/c119283ac4c1764c83eaf5a5cf450ac497e09d35/include/sp_vm_api.h#L170-L173
```
 * Unlike Execute(), this does not reset the exception state. It is
 * illegal to call Invoke() while an exception is unhandled. If it
 * returns false, an exception has been thrown, and must either be
 * handled via ExceptionHandler or propagated to its caller.
```
The new `ScriptedInvoker::Execute(cell_t* result)` will utilize it's own `default_args_` CallArgs, but here a separate `static sp::CallArgs sArgs;` is being used, which is one reason I see for it having been replaced, but maybe we do need a new `ScriptedInvoker::Execute(const CallArgs& args, cell_t* result)` overload to deal with that? Not really sure as a lot of things were changed and I wasn't even familiar with the old setup.

I had doubts about that `ThrowNativeErrorEx()` for forwards too, mostly that I'm not sure if it is ever actually desired to ignore a forward call itself failing or if that would always be a bug in the caller's plugin that should be fixed, though I'm not sure how you would reach that point without hitting some other error first.

core/logic/smn_functions.cpp
```diff
@@ -551,16 +551,15 @@ static cell_t sm_CallFinish(IPluginContext *pContext, const cell_t *params)
...
+			return SP_ERROR_NONE;
 	} else {
 		IForward *pForward = s_pForward;
 		ResetCall();
-		if (int err = pForward->Execute(local_args, result); err != SP_ERROR_NONE)
-			return pContext->ThrowNativeErrorEx(err, nullptr);
+		return pForward->Execute(local_args, result);
 	}
 
-	return 1;
```
dvander: For the invoke case, you want to declare an ExceptionHandler, then if Invoke() returns false, return the handler's error code. Then you don't need the made-up value here.

Malifox: > For the invoke case, you want to declare an ExceptionHandler, then if Invoke() returns false, return the handler's error code. Then you don't need the made-up value here.

Regarding the ExceptionHandler, it seems to currently lack a method to get the error code, only providing a way to get the exception message with Message(). Would there be a reason not to add a Code() method as well? As in:

```c++
// SourcePawn::ExceptionHandler
int Code() const {
	return env_->GetPendingExceptionCode(this);
}

// SourcePawn::ISourcePawnEnvironment
// @brief Returns the error code of the pending exception.
virtual int GetPendingExceptionCode(const SourcePawn::ExceptionHandler *handler) = 0;

// sp::Environment
int GetPendingExceptionCode(const ExceptionHandler* handler) override;

int
Environment::GetPendingExceptionCode(const ExceptionHandler* handler) {
    // Note here and elsewhere - this is not a sanity assert. In the future, the
    // API may need to query the handler.
    assert(handler == eh_top_);
    assert(HasPendingException(handler));
    return exception_code_;
}
```

With that we could do:
```c++
if (s_pFunction)
{
	IPluginFunction *pFunction = s_pFunction;
	ResetCall();

	ExceptionHandler eh(pFunction->GetParentRuntime());
	return pFunction->Invoke(local_args, result) ? SP_ERROR_NONE : eh.Code();
} else {
	IForward *pForward = s_pForward;
	ResetCall();
	return pForward->Execute(local_args, result);
}
```

Quick test (usual native setup omitted for brevity, but is 2 separate plugins)
```sourcepawn
public void OnLibraryAdded(const char []name)
{
	if (StrEqual(name, "test_call_finish", true))
		Test_CallFunction(OnFunctionCall);
}

int OnFunctionCall()
{
	int iArray[2] = {42, ...};
	int iOutOfBoundsIndex = sizeof(iArray);

	return iArray[iOutOfBoundsIndex];
}
```

```sourcepawn
void Native_TestCallFunction(Handle plugin, int numParams)
{
	int iResult;
	int iError;

	Call_StartFunction(plugin, GetNativeFunction(1));
	iError = Call_Finish(iResult);

	PrintToServer("Native_TestCallFunction: iError = %d, iResult = %d", iError, iResult);
}
```
Output:
```
[SM] Exception reported: Array index out-of-bounds (index 2, limit 2)
[SM] Blaming: testing/test_call_finish_pt2.smx
[SM] Call stack trace:
[SM]   [1] Line 24, test_call_finish_pt2.sp::OnFunctionCall
[SM]   [3] Call_Finish
[SM]   [4] Line 88, test_call_finish.sp::Native_TestCallFunction
[SM]   [6] Test_CallFunction
[SM]   [7] Line 11, test_call_finish_pt2.sp::OnLibraryAdded

Native_TestCallFunction: iError = 15, iResult = 0
```
Error code 15 is commented as `/**< Array index is out of bounds */`, so this looks like the expected behavior.

dvander: Huh, I could have sworn I added a getter for the code recently. I guess not. I'll add one.

dvander: When https://github.com/alliedmodders/sourcepawn/pull/1137 is merged you can update the sourcepawn pinned commit here.

Malifox: > Huh, I could have sworn I added a getter for the code recently. I guess not. I'll add one.

Thanks for adding!

So, this issue was the last exception error I was seeing other than https://github.com/alliedmodders/sourcepawn/issues/1135 at startup from the plugins that are loading (I haven't yet gone through the great extension recompiling). I also hadn't yet recompiled all plugins with a recent compiler, thus everything was running on the interpreter in an x64 environment. After mass compiling ~95% of plugins, I was greeted with only a single error from an old plugin:
```
[SM] Exception reported: Unknown error code 0
[SM] Blaming: zstore/store.smx
[SM] Call stack trace:
[SM]   [0] Call_Finish
[1] Line 179, zephstocks.inc::GlobalConVarChanged
```
So, what is wrong with Call_Finish() this time? Well... nothing, turns out that the function it calls is using an old float native: https://github.com/alliedmodders/sourcemod/blob/0002aa228cb0bac5891a0668e9f6d924077cda5d/plugins/include/float.inc#L162
Something interesting is that if you load the following, you'll only get `Exception reported: Unknown error code 0` with no plugin being blamed, leaving nothing to go on.
```sourcepawn
public void OnPluginStart()
{
	FloatCompare(1.0, 0.0);
}
```
I would say that this looks the most suspicious, unsure if it is a JIT exclusive issue though: https://github.com/alliedmodders/sourcepawn/commit/4bef6073118c61a33f0c7eaf3a3f016e8e0a5659
Other than 2 stock plugins, this is literally the only plugin I see using this float native from the sources I have, so I can easily just replace it, but would still remain an issue for SM's included stock plugins. I'm also certain that it wasn't compiled post that commit. For reference, the 2 stock plugins are:
https://github.com/alliedmodders/sourcemod/blob/0002aa228cb0bac5891a0668e9f6d924077cda5d/plugins/basevotes.sp#L297 https://github.com/alliedmodders/sourcemod/blob/0002aa228cb0bac5891a0668e9f6d924077cda5d/plugins/funvotes.sp#L223

Would you happen to have any ideas on that one? I can move this to it's own issue if preferred, but given how I discovered it, I figured I would go ahead and mention it here.

core/logic/smn_functions.cpp
```diff
@@ -551,16 +551,15 @@ static cell_t sm_CallFinish(IPluginContext *pContext, const cell_t *params)
 	{
 		IPluginFunction *pFunction = s_pFunction;
 		ResetCall();
-		if (!pFunction->Invoke(local_args, result))
-			return 0;
+		if (pFunction->Invoke(local_args, result))
+			return SP_ERROR_NONE;
```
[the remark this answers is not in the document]
Malifox: Following https://github.com/alliedmodders/sourcepawn/pull/1137, this appears to work as expected.

dvander:
Sorry, I'm going to gank this PR w/ #2482 so I can stop feeling guilty for having broken the API.

core/logic/smn_functions.cpp
dvander: If you can reproduce it after the current callfunction bugs are closed, please file a new bug, I'll investigate.

Kenzzer:
> Sorry, I'm going to gank this PR w/ #2482 so I can stop feeling guilty for having broken the API.

No prob. Though I'm going to repurpose this into yeeting the exception.

dvander (approved):

PR #2478, opened 2026-05-30
Update The Ship Gamedata

caxanga334:
# SDKHooks

Added gamedata for SDKHooks.

# SDKTools.

Added some missing vtable offsets.
Fixed Windows `DispatchSpawn` signature.
Added signatures for FireOutput and LookupAttachment.

Marking as draft, need to confirm a few offsets and sigs are correct and investigate a crash (https://crash.limetech.org/w4pnx2ocz57n) potentially caused by SDKHooks.


caxanga334:
Should be good.

`GroundEntChanged` had to be removed, [SDKHooks hooks](https://github.com/alliedmodders/sourcemod/blob/0002aa228cb0bac5891a0668e9f6d924077cda5d/extensions/sdkhooks/extension.cpp#L1123) the function with a `void*` parameter, which doesn't exist in the ship.

psychonic (approved):
Thanks!

PR #2479, opened 2026-05-31
feat: webclient extension

rumblefrog:
## Motivation

SourceMod plugins routinely need to talk to web services (backend, Discord, APIs), but the platform ships no first-party HTTP or WebSocket client. Authors fall back on third-party extensions that vary in TLS stack, maintenance, and safety, and there's no built-in WebSocket support at all. This PR adds a webclient extension that fills that gap: asynchronous HTTP and WebSocket client natives with a clean methodmap API. All networking runs in a Rust core (tokio + rustls) behind a thin, main-thread-only C++ shim.

## Summary
New `webclient` extension: non-blocking HTTP + WebSocket client natives for SourcePawn, backed by a Rust core (tokio/reqwest/tungstenite, rustls).  Architecture, safety defaults, and build notes are in `extensions/webclient/README.md`.
 
## Tests 
  
image

```
sm_webclient_get
[WebClient] GET dispatched (token 2)...
[WebClient] GET ok=1 status=200 length=528 url=https://example.com/?q=hello+world+%26+friends content-type=text/html
[WebClient] server returned 200 OK
[WebClient]   date: Sun, 31 May 2026 06:40:41 GMT
[WebClient]   content-type: text/html
[WebClient]   server: cloudflare
[WebClient]   last-modified: Thu, 28 May 2026 04:54:11 GMT
[WebClient]   allow: GET, HEAD
[WebClient]   age: 31
[WebClient]   cf-cache-status: HIT
[WebClient]   cf-ray: a043fd12b9148969-PDX

sm_webclient_post
[WebClient] POST dispatched...
[WebClient] POST status=200 body={
  "args": {},
  "data": "{\"hello\":\"sourcemod\"}",
  "files": {},
  "form": {},
  "headers": {
    "Accept": "*/*",
    "Accept-Encoding": "gzip,deflate",
    "Content-Length": "21",
    "Content-Type": "application/json",
    "Host": "httpbin.org",
    "X-Amzn-Trace-Id": "Root=1-6a1bd7f3-17afcf8f767a9c211337bb07"
  },
  "json": {
    "hello": "sourcemod"
  },
  "origin": "5.78.236.182",
  "url": "https://httpbin.org/post"
}

sm_webclient_download
[WebClient] download dispatched...
[WebClient] download ok=1 status=200 (saved to data/webclient)

sm_webclient_ws
[WebClient] WebSocket connecting...
[WebClient] WS connected; sending hello
[WebClient] WS message (32 bytes, text=1): Request served by 4d896d95b55478
[WebClient] WS disconnected code=1000 reason=
```

```
sm_webclient_post
[WebClient] POST dispatched...
sm exts unload 10 542
[SM] Extension webclient.ext.so is now unloaded
```
 
  
  ## Test plan
  - [x] Builds via AMBuild
  - [x] Example: GET/POST, download, WS connect/echo/close
  - [x] No crash on plugin unload mid-request


dvander:
This would be better off as a third party extension than as part of the tree. If it were in C++ I might have a different opinion, but this here is a can of worms. Virtually no maintenance path and it will hamstring the C++ API for SP.

rumblefrog:
> This would be better off as a third party extension than as part of the tree. If it were in C++ I might have a different opinion, but this here is a can of worms. Virtually no maintenance path and it will hamstring the C++ API for SP.

Fair enough, that's a reasonable call.

That said, I do think there's a real case for bringing a newer language like Rust or Zig into core eventually. It could lower the barrier to entry and help draw in a younger generation of contributors.

This does remind me of the whole Rust-in-the-Linux-kernel saga, though. Both sides have strong opinions there, and I'm not looking to argue for either one.

zer0k-z:
> It could lower the barrier to entry

I would argue quite the opposite simply by the fact that you would need to know two languages instead of one to understand the code base.

PR #2487, opened 2026-06-17
Fix SDKCall mis-decoding int64 Address parameters

Mikusch:
Remove a leftover extra dereference in `DecodeValveParam`'s `Valve_SMAddress` case that double-indirected every `Address` argument passed to SDKCall, so the call usually failed to decode.


psychonic (approved):
Thanks!

PR #2488, opened 2026-06-17
Fix x86 sign-extension of high pointers

Mikusch:
On x86 a pointer `0x90000000` sign-extends to `0xFFFFFFFF90000000`. Use `uintptr_t` to zero-extend instead.


psychonic (approved):
Thanks!

PR #2492, opened 2026-06-23
Fix up DHooks, SDKHooks & SDKTools on KHook

Mikusch:
Gets DHooks, SDKHooks and SDKTools building and running against KHook on x64, and fixes the regressions versus the old SourceHook behavior along with several leaks and hook-related bugs found while testing.

Based on the KHook migration work and depends on the matching KHook changes (https://github.com/Kenzzer/KHook/pull/13). This is a bit of a chonky PR, and I tried to break it up into separate commits as best I could.


Kenzzer (approved):
Like I've mentioned privately. I really appreciate the second pair of eyes on all of these changes. So many bugs could be easily missed, and this PR just pushes us closer to the finish line.

Now all that's really left is writing the dhooks assembly for Linux x86, Windows x86, and Windows x86_64. And we will be ready.

PR #2497, opened 2026-07-02
Add x64 sdkhooks gamedata

Mikusch:


Kenzzer:
Are we sure this is correct, which engine branch were those offsets obtained from ? What's the intent. Those offsets should be closer to `131144`. If this is to support games that are building against sdk2013-2025, the engine branch is HL2MP where the offsets are provided.

@psychonic do we want to merge this ? We already override `EntityListeners` for every diverging engine branch that have x64 support.

Here are the games that would fallback on common, but don't have x64 support (yet/ever)
- Dark Messiah
- Adrenaline Gamer 2
- Age of Chivalry
- Dino D-Day
- Fortress Forever
- Dystopia
- Empires
- Fistful of Frag
- Golden Eye: Source
- Gmod 9
- Hidden
- HL2: CTF
- Insurgency
- Kreedz
- Modular Combat
- Neotokyo
- No More Room in Hell
- Nuclear Dawn
- Open Fortress
- Pre-Fortress 2
- Stargate: The Last Stand
- The Ship
- SourceFort
- Zombie Master
- Zombie Panic Source

From my point of view, if those games (or new ones) are going to support x64. They will either : 
- make their own sauce and have diverging offsets requiring specialised gamedata anyways
- or they port to sdk2013-2025, leading them to be detected as HL2DM, so no fallback on common.txt

I would personally give this a few years, to see how many x64 source games are cropping up. Before deciding which offset for EntityListeners is common enough to take that spot.

Mikusch:
The gamedata is correct, but you are right that it is redundant. This is meant for tf2, css, dods and hl2dm (current sdk2013 games). After looking deeper into it, all of these are covered by `engine.ep2v.txt` already, which was updated in #2141. Closing.

PR #2509, opened 2026-07-14
Fix forwards overwriting result to 0 when no callback runs

Mikusch:
A regression introduced by 468760c (present on `master`). A forward with no successfully executed callbacks now overwrites the caller's result with 0, whereas previously it would leave it untouched.


Fyren (approved):
Thanks.

PR #2510, opened 2026-07-14
Fix Call_Finish propagating exceptions to the caller

Mikusch:
A regression introduced by 468760c (present on master). An exception thrown by the invoked function now propagates out of `Call_Finish` and aborts the calling function, whereas previously it was swallowed.


Fyren (approved):
Thanks.

PR #2511, opened 2026-07-14
Fix Plugin_Stop stopping ET_Event forwards early

Mikusch:
As of 468760c (present on `master`), `ET_Event` forwards violate the documented ["no mid-Stops allowed"](https://github.com/alliedmodders/sourcemod/blob/0732992c4fe8e9df418a0601d3b5867e83d83094/public/IForwardSys.h#L85) contract. One plugin returning `Plugin_Stop` in a forward skips later plugin callbacks.

I'm not entirely sure whether this was an intentional behavior change that the docs simply weren't updated to reflect, but feel free to close if so.


Fyren (approved):
Thanks.

PR #2512, opened 2026-07-15
Allow STEAM_0 in ProcessArgString

### LINKED ISSUE #1240: STEAM_X ignore
Seems like SM doesn't ignore the **x** in STEAM_X

`smrpg_addcredits "#STEAM_1:1:102325437" "1000"
SM:RPG addcredits: Toby | TFrag.dk got 1000 credits and now has 14504 credits (previously had 13504 credits)`

Compared to:

`smrpg_addcredits "#STEAM_0:1:102325437" "1000"
[SM] No matching client was found.`

https://cs.alliedmods.net/sourcemod/source/core/PlayerManager.cpp#1722

Kenzzer:
Fix #1240

### DIFF
--- a/core/PlayerManager.cpp
+++ b/core/PlayerManager.cpp
@@ -1632,7 +1632,7 @@ void PlayerManager::ProcessCommandTarget(cmd_target_info_t *info)
 
 		/* Do we need to look for a steam id? */
 		int steamIdType = 0;
-		if (strncmp(&info->pattern[1], "STEAM_1", 7) == 0 || strncmp(&info->pattern[1], "STEAM_0", 7) == 0)
+		if (strncmp(&info->pattern[1], "STEAM_", 6) == 0)
 		{
 			steamIdType = 2;
 		}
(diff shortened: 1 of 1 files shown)


core/PlayerManager.cpp
```diff
@@ -1632,7 +1632,7 @@ void PlayerManager::ProcessCommandTarget(cmd_target_info_t *info)
 
 		/* Do we need to look for a steam id? */
 		int steamIdType = 0;
-		if (strncmp(&info->pattern[1], "STEAM_", 6) == 0)
+		if (strncmp(&info->pattern[1], "STEAM_1", 6) == 0 || strncmp(&info->pattern[1], "STEAM_0", 6) == 0)
```
peace-maker: The length should be 7.

core/PlayerManager.cpp
```diff
@@ -1648,25 +1648,25 @@ void PlayerManager::ProcessCommandTarget(cmd_target_info_t *info)
 			{
 				size_t p, len;
 
-				strcpy(new_pattern, "STEAM_");
-				len = strlen(&info->pattern[7]);
-				/* Bound the copy so the "STEAM_" prefix, the copied bytes and the
+				strcpy(new_pattern, "STEAM_1");
```
peace-maker: I think a comment about why the universe in the steamid is hardcoded to 1 here would be good as it's not obvious while reading the code.

peace-maker (changes requested):

core/PlayerManager.cpp
psychonic: I don't think hardcoding STEAM_1 is right here. `pTarget->GetSteamId()` below is going to return with STEAM_0 on games that have the `UseInvalidUniverseInSteam2IDs` gamedata key enabled.

Kenzzer: okay time to rethink this then

Kenzzer: Alright so here's the new method. No more hardcoding to STEAM_1, instead the universe id gets changed based on whatever the game returns so that strcmp can succeed.

And to remain future-proof, this only happens if a STEAM_0 or STEAM_1 was provided. Say we ever support a STEAM_9 or whatever, there won't be any changes.

psychonic (approved):

peace-maker (approved):
As suggested here https://github.com/alliedmodders/sourcemod/issues/1240#issuecomment-618375690 reusing the logic to only compare the account id and ignore the universe from the admincache might simplify the logic here, but this is fine too.

Kenzzer:
If this bugs out and needs refactoring again, I'll definitively make a bigger diff and share some funcs. Otherwise I just wanted to make something small for now

PR #2513, opened 2026-07-19
Keep vtable hooks installed until extension unload

Mikusch:
Removing the last callback of a vtable hook deleted the `CVTableHookDetails` the next frame, but `KHook::RemoveHook(id, true)` only queues the actual unlink on a worker thread. Until that ran, the hook kept dispatching through freed memory. Instead of tearing hooks down at runtime, they now stay installed until the extension unloads, and dispatch just bails out early when nothing is hooked anymore. While in there, I also refactored the copy-pasted dispatch code in sdkhooks.

### DIFF
--- a/extensions/sdkhooks/extension.cpp
+++ b/extensions/sdkhooks/extension.cpp
@@ -221,8 +221,6 @@ bool SDKHooks::SDK_OnLoad(char *error, size_t maxlength, bool late)
 
 	SetupHooks();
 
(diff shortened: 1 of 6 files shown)


Kenzzer:
I'm almost okay to take this as-is, nice refactoring btw. But I will make a commit of my own on the PR to keep the unhook behaviour as it currently is, no specific reasons other than I don't want to diverge too much from SH baseline until the other PR lands. We can always revisit the unhooking process then.

PR #2516, opened 2026-07-20
Fix SlowScriptTimeout and EnableLineDebugging not being read

Mikusch:
A regression introduced by 890c711 (present on `master`). Both options are read in `logic_init`, which runs from `InitializeSourceMod` before core.cfg is parsed in `StartSourceMod`, so the lookups always return NULL.


psychonic (approved):
Thanks

PR #2518, opened 2026-07-24
Add support for dot-delimited entity property paths

Mikusch:
Current entity property functions return the first match by name, so props that appear under more than one table can't be accessed without the direct offset. For example: `m_Shared.tfsharedlocaldata.m_ScoreData.m_iCaptures` and `m_Shared.tfsharedlocaldata.m_RoundScoreData.m_iCaptures` (from TF2)

The VScript property accessors in the Source SDK 2013 solve this by resolving a dotted path. I've ported that behavior:

```sourcepawn
GetEntProp(client, Prop_Send, "m_ScoreData.m_iCaptures");
GetEntProp(client, Prop_Send, "m_RoundScoreData.m_iCaptures");
```

This is fully backwards compatible since the path resolution only kicks in when the name contains a `.` and can't be directly resolved as a prop, so this still works exactly as it has before:

```sourcepawn
GetEntProp(client, Prop_Send, "m_iCaptures");
```

### DIFF
--- a/core/HalfLife2.cpp
+++ b/core/HalfLife2.cpp
@@ -399,7 +399,70 @@ bool UTIL_FindDataMapInfo(datamap_t *pMap, const char *name, sm_datatable_info_t
 		pMap = pMap->baseMap;
 	}
 
-	return false; 
+	return false;
+}
+
+template <typename Table, typename Info>
+static bool UTIL_FindInTablePath(Table *pTable, const char *name, Info *info,
+								 bool (*search)(Table *, const char *, Info *),
+								 Table *(*descend)(const Info &))
+{
+	unsigned int baseOffset = 0;
+	const char *component = name;
+
+	while (true)
+	{
+		const char *dot = strchr(component, '.');
+		size_t len = dot ? (size_t)(dot - component) : strlen(component);
+
+		char buffer[256];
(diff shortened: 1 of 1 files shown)


MAGNAT2645:
Iirc such paths already kinda work because recently I managed to read properties like `m_ragdoll[1].pObject` and `m_ragdoll.listCount`.
But I did this with `GetEntData`. It does not work with `GetEntProp` (expects int property but these are raw data)

Mikusch:
> Iirc such paths already kinda work because recently I managed to read properties like `m_ragdoll[1].pObject` and `m_ragdoll.listCount`. But I did this with `GetEntData`. It does not work with `GetEntProp` (expects int property but these are raw data)

That already works because the dots are part of the prop name, not actually a nested table

MAGNAT2645:
> > Iirc such paths already kinda work because recently I managed to read properties like `m_ragdoll[1].pObject` and `m_ragdoll.listCount`. But I did this with `GetEntData`. It does not work with `GetEntProp` (expects int property but these are raw data)
> 
> That already works because the dots are part of the prop name, not actually a nested table

Ah, I see now. Thanks for the explanation.

psychonic (approved):

Issue #2532, opened 2026-08-03
Pass return value to WeaponSwitchPost, WeaponCanUsePost, WeaponCanSwitchToPost


SCP-076:
`SDKHook_ReloadPost` already passes the engine function's return value into the callback:
`void OnReloadPost(int weapon, bool bSuccess)`

The following hooks also return meaningful bool values, but their Post callbacks do not expose them:
`SDKHook_WeaponSwitchPost`
`SDKHook_WeaponCanUsePost`
`SDKHook_WeaponCanSwitchToPost`

Plugins need the real outcome  e.g., whether a weapon switch actually succeeded. Adding the return value as an extra callback parameter, consistent with ReloadPost, would cover this cleanly.

@Kenzzer this might be a quick addition during the khook migration, or I'm happy to submit a PR once that branch lands.

Kenzzer:
> @Kenzzer this might be a quick addition during the khook migration, or I'm happy to submit a PR once that branch lands.

I can shoulder the work, I rebase the branch every now and then so it's no biggie. But thank you for considering it!

PR #2535, opened 2026-08-03
Extend WeaponSwitch/Use callbacks

Kenzzer:
Fullfill feature request #2532 

It is indeed odd for sdkhooks to expose the return value of one post callback (weaponreload) but not weaponswitch/use. Let's correct that.

### DIFF
--- a/plugins/include/sdkhooks.inc
+++ b/plugins/include/sdkhooks.inc
@@ -274,6 +274,11 @@ typeset SDKHookCB
 	// WeaponSwitchPost
 	function void (int client, int weapon);
 
+	// WeaponCanUsePost
+	// WeaponSwitchPost
+	// WeaponCanSwitchToPost
+	function void (int client, int weapon, bool success);
+
 	// GetMaxHealth (ep2v and later)
 	function Action (int entity, int &maxhealth);
 
@@ -332,7 +337,7 @@ typeset SDKHookCB
 	// Reload
 	function Action (int weapon);
 
-	// Reload post
+	// ReloadPost
 	function void (int weapon, bool bSuccessful);
 
 	// CanBeAutobalanced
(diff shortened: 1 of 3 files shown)


Headline (approved):

plugins/include/sdkhooks.inc
peace-maker, line 270: The documentation should be updated to nudge developers to use the improved version, so the three hooks can be removed form the signature?

Kenzzer: Like adding a short comment why using the callback definition below is preferrable ?

peace-maker: I even thought it's ok to remove them from this one and only have them with the one with the `bool success` parameter.

But looking at how it was handled with other callbacks, specifically OnTakeDamage, before, we should keep it but add a comment that that parameter was an addition to sourcemod and add a proper feature provider so plugins can know whether that parameter is there or not at runtime for backwards compatibility.

psychonic:
This has the awkwardness that if they compile against a newer inc and run on older SM, the extra param will be garbage.

In the past, in at least one similar case, I think we used the feature cap system to define a capability they could detect to see if the feature was available. Another way to handle would be to add the extra param version as another hook type.

Kenzzer:
> This has the awkwardness that if they compile against a newer inc and run on older SM, the extra param will be garbage.

Given the new inc should be locked behind a bunch of new SP features, a newer plugin should fail to load on older installs.

plugins/include/sdkhooks.inc
Kenzzer: Okay I'll make a note of it now

peace-maker:
You can still compile using oldspcomp with the current includes. Adding a compatibility provider is the cleanest way to allow plugins to check if the parameter is available and notify the user if they really require it.

Kenzzer:
I'm more on about int64 etc... A plugin using old spcomp should still fail loading on 1.12

Issue #2526, opened 2026-07-31
Support MySQL 8.0 Connector/C for MySQL-DBI extension


HotoCocoaco:
## Help us help you
- [x] I have checked that my issue [doesn't exist yet](https://github.com/alliedmodders/sourcemod/issues).

## Is your feature request related to a problem? Please describe.

The MySQL-DBI extension is currently compiled against **MySQL 5.7.44** client library, which only supports the `mysql_native_password` authentication plugin.

MySQL 8.0.4+ changed the default authentication plugin from `mysql_native_password` to `caching_sha2_password` ([official MySQL blog](https://dev.mysql.com/blog-archive/mysql-8-0-4-new-default-authentication-plugin-caching_sha2_password/)). When a MySQL 8.0 server uses this default authentication, the MySQL-DBI extension fails to connect with:

```
Error: [1045]: proxy authentication failed: invalid password
```

This does **not** affect all MySQL 8.0 servers — if the server or database user is configured to use `mysql_native_password` (e.g., `IDENTIFIED WITH mysql_native_password BY '...'`), the current extension works fine. However, many users (especially those who don't control the database server) have no way to change the authentication method and are unable to connect.

Additionally, MySQL 5.7 reached End of Life in October 2023 and no longer receives security updates.

## Describe the solution you'd like

I have prepared a patch that migrates the MySQL-DBI extension from MySQL 5.7 client library to MySQL 8.0 Connector/C. The changes include:

- **Remove incompatible headers** — `my_global.h` and `my_sys.h` were deleted in MySQL 8.0
- **Replace deprecated `my_bool` type** — removed in MySQL 8.0, replaced with `bool`
- **Guard `MYSQL_OPT_RECONNECT`** — deprecated in MySQL 8.0 and removed in 8.1, guarded with `MYSQL_VERSION_ID < 80100`
- **Update `AMBuilder`** — detect and link against the correct library name on Windows: `libmysql.lib` (MySQL 8.0) vs `mysqlclient.lib` (MySQL 5.x); Linux continues to use `libmysqlclient.a` (static linking)
- **Update `configure.py`** — accept MySQL 8.0 Connector/C paths
- **Update `AMBuildScript`** — detect `MYSQL80` / `MYSQL80_64` environment variables, with MySQL 8.0 preferred over 5.x

The branch is available at: https://github.com/your-username/sourcemod/tree/dbi-mysql-8.0

## Describe alternatives you've considered

[...]

PR #2232, opened 2024-12-15
Bump MySQL to 5.7.44 on Windows too

peace-maker:
Dynamically link the CRT for the MySQL extension to be able to use libmysqlclient builds past the 5.5.54 release. This requires the [VC runtime](https://www.microsoft.com/en-us/download/details.aspx?id=48145) to be installed now while we previously linked them statically.

This allows to use the new `caching_sha2_password` default password encryption scheme from MySQL 8+ from Windows SRCDS too.

The build slaves need updating too if we want this.

Fixes #2231

## Old dll dependencies
```
dumpbin /dependents dbi.mysql.ext.dll
Microsoft (R) COFF/PE Dumper Version 14.37.32824.0
Copyright (C) Microsoft Corporation.  All rights reserved.


Dump of file dbi.mysql.ext.dll

File Type: DLL

  Image has the following dependencies:

    WSOCK32.dll
    Secur32.dll
    WS2_32.dll
    KERNEL32.dll
    ADVAPI32.dll

  Summary

      2BB000 .data
       1D000 .rdata
        7000 .reloc
        1000 .rsrc
       74000 .text
```

## New dll dependencies
```
dumpbin /dependents dbi.mysql.ext.dll
Microsoft (R) COFF/PE Dumper Version 14.37.32824.0
Copyright (C) Microsoft Corporation.  All rights reserved.


Dump of file dbi.mysql.ext.dll

File Type: DLL

  Image has the following dependencies:

    CRYPT32.dll
    WSOCK32.dll
    MSVCP140.dll
    Secur32.dll
    WS2_32.dll
    bcrypt.dll
    VCRUNTIME140.dll
    api-ms-win-crt-stdio-l1-1-0.dll
    api-ms-win-crt-convert-l1-1-0.dll
    api-ms-win-crt-heap-l1-1-0.dll
    api-ms-win-crt-runtime-l1-1-0.dll
    api-ms-win-crt-string-l1-1-0.dll
    api-ms-win-crt-environment-l1-1-0.dll
    api-ms-win-crt-math-l1-1-0.dll
    api-ms-win-crt-time-l1-1-0.dll
    api-ms-win-crt-utility-l1-1-0.dll
    api-ms-win-crt-filesystem-l1-1-0.dll
    KERNEL32.dll
    USER32.dll
    ADVAPI32.dll

  Summary

       B1000 .data
      33C000 .rdata
       1A000 .reloc
        1000 .rsrc
      1FD000 .text
```

Issue #2231, opened 2024-12-15
Build regression after upgrading MySQL version on Windows


peace-maker:
# Help us help you
  - [x] I have checked that my issue [doesn't exist yet](https://github.com/alliedmodders/sourcemod/issues).
  - [x] I have tried my absolute best to reduce the problem-space and have provided the absolute smallest test-case possible.
  - [x] I can always reproduce the issue with the provided description below.

# Environment
  * Operating System version:  Microsoft Windows Server 2022 10.0.20348
  * Game/AppID (with version if applicable): 
  * Current SourceMod version: 
  * Current SourceMod snapshot: 
  * Current Metamod: Source snapshot:
  - [ ] I have updated SourceMod to the [latest version](https://www.sourcemod.net/downloads.php) and it still happens.
  - [ ] I have updated SourceMod to the [latest snapshot](https://www.sourcemod.net/downloads.php?branch=dev) and it still happens.
  - [ ] I have updated SourceMM to the [latest snapshot](https://sourcemm.net/downloads.php?branch=dev) and it still happens.

# Description
#2192 broke building the mysql extension on Windows - but only after some time? Probably Github Actions caching the old MySQL version?
The [release notes](https://downloads.mysql.com/docs/mysql-5.5-relnotes-en.pdf) for MySQL 5.5.55 state under `Compilation Notes`:
> Windows builds now use the default runtime libraries (builds use the `/MD` flag). (Bug #25611609)

Which was the reason 5.5.54 was selected on Windows. We compile statically using `/MT` which links the libcmt.lib common runtime statically. The mysqlclient links the dynamic .dll now too which causes a conflict and would require the dynamic library to be available in the library path when loading the mysql extension.

This causes builds to fail to link in CI. The build servers don't use the checkout-deps.sh script and thus are still on the old mysql version 5.5.54.

# Problematic Code (or Steps to Reproduce)
  ```
Build sourcemod on Windows and grab dependencies using the `tools/checkout-deps.sh` script.
  ```

# Logs
```
LINK : warning LNK4098: defaultlib 'MSVCRT' conflicts with use of other libs; use /NODEFAULTLIB:library

LINK : warning LNK4286: symbol 'exit' defined in 'libucrt.lib(exit.obj)' is imported by 'mysqlclient.lib(str_alloc.obj)'

...


mysqlclient.lib(my_init.obj) : error LNK2001: unresolved external symbol __imp_getenv

[...]

PR #2539, opened 2026-08-07
Update MySQL-DBI extension to support MySQL 8.0 Connector/C

HotoCocoaco:
## Summary

Update the MySQL-DBI extension to compile against MySQL 8.0 Connector/C instead of MySQL 5.7. This enables support for the `caching_sha2_password` authentication plugin (default since MySQL 8.0.4) and allows connecting to modern MySQL 8.0+ servers without server-side configuration changes.

## Motivation

The MySQL-DBI extension is currently compiled against the MySQL 5.7 client library, which only supports the `mysql_native_password` authentication plugin. MySQL 8.0.4+ changed the default authentication plugin to `caching_sha2_password` ([official announcement](https://dev.mysql.com/blog-archive/mysql-8-0-4-new-default-authentication-plugin-caching_sha2_password/)). When a MySQL 8.0 server uses this default, the extension fails to connect with:

```
Error: [1045]: proxy authentication failed: invalid password
```

Additionally, MySQL 5.7 reached End of Life in October 2023 and no longer receives security updates.

### 1. Remove incompatible MySQL 8.0 headers (`MyDriver.h`)

MySQL 8.0 removed `my_global.h` and `my_sys.h`. Only `<mysql.h>` is needed.

### 2. Replace deprecated `my_bool` type (`MyBoundResults.h`, `MyBoundResults.cpp`, `MyDriver.cpp`)

`my_bool` was removed in MySQL 8.0. Replaced with standard `bool`.

### 3. Guard `MYSQL_OPT_RECONNECT` (`MyDriver.cpp`)

`MYSQL_OPT_RECONNECT` was deprecated in MySQL 8.0 and removed in 8.1. Protected with `#if MYSQL_VERSION_ID < 80100`.

### 4. Update build configuration (`AMBuilder`, `AMBuildScript`, `configure.py`)

- **AMBuilder**: Detect `libmysql.lib` (MySQL 8.0) vs `mysqlclient.lib` (MySQL 5.x) on Windows. Linux continues to use `libmysqlclient.a` (static linking).
- **AMBuildScript**: Auto-detect `MYSQL80` / `MYSQL80_64` environment variables, with MySQL 8.0 preferred over 5.x.
- **configure.py**: Update help text to accept MySQL 8.0 Connector/C paths.

## Backwards compatibility

The same code compiles against MySQL 5.x headers without issues. The MySQL 8.0 client library supports connecting to MySQL server versions 5.7, 8.0, 8.4 LTS, and 9.0+.

## Testing

- Compiled on Ubuntu 22.04 with `libmysqlclient-dev` (MySQL 8.0 Connector/C)
- Static linking against `libmysqlclient.a` (same approach as current builds)
- Successfully connects to a remote MySQL 8.0.40 server
- Successfully connects to MySQL 5.7 servers

## Windows 32-bit note

The official MySQL 8.0 Connector/C Windows installer no longer provides a 32-bit library. This PR has been tested on **Linux 32-bit** only. For Windows 32-bit builds, the project would need to host its own build of the MySQL 8.0 Connector/C static library, similar to the approach used for the PostgreSQL extension (see `extensions/pgsql/README.md`).

## Related

- Issue #2526 — My issue.
- PR #2232 — Previous MySQL client library bump (5.5 → 5.7.44 on Windows)
- Issue #2231 — Windows build regression that motivated PR #2232
- PR #2086 — Alternative approach using MariaDB connector

### DIFF
--- a/AMBuildScript
+++ b/AMBuildScript
@@ -144,10 +144,13 @@ class SMConfig(object):
         if builder.options.mysql_path:
           self.mysql_root['x86'] = builder.options.mysql_path
         else:
-          for i in range(10):
-            self.mysql_root['x86'] = ResolveEnvPath('MYSQL57', 'mysql-5.' + str(i))
-            if self.mysql_root['x86']:
-              break
+          # Try MySQL 8.0 first, then MySQL 5.x
+          self.mysql_root['x86'] = ResolveEnvPath('MYSQL80', 'mysql-8.0')
+          if not self.mysql_root['x86']:
+            for i in range(10):
+              self.mysql_root['x86'] = ResolveEnvPath('MYSQL57', 'mysql-5.' + str(i))
+              if self.mysql_root['x86']:
+                break
         if not self.mysql_root['x86'] or not os.path.isdir(self.mysql_root['x86']):
           raise Exception('Could not find a path to MySQL. Configure with --no-mysql to disable it.')
         self.mysql_root['x86'] = Normalize(self.mysql_root['x86'])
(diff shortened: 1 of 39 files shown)


psychonic:
Thanks!

There are a few issues here though

* 8.4 would probably be preferred over 8.0 since it's still supported as LTS
* As this doesn't touch the CI files, official releases would still be built with 5.7
* We probably don't need to worry about targeting multiple versions
* The checkout-deps scripts should be updated as well
* It looks like all touched files had their line endings changed en masse, hugely inflating the diff

I have a change locally to upgrade to 8.4 that I just didn't have a chance to test yet. I'll get to that soon, closing this one in favor of that.

HotoCocoaco:
> Thanks!
> 
> There are a few issues here though
> 
> * 8.4 would probably be preferred over 8.0 since it's still supported as LTS
> * As this doesn't touch the CI files, official releases would still be built with 5.7
> * We probably don't need to worry about targeting multiple versions
> * The checkout-deps scripts should be updated as well
> * It looks like all touched files had their line endings changed en masse, hugely inflating the diff
> 
> I have a change locally to upgrade to 8.4 that I just didn't have a chance to test yet. I'll get to that soon, closing this one in favor of that.

Fair. But I still don't have ideas about Windows 32bit build.

PR #2547, opened 2026-08-15
Fix x64 float parameter bugs

Mikusch:
Some issues I discovered while working on #2546. Decided to submit them in their own PR since these are also present on `master`.

Some of the x64 codegen helpers forgot a required SIB byte when writing to the stack, so the JIT would spit out garbage instructions (and crash the server) any time a float argument didn't fit in the registers. It's pretty uncommon and only happens when the param spills to the stack due to the first four positional params already being taken. Fixed that, plus a nearby copy-paste bug.

It's a similar problem with Windows vararg calls. A float that overflowed the register budget could end up feeding a garbage "register" straight into the generated code.


psychonic (approved):
Thanks. I added some other related fixes as well

PR #2548, opened 2026-08-15
Switch MySQL extension to use (latest) MariaDB connector (fixes 2526)

psychonic:
This switches the MySQL extension to build against (our own copy of) the MariaDB Connector/C library rather than the MySQL one.

Among other things, this brings in support for `caching_sha2_password` auth, compiled in statically rather than also requiring a separate plugin library for that. Just upgrading the MySQL connector would have gotten us the same, but with similar build hurdles for static linking.

The other reason we build our own copy is that MySQL no longer ships any x86 libs, only x86_64. MariaDB also no longer ships x86 for Linux. MariaDB was chosen over MySQL primarily because its connector code is in its own repo/build rather than being entangled with the rest of the client/server code.


peace-maker:
> Among other things, this brings in support for `caching_sha2_password` auth, compiled in statically rather than also requiring a separate plugin library for that. Just upgrading the MySQL connector would have gotten us the same, but with similar build hurdles for static linking.

The 5.7.23 client library [supports `caching_sha2_password`](
https://dev.mysql.com/doc/refman/5.7/en//caching-sha2-pluggable-authentication.html) and we built against 5.7.44 before so I don't get where this claim comes from lately. Maybe some build regression while switching to GitHub Actions releases? So I don't think this was necessary but still welcome to link against a newer client library.

psychonic:
It supports the plugin for it, but we weren't shipping the plugin

peace-maker:
It's built into libmysqlclient that we link against.
https://dev.mysql.com/doc/refman/5.7/en//caching-sha2-pluggable-authentication.html#caching-sha2-pluggable-authentication-installation

It's been a while but I'm fairly certain I tested it when I bumped the version for this exact auth compatibility. But I'll take your word for it since you tested it more recently I guess.

PR #2550, opened 2026-08-18
Prevent lock of main thread

PlayBoy31:
if the seperate thread is making a query but the SQL server is busy, and if a that time the main thread is doing a SQL_EscapeString (in a sp plugin) then the main thread is waiting for the lock. It is no necessary to do this lock here, its not a query.

### DIFF
--- a/extensions/mysql/mysql/MyDatabase.cpp
+++ b/extensions/mysql/mysql/MyDatabase.cpp
@@ -157,8 +157,6 @@ const char *MyDatabase::GetError(int *errCode)
 
 bool MyDatabase::QuoteString(const char *str, char buffer[], size_t maxlength, size_t *newSize)
 {
-	std::lock_guard<std::recursive_mutex> lock(m_FullLock);
-
 	unsigned long size = static_cast<unsigned long>(strlen(str));
 	unsigned long needed = size * 2 + 1;
 


dvander:
Have you confirmed that the internal implementation is okay with this being used concurrently with other MySQL API calls?

PlayBoy31:
> Have you confirmed that the internal implementation is okay with this being used concurrently with other MySQL API calls?

Yes, I have tested the fix and it fix the regrtession i observed since the #2528 (=> my game server was laggy when doing large operations on the  SQL server, both are not on the same machine)

I have lots of MYSQL API call in my plugins, all is fine

dvander:
That's not quite what I meant... testing empirically is different from verifying the implementation. I downloaded the client library source and asked Gemini for an analysis:

```
  ## Analysis: Is mysql_real_escape_string safe to run concurrently with mysql_query?

  No, it is not safe. Both functions operate on the same MYSQL * handle without any synchronization, and there is a data race on mysql->server_status.

  ### What mysql_real_escape_string reads

  From libmysql.cc:1059-1071:

  1. mysql->server_status — checked for SERVER_STATUS_NO_BACKSLASH_ESCAPES to decide the escape strategy
  2. mysql->charset — passed to escape_string.cc:58 or escape_quotes_for_mysql() as a read-only CHARSET_INFO *

  The escape functions themselves (escape_string.cc) are pure: they only read from the CHARSET_INFO pointer and the input buffer, writing only to the output buffer. No shared mutable state is
  touched there.

  ### What mysql_query writes

  libmysql.cc:676-678 → client.cc:8371-8386 → mysql_send_query + read_query_result

  During read_query_result, the server's response is parsed and mysql->server_status is written in multiple places, e.g. at client.cc:1156:

    mysql->server_status = uint2korr(pos);

  ### The race

   Thread A (mysql_real_escape_string)                                                           | Thread B (mysql_query)
  -----------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------
   Reads mysql->server_status to decide escape mode                                              | Writes mysql->server_status from the OK/EOF packet

  This is an unsynchronized concurrent read/write of the same unsigned int field on the same MYSQL * object — a textbook data race (undefined behavior in C/C++).

  ### Practical consequences

  • The escape function could see a torn or stale server_status, meaning it might choose the wrong escaping strategy (backslash-based vs. quote-doubling). This could lead to SQL injection if
  the server has NO_BACKSLASH_ESCAPES mode enabled but the escape function doesn't see the flag.
  • The charset pointer is set during connection setup and SET NAMES/SET CHARACTER SET operations, not during normal query execution. So the charset read is safe only if no thread is
  concurrently changing the connection's character set.

  ### Bottom line

  The MySQL C client library is explicitly not thread-safe per-connection. A single MYSQL * handle must not be used concurrently from multiple threads without external synchronization. This is
  documented, and the code confirms it: there are no locks inside these functions.
```

dvander:
You're free to sanitize strings with a function other than SQL_EscapeString BTW. you could also propose a way to do the escaping off the main thread.

PlayBoy31:
The fix introduced by #2528 fix the race but it also introduce a jitter and a deadlock with the escape string.
Perhaps in the meaning time, its more safe to revert complety #2528 until a proper fix is established.

psychonic:
`SQL_EscapeString` indeed uses the non-threaded DBI implementation.

Using the `Database.Format` methodmap function (or just `SQL_FormatString` directly) generally ends up being cleaner, without needing a call and buffer per parameter, and avoids touching the driver or threading issues.

PR #2564, opened 2026-09-05
bintools (x64): Use an absolute call when the callee is more than 2 GB away

Mikusch:
`Write_CallFunction` emits `call rel32` whenever `diff >> 32` is 0 or -1, which also accepts displacements between 2 and 4 GB. Those get sign-extended and the call lands 4 GB away from the callee.

This has been here for 9 years, but only triggers now because SP2 reserves 2 GB for virtual addresses, which pushes the code pools more than 2 GB away from server_srv.so. So currently, every direct SDKCall crashes on 64-bit Linux.

Fall back to an absolute call when the callee is out of reach, and size the stub for the longer form up front.


dvander:
Doesn't this break the two-pass size reservation? The first pass must always assume >= the space needed of the second pss.

Mikusch:
> Doesn't this break the two-pass size reservation? The first pass must always assume >= the space needed of the second pss.

That's why the `outbase` check exists. It's NULL during the sizing pass, so pass 1 always counts the 12 bytes for `movabs+call rax` while pass 2 emits 5 or 12.

PR #2566, opened 2026-09-06
GetFunctionByName: return the runtime's funcid instead of deriving it from the public index

### LINKED ISSUE #1168: `Call_Finish()` returns an incorrect result and may trigger `Invalid instruction` in SourceMod 1.13.7454
When using the dynamic call API (`Call_StartFunction` / `Call_PushCell` / `Call_Finish`) to invoke a SourcePawn function, `Call_Finish()` appears to produce an incorrect return value.

The following code should call `Func1(1, 2)` and store its return value (`3`) in `result`, but the actual output is:

```text
result = 0
```

Additionally, when an unrelated third-party header is included, the same code triggers an `Invalid instruction` exception inside `Call_Finish()`.

---

## Minimal reproduction

```sourcepawn
#include <sourcemod>

public void OnPluginStart()
{
    Function func = GetFunctionByName(INVALID_HANDLE, "Func1");

    int result;

    Call_StartFunction(INVALID_HANDLE, func);
    Call_PushCell(1);
    Call_PushCell(2);
    Call_Finish(result);

    PrintToServer("result = %d", result);
}

public int Func1(int a, int b)
{
    return a + b;
}
```

### Actual result

```text
result = 0
```

### Expected result

```text
result = 3
```

---

## Additional issue when including another header

When the plugin additionally includes a third-party header:

[log4sp.zip](https://github.com/user-attachments/files/31845350/log4sp.zip)

```sourcepawn
#include <sourcemod>

#include <log4sp>

public void OnPluginStart()
{
    Function func = GetFunctionByName(INVALID_HANDLE, "Func1");

    int result;

    Call_StartFunction(INVALID_HANDLE, func);
    Call_PushCell(1);
    Call_PushCell(2);
    Call_Finish(result);

    PrintToServer("result = %d", result);
}

public int Func1(int a, int b)
{
    return a + b;
}
```

SourceMod reports the following exception:

```text
L 09/05/2026 - 00:26:27: [SM] Exception reported: Invalid instruction
L 09/05/2026 - 00:26:27: [SM] Blaming: test.smx
L 09/05/2026 - 00:26:27: [SM] Call stack trace:
L 09/05/2026 - 00:26:27: [SM]   [0] Call_Finish
L 09/05/2026 - 00:26:27: [SM]   [1] Line 14, d:\Dev\CPP\log4sp\sourcemod\scripting\testsuite\mock\test.sp::OnPluginStart
result = 0
```

F1F88:
Resolved alliedmodders/sourcepawn#1168

[sm_GetFunctionByName()](https://github.com/alliedmodders/sourcemod/blob/d75d4101005895919308f0209deb58fa95ef2475/core/logic/smn_functions.cpp#L109) derived the function id from the public table index by reimplementing SourcePawn's (index << 1) | 1 encoding in a local [PublicIndexToFuncId()](https://github.com/alliedmodders/sourcemod/blob/d75d4101005895919308f0209deb58fa95ef2475/core/logic/smn_functions.cpp#L95) helper. That encoding is owned by the VM, and sp_public_t already carries the authoritative funcid, so the local copy duplicated it and could silently disagree with the runtime.

Fetch the public entry and return its funcid instead, and fall back to INVALID_FUNCTION if the index lookup fails rather than returning an id that was never validated against the public table.

### DIFF
--- a/core/logic/smn_functions.cpp
+++ b/core/logic/smn_functions.cpp
@@ -91,12 +91,6 @@ public:
 } g_ForwardNativeHelpers;
 
 
-/* Turn a public index into a function ID */
-inline funcid_t PublicIndexToFuncId(uint32_t idx)
-{
-	return (idx << 1) | (1 << 0);
-}
-
 /* Reset global function/forward call variables */
 inline void ResetCall()
 {
@@ -135,8 +129,13 @@ static cell_t sm_GetFunctionByName(IPluginContext *pContext, const cell_t *param
 		return pContext->GetNullFunctionValue();
 	}
 
+	/* Turn a public index into a function ID */
+	sp_public_t *func;
+	if (pPlugin->GetBaseContext()->GetBaseRuntime()->GetPublicByIndex(idx, &func) != SP_ERROR_NONE)
+		return pContext->GetNullFunctionValue();
+
 	/* Return function ID */
-	return PublicIndexToFuncId(idx);
+	return func->funcid;
 }
 
 static cell_t sm_CreateGlobalForward(IPluginContext *pContext, const cell_t *params)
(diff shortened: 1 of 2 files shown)


peace-maker (approved):

peace-maker:
Thanks!

PR #2568, opened 2026-09-08
Remove support for reloading extensions 

Headline:
There are outstanding issues relating to extension loading / reloading. The feature is trivialized by `meta reload` - which is perfectly fine for iterating in development. 

Existing extensions will need to be recompiled. Might be a good time for it considering recent V10 version. We can't simply stub everything out like I thought previously since third party plugins copied from our example would disregard `UnloadExtension()`'s return value. Leading to CExtension holding a dangling m_pAPI

AI Disclosure: Used an LLM to workshop, help find references in code, as well as update any documentation that needed changing.

### DIFF
--- a/bridge/include/BridgeAPI.h
+++ b/bridge/include/BridgeAPI.h
@@ -35,7 +35,7 @@ namespace SourceMod {
 
 // Add 1 to the RHS of this expression to bump the intercom file
 // This is to prevent mismatching core/logic binaries
-static const uint32_t SM_LOGIC_MAGIC = 0x0F47C0DE - 57;
+static const uint32_t SM_LOGIC_MAGIC = 0x0F47C0DE - 58;
 
 } // namespace SourceMod
 
--- a/bridge/include/IExtensionBridge.h
+++ b/bridge/include/IExtensionBridge.h
@@ -50,7 +50,6 @@ public:
 	virtual void CallOnCoreMapStart(edict_t *edictList, int edictCount, int maxClients) = 0;
 	virtual IExtension *GetExtensionFromIdent(IdentityToken_t *token) = 0;
 	virtual void BindChildPlugin(IExtension *ext, SMPlugin *plugin) = 0;
-	virtual void AddRawDependency(IExtension *myself, IdentityToken_t *token, void *iface) = 0;
 	virtual const CVector<IExtension *> *ListExtensions() = 0;
 	virtual void FreeExtensionList(const CVector<IExtension *> *list) = 0;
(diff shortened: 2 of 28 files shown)


Kenzzer:
Kudos on the AI disclosure. But unless I missed something, MMS doesn't have a `meta reload` command. There's `meta retry` but iirc this command is only for plugins that failed an initial load. So at the moment reloading sourcemod would involve

`meta unload X`
`meta load sourcemod`

Perhaps adding a command to unload sourcemod would be nice to have instead ? Metamod takes an ID to unload sourcemod, and I can sense this would become annoying during dev cycles.

`sm unload`
`meta load sourcemod`

would work out well, instead of having to find the ID of sourcemod through `meta list`, some sm extensions can bump the plugin ID numbers as well, so there's no reliable well to guess the ID other than checking. So I would introduce that command

Headline:
Heh I saw a meta reload mentioned somewhere, can't say I actually checked! Yeah we'll definitely need to make that thanks for the note

bottiger1:
I am against this. If this gets merged I will maintain my own fork of sourcemod. 

Doesn't seem like the author even uses reload given that he doesn't even know meta reload doesn't exist.

dvander:
diffstat is great. pre-emptive lgtm pending MM:S reload command getting added on the 1.12 branch.

bottiger1:
I don't think any of you ever reload or even tested it. 

Meta unload x, meta load sourcemod doesn't work either.

Not to mention how laggy it will be, and now you require every single plugin and extension to be built for reloading, which is way worse than the current situation.

KyleSanderson:
> I don't think any of you ever reload or even tested it.

I did frequently, unloading and reloading is absolutely a valid feature, and one worth retaining.

> Meta unload x, meta load sourcemod doesn't work either.

Yes, an ancient pain point. There were a couple PRs to help improve this, but it appears they were closed and hard deleted a couple hours ago by the author - which is a shame.

bottiger1:
> reloading is absolutely a valid feature, and one worth retaining

The project owner and another maintainer seem to disagree and seeing how this PR is approved already yet any related to fixing reload had been sidelined, it seems obvious which way this project is heading.

I'm no longer concerning myself with this. I've already hard forked sourcemod so I don't have to worry about these things.

> There were a couple PRs to help improve this, but it appears they were closed and hard deleted a couple hours ago by the author - which is a shame.

Not sure if this supposed to be a snarky reference to me, my PRs had nothing to do with meta reload, but they are still available on github, just remade the repo under an org.

PR #2576, opened 2026-09-15
Only hide statically-linked libstdc++ symbols

Mikusch:
#2533 was a bit too broad with excluding statically-linked symbols and broke SM loading on Linux. Narrow it to *only* exclude libstdc++.

I let the mock tests CI step run on a fork to ensure it passes and confirmed that the built binary only hides the intended symbols.


psychonic (approved):

Kenzzer:
> Kenzzer [:3],  — 03/08/2026 07:39
> sure, if its more than changing the visibility flag because otherwise we would need a wide audit to make sure we haven't broken anything
> Mikusch [:3],  — 03/08/2026 07:40
> visibility flag is already set in SM
> but that doesn't apply to symbols coming from static libs
> it'd be -Wl,--exclude-libs,ALL
> i already audited this and nothing relies on it, so it's fairly safe to do

This is why I noted back then these kind of changes need to be audited, because we don't know what behaviour we might have accidentally relied on for the whole of SM to load properly.

Following that, would it be possible to know the load errors encountered with the previous changes. I'd like to keep note of it.

Edit: Reported on discord, the error was a missing `CreateInterface` export.

